PR #4, opened 2020-11-13
Make keypairs

cafca:
https://github.com/p2panda/sesamoid/issues/1

Pairs with https://github.com/p2panda/beep-boop/pull/12

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,8 +3,8 @@ use rand::rngs::OsRng;
 use std::panic;
 use wasm_bindgen::prelude::wasm_bindgen;
 
-#[wasm_bindgen]
-pub fn set_panic_hook() {
+#[wasm_bindgen(js_name = setWasmPanicHook)]
+pub fn set_wasm_panic_hook() {
     panic::set_hook(Box::new(console_error_panic_hook::hook));
 }
 
@@ -27,12 +27,12 @@ impl KeyPair {
         }
     }
 
-    #[wasm_bindgen]
+    #[wasm_bindgen(js_name = publicKeyBytes)]
     pub fn public_key_bytes(&self) -> Box<[u8]> {
         Box::from(self.public.to_bytes())
     }
 
-    #[wasm_bindgen]
+    #[wasm_bindgen(js_name = privateKeyBytes)]
     pub fn private_key_bytes(&self) -> Box<[u8]> {
         Box::from(self.private.to_bytes())
     }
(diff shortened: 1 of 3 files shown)


cafca:
status: tests run but importing the lib from Javascript fails with 

```
Uncaught (in promise) ReferenceError: can't access lexical declaration '__wbg_self_1c83eb4471d9eb9b' before initialization
```

adzialocha:
> status: tests run but importing the lib from Javascript fails with
> 
> ```
> Uncaught (in promise) ReferenceError: can't access lexical declaration '__wbg_self_1c83eb4471d9eb9b' before initialization
> ```

This is an issue related to Webpack 5 + wasm-bindgen: https://github.com/rustwasm/wasm-bindgen/issues/2343 and was fixed in https://github.com/p2panda/beep-boop/pull/12/commits/f7e47706581033dd70084ccf179089d9dc9ab33e by using the (older and sadly uglier) sync wasm-import method of Webpack as a workaround.

adzialocha:
Also, I'm running into problems with the post-build steps of `wasm-opt` similar to this problem described here: https://github.com/rustwasm/wasm-pack/issues/886 the recommended steps of adding the `--enable-mutable-globals` flag did not solve the problem so I pinned `wasm-bindgen` to an older version `0.2.65` which made it work again.

src/lib.rs
```diff
@@ -1,65 +1,51 @@
-extern crate wasm_bindgen;
-extern crate console_error_panic_hook;
```
cafca: This seemed quite helpful to me in general. Why not keep it?

src/lib.rs
```diff
@@ -1,65 +1,51 @@
...
     #[wasm_bindgen(constructor)]
-    pub fn new () -> PandaKeyPair {
+    pub fn new() -> Self {
         let mut csprng: OsRng = OsRng {};
-        let key_pair: DalekKeypair = DalekKeypair::generate(&mut csprng);
-        println!("{:?}", key_pair.to_bytes().to_vec());
-        PandaKeyPair {
+        let key_pair = Ed25519Keypair::generate(&mut csprng);
+
+        Self {
```
cafca, line 25: aha!

adzialocha: Yes, here I renamed `PandaKeyPair` to `KeyPair` but I'm open to change it back or to something else.

I'm not sure yet how we should handle naming in general, I like the idea of having more-panda names for repositories but when it comes to classes, variables, methods etc. I prefer to keep it simple especially when the system gets complicated, to not make it even harder for new people to understand what is going on (think the puns disappear after you read them first and then it gets confusing). Maybe we can have a simple guideline / styleguide for naming things?

cafca: I was commenting because of the use of `Self` for the return value :D I agree with everything else! I originally named it PandaKeyPair because we already had KeyPair in the namespace when I wrote that. Live programming etc etc

cafca:
I have added the hex encoding! 

#7

PR #8, opened 2020-12-05
Use hex-encoded key method as defaults

adzialocha:
Renames the `KeyPair` methods which makes the hex encoded representation of the keys the library API default:

* `privateKeyHex` to `privateKey`
* `publicKeyHex` to `publicKey` 

This should make the API more clear for JavaScript clients as they will need to use the hex version anyways for storing private keys (as LocalStorage for example only accepts strings) or displaying the public address of an user. The bytes representation is secondary and mostly only used for Bamboo encoding or message verification (which is handled internally by sesamoid anyways).

Also, this should behave similar to our ideas to use something like `toJSON` or `fromJSON` (see https://hackmd.io/kRb3ma1AQhyhMCQKBnBTKQ) just without using actually JSON but only the private key string.

```js
const keyPair = new KeyPair();
window.localStorage.setItem('privateKey', keyPair.privateKey());

const newKeyPair = KeyPair.fromPrivateKey(window.localStorage.getItem('privateKey'));

console.log(`My address is ${newKeyPair.publicKey()}`);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,13 +12,24 @@ Utility library for p2panda applications.
 Use this in a Webpack 5 setup by enabling the [experimental `syncWebAssembly`
 option](https://webpack.js.org/configuration/experiments/).
 
-Create a key pair:
+### Create Ed25519 key pair
 
-```
+```js
 import('sesamoid').then(({ KeyPair }) => {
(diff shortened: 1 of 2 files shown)


cafca (approved):
It is good!

PR #16, opened 2020-12-07
Configure CI to automatically build and test

cafca:
Adds a Github workflow to build and test both the Rust and Javascript projects.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,73 @@
+name: tests
```
adzialocha, line 1: Right now GitHub shows me this when checking the tasks which is a little confusing and not describing whats actually running:

```
@github-actions
tests / rust build (push)

@github-actions
tests / js build (push)
```

Maybe name this `Build and test` and the jobs `p2panda-rs` and `p2panda-js` instead of `rust build` and `js build`? Then it would read:

```
Build and test / p2panda-rs
Build and test / p2panda-js
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,73 @@
...
+  js:
+    name: js build
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Read node version from .nvmrc
+        working-directory: p2panda-js
```
adzialocha, line 41: Do you know if its possible to set the working directory only once somewhere? I could find this discussion here: https://github.community/t/use-working-directory-for-entire-job/16747 and it seems to be a problem, but also in the end someone solved it like this?

```
defaults:
  run:
    working-directory: ./your_working_dir
jobs:
  ...
```

Not sure if it will help in our case, its more of a question :thinking:

.github/workflows/tests.yml
```diff
@@ -0,0 +1,73 @@
...
+            ~/.cargo/registry
+            ~/.cargo/git
+            ~/.npm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - run: npm ci
```
adzialocha, line 64: Nitpicking: Can be good to give all job steps names in the same manner as in line 49!

adzialocha (changes requested):
Yay! :birthday: I mostly have questions around naming things, otherwise it looks great!

.github/workflows/tests.yml
cafca: I looked around a bit and tried cd but that didn't work :( i can try the syntax you linked, yes!

.github/workflows/tests.yml
cafca: Agreed, a bit of explanation never hurts. I often omit the names for commands that are also used outside Github actions like `npm ci` or `npm start`.

adzialocha (approved):

PR #18, opened 2020-12-12
Replace standardx with prettier

cafca:
We had a couple of problems working with Standard + Typescript so this is switching linting to prettier.


adzialocha (approved):
Hurra!

PR #21, opened 2021-01-09
NodeJS and browser support

adzialocha:
The goal is to have a nice bundled library which runs both in NodeJS and browsers without big setup and bundler configurations:

```
├── index.d.ts (TypeScript definitions)
├── index.js (NodeJS entry)
├── index.js.map (Source-Map for NodeJS)
├── index.min.js (Browser entry, w. inline wasm)
├── index.min.js.map (Source-Map for Browser)
└── wasm
    ├── index_bg.wasm (External .wasm file which gets imported in NodeJS)
    └── index.js
```

- [x] Build an `index.js` in the library export which runs in NodeJS contexts
- [x] Build an `index.min.js` in the library export which is optimized (minified, inline wasm) for web browsers
~~- [x] Export one single TypeScript definitions file `index.d.ts` in the library~~
- [x] Run tests successfully against the NodeJS build

**Edit:** Emitting a complete and valid Typescript declaration is omitted from this PR and will be fixed separately.


adzialocha:
```
$ cat index_bg.wasm | wc -c
107526

$ cat index_bg.wasm | muter -c base64 | wc -c
143368

$ cat index_bg.wasm | muter -c hex | wc -c
215052

$ cat index_bg.wasm | muter -c hex | gzip | wc -c
69693

$ cat index_bg.wasm | muter -c base64 | gzip | wc -c
71086
```

adzialocha:
This should build two versions of the library now: One which is targeting NodeJS contexts and another for browsers. The library runs easily now in browsers by embedding the wasm code as an base64 encoded string inside the bundle. I also managed to make the tests run again by pointing them against the webpack build for node targets.

Some things I need help with:

* Typescript! I used `ts-ignore` in many places where I didn't know what to do ..
* Typescript! It exports too many definition files in the final `lib` folder, how can we only get one `index.d.ts` instead of one for each module?

p2panda-js/webpack/webpack.browser.ts
```diff
@@ -2,7 +2,7 @@ import * as webpack from 'webpack';
 import config, { getWasmPlugin, tsRule } from './webpack.common';
 
 /*
- * Extended configuration to buildesModuleInteropesModuleInterop library targeting modern browsers:
```
cafca: oopsie

cafca:
We should consider adding tests for the browser build as well to make sure we don't accidentally break that.

adzialocha:
I've removed some unneeded parts / leftovers from playing whack-a-mole and fixed an issue where the `index.d.ts` would try to import the wasm from `../wasm-node` which would not exist anymore in a library package.

I successfully used the library in `beep-boop` with typings and in NodeJS, also the tests look good!

Can you have another look @cafca? And then I'm happy to merge it (what a ride)! :balloon:

p2panda-js/webpack/webpack.common.ts
```diff
@@ -12,24 +12,38 @@ export function getPath(...args: Array<string>): string {
...
+          // will export declaration files with wrong import paths in library
+          ...(target === 'node'
+            ? {
+                compilerOptions: {
+                  paths: {
+                    'wasm-node': ['./lib/wasm'],
+                  },
+                },
+              }
+            : {}),
```
cafca, line 39: Nice. I was looking for a fix for this.

p2panda-js/src/wasm-adapter/index.ts
```diff
@@ -6,7 +6,7 @@
 declare const BUILD_TARGET_WEB: boolean;
 
 const adapter = BUILD_TARGET_WEB
-  ? (require('~/wasm-adapter/browser') as typeof import('./browser'))
-  : (require('~/wasm-adapter/node') as typeof import('./node'));
+  ? (require('~/wasm-adapter/browser') as typeof import('~/wasm-adapter/browser'))
+  : (require('~/wasm-adapter/node') as typeof import('~/wasm-adapter/node'));
```
cafca, line 10: Why do you prefer this?

adzialocha: We introduced aliases, either we use them or remove aliases at all :-D

cafca: I prefer using whichever form is shorter / easier to parse. I would take the alias over `../../../..` any time but wouldn't use it if we are just descending one directory.

p2panda-js/src/wasm-adapter/node.ts
```diff
@@ -0,0 +1,7 @@
+type WasmNode = typeof import('wasm-node');
+
+const wasm: Promise<WasmNode> = import('wasm-node').then((lib) => {
+  return lib;
+});
+
+export default wasm;
```
cafca, line 7: This whole thing feels like misdirection to me. Why not just import `wasm-node` from `index.ts` here?

p2panda-js/webpack/webpack.common.ts
```diff
@@ -0,0 +1,70 @@
...
+        },
```
cafca, line 40: I will leave it here but we have to revisit as part of #24

cafca (approved):
🐼☺️

p2panda-js/src/wasm-adapter/node.ts
adzialocha: Yes, right! I had it explicitly using the Promise before to have more control over what should get imported (there is actually loads of "garbage" in `lib` which gets public if we do it like this, originally I only deconstructed `KeyPair` to prevent this. For now we can remove this and clean up later, think this should be connected to the TypeScript stuff anyways.

Issue #2, opened 2020-11-04
Methods to encode / decode Bamboo entries


adzialocha:
https://github.com/pietgeursen/bamboo-rs/blob/master/bamboo-core/src/entry/mod.rs

PR #23, opened 2021-01-15
Add basic data types

adzialocha:
**Types to be implemented:**

- [x] Author
- [x] Entry
- [x] EntrySigned
- [x] Hash
- [x] LogId
- [x] Message
- [x] MessageEncoded
- [x] SeqNum

**Other stuff:**

- [x] `hash::to_yamf_hash` -> conversion
- [x] create an issue for validation (https://github.com/p2panda/p2panda/issues/35)
- [x] `Message` <-> `MessageEncoded` conversion traits (instead of `decode` and `from_encoded`)
- [ ] mergeeeee 🤩

**Next up:**

- [ ] Create structs for intermediate type containing: `(EntrySigned, MessageEncoded)` tuple?
- [ ] Modify errors returned so they actually contain the error type and not just Anyhow errors
- [ ] Implement sqlx Type trait for atomic structs https://github.com/p2panda/p2panda/issues/37
- [ ] Implement serde Deserialize and Serialize traits for atomic structs https://github.com/p2panda/p2panda/issues/36
- [ ] Implement `std::hash::Hash` for `Entry` and `Message` https://github.com/p2panda/p2panda/issues/32

Closes: https://github.com/p2panda/p2panda/issues/2


p2panda-rs/src/atomic/entry.rs
```diff
@@ -0,0 +1,168 @@
...
+impl Validation for Entry {
+    fn validate(&self) -> Result<()> {
+        // First entries do not contain any sequence number or links. Every other entry has to
+        // contain all information.
+        if (self.entry_hash_backlink.is_none()
+            && self.entry_hash_skiplink.is_none()
+            && self.seq_num.is_first())
+            || (self.entry_hash_backlink.is_some()
+                && self.entry_hash_skiplink.is_some()
+                && !self.seq_num.is_first())
```
cafca, line 161: Isn't this the wrong way around? Validation should throw an error for the sequence number = 1 case when backlink and skiplink **are** defined, not the other way?

adzialocha: Ha!

adzialocha: Well spotted! I guess some tests are missing 🤭

p2panda-rs/src/atomic/entry.rs
```diff
@@ -0,0 +1,220 @@
...
+            payload_size: message_size,
+            author: PublicKey::from_bytes(&key_pair.public_key_bytes())?,
+            seq_num: self.seq_num.as_u64(),
+            backlink,
+            lipmaa_link,
+            sig: None,
+        };
+
+        // @TODO: Sign BambooEntry and encode it
+        EntryEncoded::new("dummy")
```
cafca, line 113: I just looked into it and want to write down what I saw to talk about it tomorrow:

So here we want to sign the bamboo-core entry that we created just above. We should be able to use `key_pair` for that, which is a parameter for this function.

However, looking at the [dogs](https://docs.rs/ed25519-dalek/1.0.1/ed25519_dalek/) I see that a `ed25519_dalek::Keypair` instance is required to sign bytes. In our `KeyPair` implementation, the actual `KeyPair` is thrown away in `::new` and there is no way to recover it from the `public` and `private` halfs as far as I can tell.

Do we need to change our `KeyPair` implementation so that instead of serializing the private key it's serializing the whole keypair? Because a serialized keypair can be turned into `KeyPair` instance again, which we need in order to sign data.

cafca: I think `KeyPair` should be refactored to allow for a test like

```rust
#[test]
    fn signing() {
        let key_pair = KeyPair::new();
        let message = b"test";
        let signature = key_pair.sign(message);
        assert!(key_pair.verify(message, signature).is_ok())
    }
```

p2panda-rs/src/atomic/message.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    /// Basic signed `float` value.
+    Float(f64),
+
+    /// Basic signed `integer` value.
+    Integer(i64),
+
+    /// Basic `string` value.
+    Text(String),
+}
```
cafca, line 83: There's no reference type! :o

adzialocha: How does it look like?

cafca: `Relation(Hash)`? Does that make sense?

adzialocha: > Does that make sense?

Maybe its a tuple of `Relation(Hash, Schema)`?

cafca: The schema of the relation target is defined by the schema of the message this `MessageValue` would be part of. What do you have in mind when you say legacy schemas?

adzialocha: I see, then we only need `Hash` :+1:

cafca: Maybe instead of `Hash` we should also add an `InstanceID` type that wraps a hash but offers more functionality related to the instance it's referring to. I would prefer not to also do that in this PR though.

adzialocha: Ja that would make sense, better than wrapping wrapped wrappers might be to introduce a `Hash` Trait which gets implemented by `Schema`, `InstanceId` and `EntryHash` or similar?

cafca: Would that be *the* Hash trait or our own, which would also ensure a YamfHash etc?

adzialocha: Our own with `encode`, `decode` methods etc.

adzialocha: Better name needed probably.

adzialocha:
The wasm build fails now indeed because of the unexported `sign` methods. I guess we need to change the methods to return the `Signature` as raw bytes instead, then it will work to use it in a wasm context. For bamboo we need the raw bytes anyways since the Ed25519 Signature is different than the Bamboo Signature.

p2panda-rs/src/atomic/entry.rs
```diff
@@ -0,0 +1,249 @@
...
+            lipmaa_link,
+            sig: None,
+        };
+
+        let mut entry_bytes = [0u8; MAX_ENTRY_SIZE];
+        let unsigned_entry_size = entry.encode(&mut entry_bytes).unwrap();
+        let signature = key_pair.sign(&entry_bytes[..unsigned_entry_size]);
+        let sig_bytes = &signature.to_bytes()[..];
+        let signature = BambooSignature(sig_bytes.into());
+        entry.sig = Some(signature);
```
adzialocha, line 118: The wasm build fails now indeed because of the unexported sign methods. I guess we need to change the methods to return the Signature as raw bytes instead, then it will work to use it in a wasm context. For bamboo we need the raw bytes anyways since the Ed25519 Signature is different than the Bamboo Signature.

p2panda-js/test/index.test.ts
```diff
@@ -15,4 +15,14 @@ describe('KeyPair', () => {
     const keyPairSecond = KeyPair.fromPrivateKey(keyPair.privateKey());
     expect(keyPair.publicKey()).to.eq(keyPairSecond.publicKey());
   });
+
+  it('signs and validates', async () => {
+    const { KeyPair } = await p2panda;
+    const keyPair = new KeyPair();
+    const message = 'hello panda';
+    const sig = keyPair.sign(message);
+    expect(keyPair.verify(message, sig)).to.be.undefined;
```
cafca, line 24: Should we change the api to have `to.be.true` here? Testing for undefined feels very un-javascripty but I'm not sure how to have that and an acceptable rust `Result` as well.

p2panda-rs/src/keypair.rs
```diff
@@ -73,16 +76,25 @@ impl KeyPair {
         Box::from(self.0.secret.to_bytes())
     }
 
-    /// Sign data using this key pair.
-    #[cfg(not(target_arch = "wasm32"))]
-    pub fn sign(&self, data: &[u8]) -> Signature {
-        self.0.sign(data)
+    /// Sign a message using this key pair.
+    pub fn sign(&self, message: &[u8]) -> Box<[u8]> {
```
cafca, line 80: Do we want the `Box` here? I'm not sure I fully understand the consequences of that but it works and that made me happy :D

p2panda-rs/src/keypair.rs
```diff
@@ -73,16 +76,25 @@ impl KeyPair {
...
+            .verify(message, &Signature::try_from(signature).unwrap())
+    }
+
+    /// Verify a signature for a message.
+    #[cfg(target_arch = "wasm32")]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen(js_name = verify))]
+    pub fn verify_js(&self, message: &[u8], signature: &[u8]) -> Result<(), JsValue> {
+        self.0
+            .verify(message, &Signature::try_from(signature).unwrap())
+            .map_err(|_| JsValue::from_str("invalid signature"))
```
cafca, line 97: So I didn't find a way to make this work without having a separate function for the wasm target. `SignatureError` doesn't want to cross the wasm ABI. Can we implement some trait for `SignatureError` to make it compatible with wasm? Or use our own error type maybe? 

Also, failure to verify the signature from Javascript results in throwing a string instead of a proper `Error`. I didn't find a way to wrap the `JsValue` in `Err` or something like that and have it compile.

p2panda-rs/src/keypair.rs
adzialocha: I can recommend checking the wasm-bindgen documentation, there you have an overview of allowed return types (coming from Rust -> JS or vice-versa): https://rustwasm.github.io/docs/wasm-bindgen/reference/types/boxed-jsvalue-slice.html - so yeah, looks good :+1:

p2panda-rs/src/keypair.rs
adzialocha: Error handling in wasm for JavaScript seems to be a topic on itself, haven't read enough about it yet but I believe the wasm-bindgen documentation will have some words about it: https://rustwasm.github.io/docs/wasm-bindgen/reference/types/result.html?highlight=result#resultt-jsvalue

>  Can we implement some trait for SignatureError to make it compatible with wasm? Or use our own error type maybe?

For the library design in general I think having custom error types makes sense, it a) gives a clear API design of what to expect as a library consumer and b) gives us an understanding what we're dealing with developing it. So I'm in for using our own!

p2panda-js/test/index.test.ts
adzialocha: Agreed, this would be better. You can simply use `is_ok()` and it will return a boolean.

p2panda-rs/src/keypair.rs
adzialocha: But also, I think having two methods for two compiler targets is also totally fine and common. You could even name them the same :smile_cat: 

Btw, you can also write it like this as the first `cfg` attribute already sets the allowed target for the whole container:

```
/// Verify a signature for a message.
#[cfg(target_arch = "wasm32")]
#[wasm_bindgen(js_name = verify]
...
```

cafca: Well, with two functions there's the possibility that they actually do different things. At least one of them should wrap the other then instead of replicating statements.

Regarding the error handling for javascript: throwing a string means that no traceback information is available and different errors can only be distinguished by pattern matching the error message. Not a huge problem per so but not elegant either. I would like to do [this](https://github.com/rustwasm/wasm-bindgen/issues/1742) but it doesn't seem to work for me.

p2panda-rs/src/keypair.rs
cafca: Okay I was really vague, I'm sorry :D I had looked at those docs, what I meant was more like: Do we introduce a significant performance penalty by `Box`ing the return value of a function that will be called *a lot*? We are only doing it here for the Javascript side of things so if there is a disadvantage we could have separate implementations for js and rs world with only the js one returning a Box value. Now that I'm writing it down I realise that we are not at a point in time where this matters :D

p2panda-js/test/index.test.ts
cafca: You mean calling `is_ok()` on the return value? I don't know if that's possible?. I found another solution though!

adzialocha: Your solution is better! It still keeps the actual error. 

Using `is_ok()` would work, you would *ignore* the error and convert it just into a boolean. For example like this:

```rust
fn do_something(&self) -> bool {
   self.im_doing_something_which_returns_a_result().is_ok()
}
```

p2panda-rs/src/atomic/message.rs
cafca: We can continue discussing that [in this issue](https://github.com/p2panda/p2panda/issues/32) eventually.

p2panda-rs/src/atomic/entry_signed.rs
```diff
@@ -57,11 +58,12 @@ impl EntrySigned {
...
-
-    fn try_from(signed: &EntrySigned) -> Result<Self> {
+impl<'a> From<&'a EntrySigned> for BambooEntry<&'a [u8], &'a [u8]> {
+    fn from(_signed_entry: &'a EntrySigned) -> Self {
         todo!();
+        // let test = signed_entry.clone();
+        // let bytes = test.to_bytes().as_slice();
+        // TryInto::<BambooEntry<&'a[u8], &'a[u8]>>::try_into(bytes).unwrap()
     }
 }
```
adzialocha: I looked a little bit into this (with a fresher brain :dog: @cafca and @sophiiistika) and think we have to take another approach here: `bytes` is owned inside the scope of the `from` function and can never be moved out through `BambooEntry` as it is only passed on as a reference.

We can't change that as the bamboo-rs crate only allows references: https://github.com/pietgeursen/bamboo-rs/blob/master/bamboo-rs-core/src/entry/mod.rs#L75

As an alternative we can look into this trait though, assuming that `ArrayVec` allows ownership over the bytes: https://github.com/pietgeursen/bamboo-rs/blob/master/bamboo-rs-core/src/entry/mod.rs#L83

Also `into_owned` might be interesting: https://github.com/pietgeursen/bamboo-rs/blob/master/bamboo-rs-core/src/entry/mod.rs#L101.

adzialocha:
I thought about implementing the `create structs for intermediate type containing: (EntrySigned, MessageEncoded) tuple` todo but then I realized I lack a clear understanding of what we are after: Why do we want this and how would it look like / be named (favorite question)? :smile_cat:

adzialocha:
Oh wow, for some unknown reasons the tests fail now (without any code changes):

```
failures:

---- atomic::message::tests::encode_and_decode stdout ----
thread 'atomic::message::tests::encode_and_decode' panicked at 'assertion failed: `(left == right)`
  left: `Message { action: Update, schema: Hash("00402dc25d32dfb400bb295b663d4706bc47f0cb4f1edff277c737afc8a9232330ae9884fcf6d02141a785c5fd82c196b973e8427efc0c04d0444dcc3059220b9eda"), version: Default, id: Some(Hash("00405d49304e29d1d9358314e1083d05d5615a76f64c0984f51e3659a53a3e5d67a2b8e69b91e335978627e4ccafcc549211a283c6154d3af470698cfc256fbf8700")), fields: Some(MessageFields({"username": Text("bubu"), "is_admin": Boolean(false), "age": Integer(28)})) }`,
error: test failed, to rerun pass '--lib'
 right: `Message { action: Update, schema: Hash("00402dc25d32dfb400bb295b663d4706bc47f0cb4f1edff277c737afc8a9232330ae9884fcf6d02141a785c5fd82c196b973e8427efc0c04d0444dcc3059220b9eda"), version: Default, id: Some(Hash("00405d49304e29d1d9358314e1083d05d5615a76f64c0984f51e3659a53a3e5d67a2b8e69b91e335978627e4ccafcc549211a283c6154d3af470698cfc256fbf8700")), fields: Some(MessageFields({"username": Text("bubu"), "age": Float(28.0), "is_admin": Boolean(false)})) }`', src/atomic/message.rs:377:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- atomic::message_encoded::tests::decode stdout ----
thread 'atomic::message_encoded::tests::decode' panicked at 'assertion failed: `(left == right)`
  left: `Float(28.0)`,
 right: `Integer(28)`', src/atomic/message_encoded.rs:135:9


failures:
    atomic::message::tests::encode_and_decode
    atomic::message_encoded::tests::decode

test result: FAILED. 17 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
```

It looks like the deserializer interprets an `Integer` number as a `Float` .. hmn. Maybe we have to look into a better message schema sooner as we thought :smile_cat:

cafca:
oof

sandreae:
Howdy, I've been having a look into this error. I may be way off, but I think `serde_cbor` looks for target types for basic values and deserializes accordingly. So because the float value is higher up in the order `MessageValue` enum here:

https://github.com/p2panda/p2panda/blob/8d1383866bd358100b9459ae5c10ade61889c54c/p2panda-rs/src/atomic/message.rs#L69-L84

It thinks `age` is wanted as a float and deserializes it from an integer into a float.

If you swap the `Float` and `Integer` definitions around in `MessageValue` then the tests pass... :eyes: ! 

```
/// Enum of possible data types which can be added to the messages fields as values.
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
#[serde(untagged)]
pub enum MessageValue {
    /// Basic `boolean` value.
    Boolean(bool),

    /// List integer before float and the tests pass ///
    /// Basic signed `integer` value.
    Integer(i64),

    /// Basic signed `float` value.
    Float(f64),

    /// Basic `string` value.
    Text(String),
}
```

adzialocha:
> If you swap the `Float` and `Integer` definitions around in `MessageValue` then the tests pass... :eyes: ! 

Ah, super! I wonder now if it interprets Floats correctly, but I assume it's just going from the top to the bottom trying it out until serde sees fit?

In that case I'm happy with this fix for now to merge it and look for a solution in another branch, something with new message schemas ala:

```
action: "..."
schema: "..."
fields: {
   age: {
      value: 28
      type: "integer"
   }
}
```

... Or similar?

sandreae:
> But I assume it's just going from the top to the bottom in the Enum trying it out until serde sees fit?

Yeh, that's what I assumed too. It was hard to find the exact documentation about this. Mainly figured it out from people having issues where it *wasn't* predicting and converting types like this.

Yep, hacky solution for now seems appropriate 👍

sandreae (commented):
I'm not familiar enough with the repo yet to really approve this PR.... I can say the tests pass though :+1:

adzialocha:
> I'm not really familiar enough with the repo yet to approve a PR....... I can confirm the tests pass :-p

I'm sure there is always something everyone can find, like a typo or a better formulation! :-)

sandreae:
> > I'm not really familiar enough with the repo yet to approve a PR....... I can confirm the tests pass :-p
> 
> I'm sure there is always something everyone can find, like a typo or a better formulation! :-)

True, true!

p2panda-js/test/index.test.ts
```diff
@@ -15,4 +15,14 @@ describe('KeyPair', () => {
...
   });
+
+  it('signs and validates', async () => {
+    const { KeyPair } = await p2panda;
+    const keyPair = new KeyPair();
+    const message = 'hello panda';
+    const sig = keyPair.sign(message);
+    expect(keyPair.verify(message, sig)).to.be.true;
+    expect(() => keyPair.verify(message, sig)).to.not.throw();
+    expect(() => keyPair.verify('hello aquadoggo', sig)).to.throw();
```
cafca, line 26: We were talking about this earlier. Would it be more javascriptonic to make this

```
    expect(keyPair.verify('wrong message', sig)).to.be.false;
```

instead of throwing?

adzialocha: Yeah, I agree, it should only throw when something is wrong which would make validation impossible :+1:

cafca: ok, but this is not about the atomic data types, i'll pull it out

cafca: wait.. actually we implemented the key pair thing also in this pr. so i'll add it here

sophiiistika (approved):

cafca (approved):

adzialocha:
:bouquet:

sandreae (approved):

PR #33, opened 2021-02-14
Add relation type

cafca:
Adds `Relation` as a possible `MessageValue`, which contains an instance id (hash) of the related instance.


adzialocha (approved):
Woho!

Issue #31, opened 2021-01-29
Update `wasm-bindgen` crate


adzialocha:
Looks like the issue was fixed in https://github.com/rustwasm/wasm-bindgen/pull/2391 and we can now move on upgrading to the latest `wasm-bindgen` version.

Pinging @cafca here as you locked the `wasm-bindgen` version.

PR #41, opened 2021-03-23
Improve error return types

adzialocha:
**Improve error return types**

* Remove anyhow crate
* Explicitly return custom error types, not `anyhow::Error`
* Use `std::Result` instead of `anyhow::Result`
* Use thiserror crate to convert errors from external crates or other atomic error types from p2panda-rs

Closes: https://github.com/p2panda/p2panda/issues/40

**Add more error handling**

* Handle errors in `KeyPair`

**Bonus level**

* Update `wasm-bindgen` to latest version :space_invader:  (was required to use latest `js-sys` crate)

Closes: https://github.com/p2panda/p2panda/issues/31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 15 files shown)


sandreae (approved):
Looks good!

adzialocha:
> Looks good!

Thanks @sandreae! I've implemented error handling for `KeyPair` as I figured this might make sense in this PR as well, it would be great if you can have a look at it again :pray:

PR #47, opened 2021-04-05
Fix skiplink encoding

adzialocha:
Skiplinks can be omitted in the Bamboo encoding when they are the same as the backlink.


cafca (changes requested):
This should also be handled when requesting next entry args

adzialocha:
> This should also be handled when requesting next entry args

This got changed here: https://github.com/p2panda/aquadoggo/pull/21/commits/82ec424ad0412d02d05836efaa4edc6a501177ea

Issue #39, opened 2021-03-15
Implement wasm bindings for atomic structs and methods


adzialocha:

sandreae:
I'm having a go at this locally to get my head round wasm. Got a question:

Trying to implement wasm for the `Author` struct like so:

```
#[cfg_attr(target_arch = "wasm32", wasm_bindgen)]
impl Author {
    /// Validates and wraps author string into a new `Self` instance.
    #[cfg_attr(target_arch = "wasm32", wasm_bindgen(constructor))]
    pub fn new(value: &str) -> Result<Author> {
        let author = Self(String::from(value));
        author.validate()?;
        Ok(author)
    }
}
```

Which throws an error (E0401) because it doesn't like `Self`:

> use of generic parameter from outer function use a type here instead

So i tried using `Author` instead of Self which throws a different error (E0277):

> the trait `IntoWasmAbi` is not implemented for `std::result::Result<Author, anyhow::Error>`

So there is an issue with what's being returned in the `Result` object, right? Which should be something wasm likes. Am I at least on the correct error path....?

adzialocha:
@sandreae yeah, this looks good! 

Stupid with that Result type, either there is a feature flag in `anyhow` we could enable to get wasm support (I doubt it exists after checking for quickly), OR we introduce a method only for wasm contexts (see `verify` method in `key_pair.rs` as an example) but then we will need to do this a lot .. OR we have to use a different `Result` type which hurts but maybe there is an elegant solution I can't think of right now.

sandreae:
Hmmm, OK... So the problem is with the `anyhow` error rather than returning a `Result`. I'll do some digging to see how others deal with this situation, and check out the `key_pairs` method. Thanks!

sandreae:
For reference: https://rustwasm.github.io/docs/wasm-bindgen/reference/types/result.html?highlight=Result#resultt-jsvalue

sandreae:
From this issue it sounds like `anyhow` should work with `wasm`...?

https://github.com/dtolnay/anyhow/issues/115

adzialocha:
> From this issue it sounds like `anyhow` should work with `wasm`...?
> 
> https://github.com/dtolnay/anyhow/issues/115

Yeah, I was also wondering, or it can be understood as "it's not supported therefore it's not a problem"?

sandreae:
[...]

Issue #48, opened 2021-04-06
Implement non-experimental wasm bindings for signEncode and decodeEntry


sandreae:
These bindings are needed for `beep-boop` we have an [experimental version](https://github.com/p2panda/p2panda/blob/e0df7c07090d44b24c2d1c845e91f8918000781f/p2panda-rs/src/lib.rs#L83-L158) here but need to implement properly.

Related to [this](https://github.com/p2panda/p2panda/issues/39).

PR #49, opened 2021-04-06
WebAssembly module

adzialocha:
- Message encoding now has it's own method `encode_create_message()` which returns `message_encoded` as a string
- `sign_entry()` (previously `sign_encode()`) now takes the encoded message and extra argument log_id  returns `entry_encoded` and `entry_hash` 
- `decode_entry()` now returns an `entry` as lovely json object
- Nicer error handling :+1: 

Closes: #39, #48 and #51


sandreae:
This looks great! Just getting my head round what's changed..

- message encoding now has it's own method `encode_create_message()` which returns `message_encoded` as a string
- `sign_entry()` (previously `sign_encode()`) now takes the encoded message and extra argument log_id  returns `entry_encoded` and `entry_hash` 
- `decode_entry()` now returns an `entry` as lovely json object
- nicer error handling :+1: 

What was the reason for splitting the message encoding? I like that it's more explicit about what is going on, and separation of tasks. Just wondered if there were other reasons too.

adzialocha:
Thanks for summarizing, I've added it now in the description of this PR :+1: 

> What was the reason for splitting the message encoding? I like that it's more explicit about what is going on, and separation of tasks. Just wondered if there were other reasons too.

There will be different sorts of messages (Create, Update, Delete), so I'm thinking about introducing also an `encode_update_message` and `encode_delete_message` which results you can then pass over to `sign_entry`.

sandreae:
> There will be different sorts of messages (Create, Update, Delete), so I'm thinking about introducing also an encode_update_message and encode_delete_message which results you can then pass over to sign_entry.

Ok, thought there might be something deeper ;-p that makes sense!

adzialocha:
> Ok, thought there might be something deeper ;-p that makes sense!

Haha no! Also, please feel free to change the API, this is really just a draft yet.

sandreae:
Maybe we can set aside some time to chat together about it at one of the next meetups?

p2panda-rs/src/wasm.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub fn new() -> Self {
+        Self(PandaMessageFields::new())
+    }
+
+    pub fn add(&mut self, name: String, value: JsValue) -> Result<(), JsValue> {
+        // @TODO: Add more types
+        let field = match value.as_string() {
+            Some(text) => Ok(MessageValue::Text(text)),
+            None => Err(js_sys::Error::new(&format!("Invalid value type"))),
+        }?;
```
cafca, line 44: Frustrating not to have  schema definitions for this.

p2panda-rs/src/wasm.rs
```diff
@@ -0,0 +1,139 @@
...
+
+        jserr!(self.0.add(&name, field));
+
+        Ok(())
+    }
+}
+
+#[wasm_bindgen(js_name = encodeCreateMessage)]
+pub fn encode_create_message(schema: String, fields: MessageFields) -> Result<String, JsValue> {
```
cafca, line 53: I like the functional style here.

p2panda-rs/src/wasm.rs
```diff
@@ -0,0 +1,139 @@
...
+        seq_num.as_ref(),
+    ));
+
+    // Finally sign and encode entry
+    let entry_signed = jserr!(EntrySigned::try_from((&entry, &key_pair)));
+
+    // Serialize result to JSON
+    let result = jserr!(wasm_bindgen::JsValue::from_serde(&SignEntryResult {
+        entry_encoded: entry_signed.as_str().into(),
+        entry_hash: entry_signed.hash().as_hex().into(),
```
cafca, line 113: Should we use either `as_str` or `as_hex` throughout the library for consistency?

p2panda-rs/src/wasm.rs
```diff
@@ -0,0 +1,139 @@
...
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SignEntryResult {
+    pub entry_encoded: String,
+    pub entry_hash: String,
+}
+
+#[wasm_bindgen(js_name = signEntry)]
+pub fn sign_entry(
+    key_pair: KeyPair,
```
sandreae, line 69: I think this is causing some issues on the JS end of things. After passing a KeyPair instance into signEntry() it no longer exists on the JS end. I guess this is a rust ownership thing. I tried:

`key_pair: &KeyPair,`

And that works, but I don't know if it is strange Rust....

Right now I am just passing in a copy of the keyPair which also works.

cafca:
LOL

Sam Andreae ***@***.***> schrieb am Mi. 7. Apr. 2021 um 17:46:

p2panda-rs/src/wasm.rs
adzialocha: Yes, but I guess thats part of another issue?

p2panda-rs/src/wasm.rs
adzialocha: > And that works, but I don't know if it is strange Rust....

Looks like a good solution! Passing over instances via reference is even better here. Do you want to make a commit?

p2panda-rs/src/wasm.rs
cafca: That's what I was thinking

cafca: https://github.com/p2panda/p2panda/issues/50

cafca (approved):

PR #59, opened 2021-04-20
Doc code examples

sandreae:
This PR is for updating p2panda-rs rust docs to include lots of code examples. I suggest we use this example as a template:

https://github.com/p2panda/p2panda/blob/6e692522de42b849ec507e8f0c84b119195aacd0/p2panda-rs/src/key_pair.rs#L42-L56

SOME NOTES:

- when running `cargo doc` or `cargo build` we now get warnings for code documentation which doesn't have an accompanying code example. We probably won't make examples for all of these, but it's fun to see them being crossed off ;-p
- the extra wrapping around each test (hidden with a `#`) is needed so the examples can actually be run and tested during compilation.
- I quite like showing the `use` import at the top, and then either printing some result or using `assert!` to demonstrate what we are expecting. I looked through the [rand](https://docs.rs/rand/0.8.3/rand/seq/trait.SliceRandom.html) crate for inspiration. Any other thoughts on how we style these are welcome though.
- we shouldn't use `unwrap()` in code examples
- great documentation for rust docs [here](https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html)

Relevant commands:

```
# build and open the generated docs in browser
cargo doc --open

# run tests (includes testing example code)
cargo test
```

Let's try and do as much as we can this week then merge beginning of next week, just so we don't get out of date with main. ;-p :+1:

### DIFF
--- a/p2panda-rs/src/atomic/author.rs
+++ b/p2panda-rs/src/atomic/author.rs
@@ -30,6 +30,25 @@ pub struct Author(String);
 
 impl Author {
     /// Validates and wraps author string into a new `Author` instance.
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// # extern crate p2panda_rs;
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// use p2panda_rs::key_pair::KeyPair;
+    /// use p2panda_rs::atomic::Author;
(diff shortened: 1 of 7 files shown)


cafca:
Todo: Update README.md

Issue #57, opened 2021-04-19
Rename `atomic` module


adzialocha:
Is there a better name for this?

PR #62, opened 2021-04-26
Create own module for encoding, decoding and signing entries

### LINKED ISSUE #60: Consider `Entry` and `EntrySigned` conversion traits
Is this the best structure for the entry signing logic?

problems with current situation:
- conceptually confusing
- the conversion process is lossy and requires a tuplet, which go against normal rust conversion trait convention (maybe..?)

alternative ideas:
- a sign() method which lives in the `Entry` struct and accepts a `key_pair` and returns an `ÈntrySigned`
- a sign() function which lives on it's own in `sign.rs`
- something else...

adzialocha:
Refactored TryFrom traits for Entry <-> EntrySigned conversions into it's own module called `encoder` (The name isn't good... any ideas?). 

Currently these methods exist:

- encode_entry (private)
- sign_entry (private)
- sign_and_encode
- decode_entry
- validate_message

It could be expanded in the future.

**Breaking Changes**

* `signEncodeEntry` in js/wasm module expects sequence number to be of type `BigInt`
* `Entry` constructor awaits sequence number for this entry not from the previous one anymore (`panda_nextEntryArguments` response needs to be changed)
* Message field in `Entry` constructor is optional now
* Conversion traits between `Entry` and `EntrySigned` have been replaced with `encoder::sign_and_encode` method

Closes #60, #63


sandreae:
Just trying to make the changes to remove passing the entry_bytes size around but coming up against an error I don't know how to solve (because of my basic Rust knowledge :rofl:)

Refactoring like this:

```

/// Takes unsigned entry bytes and their size and a [`KeyPair`], returns a tuple containing signed and encoded entry bytes and their size.
pub fn sign_entry(entry_bytes: [u8], key_pair: &KeyPair) -> Result<[u8], EntrySignedError>{
    // Make copy of entry_bytes before passing to decode
    let mut entry_bytes_copy = entry_bytes.clone();
    
    // Decode unsigned entry bytes
    let mut entry = decode(&entry_bytes)?;
    
    // Sign and add signature to entry
    let sig_bytes = key_pair.sign(&entry_bytes_copy);
    let signature = BambooSignature(&*sig_bytes);
    entry.sig = Some(signature);

    // Get signed entry bytes
    let signed_entry_size = entry.encode(&mut entry_bytes_copy)?;

    // Return signed entry bytes
    Ok(entry_bytes_copy[..signed_entry_size])
}
```
Gives me this error:

```
error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
   --> src/encoder.rs:81:61
    |
81  | pub fn sign_entry(entry_bytes: [u8], key_pair: &KeyPair) -> Result<[u8], EntrySignedError>{
    |                                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    | 
   ::: /home/sam/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:241:17
    |
241 | pub enum Result<T, E> {
    |                 - required by this bound in `std::result::Result`
    |
    = help: the trait `Sized` is not implemented for `[u8]`
```
I understand the issue, the u8 size is unknown, but I'm not sure how to fix it...

Any clues.....?

adzialocha:
> Any clues.....?

You need a more flexible data type to pass around arrays like this. I would go for something like `Vec<u8>`, you might need to convert it around though (for example via `.to_vec()`) but I think thats alright :+1:

sandreae:
Thanks! :+1:

sandreae:
Ok, the Vec conversion was only needed because I'd split `sign_and_encode` into two different functions, as I don't think that is actually useful, I've now put them back together into one. Which means we don't need to convert back and forth.

p2panda-rs/src/atomic/entry_signed.rs
```diff
@@ -1,14 +1,12 @@
 use std::convert::{TryFrom, TryInto};
 
 use arrayvec::ArrayVec;
-use bamboo_rs_core::entry::MAX_ENTRY_SIZE;
-use bamboo_rs_core::{Entry as BambooEntry, Signature as BambooSignature};
-use ed25519_dalek::{PublicKey, Signature};
+use bamboo_rs_core::Entry;
```
adzialocha, line 4: I would keep the `BambooEntry` renaming just for readability / not confusing it with out own `atomic::Entry` struct.

p2panda-rs/src/encoder.rs
```diff
@@ -0,0 +1,282 @@
...
+
+use bamboo_rs_core::entry::{decode, MAX_ENTRY_SIZE};
+use bamboo_rs_core::{Entry as BambooEntry, Signature as BambooSignature, YamfHash};
+
+use crate::atomic::error::EntrySignedError;
+use crate::atomic::Blake2BArrayVec;
+use crate::atomic::{Entry, EntrySigned, Hash, LogId, Message, MessageEncoded, SeqNum};
+use crate::key_pair::KeyPair;
+use arrayvec::ArrayVec;
+use ed25519_dalek::PublicKey;
```
adzialocha, line 11: Small style nitpick: These imports are external and would belong to the group above the internal imports

p2panda-rs/src/test_helpers.rs
```diff
@@ -0,0 +1,51 @@
+use crate::atomic::{
+    Entry, EntrySigned, Hash, LogId, Message, MessageFields, MessageValue, SeqNum,
+};
+
+/// Returns a Message for testing
+pub fn mock_message(text: String) -> Message {
```
adzialocha, line 6: Maybe rename these `mock_*` methods to `create_*`? Mocking is more about "simulating" an API which actually does not do the real thing in the backend, but here we actually create "real" instances of something.

adzialocha (commented):
This is really cool! :star2:

I've just added some minor comments @sandreae and ran `cargo fmt` against the branch and committed the style fixes.

The `encoder` module name is maybe not ideal but I think this should be handled in a separate PR anyways, see: https://github.com/p2panda/p2panda/issues/57

sandreae:
Thanks for the comments, I'll get these done for next weeks meetup :+1:

p2panda-js/test/index.test.ts
```diff
@@ -31,6 +31,7 @@ describe('Entries', () => {
     const TEST_SCHEMA =
       '0040cf94f6d605657e90c543b0c919070cdaaf7209c5e1ea58acb8f3568fa2114268dc9ac3bafe12af277d286fce7dc59b7c0c348973c4e9dacbe79485e56ac2a702';
     const LOG_ID = 5;
+    const SEQ_NO = 1;
```
cafca, line 34: Maybe make this `SEQ_NUM` to be consistent with the codebase

p2panda-rs/src/encoder.rs
```diff
@@ -0,0 +1,286 @@
...
+            let yamf_hash: YamfHash<Blake2BArrayVec> = (&msg.hash()).to_owned().into();
+
+            if yamf_hash != entry.payload_hash {
+                return Err(EntrySignedError::MessageHashMismatch);
+            }
+            Message::from(msg)
+        }
+    };
+    Ok(message)
+}
```
cafca, line 35: We could call this `safe_decode_message`. Decoding seems to be the primary function with the addition that the hash is validated.

Is this meant to be public? It's only used in this module.

The match seems to duplicate what's happening in `decode_entry`. Would it be better to let this function reject an empty message so that it can only be used when you already know that a message exists?

p2panda-rs/src/encoder.rs
```diff
@@ -0,0 +1,286 @@
...
+
+/// Takes [`EntrySigned`] and optionally [`MessageEncoded`] as arguments, returns a decoded and
+/// unsigned [`Entry`]. When a [`MessageEncoded`] is passed it will automatically check its
+/// integrity with this [`Entry`] by comparing their hashes. Valid messages will be included in the
+/// returned [`Entry`], if an invalid message is passed an error will be returned.
+///
+/// Entries are separated from the messages they refer to. Since messages can independently be
+/// deleted they can be passed on as an optional argument. When a [`Message`] is passed it will
+/// automatically check its integrity with this Entry by comparing their hashes.
```
cafca, line 100: Well written!

p2panda-rs/src/encoder.rs
```diff
@@ -0,0 +1,286 @@
...
+        let log_id: i64 = 1;
+        let payload_bytes = "a466616374696f6e6663726561746566736368656d6178843030343063663934663664363035363537653930633534336230633931393037306364616166373230396335653165613538616362386633353638666132313134323638646339616333626166653132616632373764323836666365376463353962376330633334383937336334653964616362653739343835653536616332613730326776657273696f6e01666669656c6473a26464617465a164546578747818323032312d30352d30325432303a30363a34352e3433305a676d657373616765a164546578746d477574656e204d6f7267656e21";
+        let payload_hash = "00408960c9d4f864aef757d440bc5aa5a5c0d726312eddadad68f25d06fedd10f755d51a87565972f8c3d77ef7ac66531227131b0d8857fef749c3a98cfffae8519d";
+        let seq_num: i64 = 4;
+        let skiplink_hash = "0040944b4ae2ff31d0adc13cf94ba43b766871b4e56e96d0eebbc1b9e2b8226d448e8bc1f9507a21894579578491ff778a008688c2a3e8a409fc37522d9eabaa114c";
+        let backlink_hash = "004054f65f3ac2ccf13f5862eb7c29ac20e830e173d062416dfd03a27e8a2315b69f402cfa4ca741d243b184b1d8ff203cf1f1ec4619f44758263f19a75a3537e780";
```
cafca, line 226: We had previously used a round-trip testing model (it's still there in `entry.rs`). That is, we create an encoded data structure, decode it, re-encode it and then compare the two encoded forms to ensure that de- and encoding doesn't change the data.

This bit here introduces another approach that doesn't prevent inconsistencies within a conversion cycle but across versions, i.e. if the library is changed and these values stay here in the test, the test will only fail if the new library doesn't handle these old values properly.

I like that! Maybe we should actually make this more explicit by turning these constants into test fixtures annotated with the specific library version they were created with. Then eventually we could have tests that iterate over data produced by all previously published api versions to make sure all of them can be handled.

Of course that doesn't need to happen in this PR, just a thought.

p2panda-rs/src/test_helpers.rs
cafca: Here is a way to use fixtures for this

https://docs.rs/rstest/0.6.4/rstest/

p2panda-rs/src/wasm.rs
```diff
@@ -191,7 +189,8 @@ pub fn decode_entry(
 
     // Convert encoded entry
     let entry_signed = jserr!(EntrySigned::new(&entry_encoded));
-    let entry = jserr!(Entry::try_from((&entry_signed, message_encoded.as_ref())));
+    // Had to remove the jserr macro here to make this work, need to look at it again
```
cafca, line 192: flagging this

cafca (commented):
I like it!

p2panda-rs/src/encoder.rs
sandreae: Ah yeh, this is a really good idea, let's store it for later.

Should we bring the round trip test from entry into this module so they can be side by side and the distinction made clearer?

p2panda-rs/src/test_helpers.rs
sandreae: nice, I'll look into this :+1:

p2panda-rs/src/encoder.rs
sandreae: Yeh, you're right, there is some duplication going on in `decode_entry`. Right now that's only where it's used, but there was talk of it being useful in `aquadoggo`, mainly for validating messages though I think. Maybe the validation method should be moved (back??) into `EntrySigned` as a signed entry is always needed for the validation anyway. Then I don't think we even need this method...

p2panda-rs/src/encoder.rs
sandreae: :+1:

p2panda-rs/src/encoder.rs
sandreae: I actually had a go at implementing fixture tests with `rstest`, sure some of it needs some work still, but as a starting point I don't think it's too bad.

https://github.com/p2panda/p2panda/blob/010416d386f0bb7d94823264c14f4abd437768a8/p2panda-rs/src/encoder.rs#L167-L183

https://github.com/p2panda/p2panda/blob/010416d386f0bb7d94823264c14f4abd437768a8/p2panda-rs/src/encoder.rs#L200-L213

sandreae: and this:

https://github.com/p2panda/p2panda/blob/010416d386f0bb7d94823264c14f4abd437768a8/p2panda-rs/src/encoder.rs#L185-L199

https://github.com/p2panda/p2panda/blob/7435545d214e99c60b928c4bd6393be10ea2fe1d/p2panda-rs/src/encoder.rs#L233-L287

p2panda-rs/src/encoder.rs
```diff
@@ -0,0 +1,287 @@
...
+    let entry: BambooEntry<ArrayVec<[u8; 64]>, ArrayVec<[u8; 64]>> = entry_encoded.into();
+
+    // Message hash must match if it doesn't return an error
+    let yamf_hash: YamfHash<Blake2BArrayVec> = (&message_encoded.hash()).to_owned().into();
+    if yamf_hash != entry.payload_hash {
+        return Err(EntrySignedError::MessageHashMismatch);
+    }
+
+    Ok(Message::from(message_encoded))
+}
```
adzialocha, line 30: We can remove the decoding part and name this method `validate_message_entry_match`

adzialocha: Even better: Move this into `EntrySigned` and rename method to `validate_message`

p2panda-rs/src/atomic/entry.rs
```diff
@@ -235,46 +170,100 @@ mod tests {
...
             None,
             None,
-            Some(&SeqNum::new(1).unwrap())
+            &SeqNum::new(2).unwrap()
         )
         .is_err());
     }
+
+    #[test]
+    fn sign_and_encode_test() {
```
adzialocha, line 224: Move this into `encoder` file as the tested method lives there

p2panda-rs/src/encoder.rs
```diff
@@ -0,0 +1,287 @@
...
+        assert_eq!(
+            message_fields.get("message").unwrap(),
+            fixture_message_fields.get("message").unwrap()
+        );
+
+        assert_eq!(
+            message_fields.get("date").unwrap(),
+            fixture_message_fields.get("date").unwrap()
+        );
+    }
```
adzialocha, line 264: Newline :-D

p2panda-rs/src/encoder.rs
```diff
@@ -0,0 +1,287 @@
...
+            key_pair: KeyPair::from_private_key(String::from("31f33f8e6c262f36a0e5397348093a459d66d8cb5946798ad62d5eb8e7645bdb")).unwrap(),
+            entry: Entry::new(
+                &LogId::new(1),
+                Some(&Message::new_create(Hash::new("0040cf94f6d605657e90c543b0c919070cdaaf7209c5e1ea58acb8f3568fa2114268dc9ac3bafe12af277d286fce7dc59b7c0c348973c4e9dacbe79485e56ac2a702").unwrap(), create_message_fields(vec!["message", "date"], vec!["Guten Morgen!", "2021-05-02T20:06:45.430Z"])).unwrap()),
+                Some(&Hash::new("0040944b4ae2ff31d0adc13cf94ba43b766871b4e56e96d0eebbc1b9e2b8226d448e8bc1f9507a21894579578491ff778a008688c2a3e8a409fc37522d9eabaa114c").unwrap()),
+                Some(&Hash::new("004054f65f3ac2ccf13f5862eb7c29ac20e830e173d062416dfd03a27e8a2315b69f402cfa4ca741d243b184b1d8ff203cf1f1ec4619f44758263f19a75a3537e780").unwrap()),
+                &SeqNum::new(4).unwrap(),
+                ).unwrap(),
```
adzialocha, line 197: Formatting :-)

sandreae:
> Todo: Update README.md

What do we need to update in README wrt this PR?

adzialocha:
> > Todo: Update README.md
> 
> What do we need to update in README wrt this PR?

Also can't remember this, @cafca was this maybe meant for another PR? :sweat_smile:

Issue #55, opened 2021-04-19
Specify API for `p2panda-js`


adzialocha:
Current draft implementation here: https://github.com/p2panda/beep-boop/blob/main/src/p2panda-api.ts
Draft specification: https://hackmd.io/kRb3ma1AQhyhMCQKBnBTKQ

This draft is not ideal yet, let's meet again and think about something better! :star2:

Issue #29, opened 2021-01-18
Add method to create create, update, delete messages


cafca:
See https://gist.github.com/adzialocha/97ed6c5b4b28cb94b79bd59c3f362e98

cafca:
Bring it back

Issue #28, opened 2021-01-18
Add publishEntry method to rpc client


cafca:
Refer to https://github.com/p2panda/node/issues/12

Issue #27, opened 2021-01-18
Add nextEntryArguments method to rpc client


cafca:
Refer to https://github.com/p2panda/node/issues/3

PR #67, opened 2021-05-15
Add TypeScript API that wraps wasm bindings

cafca:
This interface can be used to connect client applications to p2panda. 

## Breaking changes

After this, the wasm-bindings will not be the default export of `p2panda-js` anymore. They are available as a named export `wasm` or the new `Session` interface can be used.

## Usage

```js
import { Session } from 'p2panda-js';

// create a key pair or restore it from storage
const keypair = getKeyPair();

// create a session with a p2panda node such as auadoggo
const session = new Session("localhost:2020").keyPair(keypair);

// query entries of a schema
const obvervationsSchema = 'CB658DE...';
session.schema(obvervationsSchema);
const entries = await session.queryEntries()

// instances cannot yet be queried

// create instances of the same schema
const fields = {
  observedBird: 'yes',
};
await session.create(fields)

// optional argument for one-off configuration
await session.create(fields, { node: 'https://welle.liebechaos.org', schema: otherSchema });
```

## Todo

- [x] Tests
  - [x] Mocking RPC calls
- [x] Docs
~~- [ ] TypeScript declaration file~~

Closes: #55, #27, #26 and #28 aaaaand #29


cafca:
For some reason `@open-rpc/client-js` can't be handled in the build process. Apparently because a relative import path is not recognized?

adzialocha:
This is a very exciting PR! 😎🌟

Is there a reason why schemas are attached to sessions? I can imagine that only having one schema is a relatively rare case for a client.

sandreae:
This is really cool!! Thanks for doing it :-)

Can I ask about the separation between a `Session` and an `Instance`? I remember talking about this at some point, but I can't see exactly where the separation of responsibilities lies.

p2panda-js/src/index.ts
```diff
@@ -1,29 +1,12 @@
...
 
-// p2panda is exported without web assembly utilities
-type P2Panda = Omit<
-  Resolved<typeof initializeWasm>,
-  'setWasmPanicHook' | 'init' | 'default'
->;
+if (!globalThis.fetch) {
+  globalThis.fetch = fetch;
+  globalThis.Headers = Headers;
+}
```
cafca, line 9: need to add a comment here

cafca:
@adzialocha about the interface you mentioned - being able to attach schema, key pair etc. to the session. Do you think we should change it or leave it as it is now? I saw you also opened an issue about discussing the api. I am open to anything, otherwise I'll just continue with this :)

adzialocha:
> @adzialocha about the interface you mentioned - being able to attach schema, key pair etc. to the session. Do you think we should change it or leave it as it is now? I saw you also opened an issue about discussing the api. I am open to anything, otherwise I'll just continue with this :)

Since our meeting last many things got clearer to me and I think it's super like that! The most important is that at all times the globally set schema and key pair can be changed AND / OR set in a one-off configuration per call. Then we should be able to flexibility handle all use cases 🤞

p2panda-js/openrpc.json
```diff
@@ -0,0 +1,150 @@
...
+          "$ref": "#/components/schemas/Entry"
+        }
+      },
+      "QueryEntriesResponse": {
+        "type": "object",
+        "properties": {
+          "entries": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/Entries"
```
sandreae, line 64: Should this actually be:

`"$ref": "#/components/schemas/Entry"`

As it is we have an array of array of entries.

sandreae:
I added example return value definitions for getNextEntryArgs in the openRPC specs but I'm not getting them back in the tests. It all looks fine to me and validates on the openRPC playground tool. The failing test is there to demonstrate.

sandreae:
I really like using openAPI/openRPC but I think we'd need to integrate some validation into our toolchain for it to be used confidently. A tool I liked when using openAPI for REST endpoints (node/express) in the past validated all req/res against the openAPI specs and would reject anything that didn't match.

p2panda-js/test/session.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  it('gets next entry args', async () => {
+    const session = new Session('http://localhost:2020');
+    const { KeyPair } = await session.loadWasm();
+    const keyPair = new KeyPair();
+    const nextEntryArgs = await session.getNextEntryArgs(
+      keyPair.publicKey(),
+      '00401d76566758a5b6bfc561f1c936d8fc86b5b42ea22ab1dabf40d249d27dd906401fde147e53f44c103dd02a254916be113e51de1077a946a3a0c1272b9b348437',
+    );
+    expect(nextEntryArgs.entryHashSkiplink).to.equal('SKIPLINK_HASH');
```
sandreae, line 36: @cafca help! I've been staring at our openRPC definition file for so long and can't see anything wrong, but this test isn't getting the correct example values back from the mock server..... Can you see anything wrong with the `panda_getEntryArguments` definition? It seems to be validating correctly in the openRPC playground site, but obviously something isn't happy.

cafca: I had a go at this yesterday. I didn't find any obvious errors as well. My approach now would be to construct a much simpler example and increase the complexity of the mock server until this error reappears.

sandreae: Thanks for checking! I found the problem now :sob:

sandreae: I didn't realise that when defining example responses, it checks against the example parameters you set. So you could have several different request and response examples (sounds so obious now I write it down, but I didn't realise it was doing this...). In our case it means we have to pass in the exact public_key string, rather than generating a new one on the fly....

cafca: Oh, I just realised that you made changes to this branch. It has been a long time. I should check those commits of yours also. Actually, the process is a bit complex now because how can I review your changes when this is my own pr? hmm..

p2panda-js/test/session.test.ts
```diff
@@ -14,7 +14,19 @@ describe('Session', () => {
...
+    let error;
+    try {
+      await session.queryEntries();
+    } catch (e) {
+      error = e;
+    }
+    expect(error.message).to.equal(
+      "Cannot read property 'length' of undefined",
+    );
+  });
```
cafca: todo: Replace this with the expected (user-friendly) error message

p2panda-js/test/session.test.ts
cafca: Actually, I think it's all good :) Glad you were able to fix it!

sandreae:
The mock server is all working well :+1: I'm quite interested in exploring templating the `openrpc.json` so that we can populate the example params and responses from a test data json file. This would be really powerful I think. I'm gunna do a POC in a new branch though so this PR doesn't get pulled away from it's main purpose.

adzialocha:
> The mock server is all working well :+1: I'm quite interested in exploring templating the `openrpc.json` so that we can populate the example params and responses from a test data json file. This would be really powerful I think

This would be super cool indeed 🤩

sandreae:
> > The mock server is all working well +1 I'm quite interested in exploring templating the `openrpc.json` so that we can populate the example params and responses from a test data json file. This would be really powerful I think
> 
> This would be super cool indeed star_struck

I have a nice flow going using [json-schema-ref-parser](https://github.com/APIDevTools/json-schema-ref-parser) to parse all `$ref`s in a source json schema file and spit out the fully formed `openRPC` spec (mostly happening [here](https://github.com/p2panda/p2panda/blob/template-openrpc-spec/p2panda-js/test/test-setup.ts)). Had to do it like this because openRPC doesn't allow using `$ref`s wherever you like (in example values, for example). The`$ref`s can be to local or remote files, which might be cool for later.

p2panda-js/openrpc.json
cafca: You are right I think!

cafca: You are right I think!

cafca:
> This is really cool!! Thanks for doing it :-)
> 
> Can I ask about the separation between a `Session` and an `Instance`? I remember talking about this at some point, but I can't see exactly where the separation of responsibilities lies.

`Session` is what developers will use, `Instance` is internal.

A `Session` instance is fixed to a p2panda node and is able to load the p2panda wasm library. It provides a low-level api for manipulating entries as well as methods to manipulate instances (`Session.create`, etc.). We want developers to ultimately use the instances api but we don't have all the parts yet (mateiralization). 

The function definitions for the instances api are in a separate `Instance.ts` module to limit the size of `session.ts`.

sandreae:
commit c33c9d1 comment typo... should be "Fix wasm module typo".....

cafca:
@sandrae that's actually an api change not a typo :)

cafca:
But I think we need a different syntax [here](https://github.com/p2panda/p2panda/blob/52fbe4fe539517de63ebb899dc4e3d9d60c9e88f/p2panda-js/src/instance.ts#L31-L31) don't we?

sandreae:
I think the `payload` byte values used in the openrpc mock server are out of date after merging the main branch into this PR. The message fields are in an older formatting.

```
panicked at 'called `Result::unwrap()` on an `Err` value: ErrorImpl { code: Message("missing field `type`"), offset: 0 }', src/message/message.rs:346:61
```

sandreae:
> @sandrae that's actually an api change not a typo :)

Yowzer, poor commit commenting :grimacing:

sandreae (approved):
Looks good to me! Thanks for all your work on this @cafca :+1:

The only thing I think can still be improved is the openrpc spec / mock server / testing workflow as currently we need to painstakingly fill in test values in the spec and test files if we want to change or include new values. But I suggest this can be dealt with in a new PR and I have something ready to go once this is merged ;-p

p2panda-js/package.json
```diff
@@ -60,12 +66,22 @@
...
+  },
+  "dependencies": {
+    "@open-rpc/client-js": "^1.6.3",
+    "@open-rpc/typings": "^1.11.3",
+    "bufferutil": "^4.0.3",
+    "debug": "^2.6.9",
+    "fetch-headers": "^2.0.0",
+    "node-fetch": "^2.6.1",
+    "utf-8-validate": "^5.0.5"
   }
```
cafca, line 86: oops?

p2panda-js/README.md
```diff
@@ -51,9 +51,34 @@ To install `p2panda-js` from the NPM package, simply run:
...
+
+const session = new Session('https://welle.liebechaos.org')
+  .keyPair(keyPair);
+
+const payload = {
+  message: 'Hi there'
+}
+const entry = await session.create(payload, { schema: CHAT_SCHEMA })
+
+```
```
adzialocha, line 76: Some smol nitpicks here since examples are important:

* Add syntax highlighting
* Consistent JavaScript style (semi or not?), empty line at the end
* Make this example actually work, so one can just copy/paste it (define `CHAT_SCHEMA`, move all imports together to the top, use JS comments instead of multiple markdown code blocks with text inbetween)
* Could be nice to write a little bit more what every line is doing (aka "is session.create sending a message"?)
* *instance* is a little mysterious for a young pandawan, maybe we can write "Send a new chat message to the server" or similar?

p2panda-js/package.json
```diff
@@ -8,10 +8,12 @@
     "build": "run-s clear wasm webpack",
     "clear": "rimraf ./lib ./wasm-web ./wasm-node",
     "lint": "eslint --ext .ts .",
-    "test": "run-s clear test:prepare test:mocha",
+    "mock-node-rpc": "open-rpc-mock-server --port 2020 > ./mock-server.log",
```
adzialocha, line 11: Storing the log into a file is cool but maybe that option should be left to the user, in case one prefers to see the logs in a terminal or anywhere else. If I need to see it in a file I can still manually add the redirect when executing the npm script?

p2panda-js/src/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+export const marshallRequestFields = (fields: Fields): FieldsTagged => {
+  const rv: FieldsTagged = {};
+  Object.keys(fields).forEach((k) => {
+    rv[k] = { value: fields[k], type: 'str' };
+  });
+  return rv;
+};
+
+/**
+ * Remove type tagging from mesasge fields on an entry received from node
```
adzialocha, line 15: "mesasge" :smile_cat:

p2panda-js/src/session.ts
```diff
@@ -0,0 +1,278 @@
...
+      // I am removing this again because it breaks the node build but it may
+      // have to go back in.
+
+      // if (this.p2panda.default != null) {
+      //   log('fallback loader from `p2panda.default`');
+      //   // @ts-expect-error only applies for node context
+      //   this.p2panda = await p2panda.default;
+      // } else {
+      //   log('loaded wasm lib');
+      // }
```
adzialocha, line 113: Dead code?

p2panda-js/src/types.ts
```diff
@@ -0,0 +1,94 @@
...
+  action: 'create' | 'update' | 'delete';
+  schema: string;
+  fields: FieldsTagged;
+};
+
+/**
+ * Object containing message fields in tagged form
+ */
+export type FieldsTagged = {
+  // currently only a schema with a text message is supported
```
adzialocha, line 84: Is this true? I thought we've added support for other types as well now :+1:

p2panda-js/test/sample-values.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * These sample values can be used in tests. They match example values in the
+ * openrpc spec, which is used to drive the mock openrpc server.
+ */
+export default {
+  // PRIVATE_KEY: '4c21b14046f284f87f1ea4be4b973664221ad483079a68ed35a6812553b41176',
```
adzialocha, line 6: Why is this commented out?

p2panda-js/src/session.ts
```diff
@@ -0,0 +1,278 @@
...
+import type { EntryArgs, EntryRecord, EncodedEntry, Fields } from '~/types';
+import { KeyPair } from 'wasm-web';
+
+const log = debug('p2panda-js:session');
+
+export default class Session {
+  // Address of a p2panda node that we can connect to
+  endpoint: string;
+
+  // An rpc client connected to the confgiured endpoint
```
adzialocha, line 18: "confgiured"

p2panda-js/src/session.ts
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  /**
+   * Cache next entry args for a given author and schema
+   *
+   * @param author public key of the author
+   * @param schema schema id
+   * @param entryArgs an object with entry arguments
+   */
+  _setNextEntryArgs(
```
adzialocha, line 158: I don't know TypeScript too well, but is there a way to protect access to these private methods by declaring them "private" or "protected" via TypeScript?

p2panda-js/src/session.ts
adzialocha: I'd suggest making an issue instead and removing this :+1:

adzialocha (changes requested):
Mostly small things and a few questions from my side!

p2panda-js/src/types.ts
adzialocha: See: https://github.com/p2panda/p2panda/blob/main/p2panda-rs/src/wasm.rs#L70

p2panda-js/README.md
cafca: Great comments! We could also add something about querying entries / instances and the current limitations there.

sandreae: Should we use "node" instead of "server" here? eg "Open a long running connection to a p2panda node", "Send a new chat message to the node"

p2panda-js/test/sample-values.ts
sandreae: My guess is that it's not used atm, but is useful to have if we want to generate more test values later on.

p2panda-js/src/utils.ts
sandreae: :+1:

p2panda-js/test/sample-values.ts
adzialocha: Then I would just uncomment it 👍

p2panda-js/README.md
sandreae: I updated the example with the recommendations. I also used "node" to describe what we were calling "server".

p2panda-js/src/types.ts
sandreae: We're limited in what we can do wrt field types until we can retrieve user schemas i think. So in terms of this PR we should just change the comment. Or did you have something more in mind?

sandreae: Change the comment, and raise an issue, of course 😉

p2panda-js/README.md
cafca: Actually, I just notice that `KeyPair`, which is used in the example, is only available by directly accessing it from the `wasm` export. But one goal of this pr was to have an API that doesn't require awaiting the import (`const { KeyPair } = await wasm;`) when you want to use `p2panda-js`.

So I added functions to create and restore key pairs directly.

p2panda-js/src/types.ts
cafca: We could try and infer the schema from the data message fields that users supply us with. Then we could use more than just `string`  fields. Would we get a good error message if the wrong message field data is supplied?

p2panda-js/src/session.ts
cafca: Yes! I never used them.

p2panda-js/src/session.ts
cafca: I don't remember exactly what the issue actually was. I think it was that we are only testing this library in a node environment with our test suite and there was a problem with the bundling in browser environments. I will make an issue to test that.

p2panda-js/src/session.ts
cafca: I have made some fields `private`, but this also prevents us from using them in other `p2panda-js` modules (e.g. `instance.ts`) so I have reverted the naming with leading underscore.

p2panda-js/src/session.ts
cafca: https://github.com/p2panda/p2panda/issues/82

p2panda-js/src/session.ts
adzialocha: Does `protected` do the job? Sounds a little like what you are describing

p2panda-js/src/types.ts
cafca: Ok, I have added support for `int` and `boolean`! `relationship` and `float` will have to wait until we have better schema support :)

PR #68, opened 2021-05-18
Implement method for iterating over MessageFields

sandreae:
We need this functionality in several places...

### DIFF
--- a/p2panda-rs/src/atomic/message.rs
+++ b/p2panda-rs/src/atomic/message.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::collections::btree_map::Iter;
 
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde_repr::{Deserialize_repr, Serialize_repr};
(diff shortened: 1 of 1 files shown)


sandreae:
I tried implementing an `Iterator` trait on `MessageFields` but came up against some issues with that. This seems like a sime solution as `BTreeMap`S are already iterable.

p2panda-rs/src/atomic/message.rs
```diff
@@ -180,6 +181,16 @@ impl MessageFields {
...
         self.0.get(name)
     }
+    
+    /// Returns an array of existing message keys.
+    pub fn keys(&self) -> Vec<String> {
+        self.0.keys().cloned().collect()
+    }
+    
+    /// Returns an iterator of existing message fields.
+    pub fn iterator(&self) -> Iter<String, MessageValue> {
```
adzialocha, line 191: More of a naming (hehe) question than a suggestion: How does one name iterators? Do you name "what" you are iterating upon (for example `keys_with_values`) or should we use the Rust std library name which is `iter`? .. or is this actual a conversion in some way, like its the same thing but in a different shape (one could see an Iterator as a different way to access the same data)? `HashMap` does it like this: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.into_keys

Hmn. I would vote for just calling this `iter` to stay consistent with the std lib.

sandreae: yeh, i think `iter`is best too :+1:

sandreae: Is this the method you would take though? I thought it would be simple to implement `Ìterable` as a trait but actually it was a little convoluted because the `Item` we were trying to return with `next()` is a tuple (and then some borrow issues, blah blah)...

adzialocha: Yes it's super like that 🌟

PR #69, opened 2021-05-18
Restructure modules

### LINKED ISSUE #57: Rename `atomic` module

adzialocha:
### Changes in this PR

* Restructures all structs and methods in `p2panda-rs` into a new module / file structure according to https://github.com/p2panda/p2panda/issues/57. :warning: **This breaks almost all imports!**
* Rename `Validation` trait to `Validate` to make it consistent with method names. :warning: 
* Split up error types into `error.rs` files in regarding modules.
* Move `decode_entry` into `decode.rs` and `sign_and_encode` into `encode.rs`.
* Move entry encoder / decoder tests into `entry/tests.rs` as they are more of integration test nature.
* Add a little bit more of documentation for modules.

### New file structure

```
├── entry
│   ├── decode.rs
│   ├── encode.rs
│   ├── entry.rs
│   ├── entry_signed.rs
│   ├── error.rs
│   ├── log_id.rs
│   ├── mod.rs
│   ├── seq_num.rs
│   └── tests.rs
├── hash
│   ├── error.rs
│   ├── hash.rs
│   └── mod.rs
├── identity
│   ├── author.rs
│   ├── error.rs
│   ├── key_pair.rs
│   └── mod.rs
├── lib.rs
├── message
│   ├── error.rs
│   ├── message_encoded.rs
│   ├── message.rs
│   └── mod.rs
├── schema
│   ├── error.rs
│   ├── message.rs
│   └── mod.rs
└── wasm.rs
```

Closes #57


sandreae (approved):
I think this looks great :+1:

The modules make sense conceptually to me. `identities` is really nice.

Also the imports make a lot of sense.

Only thing I wonder about is whether `decode` and `encode` should be in their own module. They're the only two imports which are functional instead of structs, they stand out as different. They do make sense in `entry` though as that is where they do their work. I would say leave them where they are for now and if we end up with more utilities like this we can desperate them later if it makes sense. 

Looking forward to having this in the main branch ;-p

p2panda-rs/src/entry/decode.rs
```diff
@@ -0,0 +1,48 @@
+use std::convert::TryInto;
+
+use arrayvec::ArrayVec;
+use bamboo_rs_core::Entry as BambooEntry;
+
+use crate::entry::{Entry, EntrySigned, EntrySignedError, LogId, SeqNum};
+use crate::hash::Hash;
+use crate::message::{Message, MessageEncoded};
+
```
sandreae, line 9: imports like this make so much sense now :+1:

adzialocha:
> Only thing I wonder about is whether decode and encode should be in their own module. They're the only two imports which are functional instead of structs, they stand out as different. They do make sense in entry though as that is where they do their work.

For consistency it could be better like you say indeed, though it's also nice to keep files as small as possible. I'm undecided here, what do the others think? @cafca and @sophiiistika ?

adzialocha:
I'm merging it now anyways, as this discussion will not affect the public API and we can do changes later as well!

PR #70, opened 2021-05-19
Minor API changes

adzialocha:
Some small changes for API consistency:

* Rename `as_hex` to `as_str` for consistency with similar methods
* Rename `MessageFields::iterator` to `iter` for consistency
* Return all getter values by reference when possible
* Do not return anything when validating something

Closes: #50


sandreae (approved):
Yep, all good. Thanks!

Issue #52, opened 2021-04-19
Implement all field types for adding message fields in wasm module


adzialocha:
Currently we only support fields of type `Text` to be added to `MessageFields`:

https://github.com/p2panda/p2panda/blob/main/p2panda-rs/src/wasm.rs#L53-L56

We also need to add `Float`, `Integer`, `Relation` and potentially other field types.

Related: https://github.com/p2panda/p2panda/issues/53

Issue #53, opened 2021-04-19
Improve `MessageFields` wasm export


adzialocha:
There must be a better way to write the `MessageFields` wasm export in https://github.com/p2panda/p2panda/blob/main/p2panda-rs/src/wasm.rs#L35-L68:

* All other wasm methods follow a functional style except of this one, is it possible to do the same?
* Using `JsValue` as an argument type feels strange, as we introduce the "wild type west" of the JavaScript world into Rust. Maybe we need to refactor the `add` method to be more type strict / being less generic? For example by introducing more methods like `add_text_field`, `add_float_field`, `add_integer_field` etc.?

Related: https://github.com/p2panda/p2panda/issues/52

PR #71, opened 2021-05-19
Extend `MessageFields` for WebAssembly

adzialocha:
* Add `add` method with type field to add Relation, Float, Integer and Boolean fields as well :warning: **Breaking Change!**
* Add method to read `length` of `MessageFields` instances
* Add methods to `get` and `remove` fields

Closes: #52 and #53


cafca:
@adzialocha what do you think about making a `set` method with a `type` parameter instead of the separate methods for each type?

adzialocha:
> @adzialocha what do you think about making a `set` method with a `type` parameter instead of the separate methods for each type? I was thinking about how we will be populating the fields based on type information from the schema definition later and that that might be more elegant with a type parameter.

Interesting idea, I'll have a look!

adzialocha:
Ok, I think I have it @cafca, please have a look!

adzialocha:
I've renamed the method to `add` again as it is a) consistent with how it is named in `p2panda-rs` b) `set` implies that you can just overwrite values when applying the method on the same field again which is not the case, it will actually throw an error when it is already given.

Issue #72, opened 2021-05-20
Correct format of typed message fields


adzialocha:
Finalize the specification of typed fields in messages:

## Before

```
...
fields: {
   message: {
     Text: "Hello, Panda!"
   }
   ...
}
```

## After

```
...
fields: {
   message: {
     value: "Hello, Panda!",
     type: "text"
   }
   ...
}
```

The new format also has the advantage to add future parameters to fields.

PR #73, opened 2021-05-20
Change format of typed message fields

adzialocha:
This will format messages like this:

```json
{
  "action": "update",
  "schema": "00402dc25d32dfb400bb295b6...",
  "version": 1,
  "id": "00405d49304e29d1d9358314e1083...",
  "fields": {
    "age": {
      "type": "integer",
      "value": 28
    },
    "height": {
      "type": "float",
      "value": 3.51
    },
    "is_admin": {
      "type": "boolean",
      "value": false
    },
    "profile_picture": {
      "type": "relation",
      "value": "0040cf94f6d605657e90c543b0c91..."
    },
    "username": {
      "type": "text",
      "value": "bubu"
    }
  }
}
```

**Todo:**

* [x] Changes the type tag when serializing message fields from `<Key>: { <Type>: <Value> }` to `<Key>: { type: <Type>, value: <Value> }`
* [x] Update the CDDL Message schema accordingly
* [x] Fix `p2panda-js` / wasm API

Closes: #72


sandreae:
Looks good to me. I like the new format and the shortened type strings are clear 👍

sandreae (approved):
Looks good 👍👍

cafca:
Could we also compress messages instead of trying to shorten syntax?

adzialocha:
> Could we also compress messages instead of trying to shorten syntax?

True, this will have a large effect. This is handled through the underlying protocol, like during HTTP transport we could enable brotli or gzip compression, for replication I'm sure there is similar solutions.

Are my suggestions already too short, even before compression?

cafca:
`rel` is the only one i'm not sure about. maybe `ref`? I suggest we keep it as `relation`. The other ones can be shortened or not, I don't mind. 

Compression as part of the message encoding would be interesting and could be done in `TryFrom<&Message> for MessageEncoded` and `From<&MessageEncoded> for Message`.

adzialocha:
> `rel` is the only one i'm not sure about. maybe `ref`? I suggest we keep it as `relation`. The other ones can be shortened or not, I don't mind.

Ok, I've changed and will merge now :+1:

> Compression as part of the message encoding would be interesting and could be done in `TryFrom<&Message> for MessageEncoded` and `From<&MessageEncoded> for Message`.

I think that compression should exist on another level than the actual message encoding as its is not really part of the protocol design but maybe an optimization some Client - Server communication might support, others not. We could add it as an extension of the RPC protocol (like, when a compression header flag exists the client / server will compress / decompress accordingly). Also, adding it in the conversion trait would make debugging harder as you would need to decompress all CBOR payloads before looking into them.

I've made a ticket for this: https://github.com/p2panda/p2panda/issues/76

PR #80, opened 2021-08-05
add demo to readme

nichoth:
Add a link to the demo app, `beep boop`, because it is a nice example


cafca:
Oh I like adding the demo here. And it also fits with the other example because the demo page also contains the usage example on its left hand side.

PR #85, opened 2021-08-31
Add methods to update and delete instances

### LINKED ISSUE #29: Add method to create create, update, delete messages

cafca:
Caveat: the method to delete instances is unfortunately named `remove` because delete is a reserved word in js.

Todo:

- [x] Implement methods
- [ ] Add tests

Closes #29

### DIFF
--- a/p2panda-js/src/instance.ts
+++ b/p2panda-js/src/instance.ts
@@ -85,6 +85,39 @@ const create = async (
 };
 
 /**
+ * Signs and publishes an `update` entry for the given instance id and fields
+ */
+const update = async (
+  id: string,
+  fields: Fields,
+  { keyPair, schema, session }: Context,
+): Promise<void> => {
+  const { encodeUpdateMessage } = await session.loadWasm();
+
+  // Create message
+  const fieldsTagged = marshallRequestFields(fields);
+  const messageFields = await getMessageFields(session, fieldsTagged);
+  const encodedMessage = encodeUpdateMessage(id, schema, messageFields);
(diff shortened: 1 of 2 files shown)


cafca:
@sandreae I think I need help updating/checking the openrpc schema here. I added a test for creating an instance and I think it should be exactly what's already configured in the schema but the test fails in a way that makes me think that something is off. Maybe we can have a look at this together some time soon?

sandreae:
Ok, cool, I will look at this today at some point! I'll give you a bell once I have done for a chat if there is something interesting.

sandreae:
I've had a little look into this. It looks like the `Author` public key being sent in the test request to our mock server is somehow not the correct one. Not clear why this is yet though ;-p will look into it a bit more. It seems like that's where the problem is though.

sandreae:
There were also some changes in main that needed to be merged, now it passes :-)

cafca:
Ooh nice, thanks! I guess the best way to prevent this kind of error would be generating `test-data.json` automatically using `p2panda-rs` and a more minimal number of inputs.

cafca:
I am not happy with how the tests are structured and will do some refactoring here I think.

sandreae:
> Ooh nice, thanks! I guess the best way to prevent this kind of error would be generating `test-data.json` automatically using `p2panda-rs` and a more minimal number of inputs.

Yeh, I would like to generate it from rs :+1: 

What do you mean by more minimal inputs?

PR #102, opened 2021-09-17
p2panda-js directory restructure

cafca:
Restructure p2panda-js directory structure to align closer with p2panda-rs. Previously much of the functionality was contained in `instance.ts` without a good reason. Instead, this 

- creates subdirectories in `src` for grouping code related to concepts
- creates new files `message.ts` and `entry.ts` that contain code related to these concepts
- moves `keyPair.ts` into a new directory `identity`, mirroring p2panda-rs as well
- moves test files next to the files they are testing
- leaves only test setup files in the `/test` directory


p2panda-js/src/instance/instance.ts
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import { Fields, InstanceRecord } from '~/types';
+import { marshallRequestFields } from '~/utils';
+
+import { getMessageFields } from '~/message';
+import { signPublishEntry } from '~/entry';
+import { materializeEntries } from './materialiser';
+
+import type { Context } from '~/session';
```
adzialocha, line 10: These are really small nitpicks:

* Since we group the imports via newlines I wonder what the system is behind it? All of these import paths seem to belong to the same level and I would suspect them to be in one group, except of the `type` import.
* In terms of consistency it can be cool to stick to either `~` or `./` for importing our own modules and avoid having a mix. I suggest sticking to `~` whereever we import in the codebase as it allows simple copy / paste between documents without caring about the folder structure.

Btw, this is cool: https://github.com/import-js/eslint-plugin-import

p2panda-js/src/entry/index.ts
```diff
@@ -0,0 +1,3 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+export * from './entry';
```
adzialocha, line 3: When reading this PR I found myself checking all `index.ts` files first to understand what the modules per folder export, which was really nice! Except of this case .. and the `session/index.ts` where we use a wildcard import. This is really a detail, just wanted to suggest to make exports explicit here as well for readability.

adzialocha (commented):
Thank you for this refactor! :+1: 

I only have very minor comments, only cosmetics and consistency related. I also don't really like the `'~/../lib'` path in the tests, it looks strange! But this can be tackled later when we refactor the tests using Jest.

p2panda-js/src/instance/instance.ts
cafca: > Since we group the imports via newlines I wonder what the system is behind it? All of these import paths seem to belong to the same level and I would suspect them to be in one group, except of the type import.

Good point. I also like setting up eslint to manage this, maybe in a separate PR (didn't we already do that somewhere? 🤔)

> I suggest sticking to ~ whereever we import in the codebase as it allows simple copy / paste between documents without caring about the folder structure.

That's an interesting thought. I can see how that's useful when you are copying imports. 

I like using both and choosing whatever results in a shorter import path. Usually that means, importing with `./` when things are in the same directory and using `~/ when they are elsewhere. About this I also like making an explicit distinction between imports from the same directory and imports from elsewhere in the project.

p2panda-js/src/entry/index.ts
cafca: Good point! I was also thinking about this. The downside I see is that you might `export` a new symbol from a module without updating the `index.ts` and then wonder why you can't import it elsewhere. But having an overview of what is contained in a folder is also very useful so I'll change this.

Btw: I mostly set up the `index.ts` files like this to be able to distinguish filenames without looking at their path, i.e. my editor and file search display filenames only and it's frustrating if all of the modules are `index.ts` because they all look the same. Putting the "meat" of the module in a file with a descriptive name solves this problem.

p2panda-js/src/instance/instance.ts
cafca: @adzialocha please have a look at the new grouping in `instance.ts`. Is this clearer to you?

adzialocha: Yes, thank you!!

adzialocha: > About this I also like making an explicit distinction between imports from the same directory and imports from elsewhere in the project.

This also sounds good :+1:

PR #103, opened 2021-09-19
Open-rpc improvements

sandreae:
- include descriptions and summaries for all methods
- fix regex mistake


sandreae:
I've made all these changes while leaving all the code where it is. I figured it would complicate things with the general folder structure refactoring which is going on in parallel.

adzialocha (approved):
Nice, I would maybe simplify the `README.md` a little (it doesn't need all of the fancy headings), but this we can do later :+1:

PR #104, opened 2021-09-21
Use Jest as test framework

adzialocha:
This PR replaces the mocha test framework with Jest.

* It simplifies how we test async promises.
* We can get rid of `chai`, `chai-as-promise` and `sinon` dependencies.
* All tests import directly from their modules now and not from the whole library which make the tests more like actual unit tests.
* Some types were not respected (`undefined` instead of `null` and `UInt8Array` instead of `string`), this PR fixes type errors when using the API. For conscious wrong API usage the type checker is explicity ignored.


p2panda-js/package.json
```diff
@@ -10,8 +10,8 @@
     "lint": "eslint --ext .ts .",
     "test": "run-p --race test:mock-node test:run",
     "test:prepare": "run-s --silent wasm webpack:node",
-    "test:mocha": "mocha -r ts-node/register -r tsconfig-paths/register -r @babel/register 'src/**/*.test.ts'",
-    "test:run": "run-s clear test:prepare test:mocha",
+    "test:jest": "jest",
+    "test:run": "run-s clear test:prepare test:jest",
```
cafca, line 14: I think we don't need `test:prepare` anymore or am I missing something? I was able to replace this line with

```
"test:run": "run-s clear wasm:node test:jest"
```

and then I just had to edit two lines in other files where type definitions where imported from `wasm-web` and point them at `wasm-node` instead. We can drop the whole `webpack` build and `wasm:web` as well.

p2panda-js/src/wasm.test.ts
```diff
@@ -121,37 +118,37 @@ describe('Web assembly interface', () => {
...
       const messageEncoded = encodeCreateMessage(TEST_SCHEMA, fields);
 
       // Sign and encode entry
       const { entryEncoded, entryHash } = signEncodeEntry(
         keyPair,
         messageEncoded,
-        null,
-        null,
+        undefined,
+        undefined,
```
cafca, line 130: I would be interested to hear why you changed this.

p2panda-js/jest.config.js
```diff
@@ -0,0 +1,19 @@
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
```
cafca, line 4: I am asking myself whether it would be possible to test p2panda-js in a browser environment using jsdom in the future.

cafca (changes requested):
Looks nice! I would have expected more setup to be needed for jest to work but this looks pretty straightforward.

p2panda-js/src/wasm.test.ts
cafca: (the tests still pass for me with `null`, which is more idiomatic in my understanding of JavaScript)

adzialocha: I see the following error here, not sure why this is different from what you can see?

> src/wasm.test.ts:129:9 - error TS2345: Argument of type 'null' is not assignable to parameter of type 'string | undefined'.

p2panda-js/jest.config.js
adzialocha: Checking the wasm tests against the browser can also be interesting: https://rustwasm.github.io/wasm-pack/book/tutorials/npm-browser-packages/testing-your-project.html. It does not test the p2panda-js library but at least the hairy wasm part.

```bash
wasm-pack test --headless --firefox
```

p2panda-js/package.json
adzialocha: Hmn .. this would still break the tests when running them for the first time (you still need the wasm build).

adzialocha: I've brought that change back to where it was .. there is surely some space to optimize the wasm build. This build step is only necessary a) when the repo got freshly cloned and b) when the `p2panda-rs` code changed. Currently I don't see an elegant way to only run the scripts in these cases, maybe with some smart content hash comparison?

cafca: Ah, I think that was a misunderstanding. I'll make the change and then you can have a look, hope that's okay.

cafca: I had another error yesterday that my head wouldn't really want to be wrapped around so I didn't push anything. We can do this later, yes.

p2panda-js/jest.config.js
cafca: I like that

sandreae (approved):

p2panda-js/src/wasm.test.ts
adzialocha: The error occurs when I run the tests .. it must be the `ts-jest` plugin giving me that warning? When I do `npm run build` I don't get warned, when I run `./node_modules/.bin/tsc` I do get it ..

adzialocha: Also, indeed the `index.d.ts` file in `wasm-node` declares the types as following: 

```typescript
export function signEncodeEntry(key_pair: KeyPair, encoded_message: string, entry_skiplink_hash: string | undefined, entry_backlink_hash: string | undefined, seq_num: number, log_id: number): any
```

cafca:
@adzialocha I have removed the type declarations from the wasm-adapter, they were just not working across different Typescript environments because of the dynamic import in the wasm-adapter. Hope that's okay, I would like to tackle that separately and it allows removing the unneccessary build steps from the test pipeline.

p2panda-js/src/wasm.test.ts
cafca: Strange issue, I didn't get any of these issues.

cafca (approved):

adzialocha:
> @adzialocha I have removed the type declarations from the wasm-adapter, they were just not working across different Typescript environments because of the dynamic import in the wasm-adapter. Hope that's okay, I would like to tackle that separately and it allows removing the unneccessary build steps from the test pipeline.

Super, in the `wasm-refactor` branch this is entirely gone anyhow :-)

PR #105, opened 2021-09-24
Refactor WebAssembly build pipeline

adzialocha:
This PR simplifies and unifies the WebAssembly build pipeline supporting browser and NodeJS targets.

* The `wasm-adapter` module got removed and unified to one simple interface in `wasm.ts` in the TypeScript code.
* Removed conversion to base64 string for web environment. Webpack is handling this part for us now automagically.
* Use `WasmPackPlugin` plugin instead of `wasm-pack` command line tool. This cleans up some scripts in `package.json`.
* Renamed webpack task from `browser` to `web` to be closer to Webpack and WasmPack naming.
* Use only one WasmPack target folder named `wasm` for all builds. This causes the task to be run in series instead of concurrently but on the bright side it makes the process easier to follow.

**Caveats**

* Webpack does some ugly code reformatting with the generated code by `wasm-pack` when we bundle it. This causes a bug breaking our code (`"Uncaught TypeError: getObject(...).require is not a function"`) which can only be prevented with a workaround where the `wasm` library is excluded from the Webpack build. Related: https://github.com/webpack/webpack/issues/8826 and https://github.com/rust-random/getrandom/issues/224
* Because of the above issue, we still need to make one WebAssembly build for `web` and `nodejs` targets each as we can't use webpack magic to bundle the wasm for a specific environment.

Closes https://github.com/p2panda/p2panda/issues/101
Merge this first: https://github.com/p2panda/p2panda/pull/104


p2panda-js/package.json
```diff
@@ -5,22 +5,18 @@
...
-    "wasm:copy": "mkdir -p ./lib/wasm; cp ./wasm-node/index.js ./lib/wasm; cp ./wasm-node/index.d.ts ./lib/wasm; cp ./wasm-node/index_bg.wasm ./lib/wasm",
-    "webpack": "run-p --aggregate-output webpack:browser webpack:node",
-    "webpack:browser": "webpack --mode production --config-name browser --color",
-    "webpack:node": "npm run wasm:copy && webpack --mode production --config-name node --color"
+    "test:run": "run-s clear webpack:node test:jest",
+    "test:template-openrpc": "npx ts-node ./test/generate-openrpc-spec.ts",
+    "test:watch": "nodemon --watch '../p2panda-rs/src/*' --watch './src/*' --exec 'npm test' --ext js,ts,json",
+    "webpack": "run-s webpack:web webpack:node",
+    "webpack:web": "webpack --mode production --config-name web --color",
+    "webpack:node": "webpack --mode production --config-name node --color"
```
cafca, line 19: I think it may be helpful to allow running webpack in `development` mode. What do you think about setting the mode with `cross-env NODE_ENV=development` in the `test:run` script instead of hardcoding it here? The `build` script will then implicitly use `production` mode anyway. @adzialocha

cafca (changes requested):
See comment above about dev/prod mode. Otherwise this is good to merge!

PR #108, opened 2021-09-29
Correct error in panda_queryEntries result

sandreae:


p2panda-js/openrpc.json
```diff
@@ -497,7 +495,7 @@
         {
             "name": "panda_publishEntry",
             "summary": "Publish an encoded entry and message.",
-            "description": "Publish a new `Entry` to a `p2panda` log. See the [p2panda-rs](https://github.com/p2panda/p2panda/tree/main/p2panda-rs) and [p2panda-js](https://github.com/p2panda/p2panda/tree/main/p2panda-js) libraries or the p2panda [handbook](https://github.com/p2panda/handbook) for more information.",
```
adzialocha, line 500: This will revert an earlier commit, maybe you have to pull from `main` first?

adzialocha (changes requested):

p2panda-js/openrpc.json
sandreae: hmm, it's up to date with `main` maybe the change was made in the generated spec (which gets overwritten) not the template?

adzialocha (approved):

p2panda-js/openrpc.json
adzialocha: Aahh, ok, I understand. Sorry and thank you for fixing it 😅

PR #114, opened 2021-10-12
Add methods to update and delete instances (Season 2)

cafca:
This is a second attempt at implementing and testing methods to update and delete instances ([Part 1](https://github.com/p2panda/p2panda/pull/85)), which is based off a more recent `main`.

Todo:
- [x] Implement update and delete instance functions
  - [x] Add tests
- [x] Add update and delete methods to session 
  - [x] with parameter validation
  - [x] Add tests
- [x] Build a more reliable supplier of `test-data.json`
- [x] Fix bug in materializer

### DIFF
--- a/p2panda-js/src/instance/instance.test.ts
+++ b/p2panda-js/src/instance/instance.test.ts
@@ -2,8 +2,10 @@
 
 import { recoverKeyPair } from '~/identity';
 import { Session } from '~/session';
-import { createInstance, deleteInstance, updateInstance } from '.';
 import { Fields } from '~/types';
+
+import { createInstance, deleteInstance, updateInstance } from '.';
+
 import {
   authorFixture,
   entryFixture,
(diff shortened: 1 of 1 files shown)


sandreae:
I added some tests for the test data..... lolz..... and so far they are passing fine. Not sure if I'm testing everything yet, but it is making me think the test data wasn't causing the problem we had earlier. I'll keep adding a few more tests in case I am missing something though.

adzialocha:
Also, does this have to be part of this PR? :joy_cat:

sandreae:
Phew! I worked on the test setup once again and now have now arrived at some elegant solutions which solves a few blocks (some of which I imagined).

Changes/improvements/realizations:
- test data now includes an array of `nextEntryArgs` which correspond to each past state of the log
- there are now `update` and `delete` messages` in the log
- the `nextEntryArgs` returned from the mock server are correct for the logs we have
  - this means we can test a round trip in creating a next entry, we can test the decoded values we epect to exist there, then we can publish it, we will get a response back if everything is formed correctly, but the response will contain (type valid) randomly generated values. This means we can only test if it was successful, but that's fine. 
- we can test publishing older log entries though, as we have all the previous entry args we need (writing these responses in the openRPC spec is still TODO)

Issue #97, opened 2021-09-08
Implement `Eq` trait for core structs


adzialocha:
This will enable us to use `Hash`, `Author` etc. as a key of a `HashMap` data type.

PR #129, opened 2021-11-05
Introduce materialiser module with DAG logic

sandreae:
## Create materialisation module in `p2panda-rs` with graph logic for:

- converting an array of Messages into an array of Edges
- build a DAG from scratch or from an existing array of Edges
- topologically sort the DAG

### The DAG should:
- not know anything about p2panda types
- just deal with simple String types
- sort topologically choosing leading branches by "highest" string value
 
**fun fact: writing tests for this module will be much easier once https://github.com/p2panda/p2panda/pull/116 is merged**

## 📋 Checklist


p2panda-rs/src/materializer/dag.rs
```diff
@@ -0,0 +1,300 @@
...
+    use super::DAG;
+
+    use crate::entry::{sign_and_encode, Entry, LogId, SeqNum};
+    use crate::hash::Hash;
+    use crate::identity::KeyPair;
+    use crate::materializer::{marshall_entries, Edge};
+    use crate::message::{Message, MessageEncoded, MessageFields, MessageValue};
+
+    use rstest::{fixture, rstest};
+    use std::convert::TryFrom;
```
adzialocha, line 144: External imports I would move to the top, then crate level imports and then module level.

p2panda-rs/src/materializer/utils.rs
```diff
@@ -0,0 +1,72 @@
...
+    entries: Vec<(EntrySigned, MessageEncoded, InstanceAuthor)>,
+    permitted_authors: Vec<Author>,
+) -> Vec<(EntrySigned, MessageEncoded, InstanceAuthor)> {
+    entries
+        .iter()
+        .cloned()
+        .filter(|(entry_encoded, _, instance_author)| {
+            entry_encoded.author().as_str() == instance_author.as_str()
+                || permitted_authors.iter().any(|permitted_author| {
+                    permitted_author.as_str() == entry_encoded.author().as_str()
```
adzialocha, line 68: `Author` could implement the `Eq` trait, then you would not need to convert them to comparable strings first. Maybe we should make an issue for this (doesn't need to be part of this PR)?

p2panda-rs/src/materializer/error.rs
```diff
@@ -0,0 +1,16 @@
...
+
+use thiserror::Error;
+
+/// Error types for methods of `Materializer` struct.
+#[derive(Error, Debug)]
+#[allow(missing_copy_implementations)]
+pub enum MaterializationError {
+    /// Materialization error.
+    #[error("materialization error")]
+    Error,
```
adzialocha, line 11: I can't see this error being used anywhere (also it does not describe any particular case), can this be removed?

adzialocha (changes requested):
Looking really nice! I have minor comments only regarding dead code and documentation etc.

p2panda-rs/src/materializer/utils.rs
sandreae: :+1: I think there already is one? https://github.com/p2panda/p2panda/issues/97

p2panda-rs/src/materializer/dag.rs
```diff
@@ -0,0 +1,300 @@
...
+
+        // DAG trunk A
+        graph_3.add_edge("0x2a".to_string(), "0x3a".to_string());
+        graph_3.add_edge("0x3a".to_string(), "0x4a".to_string());
+
+        assert_eq!(graph_3.topological(), ordered_dag)
+    }
+
+    #[rstest]
+    fn marshall_entries_test(entry: Entry, key_pair: KeyPair) {
```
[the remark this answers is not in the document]
sandreae: It can, but it requires quite a lot of boilerplate, which would then be duplicated. This will all be gone when we can use `test-utils` so I suggest waiting until both PRs are merged then refactoring.

sandreae:
Think all change requests have been resolved :cold_sweat:

adzialocha (approved):

sandreae:
I don't want to merge this yet actually.... Not until we've discussed the next steps a bit more, including our implementation of `Instance`. Really we need to rename and update `Operation` (`Message`) to include a `previousOperation` before implementing any of the instance/materialisation logic.

adzialocha:
> I don't want to merge this yet actually.... Not until we've discussed the next steps a bit more, including our implementation of `Instance`. Really we need to rename and update `Operation` (`Message`) to include a `previousOperation` before implementing any of the instance/materialisation logic.

I understand the reasoning but isn't it worth to still merge it? It might take a while until we understood and specified all-things-materialization-logic, and in any case I'm sure some parts implemented in this PR will be reused, even when it comes with potential refactoring.

adzialocha:
For me this PR does NOT implement a materializer, it implements helper method which might be useful for it in the future :+1:

sandreae:
Yes, both points very true, also we wanted to approach it step by step, which this definitely does do. Cool :+1:

p2panda-rs/src/materialiser/dag.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    use crate::entry::{sign_and_encode, Entry, LogId, SeqNum};
+    use crate::hash::Hash;
+    use crate::identity::KeyPair;
+    use crate::materialiser::{marshall_entries, Edge};
+    use crate::message::{Message, MessageEncoded, MessageFields, MessageValue};
+
+    use super::DAG;
+
+    // All this boilerplate to be removed once `p2panda-tests` is merged.
```
cafca, line 144: It is merged!

p2panda-rs/src/materialiser/dag.rs
```diff
@@ -0,0 +1,298 @@
...
+
+        // DAG trunk A
+        graph_3.add_edge("0x2a".to_string(), "0x3a".to_string());
+        graph_3.add_edge("0x3a".to_string(), "0x4a".to_string());
+
+        assert_eq!(graph_3.topological(), ordered_dag)
+    }
+
+    #[rstest]
+    fn marshall_entries_test(entry: Entry, key_pair: KeyPair) {
```
cafca, line 258: Should this test be placed with the marshalling function?

cafca (changes requested):
Love the clarity of the implementation. This really feels like it should become its own crate once we know it's stable. I have added some questions and comments. Also check out clippy, I just learned about it and there seem to be some helpful suggestions for the code here.

p2panda-rs/src/materialiser/error.rs
```diff
@@ -0,0 +1,12 @@
...
+
+use thiserror::Error;
+
+/// Error types for methods of `Materializer` struct.
+#[derive(Error, Debug)]
+#[allow(missing_copy_implementations)]
+pub enum MaterialisationError {
+        /// Handle errors from `EntrySignedError` struct.
+    #[error(transparent)]
+    EntrySignedError(#[from] crate::entry::EntrySignedError),
```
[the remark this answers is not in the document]
adzialocha: Thats maybe more an implicit "error handling pattern" we're using in `p2panda-rs`. We organize errors per module, so the `materialiser` module has its own Error type, the `entry` module has its own Error type and so on .. To know support handling errors coming from the outside they need to be converted. This can be done with the help of the `thiserror` crate and some magic macros. I think this separation helps a lot since Errors in Rust are relatively crazy and this gives clarity where something came from and how the dependencies are.

Errors in Rust is quite a [specific thing](https://www.lpalmieri.com/posts/error-handling-rust/) if you're interested in learning more :crying_cat_face:

p2panda-rs/src/materialiser/dag.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /// Return graph edges as array.
+    pub fn graph(&self) -> Vec<Edge> {
+        self.graph.to_owned()
+    }
+
+    /// Add a root node to the graph.
+    pub fn add_root(&mut self, node_id: Node) {
+        self.graph.insert(0, (None, node_id.into()));
```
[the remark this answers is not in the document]
sandreae: yeez, nice.

p2panda-rs/src/materialiser/dag.rs
```diff
@@ -0,0 +1,298 @@
...
+pub struct DAG {
+    /// An array of edges which make up the graph. For p2p2anda this is an array of tuples of
+    /// Entry hashes [(Some("00x42asd..."), "00x435d..."), .... ], but it can be any string.
+    /// The first string in the tuple is optional as the root of the graph has no parent.
+    graph: Vec<Edge>,
+}
+
+impl DAG {
+    /// Instantiate a new empty graph
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
sandreae: very nice!

p2panda-rs/src/materialiser/dag.rs
sandreae: Whoop, I'm using it!

p2panda-rs/src/materialiser/dag.rs
sandreae: Yeep

p2panda-rs/src/materialiser/dag.rs
```diff
@@ -0,0 +1,298 @@
...
+
+/// Wrapper type for an Edge in the graph
+pub type Edge = (Option<Node>, Node);
+
+/// A directed acyclic graph which can be ordered topologically in a depth-first sort. It is
+/// described by a list of `edges` which in turn describe connections between parent and child
+/// nodes.
+#[derive(Clone, Debug)]
+pub struct DAG {
+    /// An array of edges which make up the graph. For p2p2anda this is an array of tuples of
```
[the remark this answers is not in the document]
sandreae: moved.

p2panda-rs/src/materialiser/utils.rs
```diff
@@ -0,0 +1,72 @@
...
+            ),
+            None => (None, entry_signed.hash().as_str().to_owned()),
+        };
+        edges.push((link, id));
+    }
+    Ok(edges)
+}
+
+/// Filter entries against instance author for a single writer setting. This is needed for materializing System Logs.
+pub fn single_writer(
```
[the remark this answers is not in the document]
sandreae: Yes, these are for the future. I wrote tests for them :+1:

p2panda-rs/src/materialiser/utils.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::identity::Author;
+use crate::materialiser::{Edge, MaterialisationError};
+use crate::message::MessageEncoded;
+
+pub type InstanceAuthor = Author;
+
+/// Method for marshalling an array of (EntrySigned, MessageEncoded) into an array of Edges which can then
+/// be turned into a DAG. It is here to detach the DAG from any concepts of Entries and Messages. The DAG just sees
+/// Strings. This step will look different once have some more instance related fields in our Message struct.
+pub fn marshall_entries(
```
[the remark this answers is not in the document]
sandreae: Yes, very good point. I moved these methods into their own files with more explicit names.

p2panda-rs/src/materialiser/utils.rs
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+use crate::entry::{decode_entry, EntrySigned};
+use crate::identity::Author;
+use crate::materialiser::{Edge, MaterialisationError};
+use crate::message::MessageEncoded;
+
+pub type InstanceAuthor = Author;
```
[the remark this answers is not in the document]
sandreae: removed!

p2panda-rs/src/materialiser/error.rs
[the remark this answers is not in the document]
cafca: Thanks!

p2panda-rs/src/materialiser/marshall_entries.rs
```diff
@@ -0,0 +1,91 @@
...
+            Ok(entry) => Ok(entry),
+            Err(err) => Err(MaterialisationError::EntrySignedError(err)),
+        }?;
+
+        if entry.message().is_none() {
+            // The message has been deleted.
+            continue;
+        }
+
+        // If we have a `link` field in the Message then we can use it here.
```
cafca, line 25: I think we can't have that in create messages can we? Because we need to make the message, then make an entry from it, and only then we have the id. It's a chicken and egg situation.

If I have this mixed up don't mind this comment, otherwise this comment can be removed I think.

cafca (changes requested):
I have one question here, otherwise this is good to go. I'm sorry about taking so long ... it's year end time.

p2panda-rs/src/materialiser/marshall_entries.rs
sandreae: Yes, I think that line may have been left over after some code was removed, not sure what it's alluding to... Thanks! Removed it.

Issue #138, opened 2021-11-30
Replace YAMF hash with YASMF hash


cafca:
YASMF brings us Blake3 hashes.

# Requirements

- Replace `bamboo-rs-core` with [`bamboo-rs-core-ed25519-yasmf`](https://crates.io/crates/bamboo-rs-core-ed25519-yasmf)
- Replace `yamf-hash` with [`yasmf-hash`](https://crates.io/crates/yasmf-hash)
- Update documentation

adzialocha:
Related: https://github.com/p2panda/p2panda/issues/19

PR #139, opened 2021-11-30
Adopt YASMF

### LINKED ISSUE #19: Use BLAKE3 for entry and message hashing
Use 256bit BLAKE3.

Requires bamboo-core to allow arbitrary hashing implementations. 

Related: https://github.com/pietgeursen/bamboo-rs/issues/17

cafca:
Adopt `yasmf-hash` and the corresponding `bamboo-rs` variant `bamboo-rs-ed25519-yasmf` (what a mouth full) to switch to 256 bit Blake3 hashes.

- Increases message version to 2 (where do we record this?)
  - _or maybe it doesn't?_
  - It doesn't, issue is at https://github.com/p2panda/p2panda/issues/151

Closes #19 #138 

## Todo

- [x] Fix error in `entry_signed.rs`
- [x] Fix rs tests
- [x] Fix js tests
- [x] Update OpenRPC schema

## 📋 Checklist


p2panda-rs/Cargo.toml
```diff
@@ -44,7 +44,7 @@ serde_cbor = "0.11.2"
 serde_repr = "0.1.7"
 serde_json = "1.0.59"
 thiserror = "1.0.29"
-yamf-hash = "0.1.2"
+yasmf-hash = "0.1.0"
```
cafca, line 47: Should `yasmf-hash` version numbers track `yamf-hash` version numbers?

adzialocha:
> Increases message version to 2 (where do we record this?)

Don't think we need to do this before any official protocol release v1 :+1:

p2panda-rs/Cargo.toml
adzialocha: I can imagine `yasmf-hash` might add hashes in a different rhythm than `yamf-hash`, so maybe not?

adzialocha:
> Currently getting this error still https://github.com/p2panda/p2panda/runs/4368351011?check_suite_focus=true#step:4:287

The byte size got shorter now, from 64 to 32 bytes, so I guess the array lengths need to be adjusted :+1:

cafca:
> The byte size got shorter now, from 64 to 32 bytes, so I guess the array lengths need to be adjusted 👍

I knew I needed to adjust the array lengths but didn't remember which was which and then my IDE was driving me nuts :'D But I remembered now! I added constants for the hardcoded values and also published them so we clearly communicate what's going on here.

> Don't think we need to do this before any official protocol release v1 👍, we're not breaking anyones apps yet and everything is WIP anyhow 👶

I think it's good to already start using all our versioning mechanisms before v1. For one, I want to be able to use p2panda as soon as possible to really test it early. And then, it's important for us to see how e.g. version mismatches play out in the developer/debugging experience and if we don't use versioning now we can't adjust that. Changing the message version to 2 doesn't have any downsides as far as I can see.

adzialocha:
> I think it's good to already start using all our versioning mechanisms before v1. For one, I want to be able to use p2panda as soon as possible to really test it early. And then, it's important for us to see how e.g. version mismatches play out in the developer/debugging experience and if we don't use versioning now we can't adjust that. Changing the message version to 2 doesn't have any downsides as far as I can see.

Good points! On the other hand I think it might prematurely add complexity like backwards compatibility and checks which we currently wouldn't need to care about. I understand that it might be a good training for us to start early, but currently it really is very very early, maybe we can start doing it as soon as we specified the protocol at least? :-D

cafca:
Oh I didn't want to add backwards compatibility. I just wanted to let existing messages break at the version number instead of the hash decoding.

adzialocha:
> Oh I didn't want to add backwards compatibility. I just wanted to let existing messages break at the version number instead of the hash decoding.

This might signal future developers (for working on this or other protocol ports) that there are breaking changes. I would at least wonder when working with p2panda why this is a higher version number and I would try to solve the other cases as well when implementing a protocol as it looks like we made a  decision to bump something. Or in other words: i think we cant just bump something without either bringing in backwards compatibility or explaining somewhere clearly why p2panda version 1 has a message version 7 and there is no need to care about the former ones. I still think we're easier off with just not doing anything 🙂

cafca:
I think I got your point now! This makes sense, let me think about it some more. This whole thing sounds like it has a big _semantic versioning_ nametag attached to it but maybe we don't need to go there.

adzialocha:
> I think I got your point now! This makes sense, let me think about it some more. This whole thing sounds like it has a big _semantic versioning_ nametag attached to it but maybe we don't need to go there.

Yeah, semantic versioning got around that problem, so far I interpreted our protocol version numbers as simple incrementing numbers which indicate that something new happened and you will not be able to support it without implementation.

p2panda-rs/src/hash/hash.rs
```diff
@@ -10,8 +10,11 @@ use yasmf_hash::{YasmfHash, BLAKE3_HASH_SIZE, MAX_YAMF_HASH_SIZE};
 use crate::hash::HashError;
 use crate::Validate;
 
+/// This is the size of p2panda entries' hashes
+pub const HASH_SIZE: usize = BLAKE3_HASH_SIZE;
```
cafca, line 14: Should we publish `HASH_SIZE` to let library consumers gauge the likelihood of collisions or is this an internal that they don't need to know about?

cafca: Similar question for `SIGNATURE_SIZE`.

adzialocha: Blake3 claims to be resistent to collision attacks, so hopefully that will not happen :-D also signatures will not collide, otherwise ed25519 would be badly broken.

adzialocha: On another note: I really don't like this API where we have to keep track of the hash size of something which should be as generic as `yamf-hash` .. but I guess thats more an API issue with the crate than with us.

cafca: Yes, I agree. I will un-publish the constants but I would rather have constants than just lots of `32` and `64` all over the code 😅

cafca:
> i think we cant just bump something without either bringing in backwards compatibility or explaining somewhere clearly why p2panda version 1 has a message version 7 and there is no need to care about the former ones.

We should communicate clearly which operation version numbers are supported by a given library version anyway. I think we don't need to feel obligated to support all version numbers ever used, i.e. it's okay to publish p2panda v1 and say "there are entries flying around the internet that were created during p2panda development with earlier version numbers. p2panda v1 supports operations with version >= x".

I see big advantages to being able to use p2panda for real before a v1 release and it would be very helpful to be able to use operation versioning to make that a pleasant experience, even while we are tinkering with the protocol wildly. In my eyes the current situation where `p2panda-js` is not compatible with the Aquadoggo in the Wolke - without a helpful error message - is bad already and I really would prefer if we didn't have any more of that :)

adzialocha:
> I see big advantages to being able to use p2panda for real before a v1 release and it would be very helpful to be able to use operation versioning to make that a pleasant experience, even while we are tinkering with the protocol wildly. In my eyes the current situation where p2panda-js is not compatible with the Aquadoggo in the Wolke - without a helpful error message - is bad already and I really would prefer if we didn't have any more of that :)

I understand that point, it really makes sense but still I don't think this current approach covers the actual situation well ..

First of all: We would bump the version of `Operation` here, but actually no change has been made to operations? An Bamboo Entry will already fail during validation before we even start looking at its payload, simply because the Hash scheme got changed. How is this versioned? Not at all right now, and bumping Operations won't help here ..

You're mentioning the compatibility of Aquadoggo and p2panda-js, that was because of an JSON RPC change, again, something where changing `Operation` version would not cover and account for.

I think we could start thinking more thoroughly on this topic and would detect dozens of places where we actually have moving parts which would not be covered by versioning, or potentially can be with some work to start introducing it. My point is more: Its complicated and likely to not be solved by just bumping the operation version, because its literally not solving what you're trying to solve ..

Also, I'm still not a fan of introducing such complexity at this early stage, or worrying about it when we don't have to. I think even if we would start taking care of it we will have unforseen drastic changes anyway (replace JSON RPC with GraphQL for example) so I'm afraid we will put loads of thinking and work into every PR were doing to account for these things without it actually really having any real effects.

On top that I find it still confusing to release something with a high version number in different parts of the stack without it ever being a stable version. Maybe we reset it back to `1` before release?

adzialocha:
Or in other words: can we not bump anything in this PR and make an issue for versioning and then think it through? 😅

sandreae:
> We would bump the version of Operation here, but actually no change has been made to operations? An Bamboo Entry will already fail during validation before we even start looking at its payload, simply because the Hash scheme got changed. How is this versioned? Not at all right now, and bumping Operations won't help here ..

This was the thing I was thinking while reading through the discussion. I agree it would be good to have a conversation about how we want to use versioning (and when) but it seems outside the scope of this PR, or at least, not *only* relevant to this PR.

p2panda-rs/src/message/message_encoded.rs
```diff
@@ -110,18 +110,19 @@ mod tests {
         assert!(MessageEncoded::new("68656c6c6f2062616d626f6f21").is_err());
 
         // Valid `delete` message
-        assert!(MessageEncoded::new("a466616374696f6e6664656c65746566736368656d6178843030343032646332356433326466623430306262323935623636336434373036626334376630636234663165646666323737633733376166633861393233323333306165393838346663663664303231343161373835633566643832633139366239373365383432376566633063303464303434346463633330353932323062396564616776657273696f6e016269647884303034306366393466366436303536353765393063353433623063393139303730636461616637323039633565316561353861636238663335363866613231313432363864633961633362616665313261663237376432383666636537646335396237633063333438393733633465396461636265373934383565353661633261373032").is_ok());
+        assert!(MessageEncoded::new("a466616374696f6e6663726561746566736368656d61784430303230623137376563316266323664666233623730313064343733653664343437313362323962373635623939633665363065636266616537343264653439363534336776657273696f6e02666669656c6473a168757365726e616d65a26474797065637374726576616c75656462756275").is_ok());
```
sandreae, line 113: So short!!! :heart:

adzialocha: .. wait until we have encrypted fields with padding inside :sob: :rofl:

p2panda-rs/src/test_utils/mocks/node/utils.rs
```diff
@@ -3,16 +3,41 @@
...
+        let hash_value = Hash::new_from_bytes(vec![0, 0, 2]).unwrap();
+        assert_eq!(hash_value.as_str(), KEY_PACKAGE_SCHEMA_HASH);
+
+        let hash_value = Hash::new_from_bytes(vec![0, 0, 3]).unwrap();
+        assert_eq!(hash_value.as_str(), GROUP_SCHEMA_HASH);
+
+        let hash_value = Hash::new_from_bytes(vec![0, 0, 4]).unwrap();
+        assert_eq!(hash_value.as_str(), PERMISSIONS_SCHEMA_HASH);
+    }
+}
```
sandreae, line 43: Thanks for these tests and in general considering all the hard coded value in test_utils. I do think we can find a better approach...

sandreae (commented):
No changes requested, great to have the smol hashes implemented, thanks!

sandreae (approved):

p2panda-rs/src/message/message_encoded.rs
sandreae: booooooooo!

p2panda-rs/src/test_utils/fixtures/fixtures.rs
```diff
@@ -118,9 +118,9 @@ pub fn delete_message(schema: Hash, #[from(hash)] instance_id: Hash) -> Message
     utils::delete_message(schema, instance_id)
 }
 
-/// Fixture which injects versioned p2panda testing data into a test method.
+/// Fixture which injects p2panda testing data from p2panda version 0.3.0
 #[fixture]
-pub fn v0_2_0_fixture() -> Fixture {
+pub fn v0_3_0_fixture() -> Fixture {
```
cafca, line 123: This version is not released yet but calling that fixture `v0_2_1_fixture` would also be wrong because 0.2.1 doesn't have YASMF hashes.

sandreae: Yes, seems like a good solution.

PR #140, opened 2021-12-05
Access secret group member list

cafca:
Adds a new method `members` on secret groups that returns a vector of the group members as p2panda `Author` instances.

This will fail with a `SecretGroupError` if the MLS group has members which are not using ED25519 public keys.

## 📋 Checklist

~~- [ ] Add this PR to the _Unreleased_ section in `CHANGELOG.md`~~
~~- [ ] Link this PR to any issues it closes~~
~~- [ ] New files contain a SPDX license header~~


adzialocha (approved):
Very cool! Thank you! 😍

PR #154, opened 2021-12-13
Remove distinction of system and application logs

adzialocha:
Related: https://github.com/p2panda/handbook/issues/49

## 📋 Checklist


sandreae:
Did we decide to do this? ;-p

It does make sense (we'll have many more app schemas than system schemas, and so many empty ids).

adzialocha:
> Did we decide to do this? ;-p

Not directly maybe 🤔 somehow I thought there was a consensus but now when you ask ..

Can't think of any reason to keep it right now, especially with log per document we get much more deliberate with log ids.

sandreae:
Yeh, I agree :+1:

sandreae (approved):

cafca:
I'm also in favor of this change! But let me track back to why we introduced this system in the first place. If I remember correctly we wanted to make replication easier by having predictable log ids for some important system schemas. Then we could just request, e.g. log id 4 and we would know that this would always contain the key's group memberships. 

However, now I think that we loose nothing by instead implementing a replication mechanism that allows us to ask a remote node for all logs with documents of schema `group-membership-v1` or so and get the same result.

On the flip side we gain a reduction in complexity by treating all log ids the same, and that's nice. Also log ids will not grow as fast if we can use all of them.

PR #156, opened 2021-12-15
Use new `Operation` naming everywhere

sandreae:
We renamed "message" to "operation". Make this change in `p2panda-rs` and `p2panda-js`.

closes https://github.com/p2panda/p2panda/issues/134

## 📋 Checklist


adzialocha (changes requested):
Nice nice! I think it should still have a changelog entry though (actually literally every PR) 👍

Also, all of this smells very much like breaking changes? Probably something to note in the `CHANGELOG.md` as well :grin:

sandreae:
Yes indeed. It breaks EVERYTHING.

sandreae:
Do we have a convention for marking breaking changes in the CHANGELOG additions?

sandreae:
I have this `aquadoggo` PR in progress https://github.com/p2panda/aquadoggo/pull/48 but I think there might be other breaking changes in the mix too...

adzialocha:
> Do we have a convention for marking breaking changes in the CHANGELOG additions?

Not really yet, think for now it's enough to write it somewhere next to it, eventually when we do our first stable release we should have an own section for it.

adzialocha (approved):

cafca (approved):
I'm not going to read through all of this. But I'm really glad it's here! More clarity and niceness. Thanks @sandreae.

Btw, Some malkovich malkovich going on here:

```
test src/operation/operation.rs - operation::operation::Operation::new_create (line 226) ... ok
```

Issue #135, opened 2021-11-20
Expand `Operation` to contain `previous_operation`


sandreae:
And also maybe `instance_author`.

It should look something like this:

```
{
    action: "CREATE” | “UPDATE" | "DELETE"
    schema: <HASH>
    instance_id: <HASH>
    instance_author: <PUBLIC_KEY>
    ? previousOp: <HASH>
    ? fields: [{
        "<FIELD KEY>": <VALUE> 
        "<FIELD VALUE>": <VALUE>
    }]
}
```

Spec being drafted [here](https://laub.liebechaos.org/7cGHWPCcR8eoZBkk-X9R4w)

cafca:
I propose calling the field `previous_operation` to stick to snake case and be easy to read.

adzialocha:
What is `instance_author` doing?

sandreae:
Now we're aiming for this:

```
{
    action: "CREATE” | “UPDATE" | "DELETE"
    schema: <HASH>
    version: <NUMBER>
    ? previous_operations: <HASH>
    ? fields: [{
        "<FIELD KEY>": <VALUE> 
        "<FIELD VALUE>": <VALUE>
    }]
}
```

adzialocha:
Guess this happened during copy paste but is `previous_operations` really only a hash?

sandreae:
> Guess this happened during copy paste but is `previous_operations` really only a hash?

corrected, thanks!

PR #163, opened 2021-12-16
Update `Operation` and introduce `OperationWithMeta`

sandreae:
## `Operation` updates and new `OperationWithMeta` struct 

Introduce `previousOperations` field on `Operation` struct. This field will be used in the future to construct the causal graph of operations for each document. It is out of the scope of this PR to implement this logic fully, as it is now we would still be populating this field as if we were in a single writer situation (using the entry backlink). The `nextEntryArgs` API will need to be expanded to return the `previousOperations` for the document we are updating for this to be implemented. And before that we need to be able to build and reduce a graph of operations. So there are still a few dependencies unmet.... 

Updates required as a result of this in `wasm` and `js` code.

New `OperationWithMeta` wrapper struct for `Operation` which will be used during graph building and traversal. As part of introducing this shared `Operation` functionality has been abstracted into an `AsOperation` trait which both `OperationWithMeta` and `Operation` implement.

- [x] Refactor shared methods into `AsOperation` trait
- [x] Confirm reasoning for `id` field on `Operation` _UPDATE: (we will remove this in a later PR so pretend it isn't there)_
- [x] Introduce `previous_operations` field
- [x] introduce new struct `OperationWithMeta` 
- [x] update docs `js` + `rs` 

closes https://github.com/p2panda/p2panda/issues/135

## 📋 Checklist

- [x] Do descriptions and terminology match `handbook` content (and visa-versa)


p2panda-rs/Cargo.toml
```diff
@@ -34,6 +34,7 @@ crate-type = ["cdylib", "rlib"]
 
 [dependencies]
 arrayvec = "0.5.2"
+aquamarine = "0.1.11"
```
adzialocha, line 37: Isn't this a dev dependency?

p2panda-rs/src/wasm.rs
```diff
@@ -9,18 +9,18 @@
...
 use crate::identity::KeyPair as KeyPairNonWasm;
 use crate::operation::{
     Operation, OperationEncoded, OperationFields as OperationFieldsNonWasm, OperationValue,
 };
+use console_error_panic_hook::hook as panic_hook;
+use ed25519_dalek::{PublicKey, Signature};
+use js_sys::Array;
+use serde::{Deserialize, Serialize};
+use wasm_bindgen::prelude::wasm_bindgen;
+use wasm_bindgen::JsValue;
```
adzialocha, line 23: Think these ones belong back to before `crate`-level imports :smiley_cat:

p2panda-rs/Cargo.toml
sandreae: indeed!

p2panda-rs/src/wasm.rs
sandreae: :face_palm:

adzialocha (approved):
Exciting! Thank you for this (and taking care of the merge conflicts ..)!

PR #165, opened 2021-12-17
Update API to reflect document, hash format and JSON RPC changes

adzialocha:
* Rename *instance* to document, except of when its really an *instance* (for example after materialization)
* Update to latest aquadoggo JSON RPC API (has a `document` field now)
* Compatibility with organization of logs per document
* Update mocked node to follow new log organization as well

Related node changes:
* https://github.com/p2panda/aquadoggo/pull/47
* https://github.com/p2panda/aquadoggo/pull/48
* https://github.com/p2panda/aquadoggo/pull/44

## 📋 Checklist

- [x] Do descriptions and terminology match `handbook` content (and visa-versa)


sandreae:
:broom: :heart:

sandreae:
Hey! Just having a look at this now too. It's going to be a massive pain to re-write all the materialisation logic here to accommodate document logs. Also, it's a bit of a thankless task as it's all mostly a POC designed for the playground and mocking. We're quite far down the materialisation route in the main library, so working too hard to get this working now seems pointless. I'd suggest:

1) strip out all materialisation & document query logic from `test_utils` (it's not used anywhere for testing yet anyway)
2) import DAG logic from main lib as it is merged there now :sparkles: 
3) implement this _only_ in `getNextEntrArgs` in order to return the tip of a document (even this isn't really necessary as we can strip out multi-writer completely... but it's easy so worth doing)

This will mean we can easily make the transition to document logs without worrying about extra complications.

Then when we have everything ready in the main lib we can implement proper materialisation logic in the mocks :+1: 

I actually had a go at making these changes just now and it was quite simple. If you like it I can push a commit.

sandreae:
By "strip out materialisation logic" I mean:

1) delete `materialisation` module
2) remove query methods from `node`

adzialocha:
Sounds all like really good suggestions 👍 thank you for looking into this!

I'm also fine with not deleting the module, that can be part of the materializer v2 PR.

sandreae:
There we go, think we're back on track. Still some errors in the `js` tests but I think it's just because the test data is now slightly different from before (only one document log). We can expand it to support testing multiple document logs as things stabilise.

> I'm also fine with not deleting the module, even when it's not used like that anymore, that can be part of the materializer v2 PR.

It's cool, it was quite ugly and very much a learning thing for me at the time. The work in the next round of PRs replaces all that logic (much more nicely I hope....).

p2panda-js/src/session/session.test.ts
```diff
@@ -104,8 +104,8 @@ describe('Session', () => {
         encodedEntryFixture(4).entryBytes,
         encodedEntryFixture(4).payloadBytes,
       );
-      expect(nextEntryArgs.entryHashBacklink).toEqual(
-        entryArgsFixture(5).entryHashBacklink,
+      expect(nextEntryArgs.backlink).toEqual(
```
adzialocha, line 105: Something must have went wrong here, since the JSON RPC method awaits still `entryHashBacklink` and `entryHashSkiplink` params?

sandreae: Ah, right! Oops, I'll revert.

sandreae: reverted.

p2panda-js/src/session/session.ts
```diff
@@ -254,89 +275,109 @@ export class Session {
...
     return this;
   }
 
   /**
-   * Signs and publishes an `update` entry for the given user data and matching schema.
-   * An `update` entry references the entry hash of the `create` entry which is the root
-   * of this materialized instance.
+   * Signs and publishes an `update` entry for the given application data and
+   * matching schema. An `update` entry references the entry hash of the
+   * `create` entry which is the root of this materialized document.
```
sandreae, line 296: Should this just be:

"which is the root of this document"

p2panda-js/src/session/session.ts
```diff
@@ -254,89 +275,109 @@ export class Session {
...
+    updateDocument(documentId, fields, mergedOptions);
+
     return this;
   }
 
   /**
-   * Signs and publishes a `delete` entry for the given schema. References the entry hash of the `create` entry which
-   * is the id of this materialized instance.
+   * Signs and publishes a `delete` entry for the given schema. References the
+   * entry hash of the `create` entry which is the id of this materialized
```
sandreae, line 342: Same here (lose the "materialized"), unless I'm wrong on both counts ;-p

p2panda-rs/src/test_utils/mocks/logs.rs
```diff
@@ -2,47 +2,54 @@
...
+/// in testing and materialising instances.
 #[derive(Clone, Debug)]
 pub struct LogEntry {
-    /// The author of this entry
+    /// The author of this entry.
     pub author: Author,
-    /// The author of the instance this entry is part of
+
+    /// The author of the instance this entry is part of.
     pub instance_author: Option<String>,
```
sandreae, line 23: I think all these "instances" in `logs.rs` need to be "documents"

sandreae (commented):
This is great, thanks for doing such a naaasty job. The new language of `operation` and `document` reads really well.

I picked up a few points, but not much, I don't think they're very contentious so can do a commit myself if you like.

sandreae (approved):
niiice

Issue #174, opened 2021-12-30
Support u64


adzialocha:
BigInt support is given now in modern Safari and in SQL we can use strings (we don't do much with these values in the database).

Issue #65, opened 2021-05-13
Make runtime configurable when choosing `db-sqlx` flag


adzialocha:
Using `sqlx` attributes in `p2panda-rs` introduces an async runtime. `p2panda-rs` should be agnostic to them or at least give some options (`async-std`, `tokio`, `actix`)

PR #177, opened 2021-12-30
Remove `sqlx`, support u64 bit integers

adzialocha:
## Description

* Make all sequence numbers and log ids `u64` according to the Bamboo specification
* Re-introduce `BigInt` for the `wasm` interface for `seq_num` and `log_id`, changed APIs accordingly
* Add conversion traits for  `SeqNum`, `LogId` from `&str` and `String` to give a nice API for `aquadoggo` and others to work with string values for large numbers
* Fix issues where large numbers for operations and entries would break in `p2panda-js`, changed APIs accordingly and added tests
* Note: Adding `u64` and `i64` support is easy in the Rust world and CBOR but in JavaScript and JSON it gets complicated, this is why most of the changes here are actually towards supporting large numbers in WASM / JavaScript
* Note: I've made according changes in `beep-boop` and `aquadoggo` to support large integers. See related issues!

**Bonus level**

* Remove `sqlx` from `p2panda-rs`. Since we added `u64` support we put a little bit more work on the backends anyways and it's always been a little ugly to have a notion of databases inside this crate
* Support "float" operation field type in `p2panda-js`
* Add conversion traits from `&str` and `String` to `Hash`

Closes: #174 and #65
Related issues: https://github.com/p2panda/aquadoggo/pull/54 and https://github.com/p2panda/beep-boop/pull/32

## API changes

**Entries**

* Next entry arguments like `seq_num` and `log_id` are received as strings from the JSON RPC endpoint (will be implemented in https://github.com/p2panda/aquadoggo/pull/54). We can't send numbers since they will break for large integers in JSON. Also we can't send `BigInt` instances since they are not supported by JSON. Strings are the only option.
* `signEncodeEntry` takes `BigInt` instances for `seq_num` and `log_id`
* `decodeEntry` returns also `BigInt` instances for them

**Operations**

* Operation fields of type "int" are stored internally as i64 integers but [JS only supports max 53 bits for numbers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER#description). Adding an integer to the operation fields is done with using a string value since I could not find a way to convert from `JsValue(BigInt)` to `i64` easily. This is for the low-level wasm API.
* For the high-level JS API one can add numbers or `BigInt` instances which are then automatically converted into strings.
* When getting an integer from the operation fields one receives a `BigInt` instance for the low-level wasm API, for the high-level JS API one receives a `BigInt` or number when it is below `Number.MAX_SAFE_INTEGER`.

## 📋 Checklist


p2panda-rs/src/wasm/serde.rs
```diff
@@ -0,0 +1,27 @@
...
+use wasm_bindgen::JsValue;
+
+/// Serializes Rust type into JavaScript value (`JsValue`).
+///
+/// Note that this will NOT serialize into a JSON string but an actual JavaScript object
+/// (potentially with non-JSON values inside like "Map" or "BigInt").
+///
+/// This method uses the `serde_wasm_bindgen` serializer, both for bundle size but also for its
+/// support of large number types. With this serializer all i64 and u64 numbers get serialized as
+/// BigInt instances.
```
sandreae, line 15: Really useful and well worded description :+1:

p2panda-rs/src/wasm/tests/operation.rs
```diff
@@ -51,48 +49,78 @@ fn operation_fields() {
...
+    );
+
+    // This integer is too large and can't be represented as i64
+    let result = fields.add(
+        "really_big_number".to_string(),
+        "int".to_string(),
+        JsValue::from_str("932187321673219932187732188"),
+    );
     assert!(result.is_err());
 }
```
sandreae, line 117: Just so I'm clear, would this work if `JsValue::bigint_from_str("932187321673219932187732188")` was passed as the value instead?

p2panda-rs/src/wasm/tests/operation.rs
```diff
@@ -51,48 +49,78 @@ fn operation_fields() {
...
+    fields
+        .add("age".to_string(), "int".to_string(), JsValue::from_str("5"))
+        .unwrap();
+
+    // "int" fields always get returned as BigInt instances
+    assert_eq!(
+        fields.get("age".to_string()).unwrap(),
+        JsValue::bigint_from_str("5")
+    );
+}
```
sandreae, line 91: From reading the PR description I understood int values smaller than `Number.MAX_SAFE_INTEGER` were returned as numbers, not `BigInt`. Looking here that doesn't seem to be right though.

sandreae (commented):
This all looks good to me, well done wrapping your head round it all. It's a bit dizzying trying to understand where and how the js number limitations impact the wasm methods.

I have a couple of questios, but it may just be for my own understanding. 

Thanks!

p2panda-rs/src/wasm/tests/operation.rs
adzialocha: No, the `add` method only accepts strings for the "int" type. I've tried to use BigInts but I really can't see how to convert it into an i64 except of converting it to a string first and then to an integer which is weird, maybe you have an idea?

p2panda-rs/src/wasm/tests/operation.rs
adzialocha: Only for the "high-level" API meaning `p2panda-js`. In WASM (aka "low-level" API) its BigInts :+1:

adzialocha:
I see some bugs in the TypeScript types, will turn this into a draft again, but happy to already receive feedback!

p2panda-js/test/fixtures.ts
```diff
@@ -1,25 +1,51 @@
...
+        ] as OperationValue;
+        fields.set(key, value);
+      });
+
+      operation.fields = fields;
+    }
+
+    return operation;
+  },
+);
```
adzialocha, line 41: I have no clue on how to do this in TypeScript (convert from JSON into different type), might need some help here :cry: 

@cafca

adzialocha:
Ok, this should be ready for review now, thanks for the help and feedback so far! This should also be compatible to the regarding changes in `aquadoggo` and `beep-boop` (basically bringing it all to the same level), so please have a look there as well!

p2panda-js/src/types.ts
```diff
@@ -34,11 +34,11 @@ export type EntryRecord = Entry & {
  * Decoded entry containing optional `Operation`.
  */
 export type Entry = {
-  entryHashBacklink: string | null;
-  entryHashSkiplink: string | null;
-  logId: number;
-  operation: Operation | null;
-  seqNum: number;
+  entryHashBacklink: string | undefined;
```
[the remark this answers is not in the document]
cafca: I don't have an answer though

cafca (approved):

Issue #32, opened 2021-02-01
Implement `std::hash::Hash` for entry and operation


cafca:
There's some possible alternatives, see discussion [here](https://github.com/p2panda/p2panda/pull/23#discussion_r572798384)

sandreae:
Bumping this issue. I'll try by implementing the `Hash` trait for our own hash struct and see if this can then be returned for similar implementations on `Operation` and `EntrySigned`. 

A different approach would be needed to for `Entry`, but let's see.

PR #178, opened 2022-01-05
Implement `Hash`, `Eq` and `PartialEq` traits for several core data types

sandreae:
Implement the `Hash`, `Eq` and `PartialEq` traits for ...

- [x] `Hash`
- [x] `Operation`
- [x] `OperationWithMeta`
- [x] `OperationEncoded`
- [x] `EntrySigned`
- [x] `Author`
- [x] `SeqNum`
- [x] `LogId`

Closes: https://github.com/p2panda/p2panda/issues/97, https://github.com/p2panda/p2panda/issues/32

## 📋 Checklist


adzialocha:
Nice youre tackling this old issue! Will be very helpful to use our types in hash maps 👍

Intuitively I would expect any hashing method of an operation or entry to return yasmf + blake3 hashes and not different ones depending from which angle I come (think Rust uses SipHash by default?)

I can add some tests in the next days to maybe sketch an potential API as a suggestion?

sandreae:
Yes, sounds good. Are you thinking to implement a custom `Hasher` struct (with our own hash fn)?

adzialocha:
> Yes, sounds good. Are you thinking to implement a custom `Hasher` struct (with our own hash fn)?

Yes, exactly. When I run `Operation.hash()` etc. I expect the p2panda hash

p2panda-rs/src/operation/operation.rs
```diff
@@ -452,6 +453,20 @@ impl From<&OperationEncoded> for Operation {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.id == other.id
+    }
+}
+
+impl Eq for Operation {}
+
+impl StdHash for Operation {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.id.hash(state);
```
adzialocha, line 466: `id` is sometimes `None`, right? This would lead to the same hash for all CREATE operations.

adzialocha (changes requested):
> Yes, sounds good. Are you thinking to implement a custom `Hasher` struct (with our own hash fn)?

Yes, exactly. When I run `Operation.hash()` etc. I expect the p2panda hash over the specified values.

p2panda-rs/src/operation/operation.rs
adzialocha: Also per specification the hash of an Operation is not only on the `id` but the whole payload.

sandreae: ugh, sorry, this was an id confusion brain fart on my part (I actually already started some work to remove the deprecated `id`... how did I do this then?? :rofl:)

sandreae:
Improvement for now until we have a custom hasher....

adzialocha:
I was thinking about this a little bit more and it looks a little bit like we should maybe have a small jam / meeting on this whole hashing question? Mainly because I wonder if we have different needs with "hashes", some are for p2panda data types, some others simply just to make use of the Rust HashMap. It gets a little bit crazy especially when we start to implement a `Hash` trait for the `Hash` struct :exploding_head:

sandreae:
Yes, I agree that it gets a bit funky to have these overlapping hash implementations/uses. I'm happy to wait until we can discuss it more/have a jam and see what we want from it.

p2panda-rs/src/operation/operation.rs
```diff
@@ -455,6 +456,20 @@ impl From<&OperationEncoded> for Operation {
     }
 }
 
+impl PartialEq for Operation {
```
cafca, line 459: I think this is an important difference between `Operation` and `OperationWithMeta` that should be highlighted in their descriptions. Would you mind if I added something there?

adzialocha:
Was thinking about this whole Hashing topic a little bit more, I can share some thoughts here though they might not be too new maybe.

* We have a `Hash` struct which is simply just a wrapper around a `String`. Still, it helps us to validate that it is a valid hash and gives type safety, it also makes the code readable and assures its actually a hash arriving where we expect one to be. Aka, its nice.
* `EntrySigned` and maybe some other structs I can think of have a `hash` method which returns a `Hash` instance. This is the way currently to get the hash of an entry according to the p2panda specification, meaning: It returns an hex encoded, Blake3 hash wrapped in a YASMF container format
* There is a `Hash` trait in Rust from the std library. It is a trait allowing us to tell a method *what* to hash. So it does'nt hash by itself, its simply just returning the correct fields. In p2panda we have this clearly specified what values get hashed
* There is a `Hasher` trait one can implement in Rust from the std library. It is a trait allowing us to write something which will take a "hashable" struct which implemented the `Hash` trait to then actually run a hashing algorithm on top of the data it gave us
* "Just" implementing the `Hash` trait for a couple of structs raises some questions: Why are we doing it? What are the semantics of this? In Rust I would assume implementing the Hash trait is a) offering an API for others to hash the correct fields according to our specification *if* they want to hash something b) `HashMap` and other structs require implementing `Hash` to use them as keys. The point a) we already have covered by our own `Hash` struct (its confusing, right? Because its named exactly like the trait ..) and on top we even assure it is a Blake3 + YASMF hash which the `Hash` trait itself does not take care of
* A `Hasher` implementation does not return something of type `Hash` as we have it right now in p2panda but simply just bytes, we will loose the types here we wanted to have
* The largest issue I see currently: Implementing the `Hash` trait wants use to define a `hash` method, but it already exists for `EntrySigned` .. what happens here? Does it get overwritten? Is that even allowed? Also, it has very different outcomes: `hash` from the trait returns only the hashable fields which then are consumed by a `Hasher`. `hash` in p2panda returns an actual hash we can use

My suggestion:

1. Remove all `hash` methods from structs
2. Implement `Hash` traits for them with the correct fields according to p2panda spec
3. Implement `Hasher` returning YASMF Blake3 hashes
4. Implement a `yasmf_blake3_hash` helper method which takes in any struct which implements the `Hash` trait, hashes it with our new `Hasher`, encodes the resulting bytes as hex and wraps it all in a `Hash` struct instance

p2panda-rs/src/operation/operation.rs
adzialocha: You mean in the `PartialEq` implementation or in general for the structs? If its the latter you might be faster with making a new PR for the comments :smiley_cat:

sandreae:
Thanks for this overview @adzialocha, I agree with the suggestion. Having our own `hash()` method could definitely cause confusion in the future if people are expecting familiar rust traits.

Just to check, the only thing our helper `yasm_blake3_hash` would do on top of our implementation of the `Hash` trait (different from the yasmf/blake3 `Hasher`) would be to return a p2panda `Hash` type. Is that right?

adzialocha:
> Thanks for this overview @adzialocha, I agree with the suggestion. Having our own `hash()` method could definitely cause confusion in the future if people are expecting familiar rust traits.
> 
> Just to check, the only thing our helper `yasm_blake3_hash` would do on top of our implementation of the `Hash` trait (different from the yasmf/blake3 `Hasher`) would be to return a p2panda `Hash` type. Is that right?

Yes, right! 👍

adzialocha:
Hui! I've learned something about `Hasher` traits and don't think we can actually use it for our usecase.

Apparently the result of a `Hasher` in Rust is always a `u64` number: https://doc.rust-lang.org/stable/std/hash/trait.Hasher.html but our YASMF hashes need much larger numbers (256bit).

I guess this changes the plan then and moves all back to square 1:

We keep our own `hash` methods to generate YASMF hashes (wrapped in our `Hash` struct). We mostly just derive `Eq`, `PartialEq` and `std::hash::Hash` traits to be compatible with `HashMap` but ignore them for anything else, since they don't seem to be too powerful?

Also, this is an interesting read on duplicate method names when using traits: https://doc.rust-lang.org/stable/rust-by-example/trait/disambiguating.html

p2panda-rs/src/operation/operation.rs
sandreae: I think it would be better to have the same PartialEq implementation for both `Operation` and `OperationWithMeta`, or..?

sandreae (commented):
Thanks for getting to the bottom of this @adzialocha. Shame we can't customise the `Hasher` as we like... 

One though, maybe we should name all our own yasmf hash methods to `yasmf_hash()`, just to avoid any confusion with the Hash trait methods. I think this is actually only in `EntrySigned` where it needs renaming.

p2panda-rs/src/operation/operation.rs
cafca: Why do you think so? Which one should it be?

sandreae: Aha, we just chatted, two `OperationWithMeta` with the same field values but different authors are not equal, whereas two `Operation` with the same field values are always equal, so the `PartialEq` should be obvs different. Nice :+1:

PR #181, opened 2022-01-06
Updates in `test_utils/mocks`

sandreae:
Some tidy up:
- make `Log` and `LogEntry` structs match `aquadoggo` implementation a little more
- update operation creation methods to explicitly require `document_id` param
- remove all left over multi-writer methods (we will re-implement all this once the core lib has the necessary functionality)
- remove uses of `operation.id()` as this is deprecated
- some refactoring

## 📋 Checklist

- [x] Add tests that cover your changes
- [x] Add this PR to the _Unreleased_ section in `CHANGELOG.md`
- [x] Link this PR to any issues it closes
- [x] New files contain a SPDX license header
- [x] Check if descriptions and terminology match `handbook` content (and visa-versa)

PR #182, opened 2022-01-06
Remove deprecated `id` field from `Operation`

sandreae:
Remove deprecated `id` field from `Operation` in `rs` and `js`.

Includes changes made in https://github.com/p2panda/p2panda/pull/181

## 📋 Checklist


sandreae:
This required quite an involved re-write of the materialisation logic in `p2panda-js`:

https://github.com/p2panda/p2panda/blob/2644d634e527e5cd39460c30a87a4f27fa7bcb51/p2panda-js/src/instance/materialiser.ts#L16-L91

sandreae:
It's not the most elegant thing I've ever seen, but seeing as we will be eventually exporting this functionality from the rust (`Document` etc..) then maybe it's ok for now?

adzialocha:
> It's not the most elegant thing I've ever seen, but seeing as we will be eventually exporting this functionality from the rust (`Document` etc..) then maybe it's ok for now?

Yes, soon we should be pulling everything from one source (materialization in p2panda-rs, p2panda-js and aquadoggo), so I think its fine! As soon as we merged the new materialization methods we can remove the materializer in `p2panda-js`.

adzialocha (approved):
Wow, very cool! :+1:

PR #184, opened 2022-01-07
Create `wasm` module with tests

sandreae:
- [x] Split existing `wasm.rs` file into it's own module.
- [x] Create tests using `wasm-bindgen-test`
- [x] Run tests in browser environments
- [x] Add tests to CI pipeline

## Next

- Would be nice to add wasm code to the test coverage report... https://github.com/p2panda/p2panda/issues/185
- Do we need to run these tests in nodejs environment too?

## 📋 Checklist


adzialocha (approved):
Looking great! Thank you!

I did some changes directly, thought that was easier as they were mostly minor:

* Moved all tests into the `wasm` module again, since they still feel more like unit / integration tests for only this module
* Changed a little bit of formatting

> Do we need to run these tests in nodejs environment too?

Interesting that one can't run both for the same tests .. it would be nice indeed but as long as it is not supported we can aim for browsers, its better than nothing I guess and for Node we have the `p2panda-js` tests.

sandreae:
Thanks, very nice improvements :+1:

sandreae:
I removed the `--node` arg in the wasm test action as it currently isn't doing anything.

sandreae:
If that's ok with you we can merge!

adzialocha (approved):

PR #186, opened 2022-01-09
Bump cddl to `0.9.0-beta.0`

sandreae:
Update cddl to `0.9.0-beta.0` because it should support `wasm` targets.

(unfortunately, atm, it breaks everything though :rofl:)

## 📋 Checklist


adzialocha:
> (unfortunately, atm, it breaks everything though rofl)

Hmn :cry: I've had a quick look at it and it seems `serde_cbor` is not happy with the CBOR being passed in throwing a `CBORParsing` error. I checked the CBOR data for the failing tests in the https://cbor.me/ playground, but it works there .. not sure what we can do from here, maybe it is too early.

Another thing: Is there a reason why `document.rs` is commited in this PR?

sandreae:
> Another thing: Is there a reason why document.rs is commited in this PR?

Ah! My bad, that shouldn't be there....

> maybe it is too early.

I am watching the CI tests pass in realtime, so maybe the recent changes fixed things :+1:

sandreae:
I may start a new PR from `main`.

I already had a go at bring cddl validation into the `wasm` target. It's not completely straight forward, and there are probably several different approaches. I think it's a good candidate for more code jamming :-)

sandreae:
Hmmmm.... I just had another go at this, never came to a good solution. @adzialocha I know you said you were having issues too, I'm not sure if we're on the same problem. Where I'm at is:

- now that `cddl` targets wasm, and we compile `validate_schema` against multiple targets, the returned error types don't match what we expect in, for example, `OperationEncoded::new()`
- we have `JsValue` errors where we expect our custom type errors
- I think this suggests we need to change how we're structuring the validation code (wrt exposing it to wasm)

So yeh, maybe this whole are needs a rethink, feel like I'm hacking what was already a hack :rofl:. Interested to hear your thoughts.

sandreae:
The error handling isn't too pretty though....

sandreae:
Haha, "make it work"

sandreae:
hmmmm, looks like we're getting test failures only in the node 12 build.

p2panda-rs/src/entry/error.rs
```diff
@@ -31,6 +31,10 @@ pub enum EntrySignedError {
     #[error("operation needs to match payload hash of encoded entry")]
     OperationHashMismatch,
 
+    /// Handle errors when encoding operation.
+    #[error("error occured when encoding operation")]
+    OperationEncodingError(String),
```
cafca, line 36: Suggest renaming this to something that's less similar to `OperationEncodedError`.

adzialocha:
> hmmmm, looks like we're getting test failures only in the node 12 build.

Ouch, it is `RangeError: Maximum call stack size exceeded` in the `cddl` library. Could be a temporary bug as it is beta or actually a Node version this crate does not support (couldn't find any notes) .. I'd be happy with not supporting Node 12 anymore though.

cafca:
Curious. I have never seen a case where maximum call stack size didn't mean *infinite recursion*.

adzialocha:
We can probably close this for now as things shifted around quite a lot, or what do you think @sam and @cafca?

sandreae:
Yes, I agree! The goalposts have moved :-p

PR #188, opened 2022-01-13
Provide proper prop-testing with `prop-tests` pro prop-testing package

sandreae:
Implement `prop-tests` tests for creating and encoding operations.

## 📋 Checklist


adzialocha (approved):
Cool. Looking forward to use this more in other tests.

And: Why did noone comment on my secret PR re-naming mission? 😎

sandreae:
> And: Why did noone comment on my secret PR re-naming mission? sunglasses

Hahahah, I forgot to comment!! When I first saw it, I thought that's what I actually named the PR, took me a while t realise what had been gooooing on. :rofl: :rofl:

p2panda-rs/src/operation/operation.rs
```diff
@@ -79,7 +81,8 @@ impl Copy for OperationVersion {}
 ///
 /// An action defines the operation format and if this operation creates, updates or deletes a data
 /// document.
-#[derive(Clone, Debug, Eq, PartialEq)]
```
cafca, line 82: You removed `Eq` here?

p2panda-rs/src/tests/prop_tests.rs
```diff
@@ -0,0 +1,26 @@
...
+        } else {
+            prop_assert!(operation.is_ok());
+
+            let operation_encoded = OperationEncoded::try_from(operation.as_ref().unwrap()).unwrap();
+            let decoded_operation = Operation::from(&operation_encoded);
+
+            prop_assert_eq!(decoded_operation, operation.unwrap());
+        }
+    }
+}
```
cafca, line 26: A module docstring could be nice for this, just something short about what a prop test is.

cafca (changes requested):
I have overlooked this PR for too long even when it's actually really really nice. I think a short comment on what `prop_tests.rs` is could be nice in that file.

cafca:
Current problem is that the proptests grow too large with arbitrary hashes being generated and new complexity introduced by `Relation` and friends. I am looking for a way to limit the number of arbitrary hashes that are being tested.

sandreae:
Yeh, I agree. If we find an area where we want to apply it in the future, we can introduce it then.

Issue #191, opened 2022-01-26
More features for test_utils


sandreae:
- methods for making requests to the mock node
- introduce AuthorLogs wrapper struct
- move getters from node to associated structs
- implement logging with https://docs.rs/log/latest/log/ 
- print logs in tests with https://docs.rs/pretty_env_logger/0.4.0/pretty_env_logger/

PR #192, opened 2022-01-27
Updates in mocking structs

sandreae:
Main things for this pr:
- Refactor test_util mocks
- Implement logging for mock node methods in the test suites 

closes: https://github.com/p2panda/p2panda/issues/191
## 📋 Checklist


p2panda-rs/src/lib.rs
```diff
@@ -80,3 +80,12 @@ pub trait Validate {
     /// Validates p2panda data type instance.
     fn validate(&self) -> Result<(), Self::Error>;
 }
+
+/// Init pretty_env_logger before the test suite runs to handle logging outputs.
+/// Uses the `ctor` crate to define a global constructor function.
```
cafca, line 85: It might be good to document this a bit more. How can consumers of the logs receive them? Maybe this could go into the handbook? Or readme?

cafca: Maybe also make explicit that "global contructor function" means? Not sure, maybe only I don't know what that is :D

p2panda-rs/src/test_utils/README.md
```diff
@@ -16,6 +16,18 @@
 - A mock client
 - Methods for generating test data
 
+## Logging
+
+To enable logging in the mock node run the test suite with the following env vars set:
```
cafca, line 21: Could add a sentence about what the mock node even is :D

cafca (changes requested):
I like it! I think it would be helpful to document how to use the logging a bit more. 

I noticed that when I run `cargo t` without setting any env vars I get log messages from OpenMLS. Is there a way to suppress these by default?

```
ERROR openmls::framing::ciphertext              > Sender data decryption error
```

sandreae:
I added some comments and went overboard on the README ;-p

PR #196, opened 2022-01-31
Sketch for global operation CDDL

cafca:
Sketch to create a global representation of all valid operations in a single CDDL definition that can be used to test encoded operations for whether they are fit to be materialised.

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/schema/system.rs
+++ b/p2panda-rs/src/schema/system.rs
@@ -4,6 +4,8 @@ use crate::hash::Hash;
 use crate::schema::fields::Type;
 use crate::schema::SchemaBuilder;
 
+use super::Schema;
+
 /// Make CDDL definition for operations from a given schema
 fn make_schema_cddl(
     schema_name: &str,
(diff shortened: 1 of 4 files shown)


p2panda-rs/src/schema/system.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        {name} = {{
+            schema: "{hash}",
+            version: 1,
+            operation-body-{name}-{hash},
+        }}
+
+        hash = tstr .regexp "[0-9a-f]{{68}}"
+
+        operation-body-{name}-{hash} = (
```
cafca, line 28: I wasn't able to make this line 

```
operation-body-{name}-{hash} //= (
```

even though the spec says I should be able to do that :(

> It is not an error if a name is first used with a "/=" or "//=" (there is no need to "create it" with "=").
[Section 2.2.2](https://datatracker.ietf.org/doc/html/rfc8610#section-2.2.2)

adzialocha:
One thought: We had the idea once to have "one global CDDL" some time ago and its great we finally are getting there. Originally I understood that global CDDL as one huge CDDL definition containing all supported schemas of that node, theoretically thousands.

Now I was wondering if it might be better to only validate each incoming entry based on the regarding (much smaller) CDDL for this particular schema which just came in instead?

When all CDDL is pre-generated for all schemas (not in one global document but in multiple, separate ones), then it could be nice and fast to validate incoming data. I don't have numbers to talk about performance here and don't think this is the exact point, just wanted to raise this point here for discussion?

adzialocha:
> One thought ..

Regarding my note above I noticed one crucial difference between a "global" CDDL or a "specialized" one for each schema: The latter works only if we already deserialized the CBOR into an operation to extract the schema hash (so we can request the right CDDL schema for it). That step is not necessary if we have a "global" CDDL :+1:

adzialocha:
> I also noticed this today doing the validation graph (having to deserialize the operation to get the schema hash). I did wonder if we can actually get the schema from just the decoded entry though, as we know which log it is pointing at, and we should know the schema for that log.
> 
> As I write this, I realise that won't work for CREATE operations though... 😬

The schema is part of the operation payload, no need to look that up 👍

sandreae:
Yes, i know. I mean in the case where we haven't already deserialized the CBOR :+1:

adzialocha:
> Yes, i know. I mean in the case where we haven't already deserialized the CBOR +1

Thats still not necessary, as one can write a CDDL which matches exactly that Schema hash string

sandreae:
Yes, I realise that about the global CDDL, I was thinking within the "specialized" individual schema in memory model.

adzialocha:
Could this PR be somewhat related to this issue here? https://github.com/p2panda/p2panda/issues/216

adzialocha:
I'll close this now as there is another PR coming up.

Issue #35, opened 2021-03-01
Verify signature when decoding entries


cafca:
p2panda-rs allows decoding entries from an encoded and signed form. Currently this does not include validation of the entry signature.

adzialocha:
Closed via https://github.com/p2panda/p2panda/pull/367

PR #197, opened 2022-02-01
Add tests for size and signature method of EntrySigned

cafca:
Text coverage showed that some parts of `EntrySigned` were not covered. I added these tests but the signature seems to be broken. We don't use it anywhere.

## 📋 Checklist


adzialocha:
Very nice! :+1: 

>  I added these tests but the signature seems to be broken. We don't use it anywhere.

Doesn't sound like it was "broken", but I guess the pending issue related to "it was not used" is here: https://github.com/p2panda/p2panda/issues/35

sandreae:
> Thanks @sandreae for getting this to work!

Team effort :star:

adzialocha (changes requested):
Changelog is missing, otherwise alles nice!

adzialocha (approved):

PR #199, opened 2022-02-01
Rename OperationWithMeta

### LINKED ISSUE #198: Revisit `OperationWithMeta` and `Operation` relationship
`Operation` currently is a struct that contains the contents of a decoded `OperationEncoded`. However, as we are moving to handling documents (and further), operations become a much more central concept to p2panda than before, when we were mostly concerned with handling and processing entries. `OperationWithMeta` has already been introduced to wrap both an `Operation` as well as metadata about its author. 

At first thought we could use entries instead of `OperationWithMeta` but entries may not contain the actual operation data as that is optional.

Rename:

- `OperationWithMeta` -> `Operation`
- `Operation` -> `OperationRaw` (?)

---

_Previous issue description to provide context to discussion below_

The name `OperationWithMeta` can be improved to reflect what the role of this struct is. Conceptually, the original `Operation` doesn't carry any information about its author or the entry it was published in. It exists independent of authorship and only describes a change to a document. `OperationWithMeta` on the other side contains information about the author of the change and the entry it was published in. We also have `OperationEncoded`, which also doesn't carry author information but is encoded in CBOR. 

I think it would be good to name it `OperationSigned` to show that it represents an operation that has been signed (and published) by an author.

* **Entry** raw entry data (no author)
* **EntrySigned** Bamboo-encoded entry + signature
* **Operation** raw data change (no author)
* **OperationEncoded** CBOR-encoded operation (no author)
* **OperationSigned** raw data change + id of author & signed entry

cafca:
Closes #198

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/key_group.rs
+++ b/p2panda-rs/src/key_group.rs
@@ -1,0 +1,27 @@
+use std::collections::BTreeMap;
+
+use crate::hash::Hash;
+use crate::identity::Author;
+use crate::instance::Instance;
+
+struct Membership {
+    author: Author,
+    key_group: Instance,
+    accepted: bool,
+    can_authorise: bool,
+    can_create: bool,
+    can_update: bool,
+    can_delete: bool
+}
+
+struct KeyGroupIndex {
+    members: Vec<Membership>,
+    by_key_group: BTreeMap<Hash, Vec<Membership>>,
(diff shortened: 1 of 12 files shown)


cafca:
Closing because this needs some more discussion to unwrap how we think about operations conceptually.

PR #219, opened 2022-02-16
Implement replication functionality on mock node

sandreae:
- `replicate` method for mock nodes to exchange entries
- ~functionality to parse an array of operations into a `mermaid-js` string~

## 📋 Checklist


sandreae:
Yeh, that's a good point. For now though, rather than splitting off, I might prefer to implement all the mermaid-js stuff directly in `panda-playground` instead of here in `test_utils`. That would mean the only addition here is a few methods on the mock node. Not against splitting it out, would want to think more about it though.

p2panda-rs/src/test_utils/mocks/node.rs
```diff
@@ -200,6 +200,57 @@ impl Node {
...
+                &entry.entry_encoded(),
+                &entry.operation_encoded(),
+            );
+            self.db.insert(author.as_str().into(), author_logs);
+            true
+        }
+    }
+
+    /// Replicate entries from other_node to this node.
+    pub fn replicate_with(&mut self, other_node: &mut Node) -> Vec<Hash> {
```
cafca, line 240: Could be nice to have examples for mock node methods. I know they are also in the tests so not required, just - nice to have :-)

cafca (approved):

cafca (changes requested):
Actually :D It would be nice to cover all new lines with tests, I think not much is needed for that.

sandreae:
I'm gunna close this as most of the functionality was moved out into the playground, and soon we will do a revamp of the mock node in any case. This can be approach then if it's still needed.

Issue #211, opened 2022-02-10
Specify schema hash for application schemas


adzialocha:
So far we derive the schema hash from the first entry of a application schema document, how do we do this for system schemas?

Some options we have:

1. Use schema field to also mention some enum value like key_group_v1 instead
2. Just generate a random hash and assign it "forever" to certain system schemas
3. Use funny hashes like 0000000000...000001 or babb1ed.. and assign them forever
4. Publish schemas actually on p2panda (and the bytes of the entries / payloads in a public repository) with the system schema definitions inside. Take the hash from there and assign them forever (this does NOT mean that one needs to materialize system schemas before, they are still hardcoded), its only the symbolic act of "where did the hash come from", thats all
5. New and hot: Introduce a `type` field (similar as option 1, just separated)

sandreae:
I like option 1) too.

adzialocha:
Let's go for 1 then!

sandreae:
Hmmm...... I just looked at implementing this. I'm not so sure about option 1 anymore.... It feels like it might be adding unnecessary checking and de/serializing complexity in many places. Somehow I quite like option 3 or 4 now.....

sandreae:
We could still have an enum with option 3 or 4, checking for known system schema hashes and serdering (:wink:) correctly, but that way we still end up with a valid hash. That's quite nice.

sandreae:
Something like this: https://gist.github.com/sandreae/5417338975841ed1880c2e69cc00d0c3

sandreae:
or is it pointless doing both :thinking: ..............

sandreae:
> But wouldn't it work now if you replace the hashes (0x0000..) with human readable values like "schema" and "schema_fields"?

Yeh, that would work, it seemed strange that sometimes a `SchemaType` was represented by a strict hash string, and sometimes it was represented by.... not a hash string..... With the above then we can coerce all schema types into hashes if we like. And the strings would pass the same string validation when serialized.

sandreae:
WIP PR: https://github.com/p2panda/p2panda/pull/221

adzialocha:
Nice pages around hexspeak:

https://en.wikipedia.org/wiki/Hexspeak
https://reminiscential.wordpress.com/2008/09/03/hexspeak-word-list/

cafca:
[...]

PR #221, opened 2022-02-17
Schema hash enum

sandreae:
Introduces `SchemaId` as a new type used by the operation `schema` field.

## 📋 Checklist

  - https://github.com/p2panda/handbook/pull/148


sandreae:
Think this works well. Only thing I'm not sure about is the name.

`let application_schema = SchemaType::new(DEFAULT_SCHEMA_HASH)`

Doesn't give you a schema type, it gives you the hash for a specific schema wrapped in a type..... It would be nice just to call it `Schema` but this will be taken later. Maybe `SchemaHash`?

adzialocha:
Was stumbling upon the same thing, `SchemaHash` sounds better :+1:

p2panda-rs/src/document/document.rs
```diff
@@ -453,7 +455,7 @@ mod tests {
             )
             .unwrap(),
         );
-        let schema = Hash::new_from_bytes(vec![3, 2, 1]).unwrap();
+        let schema = SchemaHash::new(DEFAULT_SCHEMA_HASH).unwrap();
```
adzialocha, line 458: Grand Type-Tennis :yellow_heart:

sandreae: lolz

adzialocha:
Think we need a final decision on this one here before we can merge this: https://github.com/p2panda/p2panda/issues/211

sandreae:
Ah yes... true... Shame we have to actually make decisions as well as implementing them :roll_eyes:

adzialocha:
> Ah yes... true... Shame we have to actually make decisions as well as implementing them

Finally makes us make decisions :-D

I'm fine with this current solution :+1: Could be nice to have slightly more fun hashes maybe? Like: `SOLIDIFIED0000SOFA000...0001`

p2panda-rs/src/document/document.rs
adzialocha: Seriously! This makes so much sense and the API becomes clearer (like "wtf what hash do I use here for what?")

sandreae: haha, I know, right?? I think we were waiting for some hash typing this whole time :rocket:

sandreae:
`5C4E3A000000000000000000000000000000000000000000001`

SCHEMA = 5C4E3A (kind of)

cafca:
I could really use this struct for the CDDL stuff. Can I help out here? With the handbook maybe?

I think `SchemaId` would be a good name.

sandreae:
Hey! Yes, a little help would be great. I can work on the PR to get it ready now, if you don't mind doing the handbook updates that would be cool (sorry, you shouldn't always do the handbook :sob:)

Just checking, but do we already want to change `SchemaId` to contain an array of graph tips in this PR? I'm not sure.

cafca:
I would prefer we make the change for application schemas in another pr. When we do so many interdependent things in parallel its helpful to separate concerns so nothing is blocked :)

sandreae:
Great, I just suggested that in the chat as well ;-p

sandreae:
But I will change application schema variant to contain a `RelationList`, right?

sandreae:
Ughh, I'm confused, that's not a relation list, it's a DocumentViewId, right?? I think my brain is tied up in knots....

cafca:
Also have a knot situation over here and unfortunately have to take a break now and do some other stuff :S

cafca:
I will be adding changes to the handbook in this pr

https://github.com/p2panda/handbook/pull/148

sandreae:
So I changed `SchemaId::Apllication(Relation)` to contain a relation. But I basically ignored that we can include an array of graph tips to identify a specific view (it _is_ optional) as I wonder if this needs to be dealt with in several places and would be better done in a different PR.

I think this is_"OK"_, let me know what you think.

adzialocha (approved):
Wohoo!

cafca:
This was not ready to merge without the handbook changes.

Issue #216, opened 2022-02-10
Convert `Schema` struct to CDDL


adzialocha:
A method or struct which can translate a `Schema` struct to a CDDL representation.

```
; We need to generate something like this per application schema:

schema_id = ["... document view ..."]

create_fields = { some, fields, for, that, schema }
update_fields = { + (some // fields // for // that // schema) }

; This we can already generate have with our `CddlGenerator`:

some = ( some: { value_text } )
fields = ( fields: { value_int } )
for = ( for: { value_relation } )
that = ( that: { value_text } )
schema = ( schema: { value_bool } )
```

PR #239, opened 2022-03-02
Separate `cddl` from `schema` more clearly

adzialocha:
This is a preparation for #216 and https://github.com/p2panda/p2panda/pull/248. It ..

* Tries to not use the word "schema" too much in CDDL contexts, CDDL also has "schemas" but its just too confusing because it is a _very different_ thing from p2panda schemas
  * Moves all CDDL related things from `schema` into a `cddl` module
  * Renames "Operation Schema" to "Operation Format" as per discussion here: https://github.com/p2panda/handbook/pull/148
* Adds some tests around the operation format
* Renames `CDDLBuilder` to `CddlGenerator` (builder sounds like the Rust-Builder-Pattern which it is not)
* Removes the CBOR validation logic from the `Validate` trait in `OperationEncoded`. This should be handled separately a) to prepare for a cleaner way of dealing with wasm targets b) #216 will validate much more than this (Operation Format + Application / System Schema)
* Removes a couple of unused error types (interesting that this is not catched by the linter, can be good to do the same for all other errors one day)

## 📋 Checklist

- ~~Link this PR to any issues it closes~~


adzialocha:
`codecov/patch` fails about untested locs which will be removed in this PR anyhow: https://github.com/p2panda/p2panda/pull/248

sandreae (approved):
Looks great to me. Excellent to more clearly differentiate between our internal concept of schema and the cddl validation. Definitely helping me grok the schema beast. I just made a small doc string change.

PR #240, opened 2022-03-02
Update `DocumentBuilder` docs

cafca:
This is a showcase for using doctests more to document how we want library consumers to use all of this stuff (we already did this early on for `Entry` for example). Here I am using `test_utils` to generate some example values behind the scenes and then in the actual example only use the `DocumentBuilder` API.

I also did some language updates to the docstrings and moved some duplicate validation into `DocumentBuilder::build` that previously existed on both the struct itself as well as that method (but in different variations).

LMKWYT (let me know what you think)

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/document/document.rs
+++ b/p2panda-rs/src/document/document.rs
@@ -147,9 +147,9 @@ impl Document {
 /// # use p2panda_rs::test_utils::meta_operation;
 /// #
 /// # #[rstest]
-/// # fn main(meta_operation: OperationWithMeta) -> () {
+/// # fn main(#[from(meta_operation)] operation: OperationWithMeta) -> () {
 /// // You need a `Vec<OperationWithMeta>` that includes the `CREATE` operation
-/// let operations: Vec<OperationWithMeta> = vec![meta_operation];
+/// let operations: Vec<OperationWithMeta> = vec![operation];
 ///
 /// // Then you can make a `Document` from it
 /// let document = DocumentBuilder::new(operations).build();
--- a/p2panda-rs/src/document/document.rs
+++ b/p2panda-rs/src/document/document.rs
@@ -133,22 +133,36 @@ impl Document {
     }
 }
 
-/// A struct for building documents from a collection of operations. When calling `build()
-/// a document is returned wrapped in a result. The build will error if the operations passed
-/// don't follow documents validation criteria.
+/// A struct for building [documents][`Document`] from a collection of [operations with metadata][`crate::operation::OperationWithMeta`].
 ///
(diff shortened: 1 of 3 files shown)


p2panda-rs/src/document/document.rs
```diff
@@ -133,22 +133,37 @@ impl Document {
...
+/// # #[rstest]
+/// # fn main(meta_operation: OperationWithMeta) -> () {
+/// // You need a `Vec<OperationWithMeta>` that includes the `CREATE` operation
+/// let operations: Vec<OperationWithMeta> = vec![meta_operation];
+///
+/// // Then you can make a `Document` from it
+/// let document = DocumentBuilder::new(operations).build();
+/// assert!(document.is_ok());
+/// # }
+/// # }
```
sandreae, line 158: Love it! Nice succinct docs explaining a single method :+1:

p2panda-rs/src/document/document.rs
```diff
@@ -133,22 +133,37 @@ impl Document {
...
+/// # mod tests {
+/// # use rstest::rstest;
+/// # use p2panda_rs::document::DocumentBuilder;
+/// # use p2panda_rs::operation::OperationWithMeta;
+/// # use p2panda_rs::test_utils::meta_operation;
+/// #
+/// # #[rstest]
+/// # fn main(meta_operation: OperationWithMeta) -> () {
+/// // You need a `Vec<OperationWithMeta>` that includes the `CREATE` operation
+/// let operations: Vec<OperationWithMeta> = vec![meta_operation];
```
sandreae, line 152: Not sure about the name `meta_operation` suggests to me a special kind of operation which is meta, rather than a normal operation with meta data attached (I know the whole naming for this thingy is TBC but in the context of this PR..... that's my feeling ;-p)

sandreae (approved):
Niiiiice :+1: (I had one naming thing, but I am not gunna put my foot down on it, just my thoughts)

p2panda-rs/src/document/document.rs
cafca: You mean instead have `operation_with_meta` as the variable name? I could also rename the whole fixture.

sandreae: Yeh, either that, or just have it as `operation`

adzialocha: Nice! +1 for renaming the variable from `meta_operation` to `operation` in the example (not the fixture). Otherwise nice example @cafca, happy to merge it after that small change.

adzialocha (approved):

PR #248, opened 2022-03-03
CDDL for `schema_v1` and `schema_field_v1`

adzialocha:
Describe your changes here.

## 📋 Checklist

- [ ] Add tests that cover your changes
- [ ] Add this PR to the _Unreleased_ section in `CHANGELOG.md`
- [ ] Link this PR to any issues it closes
- [ ] New files contain a SPDX license header
- [ ] Check if descriptions and terminology match `handbook` content (and visa-versa)

Issue #245, opened 2022-03-02
Make `SchemaId::Application` value a `PinnedRelation` type


adzialocha:
```yaml
action: update | create | delete
version: 1
schema: [<operation id>, ...] # This will be a pinned relation
fields: ...
```

Issue #215, opened 2022-02-10
Implement `Schema` struct


adzialocha:
A cool new thing which holds: `name`, `description` and `fields`.

Sketch by Sam: https://laub.liebechaos.org/Xol8rqMjQWG04Ct6DaaiRw#

adzialocha:
Do we need a `SchemaField` struct?

sandreae:
Been looking a little into this. We talked about using new type pattern like so:

`struct Schema(DocumentView);`

Which is cool! But then I wonder about validation (as always :roll_eyes:) within `p2panda-rs`. If `Schema` is "just" a wrapper around document view, then on a node, validation should have happened when resolving the document (the schema hash is stated, operations are validated, we know it actually is a schema) but this isn't apparent in `p2panda-rs` and it feels very weird just to assume the fields in the wrapped `DocumentView` are correct.

I can see a few ways forward:

A) actually use the system schema cbor definition to validate the DocumentView (which itself is almost just a wrapper around `OperationFields`) bytes?
B) implement `TryFrom<DocumentView for Schema>` and then validate we have the correct fields.
C) Just do without these types completely :rofl: they are only document views anyway, an implementation knows the schema of the document it is resolving and performs appropriate validation in any case.

If we do "B" then it makes me wonder if we may as well be creating a richer `Schema` struct...

What do you think?

sandreae:
Ok, great, this is roughly where I ended up too. I can make some sketches :+1:

I kept tripping up on "Cool, everything is a document view, let's make thin wrappers and move on" .... later.... "Ok, let me just get my `Schema` which knows about things like fields and types and whatnot and looks a certain way.... oh....."

sandreae:
In progress here: https://github.com/p2panda/p2panda/pull/250

PR #250, opened 2022-03-05
`Schema` (one more time.... 🕺)

sandreae:
A new struct for finally bringing together a `SchemaView` with it's relevant `SchemaFieldView`s into a unified struct which can be used for:

- expressing materialised schema in CDDL format
- ~~validating encoded & decoded application operations~~ i guess we just do this with the CDDL

Both `SchemaView` and `SchemaFieldView`s are themselves materialised from published operations on the p2panda network. They are validated against expected system schema formats as part of this process. Here we are combining already validated data structures, and only validating that matching the passed fields match the schema definition.

The resulting struct is what we have for some time described as an Application Schema.

## Implementation questions:
- do we want to keep this struct as a simple definition and then take a functional approach to validating operations and generating the CDDL definitions? 
- Should this struct be called `ApplicationSchema`? I understand that to be what it is, but don't think it's strictly necessary to name it that. 

## Other musings:
~~Thinking about validation here, it made me wonder about the fact that we don't know what we are expecting each field in a `Schema` to be (`name`, `field_type` and less importantly, `description`.) This information isn't included in the `SchemaView`. Updates could happen on the field and if the schema relations were updated the schema def would just suddenly change. I know this is a whole conversation saved for migrations, we haven't figured out yet, but still, maybe we should think about explicitly limiting how schema/schema_fields can be updated for now? Until we have some migration mechanism in place.~~ will be solved with this: https://github.com/p2panda/p2panda/issues/245

## Todo
- [x] construct `Schema` from `SchemaView` and `SchemaFieldView`s
- [x] validate `Schema` struct
  - field ids match those in the schema definition
- [x] generate CDDL from `Schema`

depends on https://github.com/p2panda/p2panda/pull/248 & https://github.com/p2panda/p2panda/pull/239

## 📋 Checklist

closes: #216 #215


adzialocha:
> do we want to keep this struct as a simple definition and then take a functional approach to validating operations and generating the CDDL definitions?

Yes, lets keep `Schema` simple and have it all separated into functions :+1:

p2panda-rs/src/schema/schema.rs
```diff
@@ -0,0 +1,189 @@
...
+        )
+        .is_err());
+        // Too few fields passed
+        assert!(Schema::new(schema_view.clone(), vec![bool_field_view.clone()]).is_err());
+        // Too many fields passed
+        assert!(Schema::new(
+            schema_view,
+            vec![bool_field_view, capacity_field_view, toilet_size_field_view]
+        )
+        .is_err());
```
adzialocha, line 187: Noticed this in other parts of our code as well: We have a tendency to have very long singular tests doing many different things. Sometimes I have a hard time understanding what a test actually does. I'd be in favour of separating them into smaller tests, giving them names which describe the one thing they want to test (or add comments on top), making sure we frame every test around one thing as good as possible.

I know that usually putting all into one test comes from the work one needs to put in to bootstrap the test, but I think this can be made easier with a) using fixtures b) adding more fixtures c) writing test helpers on top of all tests to have reusable methods.

p2panda-rs/src/schema/schema.rs
```diff
@@ -0,0 +1,189 @@
...
+        #[from(random_hash)] schema_view_id: Hash,
+        #[from(random_document_id)] schema_id: DocumentId,
+        #[from(random_document_id)] bool_field_id: DocumentId,
+        #[from(random_document_id)] capacity_field_id: DocumentId,
+        #[from(random_document_id)] toilet_size_field_id: DocumentId,
+    ) {
+        // Create schema definition for "venue"
+        // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+        let mut bool_field = BTreeMap::new();
```
sandreae, line 86: It doesn't help that this variable is named completely wrong 😅

adzialocha (approved):
This looks very nice and good to me :+1: Loving the simplicity!

Issue #238, opened 2022-03-01
Validation when converting `DocumentView` into specialised `SchemaView`


sandreae:
What behaviour do we expect when trying to parse an application `DocumentView` which follows an application schema, but contains the compulsory key-values as a system schema, into a specialised `SchemaView` or `SchemaFieldView`?

Should:  
```
farm {
  name: String,
  description: String,
  fields: RelationList,
  animals: RelationList
}
```

be allowed to be coerced into a `SchemaView`?

adzialocha:
Is this related? https://github.com/p2panda/p2panda/issues/242

sandreae:
Yeh, it is related, but not exactly the same issue. 

I suppose I just don't see how allowing invalid type coercion is desirable, especially when it's very easy to stop it.

When is the case where we expect the above example to be parsed into a `SchemaView`? (I don't think there is a case, right?).

sandreae:
Ah, ok, maybe I spot the reason this is causing some mental challenge.

Currently we lose `Schema` information when creating `DocumentViews` so we can't just check if the schema is `schema_v1` or `schema_field_v1`.

sandreae:
So we end up converting just a map of values (which were validated against a schema during `Document` building, but that information is gone now).

adzialocha:
Is there a "meta" construct we can introduce which spits out a `Schema` in the end and internally uses the materializer after validating the data? Then it might not be so bad that the "DocumentViews" don't know about the schema, because the overall procedure knows what it is dealing with?

I hope I could bring that point across, not sure how to put it in better words right now :-D

sandreae:
Yeh, I think I get you.

I feel like we are heading back to `DocumentBuilder` throwing out the correct document view already, depending on the schema. I think this could be implemented quite neatly without too many changes. 

Does that match your thought?

sandreae:
Ah, sorry, maybe that doesn't quite match, as you're taking about `Schema`....

sandreae:
I mean, we know all operations have been validated on arrival against their schema though, so maybe I'm just being really psycho here......

sandreae:
(please tell me if I am)

adzialocha:
> I mean, we know all operations have been validated on arrival against their schema though, so maybe I'm just being really psycho here......

Haha, you're not a psycho lol, but indeed thats quite a good point with the validation via CDDL beforehands ..

sandreae:
[...]

PR #257, opened 2022-03-09
Dynamically resolve document depending on schema in `DocumentBuilder`

sandreae:
An attempt at implementing dynamic resolution of document views according to the schema id type. It uses trait objects which are stored on `Document` as `Box<dyn AsDocumentView>`. The trait `AsDocumentView` is implemented on all document view variants.

This would resolve issues around document view conversions, as they would happen automagically during document building. 

We work under the assumption that we are only dealing with operations which have already been validated (using their `cddl_format` definition) against their stated schema.

closes: [238](https://github.com/p2panda/p2panda/issues/238)

Notes:
- I'm not sure we want to go this direction, it somehow loosens the typing with our returned values from `DocumentBuilder`, but looking at it from another angle, it just makes explicit the fact that document views can have a variety of types.
- this is just a suggestion right now, woul dlike to discus it (both conceptually, and the implementation itself)

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/schema/schema_id.rs
+++ b/p2panda-rs/src/schema/schema_id.rs
@@ -10,38 +10,6 @@ use crate::hash::Hash;
 use crate::operation::Relation;
 use crate::schema::error::SchemaIdError;
 
-/// Struct representing a SchemaV1 id.
-#[derive(Clone, Debug, PartialEq)]
-pub struct SchemaV1(String);
-
-impl SchemaV1 {
-    fn as_str(&self) -> &str {
-        self.0.as_str()
-    }
-}
-
-impl Default for SchemaV1 {
-    fn default() -> Self {
-        Self("schema_v1".to_string())
-    }
-}
-
-/// Struct representing a SchemaFieldV1 id.
-#[derive(Clone, Debug, PartialEq)]
-pub struct SchemaFieldV1(String);
-
-impl SchemaFieldV1 {
-    fn as_str(&self) -> &str {
-        self.0.as_str()
-    }
-}
-
-impl Default for SchemaFieldV1 {
-    fn default() -> Self {
-        Self("schema_field_v1".to_string())
-    }
-}
-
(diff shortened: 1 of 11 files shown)


p2panda-rs/src/schema/schema_id.rs
```diff
@@ -10,25 +10,57 @@ use crate::hash::Hash;
...
 }
 
 impl SchemaId {
     /// Instantiate a new `SchemaId` from a hash string.
     pub fn new(hash: &str) -> Result<Self, SchemaIdError> {
         match hash {
-            "schema_v1" => Ok(SchemaId::Schema),
-            "schema_field_v1" => Ok(SchemaId::SchemaField),
+            "schema_v1" => Ok(SchemaId::Schema(SchemaV1::default())),
+            "schema_field_v1" => Ok(SchemaId::SchemaField(SchemaFieldV1::default())),
```
adzialocha, line 63: Oh, this is interesting. What is the reason for using `Default` here?

sandreae: Well, not much reason really, except it felt nicer than just sticking a string in there. As the value will be the same for now (forever?)

sandreae: Ah!! Stupid me, I started off this PR with a different approach to the type coercion and this is leftover from that. I can remove all of the changes to `SchemaId` :roll_eyes:

sandreae: They're goooone

adzialocha: Haha, alright! Yeah, I figured it would make sense if you would need the specific type of an schema enum variant :+1:

sandreae:
I'm gunna close this, I don't think it's a good direction. Trait objects are weird and hard and maybe don't give us a nice solution. lolz. I think we can take this later if we think it's still important once we have implemented some stuff in `aquadoggo`.

PR #263, opened 2022-03-11
Update error reporting for adding operation fields

### LINKED ISSUE #262: Improve wasm error messages when adding fields to an operation
e.g trying to add a hash array as a relation value throws `Invalid object`. Could be something like `Invalid value: expected a hash value for field of type relation`

cafca:
Closes #262 

## 📋 Checklist

### DIFF
--- a/p2panda-js/src/wasm.test.ts
+++ b/p2panda-js/src/wasm.test.ts
@@ -100,20 +100,22 @@ describe('WebAssembly interface', () => {
       );
 
       expect(() => fields.add('contact', 'relation', [TEST_HASH])).toThrow(
-        'Invalid object',
+        'Expected a hash value for field of type relation',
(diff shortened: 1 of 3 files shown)


p2panda-js/src/wasm.test.ts
```diff
@@ -100,20 +100,22 @@ describe('WebAssembly interface', () => {
       );
 
       expect(() => fields.add('contact', 'relation', [TEST_HASH])).toThrow(
-        'Invalid object',
+        'Expected a hash value for field of type relation',
```
sandreae, line 103: Lolz, was looking at this for a while thinking "But a relation should be just a hash, isn't this wrong?"

sandreae (approved):
Looks great :+1:

Issue #243, opened 2022-03-02
Implement hash method for `DocumentViewId`


adzialocha:
Related: https://github.com/p2panda/handbook/issues/151

PR #268, opened 2022-03-12
Introduce schema hash id as a unique string representation for schemas

### LINKED ISSUE #267: Unique string identifier for application schemas
Should we have a convention for representing all `SchemaId` variants as strings? This would make it nice and easy to store them all in a db.

For system schema this is easy, they are already defined as strings `schema_v1` and `schema_field_v1` (both these are deprecated terms though ;-p)

For application schema which contain an array of hashes (document view / pinned relation) it needs more though

## Options
- concat the array of hashes into something like `0A00_0B00_0C00`
- hash the array of hashes (then use this as an id for the list of hashes which are stored in their own table)

cafca:
The schema hash id consists of the schema name and its hashed document view graph tips, concatenated with an underscore. 

Implements `Display` for `SchemaId`.

I'll be waiting with a handbook update until I hear from you aboot what ye think.

Closes #267 and #243

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/document/document_view_id.rs
+++ b/p2panda-rs/src/document/document_view_id.rs
@@ -55,8 +55,16 @@ impl DocumentViewId {
 }
 
 impl Display for DocumentViewId {
+    /// Document view ids are displayed by concatenating their hashes with an underscore separator.
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.hash().as_str())
+        let mut is_first = true;
+        for hash in self.0.clone().into_iter() {
+            write!(f, "{}{}", if is_first { "" } else { "_" }, hash)?;
+            if is_first {
+                is_first = false;
+            }
+        }
+        Ok(())
     }
 }
 
@@ -148,7 +156,20 @@ mod tests {
 
         assert_eq!(
             format!("{}", document_view_id),
-            "0020fc76e3a452648023d5e169369116be1526f6d3fc2b7742ed1af2b55f11bca7fb"
+            "0020b177ec1bf26dfb3b7010d473e6d44713b29b765b99c6e60ecbfae742de496543"
+        );
+
+        let hash_1 = "0020b177ec1bf26dfb3b7010d473e6d44713b29b765b99c6e60ecbfae742de496543"
+            .parse::<Hash>()
+            .unwrap();
+        let hash_2 = "0020d3235c8fe6f58608200851b83cd8482808eb81e4c6b4b17805bba57da9f16e79"
+            .parse::<Hash>()
+            .unwrap();
+        let view_id_unmerged = DocumentViewId::new(vec![hash_1, hash_2]);
+
+        assert_eq!(
+            format!("{}", view_id_unmerged),
+            "0020b177ec1bf26dfb3b7010d473e6d44713b29b765b99c6e60ecbfae742de496543_0020d3235c8fe6f58608200851b83cd8482808eb81e4c6b4b17805bba57da9f16e79"
         );
     }
 }
--- a/p2panda-rs/src/document/document_view_id.rs
+++ b/p2panda-rs/src/document/document_view_id.rs
@@ -143,7 +143,7 @@ mod tests {
     }
 
     #[test]
-    fn hashes() {
+    fn string_representation() {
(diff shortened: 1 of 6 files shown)


adzialocha:
You could check out the `deserialize_str` method in the Deserializer in SchemaId and parts of my abstract query branch in aquadoggo. Can imagine that's what we need to parse the schema name in the id.

p2panda-rs/src/document/document_view_id.rs
```diff
@@ -40,6 +41,23 @@ impl DocumentViewId {
...
+            .collect();
+        Hash::new_from_bytes(graph_tip_bytes).unwrap()
+    }
+}
+
+impl Display for DocumentViewId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.hash().as_str())
+    }
 }
```
sandreae, line 61: This is interesting. I had imagined that the way we want to identify `SchemaId`s would still be by their array of hashes (for debugging purposes) but that the hash string would be a useful identifier in some specific cases. With this thinking i'd expect `Display` to show the hash list. But I'm open to thinking differently about this!

p2panda-rs/src/schema/schema.rs
```diff
@@ -60,6 +61,19 @@ impl Schema {
...
+    pub fn hash_id(&self) -> String {
+        format!("{}_{}", self.name, self.id.hash().as_str())
+    }
+}
+
+impl Display for Schema {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.hash_id())
+    }
 }
```
sandreae, line 77: Carrying on from the comment above, maybe this is the place where I would expect the hash string to be the commonly used id. Hmmm.... basically I still need a little more thought on this. Would be happy to hear yours!

p2panda-rs/src/document/document_view_id.rs
adzialocha: I agree, the `DocumentViewId` should be displayed as .. a document view id! I would suspect the hash living in something like `DocumentViewHash`

cafca: Alright, I'll add a string representation for document view ids and use that to implement `Display` in `Schema` and `SchemaId`

p2panda-rs/src/schema/schema.rs
```diff
@@ -60,6 +61,20 @@ impl Schema {
     pub fn as_cddl(&self) -> String {
         generate_cddl_definition(&self.fields)
     }
+
+    /// Returns a unique string identifier for this schema
+    ///
+    /// This has the format "<schema name>-<hashed schema document view graph tips>".
+    #[allow(unused)]
+    pub fn hash_id(&self) -> String {
+        format!("{}-{}", self.name, self.id.hash().as_str())
```
adzialocha, line 70: We could introduce something like `SchemaIdHash` here (or a better name..). It would be easy to implement "from_str" or even "Serialize" and "Deserialize" with something like this: https://github.com/p2panda/aquadoggo/blob/4f0deef606c6789b7e22d0c1e86cd3044cf12e7a/aquadoggo/src/graphql/convert/schema.rs#L38

adzialocha: The struct could look like that:

```rust
pub struct SchemaIdHash {
    /// Given name of this schema.
    name: String,

    /// Hash of the schema.
    view_hash: DocumentViewHash,
}
```

cafca: Ah, you mean because then we can use the Rust type system where we want to use it? Good idea.

adzialocha: Yeah! We will need it latest in `AbstractQuery`, could be cool to already have the functionality here :+1:

cafca: I'm not sure why we would make this a struct though. We can't really use the constituent parts for anything except checking equality and storing it as a `String` is much more straightforward to implement.

adzialocha: I'm definitely concerned about what we're doing, yeah :laughing:

adzialocha: I think it could help to have a definition somewhere WHAT this schema id tries to solve

cafca:
Closing this to split it up into multiple PRs that address different things.

PR #269, opened 2022-03-13
Let `SchemaFieldDefinition` relation types carry a `SchemaId`

### LINKED ISSUE #264: Let schemas carry a prescribed schema for relation fields
Documents referenced by relation fields must conform to the same schema. Therefore, our `Schema` struct should record this information.

cafca:
- Require setting a schema for schema field definitions of relation types
- Serialises and parses schema such field definitions using a format like `relation(schema_definition_v1)`
- Separates serialisation and display of `FieldType` values, both of which generate a string, by replacing `as_str()` with `serialise()` (which leaves room for a `Display` implementation to use `to_string()` and makes clippy happy)
- Implements conversion of `FieldType` values to `OperationValue`, which makes publishing schema field definitions a bit more straightforward. An example for this is added in the `FieldType` docstring.

Closes #264 

## Todo

- [x] Parse schema ids for relation fields 
- [x] Update schema views
- [x] Decide on a solution for the clippy `to_string` dilemma

## 📋 Checklist

  https://github.com/p2panda/handbook/pull/169


adzialocha:
Thoughts from the chat, maybe not really useful, but I paste them here in any case:

```
1. Change to Relation(String, DocumentId) and PinnedRelation(String, DocumentViewId) deserializing to the regarding string representations venue_0020... and venue_0020..._002....
2. SchemaId would then only become a string when deserialized
3. OperationValue::Relation and all the other relations variations would need to include the type before when deserializing into a string aka relation(venue_0020...) etc.
```

cafca:
Removed from milestone because this is there as issue #264 already.

p2panda-rs/src/schema/field_types.rs
```diff
@@ -19,115 +22,134 @@ pub enum FieldType {
...
+            }
+            (Some("pinned_relation_list"), Some(schema_id)) => Ok(FieldType::PinnedRelationList(
+                SchemaId::new(schema_id).unwrap(),
+            )),
+            _ => Err(FieldTypeError::InvalidFieldType(s.into())),
         }
     }
 }
 
-impl From<FieldType> for String {
```
cafca, line 69: I removed this conversion to more clearly separate serialisation and (debug) display of values.

p2panda-rs/src/schema/system/schema_views.rs
```diff
@@ -163,42 +163,44 @@ mod tests {
...
     use super::{FieldType, SchemaView};
 
     #[rstest]
     fn from_document_view(
-        #[from(random_hash)] relation_operation_id: Hash,
-        #[from(random_hash)] view_id: Hash,
+        random_operation_id: OperationId,
+        #[from(random_operation_id)] view_id: OperationId,
     ) {
-        let mut bool_field = BTreeMap::new();
```
cafca, line 178: I think this was a typo.

p2panda-rs/src/schema/field_types.rs
```diff
@@ -19,115 +32,134 @@ pub enum FieldType {
...
+        let groups = re.captures(s).unwrap();
+
+        match (
+            groups.get(1).map(|m| m.as_str()),
+            groups.get(3).map(|m| m.as_str()),
+        ) {
+            (Some("bool"), None) => Ok(FieldType::Bool),
+            (Some("int"), None) => Ok(FieldType::Int),
+            (Some("float"), None) => Ok(FieldType::Float),
+            (Some("str"), None) => Ok(FieldType::String),
```
adzialocha, line 93: It would be nice to only use regex when it is needed (for relations), for most cases we could simply just match by string.

p2panda-rs/src/schema/field_types.rs
```diff
@@ -19,115 +32,134 @@ pub enum FieldType {
...
-            "float" => Ok(FieldType::Float),
-            "str" => Ok(FieldType::String),
-            "relation" => Ok(FieldType::Relation),
-            "relation_list" => Ok(FieldType::RelationList),
-            "pinned_relation" => Ok(FieldType::PinnedRelation),
-            "pinned_relation_list" => Ok(FieldType::PinnedRelationList),
-            type_str => Err(FieldTypeError::InvalidFieldType(type_str.into())),
+        // Matches a field type name, followed by an optional group in parentheses that contains the
+        // referenced schema for relation field types.
+        let re = Regex::new(r"(\w+)(\((.+)\))?").unwrap();
```
adzialocha, line 83: Is it possible to compile this regex only once by using lazy_static or something similar?

p2panda-rs/src/schema/field_types.rs
cafca: I don't care either way, I thought it was simpler here to just do everything by regex. Feel free to make changes if you want though, because i don't have time for that this week :)

cafca: Done

p2panda-rs/src/schema/field_types.rs
cafca: Done

adzialocha (approved):
Yay! I gave it a pancake, because it has quite a significant impact on our protocol :sparkles:

PR #272, opened 2022-03-15
Introduce `OperationId`

### LINKED ISSUE #241: `OperationId` struct wrapping `Hash`
Make an `OperationId` struct analoguous to `DocumentId`.

Related discussion: https://github.com/p2panda/p2panda/pull/235#discussion_r818029183

cafca:
`OperationId` is a wrapper type around `Hash` and replaces `Hash` as the argument of all methods where the hash represents an operation. This increases type safety and developer experience.

Closes #241 

- [x] Implement `OperationId`
- [x] Replace uses of `Hash` with `OperationId`
- [x] Update js and wasm tests
- [x] Update docstrings
- [x] Check handbook (should be fine)

## 📋 Checklist

### DIFF
--- a/p2panda-js/src/session/session.ts
+++ b/p2panda-js/src/session/session.ts
@@ -303,15 +303,17 @@ export class Session {
 
   /**
    * Signs and publishes an UPDATE operation for the given application data and
-   * matching schema. An UPDATE operation references the entry hash of the
-   * CREATE operation which is the root of this document.
+   * matching schema.
+   *
+   * The document to be updated is referenced by its document id, which is the
+   * operation id of that document's initial `CREATE` operation.
    *
    * Caches arguments for creating the next entry of this schema in the given
    * session.
    *
-   * @param documentId id of the document we update, this is the hash of the root `create` entry
+   * @param documentId id of the document we update, this is the id of the root `create` operation
    * @param fields application data to publish with the new entry, needs to match schema
-   * @param previousOperations array of operation hash ids identifying the tips of all currently un-merged branches in the document graph
+   * @param previousOperations array of operation ids identifying the tips of all currently un-merged branches in the document graph
    * @param options optional config object:
    * @param options.keyPair will be used to sign the new entry
    * @param options.schema hex-encoded schema id
@@ -354,13 +356,15 @@ export class Session {
   }
 
   /**
-   * Signs and publishes a DELETE operation for the given schema. References
-   * the entry hash of the CREATE operation which is the id of this document.
+   * Signs and publishes a DELETE operation for the given schema.
+   *
+   * The document to be deleted is referenced by its document id, which is the
+   * operation id of that document's initial `CREATE` operation.
    *
    * Caches arguments for creating the next entry of this schema in the given session.
    *
    * @param documentId id of the document we delete, this is the hash of the root `create` entry
-   * @param previousOperations array of operation hash ids identifying the tips of all currently un-merged branches in the document graph
+   * @param previousOperations array of operation ids identifying the tips of all currently un-merged branches in the document graph
    * @param options optional config object:
    * @param options.keyPair will be used to sign the new entry
    * @param options.schema hex-encoded schema id
--- a/p2panda-js/src/wasm.test.ts
+++ b/p2panda-js/src/wasm.test.ts
@@ -100,21 +100,21 @@ describe('WebAssembly interface', () => {
       );
 
       expect(() => fields.add('contact', 'relation', [TEST_HASH])).toThrow(
-        'Expected a hash value for field of type relation',
+        'Expected an operation id value for field of type relation',
       );
 
       expect(() => fields.add('contact', 'relation_list', TEST_HASH)).toThrow(
-        'Exptected an array of hashes for field of type relation list',
+        'Exptected an array of operation ids for field of type relation list',
       );
 
       expect(() => fields.add('contact', 'pinned_relation', TEST_HASH)).toThrow(
-        'Expected an array of hashes for field of type relation list',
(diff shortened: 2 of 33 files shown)


p2panda-js/src/session/session.ts
```diff
@@ -303,15 +303,17 @@ export class Session {
...
+   * operation id of that document's initial `CREATE` operation.
    *
    * Caches arguments for creating the next entry of this schema in the given
    * session.
    *
-   * @param documentId id of the document we update, this is the hash of the root `create` entry
+   * @param documentId id of the document we update, this is the id of the root `create` operation
    * @param fields application data to publish with the new entry, needs to match schema
-   * @param previousOperations array of operation hash ids identifying the tips of all currently un-merged branches in the document graph
+   * @param previousOperations array of operation ids identifying the tips of all currently un-merged branches in the document graph
```
sandreae, line 316: This kind of language is much nicer now! An operation is referenced by it's `OperationId`, rather than an unspecified `Hash` :+1:

p2panda-js/src/wasm.test.ts
```diff
@@ -100,21 +100,21 @@ describe('WebAssembly interface', () => {
...
       );
 
       expect(() => fields.add('contact', 'relation_list', TEST_HASH)).toThrow(
-        'Exptected an array of hashes for field of type relation list',
+        'Exptected an array of operation ids for field of type relation list',
       );
 
       expect(() => fields.add('contact', 'pinned_relation', TEST_HASH)).toThrow(
-        'Expected an array of hashes for field of type relation list',
+        'Expected an array of operation ids for field of type relation list',
```
sandreae, line 111: This shoulf be "...of type pinned relation" right? (not related to this PR, but I just spotted it...)

p2panda-rs/src/operation/operation_id.rs
```diff
@@ -0,0 +1,79 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::hash::{Hash, HashError};
+use crate::Validate;
+
+/// Uniquely identifies an [`Operation`].
+///
+/// An `OperationId` is the hash of the [`Entry`] with which an operation was published.
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+pub struct OperationId(Hash);
```
sandreae, line 14: Haha, now that we have this struct, I wonder how we ever managed without it!

sandreae (approved):
Great addition to the library, thanks for doing these many many changes :-p

All looks good to me (except for one tiny type which is not even part of this PR), including some good doc string improvements.

p2panda-js/src/wasm.test.ts
cafca: good catch!

cafca:
@sandreae there was one thing I wasn't sure about here, which is the `send_to_node` test util. It returns a hash and I left it at that, which results in quite a bit of `.into()` when the return value is used as an operation id afterwards. I felt that this was the intention of `send_to_node` but maybe you have a better intuition of that tool's api and want to comment on that?

sandreae:
Ah yes, I see. The returned hash is the hash of the published entry which is needed for several reasons in the tests. I think it's ok to leave this as a `Hash` for now, all this mock `Node` and `Client` logic is up for a big refactor using the new `storage_provider` traits, and this method would be one thing that will be looked at.

p2panda-js/src/session/session.ts
adzialocha: One onion-step less!

p2panda-js/src/wasm.test.ts
adzialocha: If you're changing this you can also fix the old `Exptected` typo, hehe.

p2panda-rs/src/schema/schema_id.rs
```diff
@@ -87,20 +85,21 @@ impl<'de> Visitor<'de> for SchemaIdVisitor {
...
+        while let Some(seq_value) = seq.next_element::<String>()? {
+            match seq_value.parse::<OperationId>() {
+                Ok(operation_id) => op_ids.push(operation_id),
+                Err(hash_err) => {
+                    return Err(serde::de::Error::custom(format!(
+                        "Error parsing application schema id: {}",
+                        hash_err
+                    )))
+                }
+            };
```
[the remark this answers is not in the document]
adzialocha: Ah interesting, I'm not sure if I understand it correctly now, but does your change still check for the validity of a hash? Deserializing something is simply just moving a string into `Hash` without checking if it fullfils the format of YASMF. This is where `validate` comes in to prevent a random string to move in. Do you account somewhere else for this now? Or maybe I'm getting it wrong ..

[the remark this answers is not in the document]
cafca: Yes, previously the `Hash` constructor would fail for invalid values because it invokes validation. The previous code was invoking `validate()` again separately afterwards but those calls would always go through because the validity had already been checked in the constructor.

adzialocha (approved):
Very nice refactoring! I especially like the docstring changes, thank you for also doing all of this clean-up work!

cafca:
Let me fix Sams suggestions before you merge!

p2panda-rs/src/schema/schema_id.rs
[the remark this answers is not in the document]
adzialocha: `while let Some(hash) = seq.next_element::<Hash>()? {` is not calling a constructor though if I understand it correctly.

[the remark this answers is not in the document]
adzialocha: Created an issue for the general question: https://github.com/p2panda/p2panda/issues/276

Issue #223, opened 2022-02-17
Traits for storage


adzialocha:
`p2panda-rs` is currently designed without any explicit storage in mind, leaving it to the developers to come up with their own way of dealing with it. While we still want to keep the flexibility of chosing your own storage backend solution we could still provide traits around reading and writing certain types of data.

This is especially interesting for some sort of "high-level" data API and some nice code one can write in `p2panda-rs`. We could expect certain trait implementations delivering the expected data for different actions.

Examples:

* Load the key pair to sign a new entry
* Load a MLS Group state to decrypt an operation field
* Get backlink and skiplink entries from a log to check the integrity of an entry
* Load materialized document view of a schema to use it for validation

Related: https://github.com/p2panda/p2panda/issues/148

Issue #253, opened 2022-03-07
Make `p2panda-js` and `p2panda-rs` stateless?


adzialocha:
`p2panda-rs` in itself is currently designed as a stateless framework. Storage and state is something developers need to implement on top of it (we do it in `aquadoggo` and in our mocked node / client implementations). In `p2panda-js` it is slightly different, there we already have a notion of an external data provider (making requests) and a little bit of caching. Also, in `p2panda-js` we have a more "wrapper" sort of API.

Obviously to build an application on top of p2panda one needs state so the question is now where to draw the borders. Also the idea of having some sort of storage notion in `p2panda-rs` is coming up (https://github.com/p2panda/p2panda/issues/223) but it feels like a fairly large refactoring with potentially ugly API outcomes. 

Random thoughts:

* Remove all state and higher-level API from `p2panda-js`, make it as "agnostic" and low-level as `p2panda-rs`, mostly just being a simple wrapper around the WebAssembly exports
* Trying to get it to the same feature-set as `p2panda-rs` (so one could imagine building an JavaScript Node implementation one day with it)
* Stateful things will be implemented in form of a new framework, in a new repository, like a more high-level Rust and JavaScript library using `p2panda-js` and `p2panda-rs` internally. For now it could be as simple as moving everything we currently have in `p2panda-js` into this new repository
* `p2panda-js` and `p2panda-rs` become the "core" libraries. The new things are on top of that, providing everything you need to build a) a client b) a node - in both programming languages
* Storage providers can still be abstract in these high-level frameworks, in Rust we can express it in traits, in JavaScript there are surely some similarly nice API patterns we can use
* The code-bases don't get to crammed with different purposes, we work on "core specification" stuff in the core libraries and everything else happens somewhere else: State, Storage, Networking, Caching, etc.
* We don't need to refactor a lot of code to introduce storage anymore, `p2panda-rs` can stay the same, the new library will build on top of it

adzialocha:
With storage traits we could refactor p2panda to consist of 3 types of packages / crates:

1. core functionality (stateless) `p2panda-rs`
2. trait definitions `p2panda-traits`
3. framework to build clients / nodes with (including networking, storage implementations, etc.) `aquadoggo` or `shushek` or other funny names we can come up with

adzialocha:
Thoughts on why it might be nice to separate it:

* Separation of concerns, it gets very clear what a code-base should do and not should do
* Less code
* Less dependencies
[...]

PR #274, opened 2022-03-15
Storage provider traits for implementing persistent storage solutions

sandreae:
Traits which abstract over the functionality required for persisting and querying entries and logs to/from a storage solution in a p2panda implementation. It is storage method agnostic. New types `EntryWithOperation` & `Log` have been introduced which act as known representations of entries and logs when retrieved or before being stored in a database. They act as couterparts to the user defined types which will implement `AsStorageLog` & `AsStorageEntry` traits (these types may be stored in a different format to our `p2panda-rs` types, like in `aquadoggo`, we store many things a strings. `StoragePrivider` is the main trait which encapsulates all requests to the implementers storage solution. 

## `AsStorageLog` & `AsStorageEntry`
- traits traits to be implemented on types which represent logs and entries in storage

## `LogStorage` & `EntryStorage`
- traits which contain methods related to each data types storage and querying.
- most of these methods are "lower level" and should be defined for the specific storage solution an implementer is using (or if you're feeling fresh, you could use multiple storage solutions ;-p))
- _we may want to look at what goes where here, I'm not really confident about the way I've devided the methods_

## `StorageProvider`
- This is the top level trait which encapsulates all higher level storage request methods (`publish_entry`, `next_entry_args` etc..)
- Most of these have default implementations which come for free when all lower level trait requirements are satisfied.

## Notes:
- methods, and the content of the default implementations, for all storage traits are copied from our implementation in `aquadoggo`, we may want to review these, but I thought it might be good to that in a follow-up PR.

Closes: #223 

## 📋 Checklist

## Post-merge issues

- general validation review
  - validate that entries at seq nums that require skiplinks have them in publish entry requests.
- look over default implementation of publish entry method
  - there are some errors which will never trip
- consider a more general query pattern
- explore alternative crate structuring


p2panda-rs/src/storage_provider/traits/storage_provider.rs
```diff
@@ -0,0 +1,484 @@
...
+
+        // Get related bamboo backlink and skiplink entries
+        let entry_backlink_bytes = if !entry.seq_num().is_first() {
+            self.entry_at_seq_num(&author, entry.log_id(), &entry.seq_num_backlink().unwrap())
+                .await?
+                .map(|link| {
+                    let bytes = link.entry_encoded().to_bytes();
+                    Some(bytes)
+                })
+                .ok_or(PublishEntryError::BacklinkMissing)
```
sandreae, line 162: I don't think this error would ever be triggered as the one above in line 137 would always be tripped first.

p2panda-rs/src/storage_provider/traits/storage_provider.rs
```diff
@@ -0,0 +1,484 @@
...
+            // @TODO: This currently looks at the backlink, in the future we want to use
+            // "previousOperation", since in a multi-writer setting there might be no backlink for
+            // update operations! See: https://github.com/p2panda/aquadoggo/issues/49
+            let backlink_entry_hash = entry
+                .backlink_hash()
+                .ok_or(PublishEntryError::OperationWithoutBacklink)?;
+
+            self.get_document_by_entry(backlink_entry_hash)
+                .await?
+                .ok_or(PublishEntryError::DocumentMissing)?
```
sandreae, line 137: This error trips if the backlink is missing, not necessarily the actual document. With the move to checking previous operations this will change, but still relevant to keep in mind.

p2panda-rs/src/storage_provider/traits/entry_store.rs
```diff
@@ -0,0 +1,295 @@
...
+    use async_trait::async_trait;
+    use rstest::rstest;
+    use std::sync::{Arc, Mutex};
+
+    use crate::entry::{sign_and_encode, Entry, EntrySigned, LogId, SeqNum};
+    use crate::identity::{Author, KeyPair};
+    use crate::operation::{AsOperation, OperationEncoded};
+    use crate::schema::SchemaId;
+    use crate::storage_provider::errors::EntryStorageError;
+    use crate::storage_provider::traits::test_setup::{
```
adzialocha, line 84: Maybe we call that module `test_utils`? It's consistent then with how we name it in other places as well.

p2panda-rs/src/storage_provider/traits/log_store.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+}
+
+#[cfg(test)]
+pub mod tests {
+    use std::convert::TryFrom;
+
+    use async_trait::async_trait;
+    use rstest::rstest;
+    use std::sync::{Arc, Mutex};
```
adzialocha, line 60: Miau

p2panda-rs/src/storage_provider/traits/log_store.rs
```diff
@@ -0,0 +1,150 @@
...
+    use crate::test_utils::fixtures::{document_id, key_pair, schema};
+
+    /// Implement the `LogStore` trait on SimplestStorageProvider
+    #[async_trait]
+    impl LogStore<StorageLog> for SimplestStorageProvider {
+        async fn insert_log(&self, log: StorageLog) -> Result<bool, LogStorageError> {
+            let mut logs = self.logs.lock().unwrap();
+            logs.push(log);
+            // Remove duplicate log entries.
+            logs.dedup();
```
adzialocha, line 79: See entry storage implementation in my comment above

p2panda-rs/src/storage_provider/traits/models.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
+
+/// Trait to be implemented on a struct representing a stored log.
+///
+/// Should be defined within a specific storage implementation on a data struct which
+/// represents a log as it is stored in the database. This trait defines methods for
+/// reading values from the log.
+pub trait AsStorageLog: Sized + Send + Sync + TryInto<Log> + From<Log> {
+    /// Constructor method for struts
```
adzialocha, line 44: Typo

adzialocha (commented):
Didn't get through it all yet, but already very exciting! Just a first bunch of smaller comments for now.

Random thought: Why not move it all into a new `p2panda_traits` crate?

p2panda-rs/src/storage_provider/traits/entry_store.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    /// Implement `EntryStore` trait on `SimplestStorageProvider`
+    #[async_trait]
+    impl EntryStore<StorageEntry> for SimplestStorageProvider {
+        /// Insert an entry into storage.
+        async fn insert_entry(&self, entry: StorageEntry) -> Result<bool, EntryStorageError> {
+            let mut entries = self.entries.lock().unwrap();
+            entries.push(entry);
+            // Remove duplicate entries.
+            entries.dedup();
```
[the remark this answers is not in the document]
sandreae: yeh, true, this is a bit weird even for a test impl... Not sure quite what you mean by putting it into a new method, but I'd rather just swap out the vec based storage for a hashmap where duplicates would just overwrite or some kind of set where duplicates would fail. How does that sound to you?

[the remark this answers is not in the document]
sandreae: What do you think of this test setup in general? I know we will have an in-memory `StorageProvider` implementation, but for the tests I feel it's better to have something like this which is very simple and transparent. Detached from the rest of the code base.

p2panda-rs/src/storage_provider/models.rs
```diff
@@ -0,0 +1,103 @@
...
+use crate::identity::Author;
+use crate::operation::OperationEncoded;
+use crate::schema::SchemaId;
+use crate::Validate;
+
+/// Struct wrapping an entry with it's operation.
+///
+/// Used internally throughout `storage_provider` in method args and default trait definitions.
+/// The `AsStorageEntry` trait requires `TryFrom<EntryWithOperation>` & `TryInto<EntryWithOperation>`
+/// conversion traits to be present.
```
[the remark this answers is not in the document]
sandreae: I was trying to emphasise that this is a type which is only used internally within the `storage_provider` rather than a general use p2panda type. Maybe that's obvious?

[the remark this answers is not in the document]
sandreae: Slightly off topic, but does the pattern between in memory and db types make sense to you? 

`EntryWithOperation` <-> `StorageEntry` & `Log` <-> `StorageLog`

sandreae:
> Random thought: Why not move it all into a new p2panda_traits crate (living in the same repo)?

Yeh, I like the idea. Does it make sense for _everything_ to go across into something like that though. I wonder if `EntryWithOperation` and `Log` should remain in the main crate?

sandreae:
In any case, I had imagined all of this would eventually live in the higher level crate we chatted about a bit (as opposed to "core" which is what we now work on mostly) if we go in that direction.

p2panda-rs/src/storage_provider/traits/entry_store.rs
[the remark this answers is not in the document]
adzialocha: I would have introduced a method in the storage provider implementation wrapping that functionality:

```rust
Impl SProvider {
  Pub fn insert(entry) {
    Self.push
    Self.dedupe
  }
}
```

HashMaps are great ofc as well! Thought you had a reason to use vecs 👍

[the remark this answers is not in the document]
sandreae: Ah, ok, cool, yes that's nice. I didn't have a very special reason for vecs, I was having fun representing all the data as strings, then thought just pushing the in a vec was the simplest storage I could think of. ;-p

adzialocha:
> > Random thought: Why not move it all into a new p2panda_traits crate (living in the same repo)?
> 
> Yeh, I like the idea. Does it make sense for _everything_ to go across into something like that though. I wonder if `EntryWithOperation` and `Log` should remain in the main crate?

Yes, they could remain in the core library, think it's a good separation 👍

adzialocha:
> In any case, I had imagined all of this would eventually live in the higher level crate we chatted about a bit (as opposed to "core" which is what we now work on mostly) if we go in that direction. 

Maybe even in three crates I'm thinking.

1. core types
2. traits
3. frameworks (state, storage, networking)

cafca:
I would also like to review this! I have some general answers about how this should be used and how we will work with this as we change the rest of the library, maybe we can have a chat on Monday during the day?

cafca:
What advantage do you see in separating these crates out @adzialocha ? I kind of like the swiss army knife idea of having _everything a panda needs_ in one package.

sandreae:
> I would also like to review this! I have some general answers about how this should be used and how we will work with this as we change the rest of the library, maybe we can have a chat on Monday during the day?

Yep, this is definitely in the "shiny new thing" category 😅 no rush to merge and I'm very open to different approaches 👍

sandreae:
I can't come on Monday though 😞 will be travelling in the evening (visiting family in the day). Feel free to discuss / jam on it though 👍

adzialocha:
> What advantage do you see in separating these crates out @adzialocha ? I kind of like the swiss army knife idea of having _everything a panda needs_ in one package.

Let's discuss this maybe outside of this PR? This might be a good place for it: https://github.com/p2panda/p2panda/issues/253

p2panda-rs/src/storage_provider/models.rs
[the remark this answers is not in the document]
adzialocha: > I was trying to emphasise that this is a type which is only used internally within the `storage_provider` rather than a general use p2panda type. Maybe that's obvious?

Its a relatively useful struct, how about moving it into `entry` or `operations` module?

Related: https://github.com/p2panda/p2panda/issues/198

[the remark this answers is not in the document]
adzialocha: > Slightly off topic, but does the pattern between in memory and db types make sense to you?

Yes! It's great!

p2panda-rs/src/storage_provider/models.rs
```diff
@@ -0,0 +1,99 @@
...
+impl EntryWithOperation {
+    /// Instantiate a new EntryWithOperation.
+    pub fn new(entry: &EntrySigned, operation: &OperationEncoded) -> Result<Self, ValidationError> {
+        let entry_with_operation = Self(entry.clone(), operation.clone());
+        entry_with_operation.validate()?;
+        Ok(entry_with_operation)
+    }
+
+    /// Returns a reference to the encoded entry.
+    pub fn entry_encoded(&self) -> &EntrySigned {
```
adzialocha, line 31: Maybe we call this `entry_signed`? It is also encoded but knowing the p2panda API (maybe too much :sweat_smile:), I expected it to have the same name as the struct.

p2panda-rs/src/storage_provider/models.rs
```diff
@@ -0,0 +1,98 @@
...
+        self.entry_encoded().validate()?;
+        self.operation_encoded().validate()?;
+        decode_entry(self.entry_encoded(), Some(self.operation_encoded()))?;
+        Ok(())
+    }
+}
+
+/// Struct representing a bamboo append-only log structure.
+#[derive(Debug, Clone)]
+pub struct Log {
```
adzialocha, line 54: Same as `EntryWithOperation` I could imagine moving `Log` into the core modules as well, as they might be useful for someone else to work with?

adzialocha:
I'm slowly making my way through this 😅 it's very well written, it just takes some time to understand the API. That doesn't mean it's poorly designed, just need time or an introduction sometime.

sandreae:
No worries! I'm happy to do a walk-through any time, and if there are any design decisions that could be done differently I'm happy to look at it again 👍

p2panda-rs/src/storage_provider/models.rs
```diff
@@ -0,0 +1,98 @@
...
+        &self.0
+    }
+
+    /// Returns a refernce to the optional encoded operation.
+    pub fn operation_encoded(&self) -> &OperationEncoded {
+        &self.1
+    }
+}
+
+impl Validate for EntryWithOperation {
```
adzialocha, line 41: This has a nice effect of bringing validation to a more "official" level.

Looking forward revisiting this in the future and writing tests and specification around all sorts of data validation!

p2panda-rs/src/storage_provider/traits/entry_store.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    /// Returns the latest Bamboo entry of an author's log.
+    async fn latest_entry(
+        &self,
+        author: &Author,
+        log_id: &LogId,
+    ) -> Result<Option<StorageEntry>, EntryStorageError>;
+
+    /// Return vector of all entries of a given schema.
+    async fn by_schema(&self, schema: &SchemaId) -> Result<Vec<StorageEntry>, EntryStorageError>;
```
adzialocha, line 37: This could also be merged together with other query methods to one large `query(filter: EntryQuery)` (or similar) function where:

```rust
struct EntryQuery {
  schemas: Option<Vec<DocumentViewId>>,
  seq_nums: Option<Vec<SeqNum>>,
  authors: ...
  ...
}
```

I don't think it's necessary, just wanted to drop that idea here as it's reminding me of Aljoschas "ReqRes".

adzialocha: Upside with that is that it could be easier to introduce pagination like this one day for future queries like browsing all schemas etc.

adzialocha (changes requested):
I think I covered it now! Really good work wrapping your head around this, I like the approach 👍

Some formatting clean ups I comitted directly myself (hope that was alright), otherwise I have some comments on documentation and the entry API as it would be nice to be more agnostic to implementation details on this level.

p2panda-rs/src/storage_provider/traits/storage_provider.rs
```diff
@@ -0,0 +1,494 @@
...
+use crate::hash::Hash;
+use crate::operation::{AsOperation, Operation};
+use crate::storage_provider::errors::PublishEntryError;
+use crate::storage_provider::models::{EntryWithOperation, Log};
+use crate::storage_provider::traits::{
+    AsEntryArgsRequest, AsEntryArgsResponse, AsPublishEntryRequest, AsPublishEntryResponse,
+    AsStorageEntry, AsStorageLog, EntryStore, LogStore,
+};
+
+/// Trait which handles all high level storage queries and insertions.
```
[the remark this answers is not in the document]
sandreae: :pray: yeh, sorry, my bad for skipping this. I thin this PR became ready way before I expected it to be ;-p

sandreae:
Thanks for the review! I will get onto all this asap.

I have a question about the trait naming actually. I tried to follow the convention to name traits with the prefix `As` (`AsStorageEntry` etc..) however the top level store traits `EntryStore`, `Log`Store` and `StorageProvider` don't have it. This made sense at the time, but it's clearly just inconsistent... Could you give a second opinion on this naming. Is the `As` prefix nice? So should be used everywhere? Or, is it not needed, as these are all obviously traits, and should be removed everywhere?

p2panda-rs/src/storage_provider/traits/entry_store.rs
sandreae: Yeh, this would be great! There are a few things that can be raised as issues for after this PR is merged. I'll add this to the list.

sandreae:
Just so I don't forget, issues to raise after this PR is merged:
- validation review
- look over default implementation of publish entry method
- consider a more general query pattern

adzialocha:
> I have a question about the trait naming actually. I tried to follow the convention to name traits with the prefix `As` (`AsStorageEntry` etc..) however the top level store traits `EntryStore`, `LogStore` and `StorageProvider` don't have it. This made sense at the time, but it's clearly just inconsistent... Could you give a second opinion on this naming. Is the `As` prefix nice? So should be used everywhere? Or, is it not needed, as these are all obviously traits, and should be removed everywhere?

I didn't particularly stumble upon the naming of the traits but I can try to see what comes into my mind playing with it a little bit:

🤔

- `StorageProvider` offers an interface for a storage solution
- `LogStore` offers an interface to read and write logs from a storage
- `EntryStore` same but for entries
- `AsStorageEntry` / `StorableEntry` / `EntryModel` data to be stored in entry store
- `AsStorageLog` data to be stored in log store

Meh! Can't come up with anything convincing! 🤣 I think the current naming is really fine?

sandreae:
hehe, cool, if it wasn't a particular point of confusion then it must be ok. The simplest way to make it consistent would be to use `AsStorageProvider` `AsEntryStore` `AsLogStore` but somehow this doesn't seem nice to me....

p2panda-rs/src/storage_provider/traits/storage_provider.rs
```diff
@@ -0,0 +1,494 @@
...
+
+    /// Returns the related document for any entry.
+    ///
+    /// Every entry is part of a document and, through that, associated with a specific log id used
+    /// by this document and author. This method returns that document id by looking up the log
+    /// that the entry was stored in.
+    async fn get_document_by_entry(
+        &self,
+        entry_hash: &Hash,
+    ) -> Result<Option<DocumentId>, Box<dyn std::error::Error>>;
```
cafca, line 44: Rust noob question: what is this strange error type and why doesn't this use an error type defined in `errors.rs` like the other structs of the module?

cafca: Could be helpful to add a note to the docstring and describe under what conditions it would be okay to return `None`

p2panda-rs/src/storage_provider/traits/requests.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use crate::document::DocumentId;
+use crate::entry::EntrySigned;
+use crate::identity::Author;
+use crate::operation::OperationEncoded;
+use crate::storage_provider::ValidationError;
+use crate::Validate;
+
+/// Request body of `panda_getEntryArguments`.
+pub trait AsEntryArgsRequest {
```
cafca, line 11: Could name this and the other request/response traits with an imperative like `Validate` or `Copy`: `RequestEntryArgs` / `RespondEntryArgs`

p2panda-rs/src/storage_provider/traits/requests.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// Returns the Author parameter.
+    fn author(&self) -> &Author;
+
+    /// Returns the document id Hash parameter.
+    fn document(&self) -> &Option<DocumentId>;
```
cafca, line 16: Renaming to `document_id` would improve clarity

p2panda-rs/src/storage_provider/traits/responses.rs
```diff
@@ -0,0 +1,36 @@
...
+use crate::entry::{LogId, SeqNum};
+use crate::hash::Hash;
+
+/// Response when requesting arguments to create a new entry.
+#[async_trait]
+pub trait AsEntryArgsResponse {
+    /// Just the constructor method is defined here as all we need this trait for is constructing
+    /// entry arguments to be returned from the default trait methods.
+    ///
+    /// NB: We could validate that entries at seq nums that require skiplinks have them.
```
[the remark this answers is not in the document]
cafca: Does this save us from validating that in some other place?

p2panda-rs/src/storage_provider/traits/storage_provider.rs
cafca: Could be transformed to `PublishEntryError::BacklinkMissing` here.

cafca (changes requested):
I just made a first pass today as this is a lot to unpack. As I said before I like the underlying idea of this PR very much!

Even though it totally makes sense in the context of this PR I am a bit confused about all the different structs we have now of different combinations of entries and operations. I would really like to get this cleaned up a bit soon.

I am really excited how nice these traits will make extending the API in the future if we have multiple node implementations as they just have to implement the traits!

p2panda-rs/src/storage_provider/traits/models.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// Return the encoded entry.
+    fn entry_signed(&self) -> EntrySigned;
+
+    /// Returns the optional encoded operation.
+    fn operation_encoded(&self) -> Option<OperationEncoded>;
+
+    /// Returns the decoded operation.
+    fn entry_decoded(&self) -> Entry {
+        // Unwrapping as validation occurs in `EntryWithOperation`.
+        decode_entry(&self.entry_signed(), self.operation_encoded().as_ref()).unwrap()
```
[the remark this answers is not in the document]
sandreae: It's unwrapping here working under some assumptions:

- structs which impl `AsStorageEntry` don't have a constructor
- they are only created via conversion from `EntryWithOperation` (which validates) 
- stored entries and operations coming from the db should be validated already

This way we avoid having to handle error situations every time we access entry values on a stored entry struct.

Sorry, i didn't explain any of this in the code....

Not sure if it's a sound approach, but I like the goal.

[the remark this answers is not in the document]
sandreae: (this is why `EntryWithOperation` decods the entry in it's constructor validation)

p2panda-rs/src/storage_provider/traits/requests.rs
sandreae: Hmmm... yeh that is nice.

p2panda-rs/src/storage_provider/traits/responses.rs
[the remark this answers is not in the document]
sandreae: Hmmm, not at the moment, as we do this ourselves in the default entry args method which is the only place a struct with this trait is constructed..... I suppose it would be to protect against implementors making mistakes if they use different patterns when constructing their entry arg responses.

Maybe this can be looked at in "The Big Validation Review".

p2panda-rs/src/storage_provider/traits/requests.rs
adzialocha: But its not really an action, isn't it? There is nothing "returning" in that trait (like "validating" or "copying") but only definining the shape of what is returned. I'd keep the trait name as is :+1:

p2panda-rs/src/storage_provider/errors.rs
```diff
@@ -0,0 +1,90 @@
...
+    #[error(transparent)]
+    BambooValidation(#[from] bamboo_rs_core_ed25519_yasmf::verify::Error),
+}
+
+/// `LogStorage` errors.
+#[derive(thiserror::Error, Debug)]
+pub enum LogStorageError {
+    /// Catch all error which implementers can use for passing their own errors up the chain.
+    #[error("Error occured during `LogStorage` request in storage provider: {0}")]
+    Custom(String),
```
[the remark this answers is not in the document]
adzialocha: I've seen `Custom` being used in other frameworks as well. "Implementation" is also misleading, if you think that in Rust everything is an implementation :joy_cat: That being said, if we find something better than `Custom` I'm also fine with it :+1:

p2panda-rs/src/storage_provider/traits/models.rs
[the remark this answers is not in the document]
cafca: Ah, I see. But then we decode twice, once in the constructor/validation and once when we actually access the value. I made a comment on the validation method for this as well. I feel like that could be avoided?

p2panda-rs/src/storage_provider/errors.rs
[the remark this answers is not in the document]
cafca: I feel like Rust should have language for this built in.

p2panda-rs/src/storage_provider/traits/requests.rs
cafca: I'm also fine with the current name. I felt it would be nice to use a name that doesn't just wrap the name of our standard implementation.

p2panda-rs/src/storage_provider/traits/models.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::schema::SchemaId;
+use crate::storage_provider::models::{EntryWithOperation, Log};
+
+/// Trait to be implemented on a struct representing a stored entry.
+///
+/// Should be defined within a specific storage implementation on a data struct which represents an
+/// entry as it is stored in the database. This trait defines methods for reading values from the
+/// entry and it's operation and ensures the required conversion (to and from `EntryWithOperation`)
+/// are present.
+pub trait AsStorageEntry:
```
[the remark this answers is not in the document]
sandreae: Okay, I added getters to the trait `AsStorageEntry` for all values that we currently need. I really like the interface now, it results in much neater and easier to read code i think. 

Regarding your point @cafca I agree that the signed encoded entry will be what people are handling under the hood, I just really like how the trait interface spells out just exactly what is needed, and unifies decoded and signed-encoded entry values in one place. This doesn't limit implementors in adding other getters of course. See what you think, happy to add getters for the "raw" data back in if you'd like.

p2panda-rs/src/storage_provider/traits/storage_provider.rs
sandreae: This is a lovely boxed error, which means it can be anything ;-p box is a smart pointer thingy.

I'm not sure it's the right pattern here actually though, i wanted to allow for implementors to be able to return their own errors from these higher level methods. I think we could use a similar "error wrapping" technique as in the lower level traits as well, but this seemed to work a little better for me.

p2panda-rs/src/storage_provider/models.rs
```diff
@@ -0,0 +1,98 @@
...
+    type Error = ValidationError;
+
+    fn validate(&self) -> Result<(), Self::Error> {
+        self.entry_signed().validate()?;
+        self.operation_encoded().validate()?;
+        decode_entry(self.entry_signed(), Some(self.operation_encoded()))?;
```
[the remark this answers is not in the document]
sandreae: Think this is resolved as we need to do it here as we are assuming it has been validated before converting into storage data type.

p2panda-rs/src/storage_provider/traits/models.rs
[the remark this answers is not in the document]
adzialocha: The raw, encoded data will be needed for replication but I think after decoding it I imagine you move all data (seq_num, author etc.) into sql columns (or similar) and query this instead. Then you only need to decode once 👍

sandreae:
hm... all tests pass for me locally ;-p

sandreae:
It's strange, everything is failing in the CI but not locally for me. I tried re-running them but the same happened. Could someone else test this branch locally? Thanks!

adzialocha:
> It's strange, everything is failing in the CI but not locally for me. I tried re-running them but the same happened. Could someone else test this branch locally? Thanks!

You can try to re-run the CI to see if it was a hiccup or service issue, otherwise there might be a mismatch between some versions?

adzialocha (approved):
This looks good to me now! There are some more things obviously but I think they can wait until we learned more, until then this is already very exciting! Definitely pancakes time!

p2panda-rs/src/storage_provider/traits/responses.rs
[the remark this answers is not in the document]
sandreae: I'm adding this to the review issue :+1:

p2panda-rs/src/storage_provider/traits/responses.rs
```diff
@@ -0,0 +1,36 @@
...
+    ) -> Self;
+}
+
+/// Response when publishing a new entry.
+#[async_trait]
+pub trait AsPublishEntryResponse {
+    /// Just the constructor method is defined here as all we need this trait for is constructing
+    /// entry args to be returned from the default trait methods.
+    ///
+    /// NB: We could validate that entries at seq nums that require skiplinks have them.
```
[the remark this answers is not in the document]
sandreae: I'm adding this to the review issue :+1:

sandreae:
Sooo, I wonder if there should be a change to this implementation (it could happen before or after merging this PR). While implementing these traits in aquadoggo i realised maybe there's a better pattern. Right now we have `AsStorageEntry` and `AsStorageLog` which are required to be implemented on implementation structs which pass in and out of storage, they in tern require that conversion traits to and from p2panda types exist for these structs _and_ contain all getters for the fields we need when interacting with these types within the storage provider trait methods. I'm now wondering if we don't actually need these getters here... they could be moved onto eg. `EntryWithOperation` and we only assume we can convert any implementation specific storage type into our own defined types then get the fields we want there.

The problem this solves in `aquadoggo` is that we have `EntryRow` (which is all string fields) and `Entry` (which contains typed fields), the storage provider traits remove the need for both of these structs to exist and what we want to do is implement `AsStorageEntry` on `EntryRow` but then every field getter will be unwrapping string conversions, and it will be a little ugly. If we make `EntryWithOperation` a richer struct, we can simply convert into it from `EntryRow` and use the getters defined there  instead.

I'm fairly sure this is a better way to go, but wonder if there is something we lose I'm not seeing...

adzialocha:
> I'm fairly sure this is a better way to go, but wonder if there is something we lose I'm not seeing...

Where do these conversions come from exactly?

sandreae:
They're required in the `AsStorageEntry` definition here:

https://github.com/p2panda/p2panda/blob/d77caca3d577e4fdf418b3930ae3194a188776c4/p2panda-rs/src/storage_provider/traits/models.rs#L20-L21

sandreae:
Maybe you can see what I mean about the not nice pattern in the aquadoggo implementation here: https://github.com/p2panda/aquadoggo/blob/storage-provider/aquadoggo/src/db/models/entry.rs

sandreae:
I think we wouldn't need `AsStorageEntry` and `AsStorageLog` traits anymore then.... we would just require the `TryInto` and `From` conversion traits in `LogStore` and `EntryStore` :scream: (it seems to simple)

sandreae:
(don't worry if this is too vague to grok right now, i can also just make the changes and you see what it looks like after, it's not tonnes of work)

adzialocha:
I don't find it too horrible really 🙃 the ugly part of this is not the storage API but the fact that we can't store bigint in SQLite, that's why we have Entry and EntryRow.

Making `OperationWitgEntry` the main type breaks a little bit the idea of traits .. the struct becomes the contract but structs are implementations and they tend to change more often than traits. I don't want to say one can't build a nice API like that, but it feels counter intuitive in it's purpose.

adzialocha:
> (don't worry if this is too vague to grok right now, i can also just make the changes and you see what it looks like after, it's not tonnes of work)

Yeah, why not! Maybe it's worth checking it out, that was just my intuition above ✌️

adzialocha:
Is it possible to remove the conversion completely? 🤪 Now when I look at it the storage trait should be agnostic to any structs to really stay flexible?

adzialocha:
The less we tie developers to our own types the better I think 👍 that's were traits are great!

sandreae:
> Making OperationWithEntry the main type with special requirements breaks a little bit the idea of traits .. the struct becomes the contract but structs are implementations and they tend to change more often than traits, also they usually are used in multiple contexts, probably outside of storage. I also have the feeling that it will become too powerful and therefore bloated? I don't want to say one can't build a nice API like that, but it feels counter intuitive in it's purpose.

Yeh, this was a worry I had too which made me hesitate.

sandreae:
> Is it possible to remove the conversion trait requirements completely? 🤪 Now when I look at it, the storage trait should be agnostic to any structs to really stay flexible?

Yeh, let me check this, the traits evolved somewhat since the beginning of this PR, so it might be possible.

sandreae:
Thanks for the comments! It may be something we keep an eye on as we move forward and use the traits more. It's not a blocker for this PR 👍

sandreae:
Ok, had a go at this and managed to completely remove any reference to p2panda types `Log` and `EntryWithOperation`. To do this i took the following steps:

- Add constructor methods to `AsStorageEntry` and `AsStorageLog` traits
- Remove conversion trait dependencies and instead use the new constructor methods when instantiating new storage structs
- Introduce a dependency for several storage traits on p2panda `Validate` (we can call this in default trait method definitions)

sandreae:
This means we can completly remove `EntryWithOperation` and `Log` from this PR :+1: (I have them still there but commented out just for reference)

adzialocha:
Yeah! This direction looks really good! ✨👍🐇

sandreae:
Coool, this feels very good now, we are ready to go!

@cafca if you can have even a quick look over before mergy mergy that would be cool as this is a "shiny new thing"

adzialocha:
> Coool, this feels very good now, we are ready to go!

Very nice, for me it's definitely ready!
 
Interesting to see how validation unfolds as well 👍

sandreae:
> Interesting to see how validation unfolds as well

Yeh! Requiring our `Validate` trait as a dependency in the storage traits felt like a very good direction.

sandreae:
Updated implementation in `aquadoggo` can be merged after this PR: https://github.com/p2panda/aquadoggo/pull/80

p2panda-rs/src/storage_provider/traits/storage_provider.rs
cafca: But the docstring should explain the valid conditions for a `None` return value don't you think? At least it's not obvious for me from this alone.

p2panda-rs/src/storage_provider/traits/storage_provider.rs
```diff
@@ -0,0 +1,512 @@
...
+
+    /// Returns required data (backlink and skiplink entry hashes, last sequence number and the
+    /// document's log_id) to encode a new bamboo entry.
+    async fn get_entry_args(
+        &self,
+        params: &Self::EntryArgsRequest,
+    ) -> Result<Self::EntryArgsResponse, Box<dyn std::error::Error>> {
+        // Validate the entry args request parameters.
+        params.validate().map_err(|error| format!("{:?}", error))?;
```
cafca, line 68: So the errors returned are just strings?

p2panda-rs/src/storage_provider/traits/storage_provider.rs
```diff
@@ -0,0 +1,512 @@
...
+            // For any other operations which followed after creation we need to either walk the operation
+            // graph back to its `CREATE` operation or more easily look up the database since we keep track
+            // of all log ids and documents there.
+            //
+            // We can determine the used document hash by looking at what we know about the previous
+            // entry in this author's log.
+            //
+            // @TODO: This currently looks at the backlink, in the future we want to use
+            // "previousOperation", since in a multi-writer setting there might be no backlink for
+            // update operations! See: https://github.com/p2panda/aquadoggo/issues/49
```
cafca, line 130: What is stopping us?

cafca (changes requested):
Found some small things in docstrings etc. If you could fill this in Sam we can merge away :)

p2panda-rs/src/storage_provider/errors.rs
```diff
@@ -0,0 +1,96 @@
...
+    SkiplinkMissing,
+
+    /// Error returned from validating p2panda-rs `EntrySigned` data types.
+    #[error(transparent)]
+    ValidationError(#[from] ValidationError),
+}
+
+/// Errors which can occur when publishing a new entry.
+#[derive(thiserror::Error, Debug)]
+#[allow(missing_copy_implementations, missing_docs)]
```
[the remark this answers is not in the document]
adzialocha: (For whatever reason I can't reply to the comment above)

Re: `Error::Custom` I found the example in `serde` where they use a similar pattern: https://github.com/serde-rs/serde/blob/996d171461f27cd935aed473ad318bbbc6e5c2cc/serde/src/ser/mod.rs#L175=

p2panda-rs/src/storage_provider/traits/storage_provider.rs
sandreae: I don't think there's anything stopping us. I tried to change as little of the inner code in all these existing methods as possible (they all came over from `aquadoggo`) as I wanted to limit the scope of this PR to the restructuring of our API into traits rather than the implementation details. Maybe i shouldn't have done that, but it seemed like a useful limit when i started.

p2panda-rs/src/storage_provider/traits/storage_provider.rs
sandreae: In this case, yeh.. it's a string wrapped in an `Error`, I had some troubles coercing the validation error types for unknown reasons.... I'll try and fix before merging. 

In general, this method can return any implementation defined error type as long as they implement the `Error` trait.

p2panda-rs/src/storage_provider/traits/storage_provider.rs
```diff
@@ -0,0 +1,512 @@
...
+            let backlink_entry_hash = entry
+                .backlink_hash()
+                .ok_or_else(|| PublishEntryError::OperationWithoutBacklink(entry.hash()))?;
+
+            self.get_document_by_entry(&backlink_entry_hash)
+                .await?
+                // @TODO this trips if the backlink is missing not necessarily the document
+                // needs revisiting when we use previous_operations to retrieve the correct
+                // document.
```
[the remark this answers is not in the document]
sandreae: ??

p2panda-rs/src/storage_provider/traits/storage_provider.rs
sandreae: Ok, if we add some more constraints to our existing `Validate` trait definition then we can correctly coerce the returned error type. Now it looks like this:

https://github.com/p2panda/p2panda/blob/372f2721eb6721abe21b05743197da00cd671d5c/p2panda-rs/src/lib.rs#L82-L89
`

sandreae: (`Error` and `'static` added)

sandreae: Not sure if adding the 'static lifetime is a bit of a hack... I'm never sure about that.

p2panda-rs/src/storage_provider/traits/responses.rs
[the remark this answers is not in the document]
sandreae: Maybe this note is a bit redundant actually, as is it more of an implementation decision regarding what validation occurs in this constructor? I suggest opening an issue in `aquadoggo`.

p2panda-rs/src/storage_provider/traits/storage_provider.rs
sandreae: Ok... against my own advice, I've now implemented following `previous_operations` to find the doucment id instead of the backlink. I had to add some extra validation in `Operation` to do this.

p2panda-rs/src/storage_provider/traits/storage_provider.rs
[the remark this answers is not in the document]
adzialocha: Its a little bit hard to read, I think the problem might be the missing punctuation between the lines :joy_cat: 

How about?

> This error seems to occur when the backlink is missing instead of the document as stated. It should be revisisted when we use previous_operations to retrieve the correct document.

p2panda-rs/src/storage_provider/traits/storage_provider.rs
adzialocha: Looks like the standard `Error` trait _needs_ to be `'static`: https://doc.rust-lang.org/std/error/trait.Error.html

cafca (approved):

p2panda-rs/src/storage_provider/traits/storage_provider.rs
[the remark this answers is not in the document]
sandreae: Ah! So punctuation _is_ important after all 🤣 who knew?!

PR #277, opened 2022-03-17
Provide a `hash()` method on `DocumentViewId`

### LINKED ISSUE #243: Implement hash method for `DocumentViewId`

cafca:
- Returns a hash over the graph tips constituting a view id: the _document view hash id_
- Changes the constructor of `DocumentViewId` to expect a slice of operation ids instead of a vector.

## Document view id hash

Use this as a unique identifier for a document if you need a value with a limited size. The
document view id itself grows with the number of graph tips that the document has, which
may not be desirable for an identifier.

Keep in mind that when you refer to document views with this hash value it will not be
possible to recover the document view id from it.

Closes #243 

## 📋 Checklist

  - https://github.com/p2panda/handbook/pull/162

### DIFF
--- a/p2panda-rs/src/document/document_view_id.rs
+++ b/p2panda-rs/src/document/document_view_id.rs
@@ -51,9 +51,9 @@ impl DocumentViewId {
     /// Keep in mind that when you refer to document views with this hash value it will not be
     /// possible to recover the document view id from it.
     pub fn hash(&self) -> Hash {
-        let graph_tip_bytes = self
-            .0
-            .clone()
+        let mut graph_tips = self.0.clone();
+        graph_tips.sort();
+        let graph_tip_bytes = graph_tips
             .into_iter()
             .flat_map(|graph_tip| graph_tip.as_hash().to_bytes())
             .collect();
@@ -160,7 +160,9 @@ mod tests {
         #[from(random_operation_id)] operation_id_1: OperationId,
         #[from(random_operation_id)] operation_id_2: OperationId,
     ) {
-        let view_id = DocumentViewId::new(vec![operation_id_1, operation_id_2]);
-        assert!(view_id.hash().validate().is_ok());
+        let view_id_1 = DocumentViewId::new(vec![operation_id_1.clone(), operation_id_2.clone()]);
+        let view_id_2 = DocumentViewId::new(vec![operation_id_2, operation_id_1]);
+        assert!(view_id_1.hash().validate().is_ok());
+        assert_eq!(view_id_1.hash(), view_id_2.hash());
     }
 }
--- a/p2panda-rs/src/document/document_view_id.rs
+++ b/p2panda-rs/src/document/document_view_id.rs
@@ -43,6 +43,13 @@ impl DocumentViewId {
     }
 
     /// Returns a hash over the graph tips constituting this view id.
+    ///
+    /// Use this as a unique identifier for a document if you need a value with a limited size. The
+    /// document view id itself grows with the number of graph tips that the document has, which
+    /// may not be desirable for an identifier.
+    ///
+    /// Keep in mind that when you refer to document views with this hash value it will not be
+    /// possible to recover the document view id from it.
     pub fn hash(&self) -> Hash {
         let graph_tip_bytes = self
             .0
--- a/p2panda-rs/src/document/document_view_id.rs
+++ b/p2panda-rs/src/document/document_view_id.rs
@@ -178,4 +178,34 @@ mod tests {
         let view_id_2 = DocumentViewId::new(&[operation_id_2, operation_id_1]);
         assert_eq!(view_id_1.hash(), view_id_2.hash());
     }
+
+    #[test]
+    fn deserialize_unsorted_view_id() {
+        // Unsorted operation ids in document view id array:
+        //
+        // [
+        //  "0020c13cdc58dfc6f4ebd32992ff089db79980363144bdb2743693a019636fa72ec8",
+        //  "00202dce4b32cd35d61cf54634b93a526df333c5ed3d93230c2f026f8d1ecabc0cd7"
+        // ]
+        let unsorted_operation_ids = "827844303032306331336364633538646663366634656264333239393266663038396462373939383033363331343462646232373433363933613031393633366661373265633878443030323032646365346233326364333564363163663534363334623933613532366466333333633565643364393332333063326630323666386431656361626330636437";
+
+        // Construct document view id by deserializing CBOR data
+        let view_id_1: DocumentViewId =
+            ciborium::de::from_reader(&hex::decode(unsorted_operation_ids).unwrap()[..]).unwrap();
+
+        // Create the same view id via constructor API
+        let view_id_2 = DocumentViewId::new(vec![
+            Hash::new("0020c13cdc58dfc6f4ebd32992ff089db79980363144bdb2743693a019636fa72ec8")
+                .unwrap()
+                .into(),
+            Hash::new("00202dce4b32cd35d61cf54634b93a526df333c5ed3d93230c2f026f8d1ecabc0cd7")
+                .unwrap()
+                .into(),
+        ]);
+
+        // Should result in the same hash
+        assert!(view_id_1.hash().validate().is_ok());
+        assert!(view_id_2.hash().validate().is_ok());
+        assert_eq!(view_id_1.hash(), view_id_2.hash());
+    }
 }
--- a/p2panda-rs/src/document/document_view_id.rs
+++ b/p2panda-rs/src/document/document_view_id.rs
@@ -33,8 +33,8 @@ pub struct DocumentViewId(Vec<OperationId>);
 
 impl DocumentViewId {
     /// Create a new document view id.
-    pub fn new(graph_tips: Vec<OperationId>) -> Self {
-        let mut graph_tips_mut = graph_tips;
+    pub fn new(graph_tips: &[OperationId]) -> Self {
+        let mut graph_tips_mut = graph_tips.to_owned();
         graph_tips_mut.sort();
         Self(graph_tips_mut)
     }
@@ -91,7 +91,7 @@ impl IntoIterator for DocumentViewId {
 /// only consists of one graph tip hash.
 impl From<OperationId> for DocumentViewId {
     fn from(operation_id: OperationId) -> Self {
-        Self::new(vec![operation_id])
+        Self::new(&[operation_id])
     }
 }
 
@@ -101,7 +101,7 @@ impl From<OperationId> for DocumentViewId {
 /// consists of one graph tip hash.
 impl From<Hash> for DocumentViewId {
     fn from(hash: Hash) -> Self {
-        Self::new(vec![hash.into()])
+        Self::new(&[hash.into()])
     }
 }
 
@@ -113,7 +113,7 @@ impl FromStr for DocumentViewId {
     type Err = HashError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Ok(Self::new(vec![Hash::new(s)?.into()]))
+        Ok(Self::new(&[Hash::new(s)?.into()]))
     }
 }
 
@@ -135,16 +135,16 @@ mod tests {
         let document_id: DocumentViewId = hash_str.parse().unwrap();
         assert_eq!(
             document_id,
-            DocumentViewId::new(vec![hash_str.parse::<OperationId>().unwrap()])
+            DocumentViewId::new(&[hash_str.parse::<OperationId>().unwrap()])
         );
 
         // Converts a `Hash` to `DocumentViewId`
         let document_id: DocumentViewId = hash.clone().into();
-        assert_eq!(document_id, DocumentViewId::new(vec![hash.clone().into()]));
+        assert_eq!(document_id, DocumentViewId::new(&[hash.clone().into()]));
 
         // Converts an `OperationId` to `DocumentViewId`
         let document_id: DocumentViewId = OperationId::new(hash.clone()).into();
-        assert_eq!(document_id, DocumentViewId::new(vec![hash.into()]));
+        assert_eq!(document_id, DocumentViewId::new(&[hash.into()]));
 
         // Fails when string is not a hash
         assert!("This is not a hash".parse::<DocumentViewId>().is_err());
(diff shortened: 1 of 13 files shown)


cafca:
Oops!

p2panda-rs/src/document/document_view_id.rs
```diff
@@ -34,13 +34,33 @@ pub struct DocumentViewId(Vec<OperationId>);
 impl DocumentViewId {
     /// Create a new document view id.
     pub fn new(graph_tips: Vec<OperationId>) -> Self {
```
cafca, line 36: Should this parameter be a reference `&Vec<OperationId>`?

p2panda-rs/src/document/document_view_id.rs
```diff
@@ -34,13 +34,33 @@ pub struct DocumentViewId(Vec<OperationId>);
...
+    /// Use this as a unique identifier for a document if you need a value with a limited size. The
+    /// document view id itself grows with the number of graph tips that the document has, which
+    /// may not be desirable for an identifier.
+    ///
+    /// Keep in mind that when you refer to document views with this hash value it will not be
+    /// possible to recover the document view id from it.
+    pub fn hash(&self) -> Hash {
+        let graph_tip_bytes = self
+            .0
+            .clone()
```
adzialocha, line 58: I somehow assumed to find `sort` here as the specification reads that before hashing the graph tips need to be sorted, but then I realized you moved it up to make them comparable which also makes sense.

How about updating the specification that *any* document view id needs to hold operation ids which are sorted. Then this is reflected better in the code-base and brings consistency across all data.

We would still need to make sure they are sorted when coming out of deserialized data, since the constructor is *not* used in that case. I can imagine that could lead to bugs when hashing document view ids coming from CBOR.

Slightly related question: How about sorting in `previous_operations`?

It totally makes sense to assure that the graph tips in `DocumentViewId` are always sorted and not for the document view hash only.

p2panda-rs/src/document/document_view_id.rs
adzialocha: How about `&[OperationId]`?

p2panda-rs/src/document/document_view_id.rs
```diff
@@ -136,4 +156,25 @@ mod tests {
...
     }
+
+    #[rstest]
+    fn equality(
+        #[from(random_operation_id)] operation_id_1: OperationId,
+        #[from(random_operation_id)] operation_id_2: OperationId,
+    ) {
+        let view_id_1 = DocumentViewId::new(vec![operation_id_1.clone(), operation_id_2.clone()]);
+        let view_id_2 = DocumentViewId::new(vec![operation_id_2, operation_id_1]);
+        assert_eq!(view_id_1, view_id_2);
```
adzialocha, line 167: :+1:

adzialocha (approved):
Thank you!

I've added a couple of comments, especially regarding the handbook and not the code-base, I would propose to make an addition there.

p2panda-rs/src/document/document_view_id.rs
adzialocha: I've added that test now :+1:

adzialocha (changes requested):
I've added the test now to show that we need to think about sorting the operation ids (see comment).

p2panda-rs/src/document/document_view_id.rs
```diff
@@ -136,4 +166,46 @@ mod tests {
...
+        //  "00202dce4b32cd35d61cf54634b93a526df333c5ed3d93230c2f026f8d1ecabc0cd7"
+        // ]
+        let unsorted_operation_ids = "827844303032306331336364633538646663366634656264333239393266663038396462373939383033363331343462646232373433363933613031393633366661373265633878443030323032646365346233326364333564363163663534363334623933613532366466333333633565643364393332333063326630323666386431656361626330636437";
+
+        // Construct document view id by deserialising CBOR data
+        let view_id_1: DocumentViewId =
+            ciborium::de::from_reader(&hex::decode(unsorted_operation_ids).unwrap()[..]).unwrap();
+
+        assert_eq!(
+            format!("{}", view_id_1.validate().unwrap_err()),
```
cafca, line 207: I don't like having to use `format` to check that the correct error is returned. For our errors to implement `Eq` we need to add that to YASMF first.

adzialocha: Ja, matching error types in rust is surprisingly hard ..

https://zhauniarovich.com/post/2021/2021-01-testing-errors-in-rust/

cafca: Hmm, do you mean we shouldn't do that?

cafca: I made a quick pr [here](https://github.com/bamboo-rs/yasmf-hash/pull/3) but I also read [this](https://users.rust-lang.org/t/help-understanding-io-error-and-lack-of-partialeq/13212) and now I'm not sure I understand the situation :D

p2panda-rs/src/document/document_view_id.rs
sandreae: Yeh, I also expected the sorting to happen before hashing, rather than in the `DocumentViewId` constructor. What is the case for it happening there? Is there a down side to it happening in the hashing method?

sandreae (commented):
Thanks for doing this, really nice functionality which will be needing very soon I expect.

I just had a query about the hash sorting, I'm sure your approach is totally valid, just want to get my head in the same place with it as my initial thinking is that it would happen at the point where the id is hashed.

p2panda-rs/src/document/document_view_id.rs
```diff
@@ -33,20 +35,48 @@ pub struct DocumentViewId(Vec<OperationId>);
 
 impl DocumentViewId {
     /// Create a new document view id.
-    pub fn new(graph_tips: Vec<OperationId>) -> Self {
-        Self(graph_tips)
+    pub fn new(graph_tips: &[OperationId]) -> Self {
```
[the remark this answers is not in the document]
adzialocha: Good idea! Let's make a ticket for it 👍

p2panda-rs/src/document/document_view_id.rs
cafca: 1) Major reason: If we don't sort in the constructor, two document view ids are not equal if they are constructed from the same operation ids in a different order. I added an `equality` test for this, that will fail otherwise.
2) Minor reason: If we don't sort in the constructor every call to `hash()` will require sorting the ids again. If we sort in the constructor it only has to happen when the document view id is first created, after that they will never have to be sorted again, even after serialisation, because passing in an already sorted document view id into the constructor will result in `sort()` just passing over the values without doing anything (as they are already sorted).

adzialocha: Without a clear idea where this will lead, here two additional thoughts:

Re: 1) We could also implement `Eq` ourselves and sort inside of it. So then we _only_ sort if we want to compare and if we want to hash. I assume both cases are currently not really used (or in other words: We currently don't need document view hashes).

Re: 2) Graph tip arrays will always be fairly small (1-5 items) so I don't think that sorting is such a heavy penalty. Hashing in itself is probably the more expensive operation here.

adzialocha: If we sort in the constructor we need an idea how we sort for deserialization, if we sort in `hash` and `Eq` we don't.. The latter also maybe gives us a peace of mind in regards of enforcing sorting in the CBOR data as I somehow wonder if this is _really_ required?

adzialocha: Other thought: Whats more expensive, sorting whenever you use something or always checking if something _is_ sorted? :thinking:

p2panda-rs/src/document/document_view_id.rs
adzialocha: > Hmm, do you mean we shouldn't do that?

Ah, no, just wanted to link to a good article to show the problem space around error types in Rust. No real opinion on whats best to do, I'm fine with `format!` for now.

adzialocha: Most of the time we just do `.is_err()` so checking for something more is already 100% better :joy_cat:

p2panda-rs/src/document/document_view_id.rs
sandreae: Thanks for the explanations @cafca. For point 1) looking at this from the functionality of a DocumentViewId I would expect two IDs to be equal even if the hashes are listed in different orders. The order has no meaning wrt which document view you are refering too. So the custom eq approach does make sense to me.

I agree we should avoid unnecessary sorting, but I wonder if we will be constructing IDs more often than we're hashing them?

adzialocha: Ayayayay. 🤣 Now after this discussion I feel like we should implement `Eq` ourselves where we sort before, add sorting into `hash` as well while removing it from the constructor.

It seems to be the best approach plus we don't need to worry about validation too much!

adzialocha (changes requested):
Ayayayay. 🤣 Now after this discussion I feel like we should implement `Eq` ourselves where we sort before, add sorting into `hash` as well while removing it from the constructor.

It seems to be the best approach plus we don't need to worry about validation anymore.

I can implement the changes for you @cafca as you are busy next week, if you approve with this suggestion?

p2panda-rs/src/document/document_view_id.rs
cafca: Sorting is really inexpensive if the data is already sorted so I think it won't be a problem at all to have this in the constructor. And then we can just derive `Eq` without a custom implementtation.

cafca:
> 



> Ayayayay. 🤣 Now after this discussion I feel like we should implement `Eq` ourselves where we sort before, add sorting into `hash` as well while removing it from the constructor.
> 
> It seems to be the best approach plus we don't need to worry about validation anymore.
> 
> I can implement the changes @cafca as you are busy next week, if you approve with this suggestion?

Then let's have a jam on it! I still think sorting in the constructor is the way to go but maybe I just need to hear your argument with more words :)

p2panda-rs/src/document/document_view_id.rs
adzialocha: > Sorting is really inexpensive if the data is already sorted so I think it won't be a problem at all to have this in the constructor. And then we can just derive `Eq` without a custom implementtation.

Thats also alright, then we sort everywhere all the time 🤣

I'd be fine with this, as long as we remove the validation again 👍

sandreae: Yep, cool, I'm fine with this too. It doesn't need to be in the spec that `DocumentViewId` must contain sorted hashes, right? It's just that we do it in advance as it _is_ required for hashing them.

@adzialocha what do you mean by removing the validation?

p2panda-rs/src/document/document_view_id.rs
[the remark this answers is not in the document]
cafca: https://github.com/p2panda/p2panda/issues/278

p2panda-rs/src/document/document_view_id.rs
adzialocha: During our discussion we've decided to add the `Validate` trait to `DocumentViewId` and that it _needs_ to be sorted. Also there was a discussion in the chat about that topic and we've been reasoning about that any array of document view ids need to be sorted in the CBOR.

adzialocha: https://github.com/p2panda/p2panda/pull/277/files#diff-1abf4b4e983cf2c508fc8c65ce494e69129778eca08896b7ef468ed9b03c829aR71-R79

cafca: If we don't have the requirement that document view ids are sorted we can't derive `PartialEq` because it would think that two instances with differently sorted but same operation ids inside are not the same.

Then we have to implement `PartialEq` ourselves and in that implementation sorts everytime that equality of some document view ids is checked. We would also have to sort everytime `hash()` is accessed.

The alternative is to sort the ids once when they are first created and add a requirement to the spec that document view ids must always be sorted. We have a `sort` call in the constructor then but it won't do much sorting because calling `sort()` on already sorted values is inexpensive.

sandreae: It just feels like we're introducing a requirement of `DocumentViewId` into the spec which is actually a requirement of `DocumentViewHash` (which are the hash or _sorted_ document view ids). That's my only query if you both think this is ok, then lets press on :+1: I'm happy with any of the implementation solutions.

cafca: It's not just the hash, also checking equality of a set in general is more cumbersome if you don't have sorted data to begin with. I think we should have a requirement for sorted values in all places where we have sets of hashes, unless their order has some special significance.

sandreae: Rejecting unsorted ids seems like quite unexpected behaviour to me. The hash is a product of the id, so I feel it's more intuitive for the sorting to be _enforced_ at the point of hashing. I see the case for sorting in the constructor of `DocumentViewId` in our implementation though.

sandreae: > It's not just the hash, also checking equality of a set in general is more cumbersome if you don't have sorted data to begin with. I think we should have a requirement for sorted values in all places where we have sets of hashes, unless their order has some special significance.

Ok, if it's a more general thing we want in the spec, that makes more sense to me.

sandreae: Maybe we don't need to tie this PR to a spec level decision about general sorting of lists in this PR though?

cafca: It's gonna rain PRs! But yes, I can split that up as well.

sandreae: Haha, I am ready to get wet 🤣

sandreae: (I forgot my umbrella)

cafca:
I did the sorting in `hash()` and moved everything else over to [#272](https://github.com/p2panda/p2panda/pull/272).

adzialocha (approved):
Super, thank you for splitting it up :+1:

PR #279, opened 2022-03-20
Simple key group implementation

cafca:
Todo: write todo list

## 📋 Checklist


sandreae:
Just wanna say how exciting it is to start reading all these ideas in code :+1: looking great so far!

p2panda-rs/src/test_utils/mocks/node.rs
```diff
@@ -495,6 +495,18 @@ impl Node {
         document.view().to_owned()
     }
 
+    pub fn get_document_but_dont_make_it_look_like_a_view(&self, id: &Hash) -> Document {
```
sandreae, line 498: You mentioned this in the chat I think, that `DocumentView` doesn't contain everything needed when construction `KeyGroup`s. Is that right? I'd say there might be two reasons/solutions for this:

1) the mock node implementation needs changing, and should be returning `Document`s in most cases
2) actually we need to rethink what goes where in `Document` and `DocumentView`

When `Document` was being born, I made a decision at some point to contain all meta data there, and keep `DocumentView` very simple. The idea being, is that `Document` is what we pass around, and we access the document view from it. Happy to rethink this if it doesn't fit well with `KeyGroup` construction patterns, which are definitely more complex than what we've had so far.

p2panda-rs/src/operation/operation_fields.rs
```diff
@@ -56,6 +60,7 @@ impl Validate for OperationValue {
         match self {
             Self::Relation(relation) => relation.validate(),
             Self::RelationList(relations) => relations.validate(),
+            Self::Owner(relation) => relation.validate(),
```
sandreae, line 63: I often daydream about using a `PinnedRelation` here..... It seems like quite an elegant solution to some of the complexities around members moving in and out of a key group.

Is that something we've seriously considered in recent times?

p2panda-rs/src/operation/operation_fields.rs
```diff
@@ -47,6 +47,10 @@ pub enum OperationValue {
     /// Reference to a list of document views.
     #[serde(rename = "pinned_relation_list")]
     PinnedRelationList(PinnedRelationList),
+
+    /// Reference to a document's owner key group.
+    #[serde(rename = "owner")]
+    Owner(Relation)
```
sandreae, line 53: I often daydream about using a PinnedRelation here..... It seems like quite an elegant solution to some of the complexities around members moving in and out of a key group.

Is that something we've seriously considered in recent times?

sandreae: I know it throws up different challenges itself :rofl:

p2panda-rs/src/test_utils/mocks/node.rs
cafca: I think 1) would be quite nice as `Document` has everything we need here already.

cafca: https://github.com/p2panda/p2panda/issues/286

p2panda-rs/src/operation/operation_fields.rs
cafca: We could do this if we store the member list of a key group on the `key_group_v1` document instead of collecting the members from all `key_group_membership_v1` documents that point at a particular key group.

Would be interesting to compare the pros and cons of this approach vs the "transitive membership" approach we discussed on the chat/this monday meet.

sandreae: Yeh, I'd really like to look through pros and cons of these two newer approaches as well as the existing "limits" approach.

p2panda-rs/src/permissions/key_group/views/key_group.rs
```diff
@@ -0,0 +1,137 @@
...
+use crate::operation::OperationValue;
+use crate::permissions::key_group::error::KeyGroupError;
+use crate::schema::system::SystemSchemaError;
+use crate::Validate;
+
+/// Represents a root key group definition.
+///
+/// Can be used to make a [`KeyGroup`].
+#[derive(Clone, Debug)]
+pub struct KeyGroupView(Document);
```
cafca, line 15: I made this a thin wrapper, different to the schema definition view structs. I felt that document views are already so close to the data structure we want here that it's not warranted to make another struct.

p2panda-rs/src/operation/operation_fields.rs
```diff
@@ -47,6 +47,10 @@ pub enum OperationValue {
     /// Reference to a list of document views.
     #[serde(rename = "pinned_relation_list")]
     PinnedRelationList(PinnedRelationList),
+
+    /// Reference to a document's owner key group.
+    #[serde(rename = "owner")]
+    Owner(Relation),
```
sandreae, line 53: This is the main point of contention for me, I believe we want to use a `PinnedRelation` here. There are pros and cons to both, as I see it:

### owner is a `Relation`
#### pros
- any updates to the owner `KeyGroup` are automatically applied to this document so no extra actions are needed to update a documents owner once it's set.
#### cons
- updates to a documents' owner `KeyGroup` apply to the entire history of a document, this means we need to then introduce a mechanism for retaining earlier operations which might be incorrectly ignored during materialisation (this is the `limit` logic we understand well though).
- as a result of the mutable nature of the owner -> `KeyGroup` relationship as defined above, document views are no longer immutable.
  -  the same `DocumentViewId` could produce a different view on different nodes if they haven't got the same `KeyGroup` or `KeyGroupLimit` updates
  - a `DocumentViewId` could resolve to different values on the same node if the `KeyGroupLimit` changes

### owner is a `PinnedRelation`
#### pros
- once set an owner relation for a specific operation will never change, this means document views retain their "immutable" property.
#### cons
- updates to a `KeyGroup` now don't bubble up to all it'd "owned" documents.... this means we need a mechanism for noticing `KeyGroup` updates and including any changes in operations clients publish.

In terms of this PR I am happy for it to be merged as it is though, it doesn't include the `key_group_limit` logic yet, and we can explore how it is to use a `Relation` for the owner field in any case :+1: 

I have a feeling that any difference of opinion on this might boil down to whether we imagine `KeyGroups` as a moderation or authorisation layer. I see them as an authorisation layer, which to me demands the immutable properties you get from `PinnedRelation`. I can see if they were more about moderation, then the mutable nature of `Relation` and introducing `KeyGroupLimit` is more what would be needed. For me the moderation layer happens elsewhere (which we haven't defined).

p2panda-rs/src/permissions/key_group/key_group.rs
```diff
@@ -0,0 +1,297 @@
...
+        for membership in members {
+            match membership.member() {
+                // Simple case: for single key memberships just add that key to the pool.
+                Owner::Author(value) => {
+                    debug!("Adding author {:?}", value.as_str());
+                    member_pool.push((value, membership));
+                }
+
+                // When a key group is a member, recursively add that key group's members to the
+                // pool, assigned to a shared `membership`
```
sandreae, line 50: Does this account for when a key group is a member of a key group which is a member of a key group? (you know what I mean ;-p)

p2panda-rs/src/permissions/key_group/key_group.rs
```diff
@@ -0,0 +1,297 @@
...
+            id: key_group.id().clone(),
+            view_id: key_group.view_id().clone(),
+            name: key_group.name().to_string(),
+            members: member_map,
+        };
+
+        Ok(key_group)
+    }
+
+    /// Create a new key group from predecessor documents.
```
sandreae, line 95: Not sure I understand what "predecessor" means in this comment?

p2panda-rs/src/permissions/key_group/key_group.rs
```diff
@@ -0,0 +1,297 @@
...
+            // Remove requests for which we have a response
+            if let Some(request) = requests.remove(response.request()) {
+                members.push(Membership::from_confirmation(
+                    request.clone(),
+                    Some(response.clone()),
+                )?);
+            };
+        }
+
+        for request in requests.values() {
```
sandreae, line 142: Maybe a comment saying these are requests which don't have a response yet would be nice.

p2panda-rs/src/permissions/key_group/views/request.rs
```diff
@@ -0,0 +1,157 @@
...
+use crate::schema::system::SystemSchemaError;
+use crate::Validate;
+
+#[derive(Clone, Debug)]
+/// A request for key group membership.
+pub struct MembershipRequestView(Document);
+
+#[allow(dead_code)]
+impl MembershipRequestView {
+    /// Create a membership request view from its author and the request's document.
```
sandreae, line 17: > Create a membership request view from its author and the request's document.

Where is the author in this constructor?

sandreae (changes requested):
Hey @cafca! I feel awful this PR has been sat here for so long :bow: it's all really really good code and represents a huge amount of work, thought and preparation, mega thanks :clap: 


I don't mean to trigger a lot of extra development on this, but I wonder if it makes to get this merged sooner rather than later (even in a not _final_ state). We are making some quite sweeping naming and other changes across the repo and this will get harder and harder to merge as it remains open. I would be in favour of bringing this into main even if we don't fully agree on the complete path forwards for `KeyGroups`. For me (almost) everything implemented in this PR is needed in any situation anyway, so I'm very happy to merge it.

The only sticking point for me is the type of relation we use for `Owner`, I added a fairly detailed comment about that. However I am still happy for it to be merged as is and we deal with that in follow-up issues/PRs. If we want to merge an accompanying PR into the handbook, then maybe we could remove everything relating to `key_group_limits` from here:  https://github.com/p2panda/handbook/pull/118 then it would match the content of this PR.

Except for that, I just had a few small questions that should be resolved easily.

[:key: :key: :key: ]

p2panda-rs/src/permissions/key_group/key_group.rs
cafca: I'll rephrase that!

p2panda-rs/src/operation/operation_fields.rs
sandreae: > does that mean that members who joined the group after a document's owner was set to that group are not authorised to make changes to that document?

Yes, but this isn't the primary goal, it's a side-effect of having immutable permissions, the new flow would be the same as before but with one change:

1. Group G1 has members [a,]
2. Group G1 creates a blog post document P1 with itself as the owner (PinnedRelation).
3. Public key b joins the group.
4. Public key b makes a change to P1 _which includes updating the `owner` field to point at the newest document view of  Group G1 (when they are a member)_

cafca: I think I've come around on this a bit. I have come to like the idea of immutable views a lot. I would still like to avoid having changes on _all_ owned documents whenever the members of a key group change. 

Could we resolve owner fields with a mechanism similar to lenses? We would first build an immutable view from a view id. Then whenever the group's members change we store an updated materialised lens on that view, which filters operations according to group membership.

Issue #270, opened 2022-03-13
Validate schema definition requirements


cafca:
Implement `Validate` for `Schema` to check that all requirements on `name`, `description` and `fields` are met, in particular that `name` doesn't exceed the length limit and allowed character set (important for parsing a schema id string representation and for creating db table names from a schema id).

PR #282, opened 2022-03-21
Implement schema hash id

### LINKED ISSUE #267: Unique string identifier for application schemas

cafca:
The schema hash id can be used as a short identifier for schemas that doesn't grow with the amount of graph tips in the schema's document view id.

Depends on #277.

Closes #267.

Split out from #268.

Should be followed by https://github.com/p2panda/p2panda/issues/270 to make sure that no schemas can be created whose name conflicts with the hash id schema.

## 📋 Checklist

  https://github.com/p2panda/handbook/pull/168


cafca:
This is just missing a handbook update to describe the schema hash id, could even be added here maybe https://github.com/p2panda/handbook/pull/162

sandreae (approved):

sandreae (approved):
Looks great to me! Sorry took me a while to review, it slipped past me :-p

PR #288, opened 2022-03-23
Test `debug` macro calls in Github CI

cafca:
Calls to the debug macro `debug!("hey! it's a debug")` are only run when `RUST_LOG=debug` is set in the environment. In order to test that we don't introduce bugs in those calls the CI is updated here to also run them.

The effect of that is of course an enormous amount of output in the test logs. Does this bother you? I feel that it could even be useful because you would probably only log at those logs if you are trying to see what kind of error occurred and for what reason - but a shorter log is also nice.

## 📋 Checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -35,6 +35,9 @@ jobs:
         with:
           command: test
           args: --manifest-path ${{ env.cargo_manifest }}
+        # Ensure debug output is also tested
+        env:
+          RUST_LOG: debug
 
   rust-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


adzialocha:
For understanding: What does "To test debug macro calls" mean? As I understand this PR you're raising the verbosity level of our loggers, but I'm not sure what exactly is "tested" here?

For the PR: When this is about being able to see debug logging output in our CI, I'm neutral here, thats fine with me!

cafca:
When `debug` macro calls contain expressions (to generate the content of the debug log string) those are not run during testing if the log level is not raised, so we wouldn't know if they fail for some reason.

adzialocha:
> When `debug` macro calls contain expressions (to generate the content of the debug log string) those are not run during testing if the log level is not raised, so we wouldn't know if they fail for some reason.

Oha! That's good to know! Thank you!

adzialocha (approved):

PR #289, opened 2022-03-23
Generic graph and other bits and pieces for proper graphing

cafca:
Refactor for `Graph` to make it generic over both key and value of the graph nodes.

Other improvements included here:

- Docstrings formatted to max line length 100
- Fixed `add_link` implementation to prevent adding links to non-existing nodes + tests for this.
- Removed all methods that are never used.
  - They were included because they are part of the tangle api which `Graph` was implementing but as we neither use nor test them they are prone to becoming broken and non-maintained in our codebase over time. Their implementation is preserved in the branch [`graph-with-tangle-api`](https://github.com/p2panda/p2panda/tree/graph-with-tangle-api).

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/graph/graph.rs
+++ b/p2panda-rs/src/graph/graph.rs
@@ -1,6 +1,7 @@
 use log::debug;
 use std::collections::HashMap;
 use std::fmt::Debug;
+use std::hash::Hash;
 
 use super::GraphError;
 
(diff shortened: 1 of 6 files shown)


p2panda-rs/src/graph/graph.rs
```diff
@@ -95,133 +108,85 @@ impl<'a, T: PartialEq + Clone + Debug> Node<T> {
...
             return false;
         }
 
-        if let Some(from_node_mut) = self.0.get_mut(from) {
-            from_node_mut.next.push(to.to_owned());
-        } else {
+        // Check that both nodes exist
+        if self.get_node(from).is_none() || self.get_node(to).is_none() {
             return false;
         }
```
sandreae, line 171: Ouch, nice catch :+1:

sandreae (approved):
Looks great to me, very nice with the generic keys and good bug catch :rofl:

PR #292, opened 2022-03-26
String serialisation for application schema ids

cafca:
Application schema ids are serialised as strings such as `venue_0020c65567ae37efea293e34a9c7d13f8f2bf23dbdc3b5c7b9ab46293111c48fc78b`.

Precursor for https://github.com/p2panda/p2panda/pull/269

I didn't implement `Validate` because I thought it might make sense to do that in one go for all schema-related structs -> https://github.com/p2panda/p2panda/issues/270.

## Todo

- [x] Serialiser
- [x] Parser
- [x] Operation format CDDL
- [x] Fix tests
- [x] Extend tests

## 📋 Checklist

  https://github.com/p2panda/handbook/pull/167


p2panda-rs/src/document/document.rs
```diff
@@ -452,8 +451,8 @@ mod tests {
         let expected_graph_tips: Vec<OperationId> = vec![penguin_entry_3_hash.clone().into()];
         let expected_op_order = vec![
             panda_1.clone(),
-            panda_2.clone(),
             penguin_1.clone(),
+            panda_2.clone(),
```
cafca, line 455: Hashes and stuff .. had to change the order here

p2panda-rs/src/document/document.rs
```diff
@@ -573,7 +572,10 @@ mod tests {
             &update_operation(
                 schema.clone(),
                 vec![polar_entry_1_hash.clone().into()],
-                operation_fields(vec![("name", OperationValue::Text("🐼 Cafe!".to_string()))]),
+                operation_fields(vec![(
+                    "name",
+                    OperationValue::Text("🐼 Cafe!!".to_string()),
```
cafca, line 577: I was able to retain the previous ordering of the document graph by adding a "!" :D

sandreae: Haha, brilliant 🤣

p2panda-rs/src/test_utils/constants.rs
```diff
@@ -7,10 +7,9 @@
 pub const DEFAULT_HASH: &str =
     "0020b177ec1bf26dfb3b7010d473e6d44713b29b765b99c6e60ecbfae742de496543";
 
-/// The default schema hash string, used in all operation fixtures when no custom schema hash is
-/// defined.
-pub const DEFAULT_SCHEMA_HASH: &str =
-    "0020c65567ae37efea293e34a9c7d13f8f2bf23dbdc3b5c7b9ab46293111c48fc78b";
+/// Unpublished test schema id, used in all operation fixtures that don't define their own schema.
+pub const TEST_SCHEMA_ID: &str =
+    "venue_0020c65567ae37efea293e34a9c7d13f8f2bf23dbdc3b5c7b9ab46293111c48fc78b";
```
cafca, line 12: I changed this constant name to not make it sound like there is a _default_ schema in p2panda. Would like to do the same with the other constants here.

p2panda-rs/src/schema/schema_id.rs
```diff
@@ -32,20 +33,71 @@ impl SchemaId {
...
+            ));
+        }
+
+        let mut operation_ids = vec![];
+        let mut remainder = id_str;
+        while let Some((left, right)) = remainder.rsplit_once('_') {
+            // Catch trying to parse an unknown system schema
+            if right.starts_with('v') && right.len() < MAX_YAMF_HASH_SIZE * 2 {
+                return Err(SchemaIdError::UnknownSystemSchema(id_str.to_string()));
+            }
```
sandreae, line 63: So this is catching the case where a system schema which wasn't matched against an existing system schema id and was passed then passed into `parse_application_schema_str`? Coming at this cold, the "unknown system schema" confused me a little, I would export more something along the lines of "invalid application schema". Could the "unknown system schema" error be caught elsewhere and inside this method we only deal with errors related to application schema id parsing?

p2panda-rs/src/schema/schema_id.rs
```diff
@@ -174,37 +204,70 @@ mod test {
...
+    #[case(
+        "\"venue_0020c65567ae37efea293e34a9c7d13f8f2bf23dbdc3b5c7b9ab46293111c48fc7\"",
+        "encountered invalid hash while parsing application schema id: invalid hash length 33 \
+    bytes, expected 34 bytes at line 1 column 74"
+    )]
+    // this looks like a system schema, but it is not
+    #[case(
+        "\"unknown_system_schema_name_v1\"",
+        "not a known system schema: unknown_system_schema_name_v1 at line 1 column 31"
+    )]
```
sandreae, line 248: Really nice use of rstest cases!! I'd not used it like this so much so far (prefering the templates) because I felt it got hard to read, but this works so well with just strings!

p2panda-rs/src/schema/schema_id.rs
cafca: This could completely be left out, then you'd just get an error about it not being able to parse an operation id. I thought it'd be nice to catch that situation here and present a more helpful error message. If we wanted to catch it outside the method we would have to have an additional parser in the constructor (on the application schema match arm) just to check for that. I thought that wouldn't be worth it.

cafca: How about I add a comment that explains this situation?

sandreae: Maybe it could be split into a separate method and then used as a condition in the match arms?

p2panda-rs/src/schema/schema_id.rs
```diff
@@ -174,37 +204,70 @@ mod test {
...
+    fn invalid_deserialization2(#[case] schema_id: &str, #[case] expected_err: &str) {
```
sandreae, line 249: Really nice use of rstest cases!! I'd not used it like this so much so far (prefering the templates) because I felt it got hard to read, but this works so well with just strings!

sandreae (changes requested):
Looks really great, again, thanks for the work on this. I have one suggestion regarding error catching in the application schema id parsing. Otherwise all good.

p2panda-rs/src/schema/schema_id.rs
cafca: Not sure what you mean :D Want to hop on Discord?

sandreae: Yeh, maybe this is the best way. Obvs nicer error messages is optimal, I only got confused understanding the code, and a comment would clear that up for sure.

sandreae (approved):
That's great, thanks!

p2panda-rs/src/schema/schema_id.rs
adzialocha: Automatic code formatting is not working for them which is a bummer but otherwise I'll like it as well 👍

PR #296, opened 2022-03-29
Update schema impl to use new schema id functionality

cafca:
- add a `SchemaId` to `Schema`
- add getters for schema properties
- update `SchemaId`
  - add getters for name and version
  - add new `SchemaVersion` enum that represents both application and system schema versions

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/schema/schema.rs
+++ b/p2panda-rs/src/schema/schema.rs
@@ -18,7 +18,7 @@ pub struct Schema {
     /// The application schema id for this schema.
     id: SchemaId,
 
-    /// Describes the schemas indented use.
+    /// Describes the schema's intended use.
     description: String,
 
     /// This schema's field definitions.
--- a/p2panda-rs/src/schema/schema.rs
+++ b/p2panda-rs/src/schema/schema.rs
@@ -72,24 +72,26 @@ impl Schema {
 
     /// Access the schema view id.
     #[allow(unused)]
-    pub fn view_id(&self) -> &DocumentViewId {
-        if let SchemaId::Application(_, view_id) = &self.id {
-            return view_id;
-        }
-        // This statement is unreachable because it's not possible to construct a `Schema`
-        // that doesn't contain an application schema id.
-        panic!()
+    pub fn view_id(&self) -> DocumentViewId {
+        let view_id = match self.id.clone() {
+            SchemaId::Application(_, view_id) => Some(view_id),
+            _ => None,
+        };
+        // Unwrap because it's not possible to construct a `Schema` that doesn't contain an
+        // application schema id.
+        view_id.unwrap()
     }
 
     /// Access the schema name.
     #[allow(unused)]
-    pub fn name(&self) -> &str {
-        if let SchemaId::Application(name, _) = &self.id {
-            return name;
-        }
-        // This statement is unreachable because it's not possible to construct a `Schema`
-        // that doesn't contain an application schema id.
-        panic!()
+    pub fn name(&self) -> String {
+        let name = match self.id.clone() {
+            SchemaId::Application(name, _) => Some(name),
+            _ => None,
+        };
+        // Unwrap because it's not possible to construct a `Schema` that doesn't contain an
+        // application schema id.
+        name.unwrap()
     }
 
     /// Access the schema description.
@@ -209,7 +211,7 @@ mod tests {
             &SchemaId::new_application("venue_name", expected_view_id)
         );
         assert_eq!(schema.name(), "venue_name");
-        assert_eq!(schema.view_id(), expected_view_id);
+        assert_eq!(&schema.view_id(), expected_view_id);
         assert_eq!(schema.description(), "Describes a venue");
         assert_eq!(schema.fields().len(), 2);
 
--- a/p2panda-rs/src/schema/schema.rs
+++ b/p2panda-rs/src/schema/schema.rs
@@ -64,7 +64,7 @@ impl Schema {
         generate_cddl_definition(&self.fields)
     }
 
-    /// Access the schema id.
+    /// Access the schema's [`SchemaId`].
     #[allow(unused)]
     pub fn id(&self) -> &SchemaId {
         &self.id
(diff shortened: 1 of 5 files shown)


p2panda-rs/src/schema/schema.rs
```diff
@@ -60,6 +63,46 @@ impl Schema {
...
+
+    /// Access the schema view id.
+    #[allow(unused)]
+    pub fn view_id(&self) -> &DocumentViewId {
+        if let SchemaId::Application(_, view_id) = &self.id {
+            return view_id;
+        }
+        // This statement is unreachable because it's not possible to construct a `Schema`
+        // that doesn't contain an application schema id.
+        panic!()
```
adzialocha, line 81: I can see how this is hard to avoid in this case. Hmn. 🤔

How about keeping the name and document view in the struct, returning it normally here and constructing a schema id when calling `id` with these two values?

You could even keep schema id in the struct as well, I know its maybe redundant but less hacky?

cafca: That's possible of course. I don't like being that redundant :(. But I think here that's the best option, I can change it.

It's a similar situation [here](https://github.com/p2panda/p2panda/blob/c6d76afe789dbbad5cd71ace57de6f615489ef9b/p2panda-rs/src/permissions/key_group/views/key_group.rs#L47-L54) and I really like making system schema views a thin wrapper of a document view and not replicating all the data.

adzialocha: Or you just unwrap and give a note why you do it? (You can unwrap `Option`)

sandreae: Hmmm, yeh, I see the problem, that's annoying. I also really like system views being very thin wrappers, that was the original idea... but I would probably prefer to not have unreachable code like this :-( 

Could we have a `view_id()` method on `SchemaId` which returns a view id wrapped in an option? Then we just unwrap here as we know it will be some in this case?

sandreae: (didn't see your comment adz, that works too)

cafca:
I updated the codecov settings to make patch test coverage purely informational (tests don't fail anymore if the coverage % of lines changed in a pr is lower than in the project overall) because our coverage is now so high that any uncovered lines in a small to medium-sized pr will make the tests fail. Please say something if you think we should have some other setting instead!

p2panda-rs/src/schema/schema.rs
```diff
@@ -60,6 +63,48 @@ impl Schema {
...
+    #[allow(unused)]
+    pub fn view_id(&self) -> DocumentViewId {
+        let view_id = match self.id.clone() {
+            SchemaId::Application(_, view_id) => Some(view_id),
+            _ => None,
+        };
+        // Unwrap because it's not possible to construct a `Schema` that doesn't contain an
+        // application schema id.
+        view_id.unwrap()
+    }
```
adzialocha, line 83: Two new ideas 🤣

1. Same, but less lines

```rust
 ​    ​pub​ ​fn​ ​view_id​(​&​self​) -> DocumentViewId { 
 ​        ​ match​ ​self​.id.​clone​() { 
 ​            SchemaId​::​Application​(_, view_id) ​=>​ ​view_id, 
 ​            _ ​=>​ ​panic("it is not possible to construct a schema ...")​, 
 ​        }
 ​    }
```

2. Introduce a `ApplicationSchemaId(Name, DocumentViewId)` type and have `SchemaId::Application (ApplicationSchemaId)`

```rust
     pub fn id(&self) -> SchemaId {
         SchemaId::Application(self.id.clone())
     }

 ​    ​pub​ ​fn​ ​view_id​(​&​self​) -> DocumentViewId { 
 ​        ​ self.id().clone().view_id()
 ​    }
```

cafca: Oh this actually gives me another idea :D 

```rust
pub enum SchemaVersion {
    System(u8),
    Application(DocumentViewId)
}
```

adzialocha: Yes, that's cool!

The second idea is nice because it gives you the type it should have, so no weird checks required 👍

p2panda-rs/src/schema/schema.rs
```diff
@@ -15,9 +15,13 @@ type FieldKey = String;
 /// It is constructed from a `SchemaView` and all related `SchemaFieldView`s.
 #[derive(Debug, PartialEq)]
 pub struct Schema {
-    id: DocumentViewId,
-    name: String,
```
sandreae, line 19: I'm sort of feeling that it's actually neater to keep these values on the struct.... I know they technically aren't needed, but it saves some mental complexity introduced in the methods. We could then have a getter for the schema id which combined the name and document view.

sandreae: I'm happy to move forward as we are as well though...

sandreae (changes requested):
I wrote in the comment, but wondering if we need to remove name and view_id from the struct... is it worth losing some readability of code and adding mental complexity to the getter methods?

p2panda-rs/src/schema/schema.rs
cafca: Please have a look at the new changes I made today. I think now all weird checks are resolved!

cafca:
I have no idea why this fails now...

p2panda-rs/src/schema/schema.rs
adzialocha: > Please have a look at the new changes I made today. I think now all weird checks are resolved!

I like it 👍 it's tidy and nice!

p2panda-rs/src/schema/schema.rs
adzialocha: I see what you mean Sam, it's a good example how bringing concepts closer together does not necessarily mean it's "better", I would say it's more of an compromise sometimes (redundancy gives flexibility and easier to read code but is not DRY, optimization is elegant but less flexible and sometimes leads to loads of nesting).

In that case I'm fine with both how it used to be and the new update, it looks neat now ✨

sandreae (approved):
Yeh, this is nice! Thanks for sticking with it 👍

Issue #298, opened 2022-04-13
Remove `aquamarine` and `serde_json` dependencies?


adzialocha:
Both `aquamarine` (only one small diagram) and `serde_json` (in one test, we could also test against cbor data) is almost not used and there is a chance we could remove them easily?

PR #300, opened 2022-04-18
Remove `aquamarine` and `serde_json`

sandreae:
Remove `aquamarine` and `serde_json`.

In se/de tests for `SchemaId` we now de/serialise to and from cbor bytes. 

closes: https://github.com/p2panda/p2panda/issues/298

## 📋 Checklist


adzialocha (commented):
We could use the json crate completely, serialize to cbor instead and use the `cbor!` macro for readability 👍

sandreae:
aaaaaand, it's gone!

sandreae:
but I couldn't find a use for the `cbor!` macro (which I really wanted to) is there a different way you imagined this refactoring going?

adzialocha (approved):
> but I couldn't find a use for the `cbor!` macro (which I really wanted to) is there a different way you imagined this refactoring going?

I've had a go but it just creates more lines of code, so I think you're approach is better :+1:

Looks all good to me now, thanks for the clean-up!!

PR #308, opened 2022-05-05
Additions for aquadoggo dev

### LINKED ISSUE #305: Full string representation of `DocumentViewId`
There are a few places where we use `DocumentViewId` internally as a relation pointer in the db. We can use the `DocumentViewId` hash id here (and that's what I'm doing for now) but I wonder if we also want the possibility to represent a `DocumentViewId` as a string of hashes with `_` separator. This would just mean we can reconstruct the `DocumentVieId` from the string alone if needed. Not 100% sure if this is actually needed though... there have been a few times where i naturally wanted it in this format, but didn't have access to it.

sandreae:
- Implement `as_str()` method for `DocumentViewId`
- Handle multi-hash strings in `FromStr` conversion into `DocumentViewId`
- Sort hashes in `RelationList` and `PinnedRelationList` constructor
- Make `as_str()` method public in `OperationAction`

closes #305 #302

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/document/document_view_id.rs
+++ b/p2panda-rs/src/document/document_view_id.rs
@@ -207,13 +207,20 @@ impl From<Hash> for DocumentViewId {
 
 /// Convenience method converting a hash string into a document view id.
 ///
-/// Converts a hash string into a `DocumentViewId`, assuming that this document view only consists
-/// of one graph tip hash.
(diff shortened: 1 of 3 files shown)


adzialocha:
I think this misses an entry in the changelog? @sandreae

sandreae:
Yes, as soon as I did it I though the same......

Issue #309, opened 2022-05-05
Storage methods for replication


pietgeursen:
Yo,
I think I need another method to help me with replication. 

Something like:

```rs
async fn get_next_n_entries_after_seq(author_id: Author, log_id: LogId, seq_num: SeqNum, max_number_of_entries: usize) -> Result<Option<Vec<StorageEntry>>>
```

With this I think I should be able to expose a gql method that is paginated. My replication code will manage the cursor and decide what the value of `max_number_of_entries` will be.

I'm not sure if it belongs in the `entry_store` or the `storage_provider` or somewhere new. What do you think @sandreae?

sandreae:
Nice, thanks for flagging this, I'm sketching out what is needed next for the store in general here right now: https://laub.liebechaos.org/TQPRKUS4TbG0iuk28ABfwg?edit and will be working making additions today.

> With this I think I should be able to expose a gql method that is paginated. My replication code will manage the cursor and decide what the value of max_number_of_entries will be.

Fantastic, that's a big help, I have quite a lot on my plate right now ;-p

> I'm not sure if it belongs 

It's nice having responsibilities seperated out into their own stores (`EntryStore`, `LogStore`, `OperationStore` etc...) but it is super hard to know where some things should go :rofl:

pietgeursen:
> Fantastic, that's a big help, I have quite a lot on my plate right now ;-p

Are you ok picking this piece up or should I make a PR for it? Totally fine if you're too busy.

sandreae:
No, it's totally fine, this is what I'm busy with 👍

pietgeursen:
Oh, I just thought of another one:

I'm guessing we have a use case for asking for an `Entry` plus its chain of lipmaa links. 

I think we might want something like:

```rs
async fn get_all_lipmaa_entries_for_entry(author_id: Author, log_id: LogId, seq_num: SeqNum) -> Result<Option<Vec<StorageEntry>>>
```

Open to finding a better method name btw.

Which would start at `seq_num`, calculate the lipmaa links back to `seq_num` 1 and then get those entries from the db all in one go.

sandreae:
Niiice, thanks!

(I am open to not finding a better name :rofl:)

sandreae:
Ok, I started a PR for all these changes here: https://github.com/p2panda/p2panda/pull/310

sandreae:
[...]

PR #310, opened 2022-05-05
Entry store additions

sandreae:
# New methods for `EntryStore`

Additional methods required for replication logic.

```rust
  async fn get_entry_by_hash(
      &self,
      hash: &Hash,
  ) -> Result<Option<StorageEntry>, EntryStorageError>;

  async fn try_get_backlink(
      &self,
      entry: &StorageEntry,
  ) -> Result<Option<StorageEntry>, EntryStorageError>;

  async fn try_get_skiplink(
      &self,
      entry: &StorageEntry,
  ) -> Result<Option<StorageEntry>, EntryStorageError> ;

  async fn get_next_n_entries_after_seq(
      &self,
      author: &Author,
      log_id: &LogId,
      seq_num: &SeqNum,
      max_number_of_entries: usize,
  ) -> Result<Vec<StorageEntry>, EntryStorageError>;

  async fn get_all_lipmaa_entries_for_entry(
      author_id: Author, 
      log_id: LogId, 
      seq_num: SeqNum
) -> Result<Vec<StorageEntry>, EntryStoreError>;
```

## Todo:
- [x] `get_entry_by_hash()`
- [x] `try_get_backlink()`
- [x] `try_get_skiplink()`
- [x] `get_next_n_entries_after_seq()`
- [x] `get_all_lipmaa_entries_for_entry()`

## Notes:
- ~~there is a default implementation of `get_next_n_entries_after_seq` defined on the trait, however it would not be efficient if used with a database storage (anything SQL for example) as it performs a new `get_entry_by_hash()` request for each entry. This should be over-ridden in other implementations of these traits (in `aquadoggo` for example).~~ I removed the default implementation as it's misleading to leave it there even though it's not great. It can be implemented easily enough.

closes: https://github.com/p2panda/p2panda/issues/309 https://github.com/p2panda/p2panda/issues/304

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/storage_provider/traits/entry_store.rs
+++ b/p2panda-rs/src/storage_provider/traits/entry_store.rs
@@ -1,5 +1,6 @@
 use async_trait::async_trait;
 use bamboo_rs_core_ed25519_yasmf::entry::is_lipmaa_required;
+use mockall::automock;
 
 use crate::entry::LogId;
 use crate::entry::SeqNum;
@@ -13,6 +14,7 @@ use crate::storage_provider::traits::AsStorageEntry;
 ///
 /// This trait should be implemented on the root storage provider struct. It's definitions make up
 /// the required methods for inserting and querying entries from storage.
+#[automock]
 #[async_trait]
 pub trait EntryStore<StorageEntry: AsStorageEntry> {
     /// Insert an entry into storage.
--- a/p2panda-rs/src/storage_provider/traits/entry_store.rs
+++ b/p2panda-rs/src/storage_provider/traits/entry_store.rs
@@ -158,6 +158,7 @@ pub trait EntryStore<StorageEntry: AsStorageEntry> {
         entry_skiplink_hash
     }
 
+    /// Get all entries which make up the certificate pool for the given entry.
     async fn get_all_lipmaa_entries_for_entry(
         &self,
         author_id: &Author,
--- a/p2panda-rs/src/storage_provider/traits/entry_store.rs
+++ b/p2panda-rs/src/storage_provider/traits/entry_store.rs
@@ -15,6 +15,17 @@ use crate::storage_provider::traits::AsStorageEntry;
 /// the required methods for inserting and querying entries from storage.
 #[async_trait]
 pub trait EntryStore<StorageEntry: AsStorageEntry> {
+    /// Insert an entry into storage.
+    async fn insert_entry(&self, value: StorageEntry) -> Result<bool, EntryStorageError>;
+
+    /// Returns entry at sequence position within an author's log.
+    async fn entry_at_seq_num(
+        &self,
+        author: &Author,
+        log_id: &LogId,
+        seq_num: &SeqNum,
+    ) -> Result<Option<StorageEntry>, EntryStorageError>;
+
     /// Get an entry by it's hash.
     async fn get_entry_by_hash(
         &self,
@@ -96,18 +107,6 @@ pub trait EntryStore<StorageEntry: AsStorageEntry> {
         }
         Ok(expected_skiplink)
     }
-
-    /// Insert an entry into storage.
-    async fn insert_entry(&self, value: StorageEntry) -> Result<bool, EntryStorageError>;
-
-    /// Returns entry at sequence position within an author's log.
-    async fn entry_at_seq_num(
-        &self,
-        author: &Author,
-        log_id: &LogId,
-        seq_num: &SeqNum,
-    ) -> Result<Option<StorageEntry>, EntryStorageError>;
-
     /// Returns the latest Bamboo entry of an author's log.
     async fn latest_entry(
         &self,
--- a/p2panda-rs/src/storage_provider/traits/entry_store.rs
+++ b/p2panda-rs/src/storage_provider/traits/entry_store.rs
@@ -160,9 +160,9 @@ pub trait EntryStore<StorageEntry: AsStorageEntry> {
 
     async fn get_all_lipmaa_entries_for_entry(
         &self,
-        author_id: Author,
-        log_id: LogId,
-        seq_num: SeqNum,
+        author_id: &Author,
+        log_id: &LogId,
+        seq_num: &SeqNum,
     ) -> Result<Vec<StorageEntry>, EntryStorageError>;
 }
 
@@ -300,9 +300,9 @@ pub mod tests {
 
         async fn get_all_lipmaa_entries_for_entry(
             &self,
-            author: Author,
-            log_id: LogId,
-            initial_seq_num: SeqNum,
+            author: &Author,
+            log_id: &LogId,
+            initial_seq_num: &SeqNum,
         ) -> Result<Vec<StorageEntry>, EntryStorageError> {
             let seq_num = initial_seq_num.as_u64();
             let cert_pool_seq_nums: Vec<SeqNum> = get_lipmaa_links_back_to_root(seq_num)
@@ -313,7 +313,7 @@ pub mod tests {
             let mut cert_pool: Vec<StorageEntry> = Vec::new();
 
             for seq_num in cert_pool_seq_nums {
-                let entry = match self.entry_at_seq_num(&author, &log_id, &seq_num).await? {
+                let entry = match self.entry_at_seq_num(author, log_id, &seq_num).await? {
                     Some(entry) => Ok(entry),
                     None => Err(EntryStorageError::CertPoolEntryMissing(seq_num.as_u64())),
                 }?;
--- a/p2panda-rs/src/storage_provider/traits/entry_store.rs
+++ b/p2panda-rs/src/storage_provider/traits/entry_store.rs
@@ -38,7 +38,7 @@ pub trait EntryStore<StorageEntry: AsStorageEntry> {
         let expected_backlink = self
             .entry_at_seq_num(&entry.author(), &entry.log_id(), &backlink_seq_num)
             .await?
-            .ok_or(EntryStorageError::ExpectedBacklinkMissing(entry.hash()))?;
+            .ok_or_else(|| EntryStorageError::ExpectedBacklinkMissing(entry.hash()))?;
 
         // compare the expected backlink hash and the stated backlink hash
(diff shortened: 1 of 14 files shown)


sandreae:
(god damn clippy)

sandreae:
@pietgeursen any comments very welcome!

One thing in particular, I was wondering if the return values are what you would expect from eg `get_next_n_entries_after_seq`? Like, when you would expect it to return `Ok(None)`, `Ok(Some(Vec<..>))` and `Error().

pietgeursen:
> One thing in particular, I was wondering if the return values are what you would expect from eg get_next_n_entries_after_seq? Like, when you would expect it to return Ok(None), Ok(Some(Vec<..>)) and `Error().

Yeah, fair enough. That one could become `Result<Vec>`

But the other ones that get single entries should still be `Result<Option<StorageEntry>>`. 
`Result` to tell you if something unexpected went wrong (eg with the storage). The `Option` to tell you if that entry was in the db or not.

sandreae:
> Yeah, fair enough. That one could become Result<Vec>

Actually, I thought on that a while, and found that wrapping the returned vec in an `Option` was useful for showing that the requested starting entry doesn't exist. That's how I implemented it now. What do you think?

pietgeursen:
Yeah I'm fine with that too!

pietgeursen:
What are your thoughts on this one: 

```rs
async fn get_all_lipmaa_entries_for_entry(author_id: Author, log_id: LogId, seq_num: SeqNum) -> Result<Option<Vec<StorageEntry>>>
```

sandreae:
What is the exact use case you're thinking about? 

As we check the lipmaa is present before storing any entry in the db, I'm not sure when we need to go all the way back to seq number 1 in one go like this.

pietgeursen:
The use case is that a remote server wants to get a single entry (either by seq + author + log_id or by hash) and they also want to verify that it's valid.

So they want the entry they're interested in + all the lipmaa entries back to seq number 1. 

I do think we want to get them all in one query. I've seen from experience that wherever you can, you should prefer doing stuff in batches. The cost of hitting the file system + db is high.

sandreae:
Haha, lolz, sorry, brain fart. It's for the whole reason bamboo is cool.

sandreae:
Yeh, I can add that into the store for sure.

sandreae:
Yeh, I can add that into the store for sure.

p2panda-rs/src/storage_provider/traits/entry_store.rs
```diff
@@ -15,6 +15,88 @@ use crate::storage_provider::traits::AsStorageEntry;
...
+            // Or if there is no skiplink for entries at this sequence number return None
+            None => None,
+        };
+
+        // compare the expected skiplink hash and the stated skiplink hash
+        if expected_skiplink.clone().map(|entry| entry.hash()) != entry.skiplink_hash() {
+            return Err(EntryStorageError::InvalidSkiplinkPassed(entry.hash()));
+        }
+        Ok(expected_skiplink)
+    }
```
sandreae, line 98: I'd like a little feedback on this method. I see two possible approaches (maybe there are more?).

1) Fetch the expected skiplink for an entry from storage based on `Author`, `LogId` and `SeqNum`, if it is found in storage we then validate it's hash against the one encoded in the db.

2) Fetch the entry for an entries' encoded skiplink by it's hash, compare it's `Author` `SeqNum` and `LogId` to the expected values.

Option 1) is what we are currently doing, option 2) is a different approach. The reason I wonder about option 2) is because I believe in most implementations queries for fetching an entry by it's hash will be more efficient (this is an indexed value in the db). Rather than scanning across author, log_id and seq_num as would be the case I think in option 1). When we implement `get_all_lipmaa_entries_for_entry()` performance may be more of an issue.

Any thoughts on this anyone? Maybe @pietgeursen has particular insight ;-p ?

pietgeursen: > Rather than scanning across author, log_id and seq_num

At least in aquadoggo, you have: `PRIMARY KEY (author, log_id, seq_num)` which should make it very fast to search for them.

> When we implement get_all_lipmaa_entries_for_entry() performance may be more of an issue.

Are you going to try and implement `get_all_lipmaa_entries_for_entry` by calling `try_get_skiplink` multiple times?

I'd rather not do that for performance reasons. I think that can be done in a single query like:

(probably incorrect sql)
```sql
SELECT * FROM entries WHERE author = <> AND log_id = <> AND seq_num IN (1,2,3 ...)
```

sandreae: Thanks, yeh, you're right, in aquadoggo we are covered.

re: `get_all_lipmaa_entries_for_entry` I was planning to do something like your psuedo code 👍

You don't see any downside to option 2) then? I'm mostly just curious 😄

pietgeursen: I just made [this PR](https://github.com/pietgeursen/lipmaa-link/pull/3) in `lipmaa_link`. Do you reckon that's be handy for this?

pietgeursen: > You don't see any downside to option 2) then?

I've had one beer so I'm not thinking super clearly. But does one give us an advantage in detecting forks? Maybe not? Idk lol.

sandreae: Haha, I've had one too many Onsen (and beers) to be thinking clearly 🤣

But that's the kind of thing I was wondering about. I can't quite tell if the two methods are exactly the same, or subtly different....

sandreae: > Do you reckon that's be handy for get_all_lipmaa_entries_for_entry?

Ahh, thanks, yes, useful.

p2panda-rs/src/storage_provider/traits/entry_store.rs
```diff
@@ -18,29 +18,137 @@ use crate::storage_provider::traits::AsStorageEntry;
 #[async_trait]
 pub trait EntryStore<StorageEntry: AsStorageEntry> {
     /// Insert an entry into storage.
+    ///
+    /// Returns `true` if insertion was succesful, returns `false` if no error occured but an
+    /// unexpected number of insertions happened. Errors if a fatal storage error occured.
```
cafca, line 23: Which scenario would cause an unexpected number of insertions without being a fatal error? Not saying the API should be different, I am just not sure what that comment means.

p2panda-rs/src/storage_provider/traits/entry_store.rs
```diff
@@ -18,29 +18,137 @@ use crate::storage_provider::traits::AsStorageEntry;
...
     async fn by_schema(&self, schema: &SchemaId) -> Result<Vec<StorageEntry>, EntryStorageError>;
 
-    /// Determine skiplink entry hash ("lipmaa"-link) for entry in this log, return `None` when no
-    /// skiplink is required for the next entry.
-    async fn determine_skiplink(
+    /// Get all entries of a log from a specified sequence number up to passed max number of entries.
+    ///
+    /// Returns a vector of entries the length of which will not be greater than the max number
+    /// passed into the method. Fewer may be returned if the end of the log is reached.
+    async fn get_next_n_entries_after_seq(
```
cafca, line 141: Method name is a mouthful :D What do you think about something like `get_log_slice`, or `get_paginated_log_entries` (this one could go with `from_seq_num` and `page_size` as params)?

p2panda-rs/src/storage_provider/traits/entry_store.rs
```diff
@@ -65,26 +173,41 @@ pub trait EntryStore<StorageEntry: AsStorageEntry> {
 
         entry_skiplink_hash
     }
+
+    /// Get all entries which make up the certificate pool for the given entry.
+    ///
+    /// Returns a result containing vector of entries wrapped in an option. If no entry
+    /// could be found at this author - log - seq number location then an error is
+    /// returned.
+    async fn get_all_lipmaa_entries_for_entry(
```
cafca, line 182: We've been using the "skiplink" name in the API so far and I think it would be good to not add another term here.

cafca (changes requested):
Looks good! I think we should stick to the "skiplink" wording when possible.

p2panda-rs/src/storage_provider/traits/entry_store.rs
sandreae: My preferred approach would probably be be to still check, but instead of returning `true` or `false` we just error if an unexpected number of insertions happened.

p2panda-rs/src/storage_provider/traits/entry_store.rs
sandreae: haha, it is rather. I like `get_paginated_log_entries()` I'll change it to that :+1:

sandreae:
Thanks for the review @cafca :pray: !

cafca:
This looks all good! I have one more idea :'D

`get_all_skiplink_entries_for_entry` -> `get_certificate_pool`

What do you think?

p2panda-rs/src/storage_provider/traits/entry_store.rs
```diff
@@ -18,29 +18,136 @@ use crate::storage_provider::traits::AsStorageEntry;
...
+    /// Get all entries of a log from a specified sequence number up to passed max number of entries.
+    ///
+    /// Returns a vector of entries the length of which will not be greater than the max number
+    /// passed into the method. Fewer may be returned if the end of the log is reached.
+    async fn get_paginated_log_entries(
+        &self,
+        author: &Author,
+        log_id: &LogId,
+        seq_num: &SeqNum,
+        max_number_of_entries: usize,
```
cafca, line 145: I think `page_size` would make sense here to go with the idea of pagination.

sandreae: I really like `get_certificate_pool()` changed that now. Not sure about `page_size`, but maybe I'm not familiar with paginated term conventions?

sandreae:
I'm gunna merge as I understand this is signed off except for agreeing on one naming thing.

Issue #316, opened 2022-05-08
More meta data on `DocumentView` 


sandreae:
As a result of working with `DocumentView` more in `aquadoggo` I've found it would be useful to have access to a little more data on this struct. 

The wish-list includes:

- Access to the relevant `OperationId` for each field of a `DocumentView`
- ~~The views `DocumentViewId`~~ It already contains this :rofl:

The reason for needing the `OperationId`s came up largely because of the db layout we are using in `aquadoggo`, with it we can relate a document views fields directly to their relevent operations, and when updates occur, we change the id to update the view/document. However, I think it is still an interesting piece of data and may be useful in more general terms. It can imagine it being needed for sparse materialisation for example.

This first point could be achieved by creating a `DocumentViewValue` struct which contains both the field value as well as it's operation id. This would be derived during graph traversal in `DocumentBuilder`. 

```rust
pub struct DocumentViewValue {
    operation_id: OperationId,
    value: OperationValue,
}

type FieldKey = String;
type DocumentViewFields = BTreeMap<FieldKey, DocumentViewValue>;
``` 

The `DocumentViewId` already exists on `Document`, and so this would be moved/added to `DocumentView`. 

The original reason for keeping `DocumentView` rather slim was that we imagined handlind `Document` more, and maybe performing incremental updates on it on the fly at some point. This hasn't come to be yet, and as we are needing these properties on `DocumentView` it makes sense to have them there.

sandreae:
Following this direction, an interesting reflection is that the document materialisation (graph traversal) process we are employing actually doesn't need to know about the `OperationValue`s in order to materialise a view. `DocumentBuilder` could build skeleton views containing the final operation ids for each field, which are then fetched from storage. 

That's a conversation for another issue though ;-p

PR #319, opened 2022-05-11
Changes to `Document` 

sandreae:
Some additions to `DocumentView` which I feel are needed when working with this data type in `aquadoggo`. Make `DocumentView` on `Document` optional in order to represent deleted documents.

## Reasoning
- more separation wanted between document view fields / values and operation types
- a document which has been deleted generates a new document view Id but no actual view. We need a way of representing this state in `Document`.
- in `aquadoggo` we need to know which operation transported the current value on a documents fields in order to store it in the db. This is somewhat due to the current db layout, however as I thought about it more I feel it's nice metadata to have anyway (for debugging, and sparce materialisation). It also requires little extra processing overhead as we can easily add the operation id's when resolving the document view. 

### `DocumentViewFields` 
the fields of a document view as `String` -> `DocumentViewValue` mappings
### `DocumentViewValue`
struct containing the `OperationId` and `OperationValue` for a field on a document view
### `Document` changes
Make `DocumentView` optional

closes: #316 

## 📋 Checklist


p2panda-rs/src/document/mod.rs
```diff
@@ -174,7 +189,7 @@
 //! //                A
 //! //                |
 //! //                |                  +++++++++++++++++++++++++++
-//! //                -----------------  | name: "🐼 Cafe!!"        |
+//! //                -----------------  | name: "🐼 Cafe!"        |
```
cafca, line 192: :D

sandreae: :p

sandreae:
Hmmmm.....

Now I'm not completely sure about this pattern, it ends up in a strange pattern of having to insert deleted `DocumentViews` into the database, and if we want to reconstruct them then we need to persist the `is_deleted` status of each document view (also a weird pattern, if a document/document view is deleted, it doesn't exist anymore). 

I think this maybe just highlights that we (or just me?) don't have a really clear idea of how we want to handle deletion in the database. Any ideas :rofl:

sandreae:
The above comment doesn't change anything about needing field ids and the other changes here, just how/where/when/what etc.. we do about deletion wrt documents and document views.

sandreae:
I think the answer is that document views don't really know about deletion... `Documents` do, and if they are deleted, then we decide what to do at an implementation level. When a document gets deleted no new document view is created.

sandreae:
We could make `DocumentView` on `Document` optional here:

https://github.com/p2panda/p2panda/blob/19a6612affc5875953e13d7cb32ea89edbb09587/p2panda-rs/src/document/document.rs#L82-L88

sandreae:
I'm :+1: this myself, I think it's a nice solution.

p2panda-rs/src/document/document.rs
```diff
@@ -37,32 +39,30 @@ pub(super) fn build_graph(
...
 type IsDeleted = bool;
 
 /// Reduce a list of operations into a single view.
-pub(super) fn reduce<T: AsOperation>(
-    ordered_operations: &[T],
-) -> (BTreeMap<FieldKey, OperationValue>, IsEdited, IsDeleted) {
+pub(super) fn reduce(
+    ordered_operations: &[OperationWithMeta],
+) -> (Option<DocumentViewFields>, IsEdited, IsDeleted) {
     let is_edited = ordered_operations.len() > 1;
```
cafca, line 49: I just noticed a problem with this `IsEdited` logic. It will return the `IsEdited` marker for documents that were created and immediately deleted even though such a document has never been edited. I think we should initialise this as `false` and then set it to `true` when encountering the first edit instead.

cafca: This would also be a good thing to put into the docstring.

cafca (changes requested):
The changes are good! You could add a sentence about the subtleties of document deletion to the handbook https://p2panda.org/handbook/docs/organising-data/documents#document-views

p2panda-rs/src/document/document.rs
sandreae: Yes, nice, good spot!

Here are the handbook changes: https://github.com/p2panda/handbook/pull/170

cafca:
Is there something wrong with codecoverage again? I noticed the coverage changes, for example in `definitions.rs` but don't understand why those tests aren't run anymore?

https://app.codecov.io/gh/p2panda/p2panda/compare/319/changes

sandreae:
Yeh, codecov seems weird, I'm sure I got the coverage into positive too...

p2panda-rs/src/document/document.rs
```diff
@@ -37,32 +39,39 @@ pub(super) fn build_graph(
...
 type IsDeleted = bool;
 
 /// Reduce a list of operations into a single view.
-pub(super) fn reduce<T: AsOperation>(
-    ordered_operations: &[T],
-) -> (BTreeMap<FieldKey, OperationValue>, IsEdited, IsDeleted) {
-    let is_edited = ordered_operations.len() > 1;
-    let mut is_deleted = false;
+///
+/// Returns the reduced fields of a document view along with the `IsEdited and `IsDeleted boolean flags.
```
adzialocha, line 47: The backticks in the docstring are a little off 😧

p2panda-rs/src/document/document_view_fields.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+impl DocumentViewValue {
+    /// Returns a `DocumentViewValue` constructed from an `OperationId` and `OperationValue`
+    pub fn new(operation_id: &OperationId, value: &OperationValue) -> Self {
+        Self {
+            operation_id: operation_id.clone(),
+            value: value.clone(),
+        }
+    }
```
adzialocha, line 24: Newline missing

adzialocha (commented):
Sorry for commenting on these little things, they are mostly notes to myself (or someone is faster than me). I can change them myself later as soon as I'm back at the computer.

sandreae:
Regarding code-cov, it seems to have been this commit where the coverage went wonky: https://app.codecov.io/gh/p2panda/p2panda/commit/8d6185d1506a753b3dd0d8c575071f285104ca99

sandreae:
And that commit only changed doc strings....

sandreae:
I think this is just a glitch, maybe it doesn't block merging?

sandreae:
codecov sorted after merging main.

Issue #167, opened 2021-12-23
Revisit how we process MLS commits directly


adzialocha:
See: https://github.com/openmls/openmls/pull/659 and https://github.com/openmls/openmls/issues/663

Issue #147, opened 2021-12-08
Update new OpenMLS `process` API


adzialocha:
Depends on: https://github.com/openmls/openmls/pull/576

Issue #146, opened 2021-12-08
MLS: Remove unnecessary wire format checks


adzialocha:
This should be handled by `openmls` and not by us: https://github.com/p2panda/p2panda/blob/2c1883d0e03f74b8e8defed0f58db2c5d997e3fe/p2panda-rs/src/secret_group/commit.rs#L30-L36

Depends on: https://github.com/openmls/openmls/issues/584

Issue #145, opened 2021-12-08
Remove `SecretGroup` members based on `KeyPackageId` instead of leaf index


adzialocha:
Depends on: https://github.com/openmls/openmls/issues/541

PR #336, opened 2022-06-02
Update OpenMLS crates to latest `v0.4.1` version

adzialocha:
Bumps the version of the OpenMLS crate to `v0.4.0` and fixes a whole bunch of related breaking API changes:

- [x] #141 + bump related crates to latest versions
- [x] Fix new import paths and feature flags
- [x] #147 
- [x] #167 
- [x] #142 
- [x] #145 
- [x] #146 

## 📋 Checklist


adzialocha:
I've started this PR with some basic things just to see how much errors there are actually and fixed the easy ones where the paths changed or things got moved behind feature flags.

I know that you @cafca wanted to work on this, just wanted to help with the initial things which might not be obvious (changed crate names, feature flags etc.), let me know if you have any questions!

p2panda-rs/src/secret_group/group.rs
```diff
@@ -189,52 +189,13 @@ impl SecretGroup {
...
-        };
-
-        self.mls_group
-            .members()
-            .unwrap()
-            .into_iter()
-            .map(read_author)
-            .collect()
+    pub fn members(&self) -> Vec<&KeyPackage> {
+        self.mls_group.members()
```
adzialocha, line 193: The new MLS spec removes members identified by a `KeyPackage` hash. To generate the hashes we need to return the `KeyPackage` of all members here instead of `Author`.

I'm sure there is a way to make this come closer to p2panda as soon as integrate it with our operations but on that level here we have to do it like that now.

p2panda-rs/src/secret_group/mls/constants.rs
```diff
@@ -23,3 +24,24 @@ pub const MLS_PADDING_SIZE: usize = 16;
...
+/// This parameter defines how many incoming messages can be skipped in case they have been
+/// dropped, deleted or are missing.
+pub const MLS_MAX_FORWARD_DISTANCE: u32 = 1024;
+
+/// This parameter defines a window for which decryption secrets are kept.
+///
+/// This is useful in case we cannot guarantee that all application messages have total order
+/// within an epoch. Use this carefully, since keeping decryption secrets affects forward secrecy
+/// within an epoch.
+pub const MLS_OUT_OF_ORDER_TOLERANCE: u32 = 16;
```
adzialocha, line 47: These new numbers are fairly arbitrary and should be tested in a real-world use case sometime, until then they are somewhere close to the official `openmls` defaults.

p2panda-rs/src/secret_group/mls/group.rs
```diff
@@ -206,15 +239,19 @@ mod tests {
...
         let group_id = GroupId::random(&provider);
         let key_package = member.key_package(&provider).unwrap();
-        let mut group = MlsGroup::new(&provider, group_id, key_package).unwrap();
+        let mut group = MlsGroup::new(&provider, group_id, key_package.clone()).unwrap();
+
+        // Group is active and contains the owner of the group as the only member
         assert!(group.is_active());
+        assert_eq!(group.members(), vec![&key_package]);
 
+        // Owner can not decrypt its own messages (for forward secrecy)
```
adzialocha, line 250: Yes, sad! But it makes sense now!

sandreae:
Whoop!!! :fire: :fire: :fire: I will review this tomorrow :pray:

p2panda-rs/src/secret_group/mls/constants.rs
sandreae, line 47: These configurations are interesting. Am I understanding them right when I imagine we could tweak these to, eg. try and account for clients staying in an older epoch for slightly too long?

p2panda-rs/src/secret_group/mls/group.rs
```diff
@@ -206,15 +237,19 @@ mod tests {
...
```
sandreae, line 250: Hmmm, I don't understand this comment, could you expand a little?

sandreae (approved):
Hey, this all looks good to me, I just had a couple of questions for my own understanding. Cool to see how the OpenMLS team helped us out so much making the changes we needed. Really great we can make a release after this :+1: Thanks for doing all these updates, it was quite a lot and you did it super quick ;-p

p2panda-rs/src/secret_group/mls/group.rs
adzialocha: openmls deletes the session key material from this client as soon as it's been used (for encryption), other clients will do the same right after they decrypted the message. So basically you can do it only once and have to store the plaintext somewhere.

This is part of forward secrecy and improves the security around an attacker getting the private key not being able to decrypt old messages as the session keys are gone.

PR #337, opened 2022-06-02
Materialise specific document view

sandreae:
Materialise a document at a specific document view.

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/document/document.rs
+++ b/p2panda-rs/src/document/document.rs
@@ -175,19 +175,12 @@ impl Display for Document {
 pub struct DocumentBuilder {
     /// All the operations present in this document.
     operations: Vec<OperationWithMeta>,
-
-    /// The specific document view we want to materialise. If not set, then all operations
-    /// will be included.
-    view_id: Option<DocumentViewId>,
 }
 
 impl DocumentBuilder {
     /// Instantiate a new `DocumentBuilder` from a collection of operations.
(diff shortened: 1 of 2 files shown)


p2panda-rs/src/document/document.rs
```diff
@@ -162,20 +162,34 @@ impl Display for Document {
...
+        }
     }
 
     /// Get all operations for this document.
     pub fn operations(&self) -> Vec<OperationWithMeta> {
         self.operations.clone()
     }
 
+    /// Set the view id we want to materialise.
+    pub fn set_view_id(&mut self, id: DocumentViewId) -> &Self {
```
adzialocha, line 188: I read the `set_view_id` method as some sort of configuration of the next time I'm building something right now in that code. If thats the case then I would make it an (optional) argument of the the `build` method, especially since it is only used there.

If its more understood as a configuration of the whole struct, then I'd rather see it as part of the constructor (again as an optional argument).

If you don't want to change the signature of `build` one could move its logic into a private method which has that `view_id` argument and make `build` and a new `build_with_view_id` (or something like that) method call it internally.

All of these are fine with me, with `set_view_id` I think of a "builder" pattern even though it is not one.

sandreae: Yes! Nice suggestion, thanks. I'll do that now.

adzialocha (approved):
Ah, that's even better! Yay! 🤩

Issue #307, opened 2022-05-04
`RelationList` and `PinnedRelationList` equality failures


sandreae:
`RelationList` and `PinnedRelationList` fail equality comparison if they contain the same, but differently ordered ids.

sandreae:
Short term patch by sorting hashes in the constructor here: https://github.com/p2panda/p2panda/pull/308

But still needs specing out properly and dealing with in se/de cases too.

adzialocha:
> RelationList and PinnedRelationList fail equality comparison if they contain the same, but differently ordered ids.

I was thinking about it a little bit and realized that maybe this is actually not the behaviour we want? A relation list order should not be enforced as the user might want to indicate an order in them themselves, for example, if I want to relate to my favorite foods I might want to choose a priority (tastier food first). In this sense we shouldn't sort the lists in the constructor ..

In terms of `Eq` and `PartialEq` I'm not sure what to do here. Should we sort them before comparison or actually consider `[Pizza, Burger, Sushi]` something different than `[Sushi, Burger, Pizza]` when it is about a relation list to our favorite food documents?

The question is also, where was this change needed? I guess you had a blocker somewhere in `aquadoggo`, maybe this can be a hint to what the right thing here is to do.

sandreae:
> The question is also, where was this change needed? I guess you had a blocker somewhere in aquadoggo, maybe this can be a hint to what the right thing here is to do.

The issue came up when I was running tests in `aquadoggo`. I was creating an operation which had a relation list field, encoding it, storing it, retrieving it, decoding it and then comparing it to an expected operation with matching fields. However there seemed to be an indeterminate element to the relation sorting, sometimes they matched, sometimes not...

 > maybe this is actually the behaviour we want?
Not sure I feel the same on this, I don't  think I would expect meaningful application information to be encoded in the relation list ordering. I would expect two entries, which encode operations with relation lists containing matching items to be equal. If the order is important, it could be defined in the schema of the data one is relating to.
 
 > Thats connected to our spec here https://p2panda.org/handbook/docs/writing-data/operations#serialisation and one might argue that this ordering actually holds semantics.

As I said above, I would argue that there isn't semantic meaning, and so the arrays should be sorted before encoding. Actually, where is the case where item ordering is important for semantic meaning?

adzialocha:
I'm not sure if I understand your point fully, why this should _not_ be semantic. Why do you think this is not the case?

> Actually, where is the case where item ordering is important for semantic meaning?

[...]

PR #341, opened 2022-06-03
Revert relation list sorting

sandreae:
Discussion here: https://github.com/p2panda/p2panda/issues/307

## 📋 Checklist

- [x] ~~Add this PR to the _Unreleased_ section in `CHANGELOG.md`~~ The original PR which introduced this behaviour was not added to the CHANGELOG, so this removal of those changes should also be ommited.
- [x] - [x] Link this PR to any issues it closes

### DIFF
--- a/p2panda-rs/src/operation/relation.rs
+++ b/p2panda-rs/src/operation/relation.rs
@@ -252,4 +252,24 @@ mod tests {
             }
         }
     }
+
+    #[rstest]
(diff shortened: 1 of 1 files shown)


adzialocha (approved):
Super, thank you! Let's see what the issue is ..

I assume it will be async db stuff where the results arrive in different order every time, but that's just a gut feeling 😅

PR #359, opened 2022-06-15
Use TypeDoc to generate `p2panda-js` documentation

adzialocha:
This PR introduces [TypeDoc](https://typedoc.org/) to generate our TypeScript documentation in `p2panda-js`. The documentation will be built / hosted in the `handbook` repository, alongside the rest.

Closes: #356 

## 📋 Checklist


sandreae (approved):
This looks great to me!

One small question. Is it possible to change the order in which the method docs show? Only really in Session, the alphabetical ordering seems a bit strange (create, update & delete would make sense to stay together). Maybe this isn't possible, or annoying though, not a big deal.

adzialocha:
> One small question. Is it possible to change the order in which the method docs show? Only really in Session, the alphabetical ordering seems a bit strange (create, update & delete would make sense to stay together). Maybe this isn't possible, or annoying though, not a big deal.

There seems to be an option here: https://typedoc.org/guides/options/#sort I can experiment with it and see what comes out of it!

Issue #360, opened 2022-06-15
Also store the `Operation` in `publish_entry()`


sandreae:
Currently we only store the `Entry` in `publish_entry()`. It feels natural that we also store the `Operation`, let's do that.

sandreae:
We could do this in 2 lines of code now if we like....

adzialocha:
> We could do this in 2 lines of code now if we like....

Slightly related thought: https://github.com/p2panda/p2panda/issues/360#issuecomment-1166218270

I wonder if we could take out some functionality of publish entry in external, composable methods so they will be useful for other things as well?

adzialocha:
I think `publish_entry` still makes sense, maybe next to something like `batch_publish_entries`, but they should be composed of flows we can also compose differently. I'd imagine something like:

```
publish_entry(entry_encoded, operation_encoded) -> Result<()> {
  insert_entry(entry_encoded)?;
  insert_operation(operation_encoded)?;
}

insert_entry(entry_encoded) -> Result<()> {
   validate_entry_encoded(entry_encoded)?;
   store.insert(entry_encoded)?;
}

...
```

This is obviously more far from what is possible with out data types, but I just want to sketch a more "onion" approach, where there is very low level, more unsafe but composable approaches, some more mid-level ones and very high-level ones close to our GraphQL API (where I'm not sure right now if it should still be called `publishEntry`?).

PR #361, opened 2022-06-15
Implement `OperationStore` on test storage provider

sandreae:
This is required preparation for later implementing (and testing) storing `Operation`'s inside `publish_entry()`: https://github.com/p2panda/p2panda/issues/360 

- Implement `OperationStore` on the test storage provider `SimplestStorageProvider`
- use same `test_db` pattern as in `aquadoggo` [later this can be exported and made generic so it can be constructed with different backends]
- refactor tests to use new `test_db`

Follow-up: #366 

## 📋 Checklist


p2panda-rs/src/operation/traits.rs
```diff
@@ -61,7 +65,30 @@ pub trait AsOperation {
...
+    ///
+    /// Use `VerifiedOperation::new_from_entry()` instead if you want to validate that the operation
+    /// was signed by this public key.
+    fn new(
+        public_key: &Author,
+        operation_id: &OperationId,
+        operation: &Operation,
+    ) -> Result<Self, Self::VerifiedOperationError>;
+
+    /// Returns a new `VerifiedOperation` instance constructed from an `EntrySigned`
```
adzialocha, line 84: This comment somehow suggests a functionality but it is not implemented as it is a trait. I would rephrase it as a suggestion and maybe someday we can even enforce some safety by how we design the traits.

I don't want to open a whole chapter here in this PR on that topic but I wonder how strict we should be around making validation promises and if our current api is reflecting that. It's maybe something to look at again together with the whole validation topic.

(There is a small typo in "oncoded").

p2panda-rs/src/storage_provider/traits/log_store.rs
```diff
@@ -56,8 +56,14 @@ pub trait LogStore<StorageLog: AsStorageLog> {
 
 #[cfg(test)]
 pub mod tests {
+    // TEST MODULE NOTES //
+    //
+    // In this module we are implementing the `LogStore` trait on `SimplestStorageProvider` just
```
adzialocha, line 61: Im not sure if I understand this comment: aren't all tests written inside the `tests` module solely for testing purposes? 🤔

p2panda-rs/src/operation/traits.rs
sandreae: Yes, very good point.

In this particular case, these methods were only implemented on the struct before, and in this PR I moved them into the trait definition. So maybe the comments snuck across too :grimacing: .... I'll re-write the comment to be more trait-description-like. 

In general, yeh, this would be a very good conversation to have. I particularly get tripped up on where to place constructors. Sometimes they're needed in the trai (like here) but it feels a bit weird for some reason...

p2panda-rs/src/operation/traits.rs
```diff
@@ -61,7 +65,30 @@ pub trait AsOperation {
...
-pub trait AsVerifiedOperation {
+pub trait AsVerifiedOperation:
+    Sized + Clone + Send + Sync + Validate + PartialEq + std::fmt::Debug
+{
+    /// Error type for `AsVerifiedOperation`
+    type VerifiedOperationError: 'static + std::error::Error + Send + Sync;
+
+    /// Returns a new `VerifiedOperation` instance.
+    ///
+    /// Use `VerifiedOperation::new_from_entry()` instead if you want to validate that the operation
```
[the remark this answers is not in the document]
sandreae: Hmm, lolz, yeh this is weird. Is it just a naming thing? If this was called `AuthoredOperation` then it would be less weird? 

Maybe the method names could be reversed? `new()` takes entry and operation and verifies. Maybe we can place the other behind a testing flag actually, I don't think it is hugely useful in the wild. These methods aren't actually new in this PR though. So maybe I should open an issue and deal with this in a separate PR.

p2panda-rs/src/storage_provider/traits/log_store.rs
sandreae: Yeh, it clear without the comment why the traits are being implemented here in the test module? It's for us to test the methods with default implementations, and also so we can test the higher level methods on `StorageProvider`? This is what I was trying (obviously not very well) to explain.

p2panda-rs/src/operation/traits.rs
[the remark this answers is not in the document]
sandreae: I raised an issue for this: https://github.com/p2panda/p2panda/issues/366

p2panda-rs/src/storage_provider/traits/log_store.rs
adzialocha: Traits in a library crate are weird indeed. You provide an interface to the outside you don't use yourself (inside the library), except of maybe tests and partly in our some default implementations. I think its safe to assume that this is "normal" when testing traits in a library.

p2panda-rs/src/operation/traits.rs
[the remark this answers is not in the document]
sandreae: I'm gunna deal with this in a follow-up PR as the issue wasn't introduced in this PR.

p2panda-rs/src/storage_provider/traits/operation_store.rs
```diff
@@ -49,5 +49,247 @@ pub trait OperationStore<Operation: AsVerifiedOperation> {
...
+            .await
+            .unwrap()
+            .is_none());
+
+        db.store
+            .insert_operation(&create_operation, &document_id)
+            .await
+            .unwrap();
+
+        assert_eq!(
```
cafca, line 237: Wondering why this line is reported as not being hit 🤔

p2panda-rs/src/operation/verified_operation.rs
```diff
@@ -27,12 +27,14 @@ pub struct VerifiedOperation {
...
-impl VerifiedOperation {
+impl AsVerifiedOperation for VerifiedOperation {
+    type VerifiedOperationError = VerifiedOperationError;
+
     /// Returns a new `VerifiedOperation` instance.
     ///
     /// Use `VerifiedOperation::new_from_entry()` instead if you want to validate that the operation
     /// was signed by this public key.
-    pub fn new(
+    fn new(
```
cafca, line 37: I'm still a bit unsure about documenting traits vs documenting structs but maybe trait items like this should have a description of what implementations need to verify - or don't need to verify. In this case it could say something like:

> Implementations should document that this method doesn't validate whether it's parameters match each other and that consumers must ensure that parameters were validated before using this constructor.

cafca (approved):
I had one suggestion for a trait item docstring but it's good!

p2panda-rs/src/operation/verified_operation.rs
sandreae: Yes, I like that!

sandreae: I assumed you meant changing the docs in the trait definition?

sandreae: Hmm, also, thinking again. Not validating here isn't _compulsory_, if an implementation can do some validation somehow, this is fine. The thing we want to strongly encourage is that `new_from_entry)` implementations _do_ perform validation.

p2panda-rs/src/storage_provider/traits/operation_store.rs
sandreae: Yeh, I don't know either... I've seen this behaviour before too...

p2panda-rs/src/operation/verified_operation.rs
sandreae: What do you think @cafca ? Is this good to go as is?

cafca: Yes!

PR #362, opened 2022-06-15
Untagged operation serialisation

cafca:
Introduces a new intermediary form for operations when they are serialised - `PlainOperation`, which represents operation fields as `ciborium::value::Value`. 

Previous:

```json
{
    "action": "create",
    "schema": "chat_0020c65567ae37efea293e34a9c7d13f8f2bf23dbdc3b5c7b9ab46293111c48fc78b",
    "version": 1,
    "fields": {
        "message": {
            "type": "str",
            "value": "Ohh, my first message!"
        }
    }
}
```

after

```json
{
    "action": "create",
    "schema": "chat_0020c65567ae37efea293e34a9c7d13f8f2bf23dbdc3b5c7b9ab46293111c48fc78b",
    "version": 1,
    "fields": {
        "message": "Ohh, my first message!"
    }
}
```

I went a bit fast and loose implementing this so there are some todos left if we want to go forward with this.

(Rust) tests should be working all, though! 👯‍♂️

- [ ] Document new items
- [ ] Should `PlainOperation` be public or an internal tool only? Could be helpful for testing.
- [ ] Consider adding additional tests around conversion
- [ ] refactor schema fixtures, especially their naming 
- [ ] Should `v0_3_0_fixture` have a hardcoded way to access the test schema used?
- [ ] Some validations methods were decoding an operation payload in order to validate it. I wasn't sure if that was intended but I replaced those with calls to `EntryEncoded::validate_operation`
- [ ] Should `AsStorageEntry::operation` need to be able to decode the operation? (which requires access to the schema) It seems like that is more due to the way `SimplestStorageProvider` is implemented rather than the trait design itself but I may be wrong.
- [ ] I changed the `EntryStore::get_entries_by_schema` implementation so that it requires there to be logs for all entries. Seems reasonable but that was not the case before, just wanted to highlight that.
- [ ] I added a way to create `Schema` instances on the fly with `Schema::new_definition` but that should maybe become a test utility instead. Debatable!

## 📋 Checklist


adzialocha:
Woah 🤩✨!

sandreae:
I just had a really brief skim so far, but I like it! The reason for `PlainOperation` and how to use it is quite intuitive, and I can imagine having this `PlainOperation` into `Operation` conversion doesn't actually add any more complexity on say, the node, as we would be validating the operation with the schema in any case. I actually really like that this change brings schema more into use/visible in `p2panda-rs`, rather than something that we just somehow fake everywhere, knowing that "somewhere" (on the node) it will be used massively. Obvs having smaller payloads is great too :+1:

adzialocha:
> Obvs having smaller payloads is great too +1

And it looks cool as :cool: :cookie: :sunglasses:

adzialocha:
Closing this now as we will start from scratch, we keep it here as a reference for the future implementation :+1:

PR #367, opened 2022-06-18
Improve validation in entry signed

sandreae:
We don't perform validation of the encoded entry when constructing an `EntrySigned` from a string (just if it is valid hex). this should be improved and tested.

related: https://github.com/p2panda/aquadoggo/issues/157

- [x] Catch empty string
- [x] Validate that the entries' properties are correct (follow the bamboo spec)
- [x] fixtures and test constructors to make this easier to test
- [x] test the hell out of it (with some prop-testing :cake:)

## 📋 Checklist


p2panda-rs/src/entry/entry_signed.rs
```diff
@@ -176,10 +188,174 @@ mod tests {
...
+            entry_unvalidated(
+                14,
+                1,
+                Some(random_hash()),
+                Some(DEFAULT_HASH.parse().unwrap()),
+                None
+            ),
+            random_key_pair()
+        ), "payload hash and byte size missing" // TODO
+    )]
```
sandreae, line 338: All of these fail :star:

p2panda-rs/src/entry/tests.rs
```diff
@@ -65,6 +70,104 @@ fn sign_and_encode_roundtrip(#[case] entry: Entry, key_pair: KeyPair) {
...
+#[case::skiplink_and_backlink_should_not_be_the_same(entry_signed_encoded_unvalidated(
+    entry_unvalidated(
+        13,
+        1,
+        Some(DEFAULT_HASH.parse().unwrap()),
+        Some(DEFAULT_HASH.parse().unwrap()),
+        Some(operation(Some(operation_fields(default_fields())), None, None))
+    ),
+    random_key_pair()
+))]
```
sandreae, line 157: Would we expect this case to panic? 

It's an entry which MUST have a backlink and skiplink, and they should be different (but they're the same).

p2panda-rs/src/entry/entry_signed.rs
```diff
@@ -176,10 +192,160 @@ mod tests {
...
+    //     entry_signed_encoded_unvalidated(
+    //         1,
+    //         0,
+    //         None,
+    //         None,
+    //         Some(operation(Some(operation_fields(default_fields())), None, None)),
+    //         key_pair(DEFAULT_PRIVATE_KEY)
+    //     ),
+    //     "Log id must be larger than 0 but was 0"
+    // )]
```
sandreae, line 260: :point_up:

p2panda-rs/src/entry/tests.rs
```diff
@@ -65,6 +69,125 @@ fn sign_and_encode_roundtrip(#[case] entry: Entry, key_pair: KeyPair) {
...
+//     entry_signed_encoded_unvalidated(
+//         1,
+//         0,
+//         None,
+//         None,
+//         Some(operation(Some(operation_fields(default_fields())), None, None)),
+//         key_pair(DEFAULT_PRIVATE_KEY)
+//     ),
+//     "Log id must be larger than 0 but was 0"
+// )]
```
sandreae, line 101: :point_up:

p2panda-rs/src/entry/tests.rs
```diff
@@ -65,6 +69,125 @@ fn sign_and_encode_roundtrip(#[case] entry: Entry, key_pair: KeyPair) {
...
+// #[should_panic(expected = "SOME PANIC MESSAGE")]
+// #[case::skiplink_and_backlink_should_not_be_the_same(entry_signed_encoded_unvalidated(
+//         13,
+//         1,
+//         Some(DEFAULT_HASH.parse().unwrap()),
+//         Some(DEFAULT_HASH.parse().unwrap()),
+//         Some(operation(Some(operation_fields(default_fields())), None, None))
+// ,
+//     key_pair(DEFAULT_PRIVATE_KEY)
+// ))]
```
sandreae, line 173: :point_up:

sandreae: https://github.com/p2panda/p2panda/issues/368

p2panda-rs/src/entry/entry_signed.rs
sandreae: https://github.com/p2panda/p2panda/issues/369

sandreae: This is valid, we want log_id = 0, but we need to update the spec to reflect this.

p2panda-rs/src/entry/tests.rs
```diff
@@ -65,6 +69,111 @@ fn sign_and_encode_roundtrip(#[case] entry: Entry, key_pair: KeyPair) {
     assert!(sign_and_encode(&entry_second, &key_pair).is_ok());
 }
 
+#[rstest]
```
sandreae, line 72: These tests are actually duplicating what we have in `SignedEntry` now that the errors are being caught there. I would say they can be removed, they don't bring anything new.

p2panda-rs/src/entry/entry_signed.rs
```diff
@@ -191,4 +363,41 @@ mod tests {
         let key_value_retrieved = hash_map.get(&entry_first_encoded).unwrap().to_owned();
         assert_eq!(key_value, key_value_retrieved)
     }
+
+    proptest! {
+        #[test]
+        fn non_standard_strings_dont_crash(ref s in "\\PC*") {
+            println!("{s}");
```
adzialocha, line 370: Is this `println` statement required for this test?

sandreae: I just wanted to see the funny strings.....

cafca (changes requested):
Very cool! Had a question about one error message.

p2panda-rs/src/entry/error.rs
```diff
@@ -39,6 +39,10 @@ pub enum EntrySignedError {
     #[error("entry requires skiplink for encoding")]
     SkiplinkMissing,
 
+    /// Backlink and skiplink hashes should be different.
+    #[error("entry requires unique skiplink and backlink hashes")]
```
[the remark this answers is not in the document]
adzialocha: I agree, sounds better!

> Should this behaviour also be added to the handbook page about how we are using bamboo?

I don't see it as a special thing we demand here, I see this as a "friendly" protection from something which will definitely be faulty in terms of the Bamboo spec (there will never be an identical back and skiplink in one entry).

p2panda-rs/src/entry/entry_signed.rs
```diff
@@ -191,4 +362,40 @@ mod tests {
...
+                + log_id.to_string().as_str()
+                + seq_num.to_string().as_str()
+                + skiplink
+                + backlink
+                + payload_size.to_string().as_str()
+                + payload_hash
+                + signature;
+            let result = EntrySigned::new(&encoded_entry);
+
+            assert!(result.is_err())
```
cafca, line 398: I thought of something else with prop testing: If these eventually crash due to some randomly generated value it might be very hard to recreate that value if we don't have a record of it. Maybe all prop tests can include a logging call when they error?

cafca: (in this case - when they don't error?)

cafca: No wait ... in this case when this panics, which we can't predict ... so probably everytime.

Issue #183, opened 2022-01-06
Generate more complex `test-data` 


sandreae:
Currently the test data we generate in `test_utils` and use in `p2panda-js` contains one author, who has one log (document) which has 4 entries (create -> update -> update -> delete). If we want to test more complex situations (multiple documents, with different collections of entries, multiple authors) we need to rethink how the test data is composed and consumed.

adzialocha:
This can be closed now since we left JSON RPC, right @sandreae?

PR #373, opened 2022-06-22
Remove test-data generator

adzialocha:
We moved from JSON RPC to GraphQL which makes generating OpenRPC data obsolete. `p2panda-js` is still running with this, but I assume this will be fixed very soon with https://github.com/p2panda/p2panda/issues/345.

Closing: https://github.com/p2panda/p2panda/issues/183, https://github.com/p2panda/p2panda/issues/374

## 📋 Checklist


sandreae (changes requested):
Yes! Nice to remove all this. 

It would be nice to replace it with something else though, for when we just need some quick test values (I have been using this to generate a collection of entries when I need them). Did you have anything in mind for that? (obviously it can be simpler, and we don't need it in some special format).

adzialocha:
> It would be nice to replace it with something else though, for when we just need some quick test values (I have been using this to generate a collection of entries when I need them). Did you have anything in mind for that? (obviously it can be simpler, and we don't need it in some special format).

I'm not sure if this would be something different though. Currently I'd pull test data through fixtures into our codebase and I think anything related to p2panda-rs development should do this because it couldn't be more ergonomic than that.

We had this tool to work with a rpc mock node in p2panda-js with graphql this will obviously be different but it's too early to say what it is but I hope it will be something more integrated. The dependency of generating test data through our rs crate for js was quite complicated.

That's my two cents for why I don't think we need it anymore. When it comes to some general generator tooling I'd be up for that though it sounds very different from what this utility provided so far?

I could see a separate repository for a cargo toolbelt app or a standalone binary generating data maybe? Though, not a priority for now probably from my perspective at least.

adzialocha:
.. or your playground seems to already be helpful if one just needs to generate data?

PR #377, opened 2022-06-27
Update p2panda-js session to use GraphQL instead of RPC

### LINKED ISSUE #345: Remove RPC api from p2panda-js

cafca:
- `Session.nextEntryArgs` and `Session.publishEntry` now use GraphQL
  - Of course this can only work once we have https://github.com/p2panda/aquadoggo/pull/141
- Updated fixtures use `backlink` instead of `entryHashBacklink` etc.
- Removed materialiser and all references to instances
- Updated the example in the readme

Closes #345

## 📋 Checklist

  the changed `backlink` field is also in https://github.com/p2panda/handbook/pull/179/files


p2panda-js/README.md
```diff
@@ -74,16 +75,39 @@ const session = new Session('https://welle.liebechaos.org').setKeyPair(keyPair);
...
 await session.create(payload, { schema: CHAT_SCHEMA });
 
 // Query instances from the p2panda node
-const instances = await session.query({ schema: CHAT_SCHEMA });
-console.log(instances[0].operation);
-// -> 'Hi there'
+import { gql, useQuery } from '@apollo/client';
+
+const GET_CHAT_MESSAGES = gql`
+  query GetChatMessages {
```
adzialocha, line 88: I think there is a shorthand form where you can even skip the query bracket, looks even simpler then!

p2panda-js/src/entry/entry.ts
```diff
@@ -38,8 +38,8 @@ export const signPublishEntry = async (
   const { entryEncoded, entryHash } = signEncodeEntry(
     keyPair,
     operationEncoded,
-    entryArgs.entryHashSkiplink,
-    entryArgs.entryHashBacklink,
+    entryArgs.skiplink,
```
adzialocha, line 41: Cool, this is also updated in the handbook, but not implemented yet in aquadoggo

p2panda-js/src/session/session.ts
```diff
@@ -26,6 +26,32 @@ export type Context = {
...
+    }
+  }
+`;
+
+// GraphQL mutation to publish an entry and retrieve arguments for encoding the
+// next operation on the same document (those are currently not used to update
+// the next entry arguments cache).
+export const GQL_PUBLISH_ENTRY = gql`
+  {
+    publishEntry(entryEncoded: $entry, operationEncoded: $operation) {
```
adzialocha, line 46: Do we have to mention here that this isa mutation?

sandreae (approved):
Looks good to me!

p2panda-js/src/entry/entry.ts
sandreae: https://github.com/p2panda/aquadoggo/pull/181

adzialocha: In publishEntry it needs to be renamed as well I think 👍

adzialocha: Ah, but this is also not reflected here, only in the handbook so far

p2panda-js/src/session/session.ts
adzialocha: Ah, I see the apollo package solves that.

PR #383, opened 2022-07-03
Refactor mock node using storage provider

sandreae:
- [x] move `SimplestStorageProvider` trait implementations to own module
- [x] move storage type structs and trait implementations to own module
- [x] implement `DocumentStore` on `SimplesetStorageProvider`
- [x] remove not needed `AuthorLog` struct
- [x] use `HashMap` and `HashSet` collections on `SimplestStorageProvider` instead of vecs
- [x] ~~ keep `Node` sync by blocking on async tasks and offering sync wrapper methods ~~
- [x] API tidy up
- [x] bring across nice new storage test helper methods from `aquadoggo`
- [x] ~~fix issues with async and wasm .... :cry:~~
  - [x] exclude `Node` and related tests from wasm target
- [x] info logs for store insertions, publish entry and next entry args
- [x] method and module doc strings
- [x] rename `SimplestStorageProvider`? (`TestStore`, `MemoryStore`?)
- [x] methods to expose stored entries, operations, documents and views
- [x] improve template export pattern
- [x] more sane result type pattern
 
## Next steps
- [ ] method level doc string examples
- [ ] do we want `MemoryStore` to be used outside of tests?

## 📋 Checklist


p2panda-rs/src/test_utils/db/stores/entry.rs
```diff
@@ -0,0 +1,355 @@
...
+    use crate::schema::SchemaId;
+    use crate::storage_provider::traits::test_utils::{test_db, TestStore};
+    use crate::storage_provider::traits::{AsStorageEntry, EntryStore};
+    use crate::test_utils::db::{SimplestStorageProvider, StorageEntry};
+    use crate::test_utils::fixtures::{
+        entry, entry_signed_encoded, key_pair, operation_encoded, random_key_pair, schema,
+    };
+
+    #[rstest]
+    #[async_std::test]
```
adzialocha, line 154: When you embed this in aquadoggo you would end up having two runtimes in the tests (async_std and tokio), which is something you want to avoid.

Could we change this to `tokio-test`? Its an own crate dedicated to testing :+1:

adzialocha:
:+1: for `MemoryStore`, because this is what it is! I even like the idea of having outside of the testing utilities? Its something you maybe want to use in some code?

p2panda-rs/src/test_utils/mocks/utils.rs
```diff
@@ -3,4 +3,4 @@
 //! Utilities for the `mocks` module.
 
 /// A custom `Result` type to be able to dynamically propagate `Error` types.
-pub type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
+pub type Result<T> = std::result::Result<T, Box<dyn std::error::Error + Send + Sync>>;
```
adzialocha, line 6: Isn't `std::error::Error` always `Sync` and `Send`? https://doc.rust-lang.org/src/std/io/error.rs.html#496

p2panda-rs/src/test_utils/db/stores/entry.rs
sandreae: How do you suggest using `tokio-test`? It doesn't use macros in the same way as above. I was thinking to just swap to `tokio` and use `#[tokio::test]` (or whatever it is).

p2panda-rs/src/test_utils/mocks/utils.rs
sandreae: hmm...

I get this error without it:

```
error[E0277]: the size for values of type `dyn StdError + Sync + std::marker::Send` cannot be known at compilation time
   --> src/test_utils/mocks/node.rs:262:15
    |
262 |         .await?
    |               ^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `dyn StdError + Sync + std::marker::Send`
    = help: the following other types implement trait `FromResidual<R>`:
              <std::result::Result<T, F> as FromResidual<Yeet<E>>>
              <std::result::Result<T, F> as FromResidual<std::result::Result<Infallible, E>>>
    = note: required because of the requirements on the impl of `StdError` for `Box<dyn StdError + Sync + std::marker::Send>`
    = note: required because of the requirements on the impl of `From<Box<dyn StdError + Sync + std::marker::Send>>` for `Box<dyn StdError>`
    = note: required because of the requirements on the impl of `FromResidual<std::result::Result<Infallible, Box<dyn StdError + Sync + std::marker::Send>>>` for `std::result::Result<(), Box<dyn StdError>>`
```

p2panda-rs/src/test_utils/db/stores/entry.rs
adzialocha: Aha! Yeah, I just found that in a random example on the internet, haven't looked deeper, but maybe its not providing what we need.

sandreae:
I made it `async` I like it!

p2panda-rs/src/test_utils/mocks/utils.rs
sandreae: Maybe this is resolved now? Unless you are thinking of a larger review of the error handling patterns? I think it's fairly sensible now that this issue with the `Result` type is sorted.

sandreae: Very happy to raise a separate issue if there are still things you'd like to be looked at.

adzialocha (changes requested):
Thats a biggie! Well done! I need some more time to review all of this, but here is a first batch :smile:

sandreae:
Thanks for the comments so far :+1: I totally forgot about switching to `tokio` I will do that before merging for sure.

p2panda-rs/src/document/document.rs
```diff
@@ -358,7 +361,8 @@ mod tests {
     }
 
     #[rstest]
-    fn resolve_documents(schema: SchemaId) {
+    #[async_std::test]
+    async fn resolve_documents(schema: SchemaId) {
```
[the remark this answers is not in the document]
sandreae: There are other places where we use `schema` to mean `schema_id` (like in `Document`), maybe this could be an issue to correct/decide on what we want, in all of these places?

p2panda-rs/src/test_utils/mocks/utils.rs
adzialocha: > Maybe this is resolved now? Unless you are thinking of a larger review of the error handling patterns? I think it's fairly sensible now that this issue with the `Result` type is sorted.

Okay I think I understand now, we don't use `anyhow::Error` in `p2panda_rs` (which requires the inner `E` type to be `Sync`, `Send` and `'static`), but use `thiserror` to implement the `std::error::Error` trait which doesn't. We need them to be thread-safe though, so we introduce our own generic error type. So yeah, I think this is fine as is .. a nice refactoring maybe (in a separate issue) would be to define a lib-level export of our `Error` type which is then used across the crate, but thats a minor change.

p2panda-rs/src/document/document.rs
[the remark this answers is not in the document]
adzialocha: Oh ja, that would be nice!

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ Highlights are marked with a pancake 🥞
 
 ## [Unreleased]
 
+## Added
```
adzialocha, line 12: Needs to be heading level 3 (`###`)

p2panda-rs/Cargo.toml
```diff
@@ -75,6 +79,7 @@ pretty_env_logger = "^0.4.0"
 proptest = { version = "^1.0.0", default-features = false, features = ["std"] }
 rstest = "^0.15.0"
 rstest_reuse = "^0.3.0"
+tokio = { version = "1.19.2", features = ["rt", "macros", "rt-multi-thread"] }
```
adzialocha, line 82: Do we have to mention `tokio` again as a dev dependency? I think if its already a dependency it will also be accessible for the `tests` target or is there something special when it is `optional`?

p2panda-rs/src/test_utils/db/provider.rs
```diff
@@ -0,0 +1,73 @@
...
+use crate::storage_provider::traits::{AsStorageEntry, AsStorageLog};
+use crate::storage_provider::utils::Result;
+use crate::test_utils::db::{
+    EntryArgsRequest, EntryArgsResponse, PublishEntryRequest, PublishEntryResponse, StorageEntry,
+    StorageLog,
+};
+
+type AuthorPlusLogId = String;
+
+/// An in memory implementation of p2panda storage provider traits.
```
adzialocha, line 22: "in-memory" reads better

adzialocha (approved):
Only a few mini-changes left but otherwise it is good to go! Thank you for this large refactoring, `test-utils` becoming even more fancy!

It would be nice to already tackle the log-levels here (as it is new code), but thats up to you!

p2panda-rs/Cargo.toml
sandreae: I think we do as it's optional in `[dependencies]` which means it wouldn't be included for the `tests` target if we didn't include it again here. Maybe we can do it differently though, like how we are doing for `testing`.

p2panda-rs/src/storage_provider/traits/storage_provider.rs
```diff
@@ -59,17 +61,24 @@ pub trait StorageProvider<
...
+    async fn get_document_by_entry(&self, entry_hash: &Hash) -> Result<Option<DocumentId>>;
 
     /// Returns required data (backlink and skiplink entry hashes, last sequence number and the
     /// document's log_id) to encode a new bamboo entry.
     async fn get_entry_args(
         &self,
         params: &Self::EntryArgsRequest,
-    ) -> Result<Self::EntryArgsResponse, Box<dyn std::error::Error + Send + Sync>> {
+    ) -> Result<Self::EntryArgsResponse> {
+        info!(
```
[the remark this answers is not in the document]
sandreae: Thanks, yes, let's take that as it's own subject.

p2panda-rs/Cargo.toml
adzialocha: Okay, that makes sense, let's keep it like this then :+1:

Issue #327, opened 2022-05-19
Consistent string slice return type for `as_str` methods


cafca:
Make sure that `as_str` methods return a string slice and, if the method requires allocating new memory for the return value, name that method `serialise` instead.

# Rationale

The `as_str()` method returns a string slice by convention. However, some of our `as_str` implementations return `String`. Now we could just add `to_string` methods that return `String`, but Rust style (Clippy) wants us to to implement `Display` instead of just having `to_string` methods. In the past we decided to call String-returning methods `serialise` instead (in https://github.com/p2panda/p2panda/pull/269/). Personally, I would rather disregard the Clippy rule and use the `to_string` name but I think between us there was some disagreement about that and we settled on `serialise`.

e.g.

```rust
pub fn serialise(&self) -> String {...}

pub fn as_str(&self) -> &str {...}
```

adzialocha:
That's maybe helpful: https://loige.co/how-to-to-string-in-rust/

adzialocha:
I've stumbled now a couple of times over implementation bugs (could discover them through tests, but otherwise unnoticed), where I accidentally used `to_string` and that led to an invalid hash .. Is there a way to decouple the short string methods even more? I understand they have a nice logging benefit, but I'm afraid it might also be a source of many bugs which are hard to track down.

adzialocha:
> I've stumbled now a couple of times over implementation bugs (could discover them through tests, but otherwise unnoticed), where I accidentally used `to_string` and that led to an invalid hash .. Is there a way to decouple the short string methods even more? I understand they have a nice logging benefit, but I'm afraid it might also be a source of many bugs which are hard to track down.

Meh, but it doesn't really make sense with how Rust wants it to be 😥

This is another resource on that topic: https://github.com/rust-lang/rfcs/blob/master/text/0565-show-string-guidelines.md#user-facing-fmtdisplay

At the end they talk about an "adapter pattern" which looks interesting?

PR #389, opened 2022-07-06
Consistent `to_string` and `as_str` functions, introduce `Human` trait with `display` method

adzialocha:
This brings consistency into `as_str` and `to_string` functions and introduces a new lib-level `Human` trait with a `display` method which can be used for implementing shorter string representations of our types.

**Summary**

* Remove short string representation in `Display` and introduce `Human::display` trait method instead
* Renames `as_str` to `to_string` as we return an owned `String` in `DocumentViewId` and `SchemaId`
* Refactors `serialise` to `Display` implementation (which gives us `to_string`) in `FieldType`
* Implements `Display` for `CddlType`, `DocumentViewHash`, `EntrySigned` and `OperationEncoded`
* Adds more tests around string representation, moves some assertions around as the tests were doing too many different things at once

**Bonus level**

* Adds a test to check conversion from `PublicKey` to `Author`
* Removes derived `Serialize` and `Deserialize` traits from `Schema` and `FieldType` as it seems to be unused and untested? (If we need it back, lets make a ticket and write a test for it as I think it would maybe be broken)

Closing: #327 

## 📋 Checklist


cafca:
Just so I understand correctly:

- `to_string` returns the verbatim value of a data type formatted as `String`
- `as_short_str` returns a human-readable and potentially shortened text representation of the value that is suitable for being included in log outputs etc.
  - the preferred format is `<Type value>` e.g. `<LogId 5>`
  - if the value identifying a type has a short string representation itself, the inner value doesn't use this format. e.g. document id is `ab6e98` and not `<DocumentId ab6e98>` so that document can be `<Document ab6e98>` instead of `<Document <DocumentId ab6e98>>`
- `Display` implementations call `as_short_str` internally against clippy's iron will

?

adzialocha:
> Just so I understand correctly (correct me if i'm wrong):
> 
>     * `to_string` returns the verbatim value of a data type formatted as `String`
> 
>     * `as_short_str` returns a human-readable and potentially shortened text representation of the value that is suitable for being included in log outputs etc.
>       
>       * the preferred format is `<Type value>` e.g. `<LogId 5>`
>       * if the value identifying a type has a short string representation itself, the inner value doesn't use this format. e.g. document id is `ab6e98` and not `<DocumentId ab6e98>` so that document can be `<Document ab6e98>` instead of `<Document <DocumentId ab6e98>>`
> 
>     * `Display` implementations call `as_short_str` internally against clippy's iron will
> 
> 
> Could be something for our code style guidelines.

Exactly that is realised in this PR, right. Though, I would vote for making `Display` actually implement `to_string` again (that would be in Rust's will and also avoid bugs).

cafca:
I think `Display` should always give you the human-readable version, that's what everyone will expect:

> Display is similar to [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html), but Display is for user-facing output, and so cannot be derived.

If I understand you correctly you would be surprised when `format!("{}", value)` gives you something like `<Value 7829038>` instead of the full value as a string? Because for me it's the exact opposite.

adzialocha:
> I think `Display` should always give you the human-readable version, that's what everyone will expect:
> 
> > Display is similar to [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html), but Display is for user-facing output, and so cannot be derived.
> 
> If I understand you correctly you would be surprised when `format!("{}", value)` gives you something like `<Value 7829038>` instead of the full value as a string? Because for me it's the exact opposite.

I think the question is what we mean with "Human Readable" output .. in Rust its the difference between for example:

`{:?}` prints .. `"Some("Lalala")"` (including the quotation marks around even)`
`{}` vs. `Lalala` (without quotation marks)

I think this is what the Rust specification means by "user facing" .. it doesn't print the underlying structs and doesn't wrap strings in quotation marks etc. but this does not mean it needs to be shortened or changed in its value.

cafca:
What I understand is:

- For you it's important to be able to call `to_string()` and get the full string-serialised value without any modifications etc. You would rather not have to remember any other function name than `to_string()` for this.
- For me it's important to be able to quickly write `debug!("{}", value)` and get a friendly, human-readable output in the console that doesn't span multiple lines and is easy to compare with other values. I would rather not have to use any more verbose syntax than this.

Is there a way we can have both?

adzialocha:
This might also help: https://doc.rust-lang.org/std/fmt/index.html#fmtdisplay-vs-fmtdebug

cafca:
I now also ran into this `to_string()` trap. I just didn't understand what your issue with this was before lol. Let me have another look tomorrow or so.

adzialocha:
I've removed the short string representation for `Display` now (and the clippy rule, yay) so we get the same value with `to_string` and `as_str`.

For the short string representation all structs have a method named `short_repr` now. Happy to hear any suggestions for better names, but I thought thats better as `as_short_str` is not correct anymore (it returns a `String`).

cafca:
`short_repr` doesn't exactly roll of the tongue :D I would prefer just `display()` as the method name.

adzialocha:
> `short_repr` doesn't exactly roll of the tongue :D I would prefer just `display()` as the method name.

Hmn .. `display` is very abstract, "what" does it display?

cafca:
Think about a computer display that displays things. You display the value, then you can see it!

adzialocha:
> Think about a computer display that displays things. You display the value, then you can see it!

I'm not convinced yet :smiley_cat: .. its like naming a method "show" etc., I'd say let's explore some more alternatives? I think `short_repr -> String` does exactly what it promises (it returns a short representation of a thing as a string), but I'm happy to change it.

cafca:
I asked this question on StackOverflow and liked this idea from one of the replies:

> > How should I structure my code and name my methods if I need to implement both of those?

> For convenient use in format strings, you can write one or more methods which return wrapper types that implements Display (like [Path::display()](https://doc.rust-lang.org/std/path/struct.Path.html#method.display) in std).

https://stackoverflow.com/questions/72984552/what-should-an-idiomatic-display-implementation-write

adzialocha:
Ah, thats interesting! It looks like though that for every struct we would need a `Display` type then? Or as I understand, this is how they do it at least for `Path`?

adzialocha:
Let's call it `ShortRepr`? :rofl:

adzialocha:
I did it and I like it very much!

Named the trait `Human` with a `display` method inside. It needed to be another name than `Display`, otherwise it would collide with our current `std::fmt::Display` implementations or you would have needed to explicitly specify which `Display` you mean, but that doesn't remove that it looks confusing.

But I like the name `Human` as well, also it opens the door for more human-facing methods on that trait, who knows!

In any case, this is _finally_ ready for review now, please have a look!

p2panda-rs/src/lib.rs
```diff
@@ -92,6 +92,16 @@ pub trait Validate {
     fn validate(&self) -> Result<(), Self::Error>;
 }
 
+/// Trait used by p2panda structs for human-facing functionality, like better readability.
```
cafca, line 95: I like the trait name :D We could use the trait's docstring to differentiate it more explicitly from `Display` and `Debug`. `Display` in particular also claims to be made for human-facing display but we have found out that it's not very well suited for that.

sandreae: I really like the name, and also in general how we're using a trait here to better explain and contain a functionality. Along with understanding how we use `Validate` better I think this may be opening the door to some nice p2panda trait inventions ;-p

sandreae (approved):
Great work @adzialocha, it was quite a knot to untie but we definitely landed at a very elegant solution, so thanks for leading it there.

PR #402, opened 2022-07-08
Add more tests for `publish_entry` and `get_next_entry_args`

sandreae:
- [x] test for update of a different authors document (multi-writer update)
- [x] test for passing non-existent document id

This just adds some more tests for an area where there seems to be quite a few edge cases/tricky behaviour. We don't think `get_entry_args` is performing as we want when being passed non-existent `DocumentId`s atm. 

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/storage_provider/traits/storage_provider.rs
+++ b/p2panda-rs/src/storage_provider/traits/storage_provider.rs
@@ -267,19 +267,22 @@ pub mod tests {
     use rstest::rstest;
 
     use crate::document::DocumentId;
-    use crate::entry::{sign_and_encode, Entry, LogId};
-    use crate::identity::KeyPair;
+    use crate::entry::{sign_and_encode, Entry, LogId, SeqNum};
+    use crate::identity::{Author, KeyPair};
     use crate::operation::{
-        AsOperation, OperationEncoded, OperationFields, OperationId, OperationValue,
+        AsOperation, Operation, OperationEncoded, OperationFields, OperationId, OperationValue,
     };
     use crate::storage_provider::traits::test_utils::{test_db, TestStore};
     use crate::storage_provider::traits::{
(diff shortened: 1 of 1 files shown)


adzialocha:
I think we should add a changelog entry for anything which can't be expressed as a single commit directly into `main` 😅

p2panda-rs/src/storage_provider/traits/storage_provider.rs
```diff
@@ -541,6 +536,114 @@ pub mod tests {
...
+        let expected_entry_args_response =
+            EntryArgsResponse::new(None, None, SeqNum::new(1).unwrap(), LogId::new(1));
+
+        // Correct request returns the expected next entry args.
+        assert_eq!(entry_args_response, expected_entry_args_response);
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn get_entry_args_non_existand_document_id(
```
adzialocha, line 620: "existand" typo

p2panda-rs/src/storage_provider/traits/storage_provider.rs
```diff
@@ -541,6 +536,114 @@ pub mod tests {
...
+        let entry_args_request = EntryArgsRequest {
+            public_key: Author::try_from(key_pair.public_key().to_owned()).unwrap(),
+            document_id: Some(document_id),
+        };
+
+        let entry_args_response = db.store.get_entry_args(&entry_args_request).await.unwrap();
+
+        // The response contains args for a new log at sequence number 1.
+        //
+        // @TODO: This is the current behaviour but it may not be what we want.
```
adzialocha, line 640: Could be nice to clarify what is the expected behavior. Is there a ticket?

adzialocha (changes requested):
Small things plus the changelog comment above ✌️

PR #408, opened 2022-07-19
Remove generic parameters from StorageProvider

sandreae:
Remove generic parameters and replace with _associated types_. This makes the implementation of StorageProvider much cleaner and means we can work with the trait as generic parameter more easily. It is also a more correct way to define the types used inside the different stores, they are not actually generic, they are defined once per implementation (this is what associated types are used for). Later we can move the types inside their respective stores for even more neat clarity.

Follow-up steps for later on:
- deprecate `publish_entry` and `next_entry_args`
- deprecate req-res types only needed in these mthods

needed for: https://github.com/p2panda/aquadoggo/pull/204 and includes breaking changes so should be coordinated.

## 📋 Checklist


adzialocha (approved):
This is really much better :+1:

sandreae:
> This is really much better +1

Haha, I know right, I wonder what I was thinking before.... On the long hard rusty journey ;-p

PR #414, opened 2022-07-27
Update `Human` for `SchemaId` and `DocumentViewId`

cafca:
When we want to include `SchemaId` or `DocumentViewId` in log lines etc. they will represent the respective schema or document view themselves and not just the ID. The current implementation of `Human` for these traits just returns their shortened ids, which leads to log lines like 

```
Dispatch dependency task for view with id: 30e31e
```

when instead we would rather like to have

```
Dispatch dependency task for <DocumentView 30e31e>
```

This PR changes the `Human` impl on `SchemaId` and `DocumentViewId` to represent these items like the respective `Schema` and `DocumentView`.

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/document/document_view_id.rs
+++ b/p2panda-rs/src/document/document_view_id.rs
@@ -59,6 +59,22 @@ impl DocumentViewId {
         graph_tips.sort();
         graph_tips
     }
+
+    /// Can be used as a human-readable representation of a document view id.
+    ///
+    /// The return value contains all view graph tips' last 6 characters
+    /// concatenated with an underscore.
+    pub fn to_short_string(&self) -> String {
+        let mut result = String::new();
+
+        let offset = yasmf_hash::MAX_YAMF_HASH_SIZE * 2 - 6;
+
+        for (i, operation_id) in self.0.clone().into_iter().enumerate() {
+            let separator = if i == 0 { "" } else { "_" };
+            write!(result, "{}{}", &separator, &operation_id.as_str()[offset..]).unwrap();
+        }
+        result
+    }
 }
 
 impl Display for DocumentViewId {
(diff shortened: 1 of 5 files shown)


adzialocha:
Ah, good find! Thank you! :+1: 

Could you rebase this PR to make the commit history clean?

cafca:
Oof, is that necessary? It will be squashed anyway and I never liked using
rebase because of the ux. I think it’s a headache :(

Andreas Dzialocha ***@***.***> schrieb am Mi. 27. Juli 2022
um 21:16:

adzialocha:
> Oof, is that necessary? It will be squashed anyway and I never liked using
rebase because of the ux. I think it’s a headache :(

Hmn, its not thaaaaat hard, let me do it :+1:

p2panda-rs/src/document/document_view_id.rs
```diff
@@ -59,6 +59,22 @@ impl DocumentViewId {
         graph_tips.sort();
         graph_tips
     }
+
+    /// Can be used as a human-readable representation of a document view id.
+    ///
+    /// The return value contains all view graph tips' last 6 characters
+    /// concatenated with an underscore.
+    pub fn to_short_string(&self) -> String {
```
adzialocha, line 67: Not sure if it will ever be used outside of `Human`, should we make this only visible on a crate level?

cafca:
I just like not changing commits once they are committed. I also almost never force push. I think in general it's preferable to use tools like squash merging instead of rewriting history.

p2panda-rs/src/document/document_view_id.rs
cafca: Yes, good catch!

cafca:
This is a change in intended behaviour and not really a fix I think.

adzialocha:
> This is a change in intended behaviour and not really a fix I think. We haven't said that `Human` always has to follow the `<Type id>` schema (although I think we should).

Okay, I've changed it as the PR and branch was named "fix" before and it felt like something which broke after my PR before.

Issue #225, opened 2022-02-18
Test invalid operation and entry data


adzialocha:
With https://cbor.me/ it is easy to generate weird or invalid operations, like:

```
A366616374696F6E666372656174656776657273696F6E63626C61666669656C6473A16568616C6C6FA26576616C75650C647479706563737472
```

```json
{"action": "create", "version": "bla", "fields": {"hallo": {"value": 12, "type": "str"}}}
```

We should have some invalid operations and entries in our test fixtures and make sure we account for them in the code.

adzialocha:
For entries this was completed now via https://github.com/p2panda/p2panda/pull/367

Issue #406, opened 2022-07-18
Depricate `publish_entry`, `next_entry_args` and associated req/res structs in `StorageProvider`


sandreae:
Once this PR is complete: https://github.com/p2panda/aquadoggo/pull/204

Issue #405, opened 2022-07-13
Align names of field types with names of operation value types


cafca:
Now we have for example `FieldType::Bool` and `OperationValue::Boolean`. It would be nice if `FieldType` uses the same names as `OperationValue`.

Issue #399, opened 2022-07-07
Validate ordering of keys and values in CBOR


adzialocha:
We define in our handbook that all arrays and keys _need_ to be ordered in operations aka CBOR payload. Do we validate that?

Keys could be handled through following the canonical specification of CBOR though I'm not sure if any Rust crate supports that 😭

Arrays need to be checked by us I guess.

PR #415, opened 2022-07-27
Untagged operation format, schema validation, new operation and entry API

adzialocha:
This PR changes the operation encoding format to untagged values since we have schemas now! Also it introduces schema validation of operations, refactors the validation flows and changes the whole API around operations and entries.

> :warning: It was not possible for me to tackle this PR without doing larger refactorings all around the place (test utilities, entries, validation flows, storage provider), literally everything touches upon it somewhere somehow, this is why this PR is fairly large.

## Where to start?

The `entry` and `operation` module-level documentation and diagrams should help with some introduction. They contain the most changed bits, `schema::validate` as well. The rest is mostly necessary refactoring for breaking APIs.

## Todo

- [x] Introduce a `next` module which will be a temporary place for future things 
- [x] New `entry`, `schema` and `operation` API
- [x] Untagged operation format
- [x] Application schema validation
- [x] `Validation` trait clean ups (evaluate where invalid data might come in)
- [x] Fix issues with breaking API changes between `next` module and test utilities
- [x] Finish implementing entry validation methods
- [x] Look at removed tests (including `cddl`) in `next` and check if we can still use them
- [x] Write documentation for new API
- [x] System schema validation
- [x] Loads of new tests for `next`
- [x] Fix js and wasm tests
- [x] Re-integrate `next` into the library (large refactoring), this is left:
  - [x] `wasm`
  - [x] `storage_provider`
  - [x] `test_utils`
    - [x] `constants`
    - [x] `fixtures`
    - [x] `templates`
    - [x]  `db`
    - [x] `mocks`
    - [x] `utils`
- [ ] Re-implement `send_to_store` and `populate_test_db` methods in store test utils
- [ ] Bring back tests which depend on the above
- [ ] update diagrams

### Encoding Entries

```
             ┌────────────┐                         ┌─────┐
  bytes ───► │EncodedEntry│ ────decode_entry()────► │Entry│
             └────────────┘                         └─────┘
 ┌───────┐                                             ▲
 │KeyPair│ ──────────┐                                 │
 └───────┘           │                                 │
                     │                                 │
 ┌────────────┐      ▼                                 │
 │EntryBuilder├────sign()──────────────────────────────┘
 └────────────┘
```

### Decoding Entries

```
 ┌─────┐                     ┌────────────┐
 │Entry│ ──encode_entry()──► │EncodedEntry│ ─────► bytes
 └─────┘                     └────────────┘
```

### Encoding Operations

```
             ┌────────────────┐
             │OperationBuilder├──────────build()──────────────┐
             └────────────────┘            ▲                  │
                                           │                  │
                                           │                  │
                                           │                  ▼
                                        ┌──┴───┐          ┌─────────┐
                                        │Schema│          │Operation│
                                        └──┬───┘          └─────────┘
                                           │                  ▲
                          Lookup Schema    │                  │
                                           │                  │
             ┌──────────────┐              ▼                  │
             │PlainOperation├───────validate_operation()──────┘
             └──────────────┘
                    ▲
                    │
            decode_operation()
                    │
            ┌───────┴────────┐
bytes ────► │EncodedOperation│
            └────────────────┘
```

### Decoding Operations

```
 ┌─────────┐                           ┌────────────────┐
 │Operation│ ───encode_operation()───► │EncodedOperation│ ────► bytes
 └─────────┘                           └────────────────┘
```

### Complete operation and entry validation

The flow for someone using this will be: Receive `EncodedEntry` and `EncodedOperation` via GraphQL API (`publish` request), decode them both, get decoded `PlainOperation` to already read the `SchemaId`, make a look-up request in schema provider to get the `Schema` instance. Get decoded `Entry` to look up the regarding entries in database. Feed both links, schema, entry and operation into `validate_operation_and_entry` to finally verify everything.

```
                                                                Look-Up
                                                                               
           ┌────────────┐                       ┌─────┐    ┌─────┐    ┌─────┐
bytes ───► │EncodedEntry├────decode_entry()────►│Entry│    │Entry│    │Entry│
           └──────┬─────┘                       └──┬──┘    └─────┘    └─────┘
                  │                                │
                  └───────────────────────────┐    │       Skiplink   Backlink
                                              │    │          │          │
           ┌────────────────┐                 │    │          │          │
bytes ───► │EncodedOperation├─────────────┐   │    │          │          │
           └───────┬────────┘             │   │    │          │          │
                   │                      │   │    │          │          │
           decode_operation()             │   │    │          │          │
                   │            Look-Up   │   │    │          │          │
                   ▼                      │   │    │          │          │
            ┌──────────────┐    ┌──────┐  │   │    │          │          │
            │PlainOperation│    │Schema│  │   │    │          │          │
            └──────┬───────┘    └──┬───┘  │   │    │          │          │
                   │               │      │   │    │          │          │
                   │               │      │   │    │          │          │
                   │               │      │   │    │          │          │
                   │               │      │   │    │          │          │
                   │               ▼      ▼   ▼    ▼          ▼          │
                   └───────────►  validate_operation_and_entry() ◄───────┘
                                               │
                                               │
                                               │
                                               │
                                               ▼
                                       ┌─────────────────┐
                                       │VerifiedOperation│
                                       └─────────────────┘
```

### Validation and the `Validate` trait

I've removed all `Validate` implementations on more complex structs and shifted that into the regarding `validate_` methods which are used by different higher-level methods (like `encode`, `decode` or the builders).

`Validate` still makes sense for a couple of structs where we need to check against a) format and b) canonic format, like `Hash` and therefore `OperationId` (just a hash), `DocumentId` (just a hash). It gets more interesting in `DocumentViewId`, its the only place where we require a special canonic representation (no duplicates, sorted operation ids).

All `Deserialize` implementations use `Validate` now internally, which means that probably developers will not need to use it anymore themselves as everything gets also checked using the public programmatic API. Similar to all other structs there is simply no way to enter invalid data from the outside anymore, expect within the crate or in tests.

Most structs have _not_ been checked against their format during deserialization, this is fixed now by that change.

### Minor changes

General simplifications of conversion traits where we return a `Result` for no reason (it would panic on a more serious io issue). Most of the time these conversions have already been unwrapped on higher-level methods.

* `impl<T: core::borrow::Borrow<[u8]> + Clone> TryFrom<YasmfHash<T>> for Hash` ↦ `impl<T: core::borrow::Borrow<[u8]> + Clone> From<&YasmfHash<T>> for Hash`
* `impl TryFrom<PublicKey> for Author` ↦ `impl From<&PublicKey> for Author`

Consistency around using borrowed types (similar to changes above):

* `impl From<Hash> for YasmfHash<Blake3ArrayVec>` ↦ `impl From<&Hash> for YasmfHash<Blake3ArrayVec>`

Similar reason, as above, but a method where it is nicer to take a slice and ignore io errors:

* `Hash::new_from_bytes(value: Vec<u8>) -> Result<Self, HashError>` ↦ `Hash::new_from_bytes(value: &[u8]) -> Self`

New trait implementations:

* `impl From<&Author> for PublicKey`

Renamed methods:

* `OperationFields::add` ↦ `OperationFields::insert` (to match the BTreeMap naming)

## Tasks

**New API**

* Renames `EntrySigned` to `EncodedEntry` and `OperationEncoded` to `EncodedOperation`
* Moves most funcionality of `EntrySigned` into `Entry`, so we finally have everything in one place
* Makes `Operation` and `Entry` immutable and removes serialization from them, nothing can enter it without strict validation
* Introduces `OperationBuilder` and `EntryBuilder` to have a tool to build new `Entry` and `Operation` instances programmatically https://github.com/p2panda/p2panda/issues/131
* `VerifiedOperation` holds `entry`, `operation` and `operation_id` now
* `Operation` holds `schema` now
* Introduces `PlainOperation`, its the intermediary type between `EncodedOperation` and `Operation` and allows a) the developer to access the `schema_id` to retreive a `Schema` instance, which can then be used to finally get the `Operation` b) allows schema-less deserialization from raw bytes
* Introduces `Schematic` and `Actionable` traits which are important for schema- & operation validation. They allow us to validate the schemas of both `PlainOperation` and `Operation`. This might be the first steps towards https://github.com/p2panda/p2panda/issues/320 (making p2panda-rs agnostic of the encoding formats)
* Made `error`, `validate`, `traits`, `decode` and `encode` modules public as the modules `operation`, `entry` and `schema` get fairly complex now, I'd be up for doing this consistent later on for all other modules as well

**New encoding format**

* Remove tagged operation fields #291 :sparkles: 
* Use array instead of map for operation header format (this is still a pending decisions based on https://github.com/p2panda/handbook/issues/200) :warning:

**Improved validation flows**

* Many new validation methods in `entry` and `operation`, most of them will not be used for high-level development, but we can still export them in case someone needs it. The main method which validates _everything_ according to our [specification](https://whimsical.com/4X53TdGvPLFUtmuBFvriW5) is `operation::validate::validate_operation_with_entry` :+1: I imagine this will be used in aquadoggo
* Started using some sort of code system for validation steps in the documentation (`#E1`, `#E2`, etc.), this can be used in the handbook as well
* Stricter / "more canonic" CBOR validation #399 
* Checking against duplicate field names and duplicate hashes in relation lists, document view ids, etc. https://github.com/p2panda/p2panda/issues/395
* You can only create a `VerifiedOperation` instance when _everything_ has been validated #366 
* This should also finally close https://github.com/p2panda/p2panda/issues/225
* Refactored all `Validate` trait implementations and made sure that everything gets checked both going through programmatic API but also deserialization #276

**Schema validation**

* Validate operations against `Schema` with nice error messages :sparkles: 
* Validate also special formatting requirements for system schemas as per specification
* Remove CDDL validation and do it manually instead (we get nicer error messages, it makes validation more explicit, typed and easier to test). This removes a bunch of CDDL-related tickets:
  * #396 
  * #385 

**Extras**

* Rename `schema` fixture to `schema_id`
* https://github.com/p2panda/p2panda/issues/394
* #397 
* #405 

## Follow ups

* Probably the benchmarks will break now, they need refactoring as well https://github.com/p2panda/p2panda/issues/420
* A couple of nice doc-string examples might have been deleted on the way, we should bring them back https://github.com/p2panda/p2panda/issues/424
* Even nicer validation error messages?
* Removed prop tests, we should bring them back https://github.com/p2panda/p2panda/issues/421
* `p2panda-js` changes
    *  Test data is not really correct (but also not really used either ..), we probably need a new testing approach here if we want to use fixtures https://github.com/p2panda/p2panda/issues/422
    * WebAssembly and JavaScript uses plain operations now, we have to bring operations and schemas in there as well https://github.com/p2panda/p2panda/issues/423
    * Relation support? https://github.com/p2panda/p2panda/issues/244
* bring `validation` and `domain` modules across from `aquadoggo` and use them to replace deprecated store methods for publishing entries and getting next args needed in `test_utils` and the mock `Node`. 
  * https://github.com/p2panda/p2panda/issues/418
  * https://github.com/p2panda/p2panda/issues/406 
* after the above, we can bring back tests in `document` module and the store which were commented out 
* Validate schema specifications when going through programmatic API:
  * https://github.com/p2panda/p2panda/issues/426
  * https://github.com/p2panda/p2panda/issues/425

## 📋 Checklist

    - https://github.com/p2panda/handbook/pull/208


p2panda-rs/src/next/entry/validate.rs
```diff
@@ -0,0 +1,301 @@
...
+use crate::next::entry::error::ValidateEntryError;
+use crate::next::entry::{EncodedEntry, Entry, Signature};
+use crate::next::hash::Hash;
+use crate::next::identity::{Author, KeyPair};
+use crate::next::operation::EncodedOperation;
+
+/// Checks if backlink- and skiplink are correctly set for the given sequence number (#E3).
+///
+/// First entries do not contain any links. Every other entry has to contain a back- and skiplink
+/// unless they are equal, in which case the skiplink can be omitted.
```
cafca, line 16: for this implementation:

> in which case the skiplink _must_ be omitted.

p2panda-rs/src/next/operation/verified_operation.rs
```diff
@@ -0,0 +1,144 @@
...
+#[derive(Debug, Clone)]
+pub struct VerifiedOperation {
+    /// Identifier of the operation.
+    pub(crate) operation_id: OperationId,
+
+    /// Operation, which is the payload of the entry.
+    pub(crate) operation: Operation,
+
+    /// Entry which was used to publish this operation.
+    pub(crate) entry: Entry,
```
sandreae, line 25: I understand that a `VerifiedOperation` is now _actually_ verified ;-p but could you expand on the reason for having the whole entry accessible here now as well.

adzialocha: Good question! No clear reason, other than we have the object, .. I think we can totally change this if it doesn't make sense

p2panda-rs/src/next/operation/operation.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    /// Optional document view id containing the operation ids directly preceding this one in the
+    /// document.
+    pub(crate) previous_operations: Option<DocumentViewId>,
+
+    /// Optional fields map holding the operation data.
+    pub(crate) fields: Option<OperationFields>,
+}
+
+impl AsOperation for Operation {
```
sandreae, line 126: Did you have a "next step" in mind here? I see that the implementations of `Schematic` and `Actionable` duplicate the methods implemented in `AsOperation`. Should we favour the new traits and remove the old one?

sandreae: Hmm... I see the `fields()` in `Actionable` return `PlainFields` maybe this is important...

adzialocha: They are close but still for different purposes, `AsOperation` was written with storage in mind. I think it's fine to leave it as is even when its partly redundant, otherwise we might end up with too strong ties in the next refactoring when suddenly storage needs new things, no shame in ignoring DRY 😅✌️

At least not after we learned a bit about the API?

p2panda-rs/src/next/operation/validate.rs
```diff
@@ -0,0 +1,338 @@
...
+
+/// Checks the fields of an operation-like data type against a schema.
+///
+/// This method checks against:
+///
+/// 1. Correct operation format (#OP2)
+/// 2. Correct canonic operation field values, like document view ids of pinned relations (no
+///    duplicates, sorted) (#OP3)
+/// 3. Schema matches the given operation fields (#OP4)
+pub fn validate_operation<O: Actionable + Schematic>(
```
sandreae, line 128: Is this the only place these traits are used?

p2panda-rs/src/next/operation/operation.rs
sandreae: Yeh, I like how traits can explain the properties of a struct, want to understand the implications of splitting it up like this though. Not just a question of DRYness :desert: ;-p

sandreae: > AsOperation was written with storage in mind

This isn't true actually, `AsOperation` wasn't about storage, it was a way to abstract out the properties of an `Operation`, these are present on `Operation` & `VerifiedOperation`.

p2panda-rs/src/next/operation/validate.rs
adzialocha: Yes!

p2panda-rs/src/next/operation/operation.rs
adzialocha: Ah I see! Yeag, if you have a suggestion to bring stuff together please go ahead 👍

adzialocha: If you're interested in the design / background thinking, this might help: https://github.com/p2panda/handbook/issues/187

sandreae:
I did some work on this PR:

- removed `publish_entry` and `next_entry_args` on `StorageProvider`
- removed this functionality from the mock `Node`
- commented out all tests which requires this functionality in the `TestStore` (to be reinstated later)
- removed `operation` from `StorageEntry` + related refactoring
- moved `storage_provider` into `next/`
- moved `db` and `mocks` into `next/test_utils`

sandreae:
Any dependencies on what was previously the high level storage methods `publish_entry` and `next_entry_args` has been removed as this functionality is being refactored into `validate` and `domain` modules (currently in `aquadoggo`).

I suggest we deal with bringing in those modules and reinstating the needed functionality in a follow-up PR. Open to other approaches though :+1:

p2panda-rs/src/next/operation/operation.rs
sandreae: Thanks, that's good background.

I see the need for the traits in the new validation patterns, but the existing `AsOperation` trait is performing the same function (abstracting the same methods). All I'd suggest is removing one or the other. So `PlainOperation`, `Operation` and `VerifiedOperation` all implement the same trait(s) related to being an operation. 

The more specific naming `Actionable` and `Schematic` are cool how they define the specific functionality they provide. My instinct is that it's a little too specific though, as these properties are an inherent part of operations themselves, we can't have an un-actionable or un-schematic operation (i don't think...?), so I'm not sure the benefit of splitting them up is (although I actually like it in some way, but maybe that's just cos I'm so deep in it though...).

adzialocha: > The more specific naming Actionable and Schematic are cool how they define the specific functionality they provide. My instinct is that it's a little too specific though, as these properties are an inherent part of operations themselves, we can't have an un-actionable or un-schematic operation (i don't think...?), so I'm not sure the benefit of splitting them up is (although I actually like it in some way, but maybe that's just cos I'm so deep in it though...).

Currently that's (kinda) true and one could see the new traits as something too far in the future? I don't really mind going in either direction, the difference is maybe that the new traits open more doors, but again, I don't know if their design really reflects our protocol yet (we didn't reflect upon what operations are yet):

* DELETE operations are currently only half "schematic", they contain a schema id but nothing to validate against (fields), I think there is a hole in our current thinking
* We probably want non-operations to implement some sub features of operations, like `DocumentView` to implement `Schematic`, as we need to validate materialized views against schemas after decrypting their fields
* Not sure about some future system schemas, they might not need a schema id at all?

You can see, all of these are a little too far in the future, I'm happy to look into this later or keep both worlds as it us in this PR right now to explore that direction more.

adzialocha: Designing traits is special I realize and learning right now: you _can_ reflect more abstract (like protocol-level) designs with it but also want to offer some pragmatic API guarantees (like structs which have the same fields, even though they are different).

adzialocha: The `AsOperation` concept can maybe be described as: we have `Operation`, `VerifiedOperation` and `PlainOperation` and all have an action and version field and theoretically can be validated against a schema etc.

adzialocha: The "new" direction somewhat hints on having distinct structs for each action, ala `DeleteOperation` etc. instead of a single Operation struct which accommodates everything. The more different new features (Key group, secret group, blobs, ...) we introduce the clearer we might need to separate them with types and traits. That's future thinking again but it was tempting to go there during implementation of this PR.

adzialocha:
How should we deal with dead code / tests in this? Usually it's something you would simply just remove / clean up, it feels weird to keep it. On the other hand we might loose track of what there was before after merging?

Just commenting things out also doesn't help really, think we need searchable TODO markers?

p2panda-rs/src/next/test_utils/db/types/entry.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub entry_bytes: EncodedEntry,
+
+    /// Hash of Bamboo entry data.
+    pub entry_hash: Hash,
+
+    /// Used log for this entry.
+    pub log_id: LogId,
+
+    /// Hash of payload data.
+    pub payload_hash: OperationId,
```
adzialocha, line 27: Think the type is wrong here, this should be `Hash` instead of `OperationId` (the payload hash is not the operation id, its the hash of the entry) :+1:

adzialocha:
> I suggest we deal with bringing in those modules and reinstating the needed functionality in a follow-up PR. Open to other approaches though +1

Sounds good to me, can you add these concrete steps in the "Follow up" section of this PR description? Thank you!

sandreae:
> How should we deal with dead code / tests in this? Usually it's something you would simply just remove / clean up, it feels weird to keep it. On the other hand we might loose track of what there was after merging?
> 
> Just commenting things out also doesn't help really, maybe we use searchable TODO markers and delete all dead code?

Yeh, TODOs sounds good to me.

I figured we'd also merge into a `development` branch until everything was in place. Do you agree?

adzialocha:
> I figured we'd also merge into a development branch until everything was in place. Do you agree?

Yeah, definitely!

> Yeh, TODOs sounds good to me.

Cool, you want to add yours?

sandreae:
> Cool, you want to add yours and remove the dead code bits?

Sure thing

adzialocha:
I'm refactoring the wasm module currently, it will use plain operations for now to not open the big door of bringing schemas into the js world (yet)

sandreae:
> > Cool, you want to add yours and remove the dead code bits?
> 
> Sure thing

Hmm, I can add `TODO`s wherever commented out tests need to be refactored and reinstated in the next PR, but I don't _really_ wanna delete them entirely, is that what you meant? Seems unnecessary as they need to come back v soon with only slight tweaks to make them work again.

adzialocha:
> > > Cool, you want to add yours and remove the dead code bits?
> > 
> > Sure thing
> 
> Hmm, I can add `TODO`s wherever commented out tests need to be refactored and reinstated in the next PR, but I don't _really_ wanna delete them entirely, is that what you meant? Seems unnecessary as they need to come back v soon with only slight tweaks to make them work again.

I would delete them but only to not have too much dead code flying around (there will be many follow ups and don't know if writing examples will be the first thing to do), but I'm also fine if you think it's good to keep them.

p2panda-rs/src/next/operation/operation.rs
adzialocha: Your todo commit decided for AsOperation I can see 😅

sandreae:
Poop, I pushed some dirty changes I forgot were there from me playing around with the operation traits :grimacing: I need to revert the last commit but don't wanna right now in case you're working in something in parallel and I mess you up @adzialocha

adzialocha:
> Poop, I pushed some dirty changes I forgot were there from me playing around with the operation traits :grimacing: I need to revert the last commit but don't wanna right now in case you're working in something in parallel and I mess you up @adzialocha  

That's fine, you can just commit the change, no need to force push ✌️

p2panda-rs/src/document/document_view_id.rs
```diff
@@ -37,147 +39,113 @@ use super::error::DocumentViewIdError;
...
+    ///
+    /// The given operation ids will automatically be sorted and de-duplicated.
+    pub fn new(graph_tips: &[OperationId]) -> Self {
+        let mut graph_tips = graph_tips.to_owned();
 
-    /// Get sorted graph tips for this view id.
-    pub fn sorted(&self) -> Vec<OperationId> {
-        let mut graph_tips = self.0.clone();
+        // Make sure all operations are sorted
         graph_tips.sort();
```
cafca, line 53: I wanted to highlight this change in the behaviour of the `DocumentViewId` constructor. I seem to remember that @sandreae made sure that we did not sort operation ids in document view ids because their order represented information about the graph structure. Do I misremember this? Is that actually not the case?

adzialocha: You're looking at a DocumentViewId where indeed the operation ids need to be sorted and de-duplicated. The semantic meaning of arrays you are referring to can be found in relation lists and pinned relation lists :+1:

cafca: No, I mean `DocumentViewId`!

adzialocha: > No, I mean `DocumentViewId`!

Hmn, I do not remember any meaning in their ordering. Don't think there is anything but surely @sandreae knows more.

p2panda-rs/src/operation/operation_value.rs
```diff
@@ -1,69 +1,44 @@
...
     Integer(i64),
 
     /// Floating point value.
-    #[serde(rename = "float")]
     Float(f64),
 
     /// String value.
-    #[serde(rename = "str")]
-    Text(String),
+    String(String),
```
cafca, line 19: Hm, I always thought it was so friendly that we call this "text" and not "string" and was hoping we could use "text" everywhere, also in schema field definitions. Do you think this is better?

p2panda-rs/src/operation/encode.rs
```diff
@@ -0,0 +1,41 @@
...
+//!
+//! Encoding an operation does not require a schema, the `Operation` will be serialized into bytes,
+//! represented as a `EncodedOperation` which is then ready to be sent to a p2panda node (alongside
+//! an entry).
+//!
+//! ```text
+//! ┌─────────┐                           ┌────────────────┐
+//! │Operation│ ───encode_operation()───► │EncodedOperation│ ────► bytes
+//! └─────────┘                           └────────────────┘
+//! ```
```
cafca, line 13: Gotta love these diagrams :)

p2panda-rs/src/document/document_view_id.rs
sandreae: Hmm, I can't remember there being meaning to their ordering either 😅

p2panda-rs/src/operation/operation_value.rs
adzialocha: Ah, I understood that we like `String` more, this is why I changed it, it also reflects the `str` type definition maybe :-D But feel free to bring it back to `Text`, I don't have a preference :+1:

p2panda-rs/src/document/document_view_id.rs
adzialocha: TELL US WHAT YOU SEE SAM!

cafca: It's hard to piece this together from just pr comments buuut

https://github.com/p2panda/p2panda/pull/277#issuecomment-1073104989

Here I was advocating for sorting document view id hashes in this constructor and we agreed not to do it?

adzialocha: > Here I was advocating for sorting document view id hashes in this constructor and we agreed not to do it?

Seems to be a different topic though, this is related to the hashing method and not what we consider a valid document view id (if the sorting has implications for being rejected or not).

Still its an interesting old discussion which also made it back into this PR .. I'm sorting now everything in the constructor when it comes through `new` as doing it yourself before is too complicated. Nice thing is also, that we don't need to return a `Result` then. There is this `from_untrusted` method which is not public which is used by all places where we come from the outside. This one doesn't sort but instead validates the input.

sandreae: That jogs my memory a little. I think my position was coming from implementing `Eq` and that two `DocumentViewId`s which contain the same `OperationId`s should be equal regardless of the order.

> Thanks for the explanations @cafca. For point 1) looking at this from the functionality of a DocumentViewId I would expect two IDs to be equal even if the hashes are listed in different orders. The order has no meaning wrt which document view you are refering too. So the custom eq approach does make sense to me.

p2panda-rs/src/schema/schema.rs
```diff
@@ -253,43 +196,95 @@ impl Schema {
...
+    ///         ("name", FieldType::String),
+    ///         ("grow_cycle_days", FieldType::Int),
+    ///         ("flavor_rating", FieldType::Int),
+    ///     ]
+    /// );
+    /// assert!(schema.is_ok());
+    /// # }
+    /// # }
+    /// ```
+    pub fn new(
```
adzialocha, line 261: I'd be up for making this public to be able to create schema instances without a persistence and materialization layer. I'm afraid that there is otherwise quite a bunch to do before you can write an operation. What do you think?

p2panda-rs/src/next/operation/operation.rs
sandreae: Thanks for all the explanations here.

I see that these traits are a step towards some future coolness, and even though there are some possibilities for refactoring them to work more nicely with the existing operation traits (just sugar really), it's fine with me to just leave them all existing as they are now.

p2panda-rs/src/schema/schema.rs
sandreae: Yeh, I'd be up for that.

p2panda-rs/src/operation/plain/plain_value.rs
```diff
@@ -123,3 +123,125 @@ impl From<Vec<DocumentViewId>> for PlainValue {
...
+            deserialize_into::<PlainValue>(&serialize_value(cbor!(f64::MAX))).unwrap(),
+            PlainValue::Float(f64::MAX)
+        );
+
+        // It deserializes a too large integer into a float and passes which is not the expected
+        // behaviour, latest when checking against a schema it should fail though!
+        let bytes = serialize_value(cbor!(u64::MAX));
+        let value = deserialize_into::<PlainValue>(&bytes);
+        assert!(value.is_ok());
+        assert_eq!(value.unwrap().field_type(), "float");
```
adzialocha, line 245: This is pretty funny :scream:

p2panda-rs/src/schema/schema.rs
adzialocha: Okay, I think I can add it using the new validation methods :+1:

adzialocha:
As you can see I ignored our "rule" a little and choose `pub(crate)` visibility over some schema fields (`Operation`, `Entry`, `Schema` etc.) instead of using a `new` constructor. I think this is a good approach to separate the concerns: `new` is publicly facing while the structs can only be constructed internally via the exposed fields from all directions (test fixtures, deserialization, etc.).

adzialocha:
I've made `Schema::new` public for now, think its a good idea to expose the functionality to developers to create `Schema` instances easily, even when it is not derived from "real" data.

If I would like to write a client in `p2panda-rs` currently I _need_ a `Schema` instance to write Operations. Its not like its horrible, its cool to check your data against a schema on the fly, but you need to get the instance from somewhere. Since you're writing a client you don't have any materialization logic or a schema provider or anything, so you need to create a `Schema` instance programmatically. If you do it right, it will work well, if you didn't the node will reject the operation because the schema id didn't match or you made the fields wrong, so I don't see a "risk" in this approach.

p2panda-rs/src/operation/validate.rs
```diff
@@ -0,0 +1,338 @@
...
+///                                                 │
+///                                                 │
+///                                                 │
+///                                                 ▼
+///                                         ┌─────────────────┐
+///                                         │VerifiedOperation│
+///                                         └─────────────────┘
+/// ```
+#[allow(clippy::too_many_arguments)]
+pub fn validate_operation_with_entry(
```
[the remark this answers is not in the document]
sandreae: Things are sinking into my brain a little more and now I wonder about a more general point this threw up for me. It's a bit of an update on the previous comment, but I'll leave that there for posterity in any case.

Maybe the API for creating operations is a little too tied to `Schema` now. Obviously I understand the place this is coming from (making guarantees about the validity of data on it's creation), but I it could be ok to create operations optimistically (without their schema) as well. Of course we need methods for actually verifying them against their stated schema, but maybe this doesn't need to be enforced at this level. Operations which actually get _published_ will for sure be rejected by a node if they don't match their schema, but on the client side, maybe nice to just chuck out operations while only knowing the schema id. 

I could imagine 3 different useful APIs:

- create operations (not validated against a schema) with `Operation` or an `OperationBuilder`
- validate these operations against a schema with `validate_operation(&operation, &schema)`
- create and validate an operation with `Schema::create()`, `::update()` & `::delete()` (sure there are a million variations on this, but here's one idea)

Splitting up the operation creation into validated/un-validated would allow for a flow where we optimistically create operations without the schema, and the one we have, where we validate at the time of creation. I _think_ both these flows are valuable.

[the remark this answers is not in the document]
sandreae: If these flows are split up, then `VerifiedOperation` could become the one which has all guarantees, and `Operation` could still be useful (you can publish it) but not with quite the same guarantees as `VerifiedOperation`.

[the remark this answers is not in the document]
sandreae: related to this: https://github.com/p2panda/p2panda/pull/415#issuecomment-1209490662 which approaches the same issue by making `Schema` very easy to create :+1:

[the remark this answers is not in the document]
sandreae: This conversation reminds me of front-end data validation and back-end data validation, like, validating data in a form before sending it to the server is _better_ but you don't have to :rofl: it'll get rejected anyway.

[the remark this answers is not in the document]
adzialocha: > I could imagine 3 different useful APIs:

Yeah, thats also a good idea, I'd be up for going with 1. and 2.! 3. is cool, but could be done later :+1:

p2panda-rs/src/operation/mod.rs
```diff
@@ -3,26 +3,148 @@
...
+//! operation fields.
+//!
+//! To programmatically create an `Operation`, use the `OperationBuilder`, when working with
+//! operations coming in as bytes, you can use the `decode_operation` method to first deserialize
+//! it into a `PlainOperation` instance, which is a schemaless object giving you already access to
+//! the "header" data, like the schema id.
+//!
+//! Knowing the schema id you can look up your internal database for known schemas and derive a
+//! `Schema` instance from there. Now together with the `PlainOperation` and `Schema` you can
+//! finally validate the operation (via `validate_operation`) to arrive at the final, verified
```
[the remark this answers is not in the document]
adzialocha: Maybe use "checked"? :laughing:

p2panda-rs/src/operation/validate.rs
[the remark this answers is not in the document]
sandreae: Yeh, I was imagining 3 as a "future" thing too.

p2panda-rs/src/entry/validate.rs
```diff
@@ -0,0 +1,301 @@
...
+pub fn validate_log_integrity(
+    entry: &Entry,
+    skiplink_entry: Option<&Entry>,
+    skiplink_entry_hash: Option<&Hash>,
+    backlink_entry: Option<&Entry>,
+    backlink_entry_hash: Option<&Hash>,
+) -> Result<(), ValidateEntryError> {
+    // First make sure we didn't pass wrong arguments
+    if backlink_entry.is_some() != backlink_entry_hash.is_some() {
+        panic!("Both encoded and decoded backlink entry must be the same")
```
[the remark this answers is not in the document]
adzialocha: I'm fine with either, this post I read today and actually I think still panicking would fit: https://blog.burntsushi.net/unwrap/ as it would be a buggy implementation but using the tuples is also nice!

p2panda-rs/src/entry/validate.rs
```diff
@@ -0,0 +1,301 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+//! Collection of low-level validation methods for entries.
+//!
+//! You will not find methods here to check the encoding of Bamboo entries, as this is handled
+//! inside the external bamboo-rs crate.
```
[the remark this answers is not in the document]
adzialocha: Yes, thats right, but also we never used them before either as far as I know. I like that our methods are more granular :+1:

p2panda-rs/src/operation/validate.rs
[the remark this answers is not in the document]
sandreae: Do you want me to implement this for you to check after?

p2panda-rs/src/entry/mod.rs
```diff
@@ -4,24 +4,88 @@
...
+//! │Entry│ ──encode_entry()──► │EncodedEntry│ ─────► bytes
+//! └─────┘                     └────────────┘
+//! ```
+//!
+//! ## Validation
+//!
+//! The above high-level methods will automatically do different sorts of validation checks. All
+//! low-level methods can also be used independently, depending on your implementation:
+//!
+//! 1. Correct hexadecimal encoding (when using human-readable encoding format) (#E1)
```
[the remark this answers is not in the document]
adzialocha: Should we add a note that the codes refers to the specification? I imagine it will be written there, also there is a list in `entry/mod.rs`

p2panda-rs/src/entry/validate.rs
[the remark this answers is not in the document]
sandreae: We're using the bamboo verify method in `aquadoggo` but can replace it with these now :+1:

adzialocha:
> Hey, first up, amazing you've managed to take all this on, it is great and dealing with like a million issues we have open. My general comment would be "Whoop!"

Cool, thanks so much! I'm up for you or @cafca doing direkt changes instead of me going through them if thats alright? I love all your feedback :+1:

sandreae:
Yep, cool, I'm up for direct changes, just wanted to check I wasn't missing something important first :rofl: (or if it was a terrible idea)

p2panda-rs/src/operation/validate.rs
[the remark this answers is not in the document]
adzialocha: > Do you want me to implement this for you to check after?

Yes, that's great! We can check with @cafca maybe first to get consensus, but I'm for it ✌️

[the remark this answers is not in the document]
sandreae: I made a very light touch implementation in this PR: https://github.com/p2panda/p2panda/pull/427

p2panda-rs/src/entry/mod.rs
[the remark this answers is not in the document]
cafca: I looked into adding this to the spec and I think it will require some changes to the format. Hacking on that now...

p2panda-rs/src/operation/validate.rs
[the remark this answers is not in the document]
sandreae: @cafca do you have any impression on this change (I know it may be hard without looking at _everything_)?

The key takeaway I think, is that you can now construct an `Operation` without a `Schema` and without validating the fields. All the original methods for "validation max" remain where everything _is_ validated. We just opened a back door for easy operation creation.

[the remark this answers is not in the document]
sandreae: @adzialocha an alternate approach to what I implemented would be to offer an API for constructing `PlainOperation`s. This would then leave the `PlainOperation` -> `Operation` -> `VerifiedOperation` hierarchy un-touched and still offer a way to create/publish operations without needing the schema.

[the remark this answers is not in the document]
sandreae: That's what you did in `wasm` world... maybe that's nicer....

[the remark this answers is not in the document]
sandreae: An important case to consider in this convo is operations coming from the db in `aquadoggo`. We just want to instantiate an operation in this case, without the schema validation, as we trust the data.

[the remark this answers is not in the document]
sandreae: > @adzialocha an alternate approach to what I implemented would be to offer an API for constructing `PlainOperation`s. This would then leave the `PlainOperation` -> `Operation` -> `VerifiedOperation` hierarchy un-touched and still offer a way to create/publish operations without needing the schema.

I'm :+1: ing my own suggestion, think it's better. (partly cos then we don't need to change any of the diagrams :rofl: )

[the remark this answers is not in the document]
sandreae: > I think we still should go for Operation being without schemas.

Ok, coool

p2panda-rs/src/operation/verified_operation.rs
```diff
@@ -2,178 +2,84 @@
...
 }
 
 impl AsVerifiedOperation for VerifiedOperation {
-    type VerifiedOperationError = VerifiedOperationError;
-
-    /// Returns a new `VerifiedOperation` instance.
-    ///
-    /// Use `VerifiedOperation::new_from_entry()` instead if you want to validate that the operation
-    /// was signed by this public key.
-    fn new(
```
sandreae, line 37: It's only possible to get a `VerifiedOperation` via `validate_operation_with_entry()` which requires a whole bunch (schema, entry etc...) however when we get operations from the `OperationStore` they come out as a load of strings and need to be parsed into a `VerifiedOperation`. Any thoughts on how we might do this nicely @cafca & @adzialocha ?

adzialocha: It sounds a little bit like you need a different data type in aquadoggo, something which got verified in the past but that's it, from that point you don't need it anymore. Maybe we introduce a new struct in the doggo which just represents what you have in the database?

adzialocha: Isn't it funny how the separation of entry and operation and it's dependency creeps back at us all the time? 😅 I can see how it is fine to combine objects without checking but I'm not sure yet if the core library needs to provide such a thing?

sandreae: > It sounds a little bit like you need a different data type in aquadoggo, something which got verified once in the beginning. Verifying it a second time doesn't make sense after you already stored it in the database. Maybe we introduce a new struct in the doggo which just represents what you have in the database? Like a simple struct which just holds the required fields?

yeh, I'm considering this, it's actually how things were earlier in the storage provider story, but the structs got merged at some point. Slight worry here is that we may still need a `VerifiedOperation` at some point, but this is what we have the traits for, so it could hopefully work out.

p2panda-rs/src/operation/operation_value.rs
cafca: Yeah, maybe "string" is actually less ambiguous, every hacker will know what it means. 

For "string" I always think about a short, maybe one-word value, and also I always wonder "does it support unicode?". For "text" I thought maybe it's more obvious that there is no, i don't know, 255 character limit and you can't put äëüö in it. But maybe that's also just me :)

p2panda-rs/src/operation/operation_action.rs
```diff
@@ -0,0 +1,126 @@
...
+use std::fmt::Display;
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+/// Operations are categorised by their action type.
+///
+/// An action defines the operation format and if this operation creates, updates or deletes a data
+/// document.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum OperationAction {
```
cafca, line 13: In the spec this is called _operation type_, which is a bit more generic. Would you mind if I rename it?

adzialocha: Oh! I have the feeling we've been always talking about "operation action", so its hard for me to think about anything else. Its a relatively big change to change it to "type" now as many other places use the word "action" .. what do you think @sandreae?

sandreae: Hmm, I'm scared to introduce more renaming in the code right now... :grimacing:

p2panda-rs/src/document/document_id.rs
```diff
@@ -23,13 +23,13 @@ use crate::{Human, Validate};
 ///                         \
 ///                          \__ [UPDATE] (Hash: "eff..")
 /// ```
-#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
```
sandreae, line 26: `Serialize` and `Deserlialize` traits are gone in this PR but they are needed for gql scalar conversions in `aquadoggo`. Any reason they should be gone? @adzialocha @cafca ?

adzialocha: I've removed all serde implementations which are not required directly, of course without looking at aquadoggo though .. if we need them we can bring them back but bear in mind to not just derive `Deserialize` as it would just throw anything unchecked into the underlying string type.

That being said, is it really required in aquadoggo? I understand that the scalar trait actually doesn't use serde, the derives there are unused and misleading, should probably be removed.

I'm not sure all of this is correct, haven't used it yet, I just have a feeling 😅

p2panda-rs/src/operation/operation_action.rs
cafca: I can understand that :)

p2panda-rs/src/document/document_id.rs
cafca: My understanding is that Piet used these for replication but `async_graphql` actually has its own system for scalar conversions (`ScalarType`) that _can_ use serde but doesn't have to.

sandreae: Ok, thanks for the comments, I'm unfamiliar with this particular corner of the doggo. I'll check if the traits are actually needed and remove or re-look at it.

sandreae:
@adzialocha I've worked in `aquadoggo` on integrating these changes and am confident there are no nasty surprises up ahead :+1: so now I want to come back to this PR and get the tests we are missing back in and continue the review process. I made these changes https://github.com/p2panda/p2panda/pull/434 while working on the `aquadoggo` integration though, and wonder how best to proceed. I see these changes as coming out of my review of this PR including things that I think are needed for nicely integrating it into `aquadoggo`. So with that in mind, I'd like to proceed in the following order:

1) merge https://github.com/p2panda/p2panda/pull/434 into this PR (after you looked at it)
2) bring back all the missing tests
3) open this up for review again

From 2) in this list, I would say we are back to parity with `main`, we still have some things in the list of "Follow-ups" but i recon it's ok for us to take these a little later. Nothing stops us from finishing the review process for this PR and implementing in `aquadoggo` then. Does this all sound ok to you or would you prefer a different approach?

p2panda-rs/src/storage_provider/traits/test_utils.rs
```diff
@@ -209,192 +200,79 @@ pub struct TestData {
...
-                }
-            }
-        }
-    }
+    // TODO: Needs reinstating when we deal with https://github.com/p2panda/p2panda/issues/418
+    todo!()
 }
 
 /// Helper method for publishing an operation encoded on an entry to a store.
 pub async fn send_to_store(
```
sandreae, line 208: This is the helper methods that needs re-implementing and is blocking us from bringing the tests back.

In the comment I said it's tied to https://github.com/p2panda/p2panda/issues/418 but actually, that is just ideal scenario, we can bring it back without that. But changes made in my follow-up PR will impact here.

adzialocha:
> Does this all sound ok to you or would you prefer a different approach?

Thank you! I think this sounds good, maybe we even merge it to `development` before 2.?

sandreae:
> Thank you! I think this sounds good, maybe we even merge it to development before 2.?

Ok, great! yep, can merge across to `development` then, good call. Shall I copy all of your opening comments across? Wouldn't want to not have them..

adzialocha:
> Ok, great! yep, can merge across to `development` then, good call. 

Super!

> Shall I copy all of your opening comments across? Wouldn't want to not have them.

Yeah, why not! Or link to them, whatever is best.

sandreae:
Do we wanna squash merge into `development` or retain all commits :thinking: ?

adzialocha:
> Do we wanna squash merge into `development` or retain all commits :thinking: ?

Think it's fine to squash .. we can make sure we nicely link to everything in the CHANGELOG.md

PR #416, opened 2022-07-28
Easy operations and schemas

cafca:
Writing tests for the GraphQL api I kept thinking about how much I wanted to use all the test utils even for things that should be easy even when not writing tests. Like creating operations without having to initialise `OperationFields` instances or making operations that create schemas when sent to a node without jumping through all kinds of hoops. 

When we write examples (and p2panda apps!) we don't want to have to use test utils all the time I guess sooo..

So here I have implemented 

**A) conversion implementations for `OperationValue` and `OperationFields`:**

for example:

```rust
assert_eq!(OperationValue::Text("pingu".to_string()), "pingu".into());
```

these can now be used with the `Operation::new_create` etc. methods we already had:

```rust
let op = Operation::new_create(
    SchemaId::SchemaFieldDefinition(1),
    vec![ // <-- No `OperationFields::new()` here
        ("name", "field_name".into()), // <-- or here
        ("type", FieldType::String.into()),
    ]
    .try_into()
    .unwrap(),
);
assert!(op.is_ok());
```

**B) helper to create schema `create` operations**

(the same exists for schema field definitions on `Schema::create_field`)

```rust
let operation = Schema::create_field("title", FieldType::Bool).unwrap();
let (field_entry, _) = send_to_node(&mut node, &client, &operation).await.unwrap();

let field_view_id: DocumentViewId = field_entry.as_str().parse().unwrap();

// Create schema definition
let operation = Schema::create(
    "books",
    "books in my library",
    vec![field_view_id].into(),
)
.unwrap();

let (schema_view_id, _) = send_to_node(&mut node, &client, &operation).await.unwrap();
```

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/operation/operation_fields.rs
+++ b/p2panda-rs/src/operation/operation_fields.rs
@@ -2,6 +2,7 @@
 
 use std::collections::btree_map::Iter;
 use std::collections::BTreeMap;
+use std::convert::TryFrom;
 
 use serde::{Deserialize, Serialize};
 
@@ -120,12 +121,25 @@ impl Validate for OperationFields {
     }
 }
 
+impl TryFrom<Vec<(&str, OperationValue)>> for OperationFields {
+    type Error = OperationFieldsError;
+
+    fn try_from(spec: Vec<(&str, OperationValue)>) -> Result<Self, Self::Error> {
+        let mut operation_fields = OperationFields::new();
+        for field in spec {
+            operation_fields.add(field.0, field.1)?;
+        }
+        Ok(operation_fields)
+    }
+}
+
 #[cfg(test)]
 mod tests {
+    use std::convert::TryFrom;
+
     use rstest::rstest;
 
     use crate::document::DocumentViewId;
-
     use crate::operation::{OperationId, PinnedRelationList};
     use crate::test_utils::fixtures::random_operation_id;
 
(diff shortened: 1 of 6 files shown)


cafca:
I see I have to do something about Node not being available in wasm. Not today, I'll just revert this to draft. Would like to hear from you anyway about what you think.

sandreae:
Ooh, this is just great!

adzialocha:
Yeah, this is totally integrateable into the monster PR 👍

adzialocha (approved):
Very helpful 🤩

p2panda-rs/src/operation/operation_fields.rs
```diff
@@ -120,12 +121,25 @@ impl Validate for OperationFields {
     }
 }
 
+impl TryFrom<Vec<(&str, OperationValue)>> for OperationFields {
```
adzialocha, line 124: In the refactor this is not required anymore as the `OperationBuilder` has a similar API (it takes in `.fields(&[(&str, OperationValue)])`)

p2panda-rs/src/schema/schema.rs
```diff
@@ -575,3 +646,46 @@ mod tests {
         .is_err());
     }
 }
+// Exclude wasm which doesn't support `Node`
+#[cfg(not(target_arch = "wasm32"))]
+#[cfg(test)]
+mod tests_wasm_not_invited {
```
adzialocha, line 652: In the refactor I'll move this in another place as this is already much more of an "integration" test, I will write some more unit tests here instead :+1:

Issue #24, opened 2021-01-16
Create Typescript declarations with build


cafca:
- [ ] Create a single `index.d.ts` in the build folder with types for the exported module
- [ ] No import paths in the declaration that don't resolve independent of the p2panda repository
- [ ] Unset env var in `npm run test` that disables type checks

adzialocha:
Creating a single automated definition file seems to be tricky, I found a few experimental webpack plugins (https://github.com/eliraz-refael/dts-webpack-bundler) but they were all not working.

adzialocha:
I remember you had a simple solution for this in the end @cafca (as all the others seemed very complicated), but I can't find it in the code-base. Did we merge it in the end or is this still pending?

It could be nice to have it in for the next release (with our nice and shiny new API) :smile:

PR #429, opened 2022-08-11
Fix WebAssembly builds for bundlers and browsers

### LINKED ISSUE #126: Using webpack with `p2panda-js` (v0.2.1) is not working "out of the box"
When having `p2panda-js` (v0.2.1) as a dependency in a project using Webpack as a bundler (for example `beep-boop`), the app breaks stating that it could not find the `.wasm` file. Webpack did not consider it since it is loaded via `fetch` inside the `index.min.js` file (this is the file created for browser targets in our build process of `p2panda-js`) and is therefore missing in the dependency tree.

## Possible Solution

To solve the issue with Webpack not importing the `.wasm` file of `p2panda-js` we just found a solution (not elegant, but working for now):

1. Install the CopyWebpackPlugin as a dev-dependency
2. Add the following config under `plugins` in the webpack file (also import the CopyWebpackPlugin):

```js
new CopyPlugin({
  patterns: [
    { from: 'node_modules/p2panda-js/lib/*.wasm', to: '[name][ext]' },
  ],
}),
```

3. Webpack will now copy the needed `.wasm` file into the `dist` folder.

## Another possible solution

Roll back this change: https://github.com/p2panda/p2panda/pull/105 and use wasm inside the bundled file as an base64 encoded string again ..

adzialocha:
**This PR ..**

* Fixes support of `p2panda-js` in Browsers and NodeJS (#126 
* Exports TypeScript definitions (#24) 
* Optimizes WebAssembly build for size and speed
* Requires now `wasm-bindgen` and `wasm-opt` instead of `wasm-pack` to be installed

**New builds**

It changes `p2panda-js` to export three build targets:

1. `node` (for NodeJS environments, no async initialization required)
    ```javascript
    const p2panda = require('p2panda');
    const keyPair = p2panda.createKeyPair();
    console.log(keyPair.publicKey());
    ```
2. `web` (running in Browser and Web-Apps, async one-time initialization required, developers need to manually serve and import `.wasm` file, but faster initialization and smaller files)
    ```javascript
    const { initWebAssembly, createKeyPair } = p2panda;
    await initWebAssembly('/path/to/p2panda.wasm');
    const keyPair = createKeyPair();
    console.log(keyPair.publicKey());
    ```
3. `inline` (running in Browsers and Web-Apps, very easy to get started, nice for quick hacks, async one-time initialization required, larger files)
    ```javascript
    const { initWebAssembly, createKeyPair } = p2panda;
    await initWebAssembly();
    const keyPair = createKeyPair();
    console.log(keyPair.publicKey());
    ```

Closes #126 and #24 (our oldest task 👴)

## 📋 Checklist


p2panda-js/scripts/compile-wasm.sh
```diff
@@ -0,0 +1,127 @@
+#!/bin/bash
```
cafca, line 1: I think this is the most usable non-trivial shell script I have ever seen 👏

p2panda-js/README.md
```diff
@@ -108,15 +152,32 @@ const Chat = ({}) => {
 };
 ```
 
-## Development Setup
+### Manually load `.wasm`
```
sandreae, line 155: Very cool we offer this as well :+1:

And these multiple examples are great to have, very developer friendly!

p2panda-js/scripts/compile-wasm.sh
sandreae: wow, nice script!

p2panda-js/src/index.ts
```diff
@@ -3,14 +3,20 @@
 import fetch from 'node-fetch';
 import Headers from 'fetch-headers';
 
+import { setWasmPanicHook } from '../wasm/node';
+
+// @TODO: Remove `fetch` together with `apollo/client`
```
sandreae, line 8: :+1:

p2panda-js/src/wasm.test.ts
```diff
@@ -1,28 +1,32 @@
...
+} from './wasm';
 
 const TEST_HASH =
   '0020ddc99aca776df0ca9d1b5871ba39d4edacc752a0a3426b12c3958971b6c847ac';
 const TEST_SCHEMA_ID = `test_${TEST_HASH}`;
 
 describe('WebAssembly interface', () => {
   describe('KeyPair', () => {
-    it('creates a key pair', async () => {
-      const { KeyPair } = await wasm;
```
sandreae, line 12: Nice without `await` everywhere

sandreae (approved):
This looks great to me! Thanks for doing all this important work :+1:

I really like having the different import methods available and well documented.

cafca:
This is working?? 😍😍 can't wait to see it.

PR #437, opened 2022-08-14
Bring back all the tests (of the tests)

sandreae:
Brings back tests for `Document` and `MemoryStore` in `test_utils` re-implementing needed testing utilities. Several steps were taken in order to do this neatly and along the way I deprecated the mock structs `Node` and `Client`. This was possible because their functionality has been overtaken by the `test_db` methods we developed in `aquadoggo` as well as having `publish` and `next_args` as stand alone methods. They've been introduced now in this pr but only in the `test_utils` module, we can leave them there for a while.

- [x] deprecate mock `Node` and `Client`, all cases where they were used (only in document tests) have been replaced by `MemoryStore` and the test db methods.
- [x] introduce the `validation` and `domain` modules to the `db` module in `test_utils` and use these to re-implement other testing methods
- [x] re-implement `send_to_node`, `populate_test_db` and related methods
- [x] refactor all tests using these methods
- [x] deprecate `find_next_log_id()` on LogStore

## Comments
`MemoryStore` is super powerful now, it has parity with `SqlStore` in `aquadoggo`. It's also massively OTT for what we need it for in `p2panda-rs`*..... We could actually refactor it out of the remaining few tests that need it (see `document` tests, they are much neater without `Node` :rofl:). It is ripe for being stripped out into it's own crate, but for now it was easier just to fix all the tests as has been done in this PR. I wondered if we could just pull it out into it's own PR but we need it for some quite important tests in `aquadoggo` where we test behaviour where entries are missing from a log, which is hard to do with the sql db. If anyone has thoughts on this (stripping it out) then I'm all ears :+1:.

(*) Some context: the reason we needed the ability to test the stores in `p2panda-rs` is because originally we implemented a load of default trait methods which needed a testing setup, this has all moved over to `aquadoggo` now and the `domain` and `validation` modules. It leaves the trait definitions as just that, definitions, no default implementations, so we don't really need the ability to test these anymore.

## Extras
There are a couple of thousand lines of tests which should be brought across for the `domain` and `validation` modules, they should be re-instated, could be in this PR (they already work) but I didn't want to freak you all out with the diff change :rofl:.

## 📋 Checklist


p2panda-rs/src/document/document.rs
```diff
@@ -309,169 +314,171 @@ mod tests {
...
+            .fields(&[(
+                "name",
+                OperationValue::String("Polar Bear Cafe".to_string()),
+            )])
+            .previous_operations(&DocumentViewId::new(&[
+                penguin_entry_1.hash().into(),
+                panda_entry_2.hash().into(),
+            ]))
+            .build()
+            .unwrap();
```
adzialocha, line 404: This looks so nice! We could even make use of the "simple operation" conversion traits for operation values to make it even nicer, but thats not necessary.

p2panda-rs/src/test_utils/constants.rs
```diff
@@ -4,8 +4,11 @@
 use crate::operation::{
     OperationValue, PinnedRelation, PinnedRelationList, Relation, RelationList,
 };
+use crate::schema::Schema;
 use crate::test_utils::fixtures::{document_id, document_view_id};
 
+use super::fixtures::{schema_fields, schema_id};
```
adzialocha, line 10: Use the `crate::test_utils::fixtures` import above instead

p2panda-rs/src/test_utils/db/domain.rs
```diff
@@ -0,0 +1,1246 @@
...
+    //////////////////////////////////
+    // VALIDATE ENTRY AND OPERATION //
+    //////////////////////////////////
+
+    // TODO: Check this validation flow is still correct.
+
+    // Verify that the claimed seq num matches the expected seq num for this author and log.
+    let latest_entry = store.get_latest_entry(author, log_id).await?;
+    let latest_seq_num = latest_entry.as_ref().map(|entry| entry.seq_num());
+    is_next_seq_num(latest_seq_num, seq_num)?;
```
adzialocha, line 230: Interesting! This should be validated implicitly by `validate_log_integrity` inside of `validate_operation_with_entry` as the hash will otherwise be wrong of the backlink entry. I wonder if we can add a test for that just to be sure!

p2panda-rs/src/test_utils/db/domain.rs
```diff
@@ -0,0 +1,1246 @@
...
+                .map_err(|_| {
+                    "You are trying to update or delete a document which has been deleted"
+                })?;
+
+            document_id
+        }
+    };
+
+    // Verify the claimed log id against the expected one for this document id and author.
+    verify_log_id(store, author, log_id, &document_id).await?;
```
adzialocha, line 300: Maybe we can call this `verify_document_log_id` to not confuse it with entry / log id validation?

p2panda-rs/src/test_utils/db/validation.rs
```diff
@@ -0,0 +1,432 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+//! Methods for validating entries and operations against expected and stored values.
```
adzialocha, line 3: So all of this is here is tooling for the massive storage tests? :rofl: Its amazing stuff! I'm curious to see how we can integrate it into the library at one point.

adzialocha (approved):
Just small style comments. Otherwise woho!

p2panda-rs/src/test_utils/db/validation.rs
sandreae: Well, it's the module we use in `aquadoggo`, both `domain` and `validation` are directly picked up from there. I brought them across as then I could recycle the existing code for the test utils and we'd be fairly confident the behaviour of `MemoryStore` and `SqlStore` are equal.

But in the context of `p2panda-rs`, yes, is all just test tooling :rofl: I had imagined they would come into the main lib eventually, but thought at this point it wasn't necessary and just having them here was enough.

p2panda-rs/src/test_utils/db/domain.rs
sandreae: Aha, yes, this is exactly what I was asking you to look at this morning :+1: these methods are exactly what we are using in `aquadgoog` too. I'll check this and remove if it's not needed.

p2panda-rs/src/test_utils/db/validation.rs
sandreae: (i wouldn't have written all this just for test utils.... honest)

p2panda-rs/src/test_utils/db/domain.rs
sandreae: `aquadgoog`! amazing word

sandreae: I'll leave this here now as I think there may be more which we will catch in the `aquadoggo` pr, then I'll bring all that over here in one go.

p2panda-rs/src/document/document.rs
sandreae: I'll look into that!

PR #461, opened 2022-09-03
Rename `Author` -> `PublicKey` & `previous_operations` -> `previous`

sandreae:
Rename `Author` struct to `PublicKey` following these conventions:
- when refering to bamboo entries specifically it is still called an `author`
- if we are talking about a high level "author" it is still an author
- a `Document` still has an `author` field which contains a `PublicKey` (this just felt right as it is a higher-level data type and the author will be a KeyGroup later i guess)
- everywhere else `author` -> `public_key`

Rename `previous_operations` -> `previous`

## 📋 Checklist


p2panda-rs/src/document/document.rs
```diff
@@ -56,8 +56,8 @@ impl Document {
         &self.view_id
     }
 
-    /// Get the document author.
-    pub fn author(&self) -> &Author {
+    /// Get the document author's public key.
+    pub fn author(&self) -> &PublicKey {
```
adzialocha, line 60: Maybe we should also rename the method to `public_key` (and struct fields etc.)

adzialocha (changes requested):
Thank you for doing this (historic) change 🤩 It looks all good, just have this question and the one above:

It's slightly confusing now that we have two `PublicKey` instances, one coming from the ed25519 crate and the other from us. My suggestion would be: 

1. Wrap the ed25519 struct in ours instead of `String`
2. Return our struct when calling `KeyPair.public_key()`

p2panda-rs/src/document/document.rs
adzialocha: Ah yeah, Author actually makes sense on a document as it could be many public keys

sandreae:
> Wrap the ed25519 struct in ours instead of String (it's strange anyhow, basically an early sqlx remaining, and we already do this with Signature)

> Return our struct when calling KeyPair.public_key()

Nice, I'll make these changes.

sandreae:
Hmm, one hitch, `ed25519_dalek` PublicKey doesn't implement `Hash`, soooo I guess we need to manually implement that ourselves on our own `PublicKey` wrapper struct?

adzialocha (approved):
This looks great!

cafca (approved):
Looks good! I think we would differentiate information about public keysassociated with a document furthermore anyway dot there would be a key group which is the owner of the document, a public-key who has started the document, and A setof the public keys who have participated in the document in the past, regardless of whether they are members of the key group who is currently owning the document.

Issue #426, opened 2022-08-09
Validate description, fields and name in `Schema`


adzialocha:
Whenever we create a `Schema` instance we should check if the description, schema name and field names are following the specification.

adzialocha:
I suggest we introduce new `FieldName` and `SchemaDescription` and `SchemaName` structs which in themselves have `Validate` implementations checking the necessary things

sandreae:
Hmm... I introduced a `SchemaName` struct in the above linked PR, it makes sense and easy to see what validation occurs for the name part of a schema id. However doing the same for `FieldName`, `SchemaDescription`, `SchemaFields` and `FieldType` now looks a bit overkill.... Maybe the benefits of having more consistency across the code base is lost in adding so many extra "single use" types.... @adzialocha any preference? Right now I'm preferring just leaving as is and not adding any new types. It's a slightly different more functional pattern in this module but equally easy to read and understand.

adzialocha:
> Hmm... I introduced a `SchemaName` struct in the above linked PR, it makes sense and is easy to see what validation occurs for the name part of a schema id. However doing the same for `FieldName`, `SchemaDescription`, `SchemaFields` and `FieldType` now looks a bit overkill.... Maybe the benefits of having more consistency across the code base is lost in adding so many extra "single use" types.... @adzialocha any preference? Right now I'm preferring just leaving as is and not adding any new types. It's a slightly different more functional pattern in this module but equally easy to read and understand. 

Sounds good to me 👍🏻 as long as all validation is handled from all ends it's good to not over engineer it.

adzialocha:
> Whenever we create a `Schema` instance we should check if the description, schema name and field names are following the specification.

Just checking, is this implemented?

sandreae:
Ugh, no, it is not. I was way to focused on `SchemaId` thinking the problem was there.... 

https://github.com/p2panda/p2panda/blob/0fdb7d05fb971e0d6fb9910762119f93a064a0af/p2panda-rs/src/schema/schema.rs#L83-L105

sandreae:
https://github.com/p2panda/p2panda/blob/0fdb7d05fb971e0d6fb9910762119f93a064a0af/p2panda-rs/src/schema/schema.rs#L38-L52

https://github.com/p2panda/p2panda/blob/0fdb7d05fb971e0d6fb9910762119f93a064a0af/p2panda-rs/src/schema/schema.rs#L410-L424

sandreae:
[...]

PR #481, opened 2023-02-19
Introduce `SchemaName` `SchemaDescription` and `SchemaFields` structs

sandreae:
This PR introduces three new types which provide validation, checking adherence to the specification requirements, in the construction of `Schema`s. The new types are `SchemaName`, `SchemaDescription` and `SchemaFields` each of which has a `validation` method which performs all required validation steps and is called in each structs constructor. Using these types in the constructor of `Schema` ensures validation occurs as expected.

An additional change included in this PR is to add checks for any additional fields which might exist (but currently be ignored) when converting generic `DocumentViews` into specialised `SchemaDefinitionView` and `SchemaFieldDefinition` views.

NOTE: There are three distinct validation flows all relating to validating the fields of schema-like objects. One is dealing with `PlainOperation`s, one is dealing with the construction of `Schema` and one is dealing with conversion of `DocumuntView`s into system schema views. Each flow has their own errors and validation methods. This is a little confusing and could maybe be simplified if the methods became more generic. It seemed out of scope to tackle all that (the operation validation flow in particular is dense) but I raise a refactoring issue https://github.com/p2panda/p2panda/issues/482.

closes: https://github.com/p2panda/p2panda/issues/426 https://github.com/p2panda/handbook/issues/220 #238 

## 📋 Checklist


sandreae:
Closing this as we are performing all necessary validation and this PR leads us down a purely cosmetic round of refactoring when the code is already very nice 👍

sandreae:
Resurrecting this PR as introducing this struct is indeed the way to go (I realise now that I more completely understand the problem at hand ;-p)

p2panda-rs/src/schema/schema_description.rs
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+use std::fmt;
+use std::fmt::Display;
+use std::str::FromStr;
+
+use crate::schema::error::SchemaDescriptionError;
+use crate::schema::validate::validate_description;
+
+/// A human readable schema name, used in the construction of `SchemaId`.
```
sandreae, line 10: Fix doc string.

p2panda-rs/src/schema/schema_description.rs
```diff
@@ -0,0 +1,72 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct SchemaDescription(String);
+
+impl SchemaDescription {
+    /// Construct and validate new schema name from a string.
```
sandreae, line 15: Fix doc string.

p2panda-rs/src/schema/schema_description.rs
```diff
@@ -0,0 +1,72 @@
...
+    #[should_panic]
+    #[case(
+        "In common use the term is used to describe the largest species from this
+           family, the red kangaroo, as well as the antilopine kangaroo, eastern grey
+           kangaroo, and western grey kangaroo! Kangaroos have large, powerful hind legs,
+           large feet adapted for leaping, a long muscular tail for balance, and a small
+           head. Like most marsupials, female kangaroos have a pouch called a marsupium
+           in which joeys complete postnatal development."
+    )]
+    fn check_name_field(#[case] description_str: &str) {
```
sandreae, line 69: Fix test name.

p2panda-rs/src/schema/schema_fields.rs
```diff
@@ -0,0 +1,138 @@
...
+use super::error::SchemaFieldError;
+use super::validate::validate_field_name;
+use super::FieldType;
+
+/// The fields definitions of a [`Schema`].
+#[derive(Clone, Debug, PartialEq, Default, Eq)]
+pub struct SchemaFields(BTreeMap<String, FieldType>);
+
+impl SchemaFields {
+    /// Creates a new schema fields instance to add data to.
```
sandreae, line 15: Fix doc string.

p2panda-rs/src/schema/schema_fields.rs
```diff
@@ -0,0 +1,138 @@
...
+        if keys.len() != fields.len() {
+            return Err(SchemaFieldError::DuplicateFields);
+        }
+
+        let mut schema_fields = BTreeMap::new();
+        for (key, value) in fields {
+            schema_fields.insert(key.to_string(), value.to_owned());
+        }
+
+        let schema_fields = Self(schema_fields);
```
sandreae, line 30: Some more comments here

p2panda-rs/src/schema/error.rs
```diff
@@ -5,6 +5,48 @@ use thiserror::Error;
...
+
+/// Custom errors related to `SchemaDescription`.
+#[derive(Clone, Error, Debug)]
+pub enum SchemaDescriptionError {
+    /// Encountered a too long schema description.
+    #[error("Schema description contains more than 256 characters")]
+    TooLongSchemaDescription,
+}
+
+impl Copy for SchemaDescriptionError {}
```
adzialocha, line 26: I'm not sure about this particular case, but I realized that the `Copy` trait is often not required and implemented because of copy & pasting ☺️

Think it's only required if the error types contain additional data?

Just saying, if you already checked and it's required then all good.

p2panda-rs/src/schema/schema_description.rs
```diff
@@ -0,0 +1,73 @@
...
+    ///
+    /// 1. It consists of unicode characters
+    /// 2. ... and must be at most 256 characters long
+    pub fn validate(&self) -> Result<(), SchemaDescriptionError> {
+        if !validate_description(&self.0) {
+            return Err(SchemaDescriptionError::TooLongSchemaDescription);
+        }
+        Ok(())
+    }
+}
```
adzialocha, line 33: Is there a reason you didn't use the validation trait here (and in other places)?

p2panda-rs/src/schema/schema_fields.rs
```diff
@@ -0,0 +1,142 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+use std::collections::btree_map::Iter;
+use std::collections::BTreeMap;
+
+use super::error::SchemaFieldError;
+use super::validate::validate_field_name;
+use super::FieldType;
```
adzialocha, line 8: Let's not use super when it's not somewhat required because of visibility or tests

adzialocha (changes requested):
Looking good! I have some questions but in general all nice 👍🏻

It would be good indeed to streamline this with the other flows, especially deserialization but I agree this should be handled in another PR.

p2panda-rs/src/schema/error.rs
sandreae: I did it cos clippy told me to 😅 it's not needed, but it also wasn't a copy pasty error.... I guess we can switch that clippy flag off if we like.

p2panda-rs/src/schema/schema_description.rs
sandreae: No! I totally forgot about it 😱 I'll use it 👍

p2panda-rs/src/schema/schema_fields.rs
sandreae: 👍

Issue #494, opened 2023-05-18
`PublicKey` serialization generates bytes while deserialization expects string 


sandreae:
With this error:

```
---- identity::public_key::tests::deserialize stdout ----
thread 'identity::public_key::tests::deserialize' panicked at 'called `Result::unwrap()` on an `Err` value: Semantic(None, "invalid type: bytes, expected string")', src/identity/public_key.rs:179:88
```

PR #495, opened 2023-05-18
Serialize `PublicKey` as bytes

sandreae:
closes: #494 

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@ Highlights are marked with a pancake 🥞
 - Implement serde for `SchemaName` and `SchemaDescription` [#487](https://github.com/p2panda/p2panda/pull/487) `rs`
 - Implement `Ord` for `SchemaId` [#492](https://github.com/p2panda/p2panda/pull/492) `rs`
 
-### Cha
+### Changed
+
+- Serialize `PublicKey` as bytes [#495](https://github.com/p2panda/p2panda/pull/495) `rs`
 
 ### Fixed
 
--- a/p2panda-rs/src/identity/public_key.rs
+++ b/p2panda-rs/src/identity/public_key.rs
@@ -11,7 +11,7 @@ use crate::identity::error::PublicKeyError;
 use crate::Human;
 
 /// Authors are hex encoded Ed25519 public key strings.
-#[derive(Clone, Debug, Serialize, Copy)]
+#[derive(Clone, Debug, Copy, Serialize)]
 pub struct PublicKey(Ed25519PublicKey);
 
 impl PublicKey {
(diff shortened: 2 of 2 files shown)


adzialocha:
Interesting issue! I wonder if actually serializing to a string is the thing we want to do ..

The steps (I think) are currently like that:

1. Have a `PublicKey` which is a wrapper type around the `ed25519-dalek` struct equivalent, this one serializes to bytes
2. Convert it to string representation for serialization, by encoding bytes as hex string, now the size doubles (64 bytes)
3. Encode it as CBOR, which will be a CBOR string

But what we want:

1. Same as above
2. Serialize it as bytes (still 32 bytes)
3. Encode it as CBOR, still bytes

IF we want a human readable representation of the CBOR bytes, we can always still encode it as a hex string afterwards, but the machine representation it doesn't matter (and we save bandwith).

adzialocha:
The issue here is that we Serialize already as in the second scenario above (32 bytes), but the Deserializer is wrong ..

adzialocha:
Here is an example on how this can be automatically done depending on if the underlying encode is human readable or not: https://github.com/p2panda/p2panda/blob/679107b636eea39e7713d9b21ac69df28aaad97d/p2panda-rs/src/entry/encoded_entry.rs#L26

sandreae:
Fixing `Deserialize` so it's expecting bytes sounds good to me 👍

sandreae:
@adzialocha how is this looking to you? Not sure I totally understood the best way to use the existing helper methods....

adzialocha:
> @adzialocha how is this looking to you? Not sure I totally understood the best way to use the existing helper methods....

You're serializing to hex which still doubles the byte size, for CBOR it's enough to just encode the bytes directly from the public key (and vice versa for decoding).

Check out the example I commented earlier on, we should have methods which only serialize to hex if it's a human readable encoding (for example JSON).

sandreae:
> 
@adzialocha how is this looking to you? Not sure I totally understood the best way to use the existing helper methods....

You're serializing to hex which still doubles the byte size, for CBOR it's enough to just encode the bytes directly from the public key (and vice versa for decoding).

Check out the example I commented earlier on, we should have methods which only serialize to hex if it's a human readable encoding (for example JSON).

I actually thought this is what the methods i used were doing. This one for example: https://github.com/p2panda/p2panda/blob/4dd6d8e199d9451951d1e9d454e80cb1be9b1fc1/p2panda-rs/src/serde/hex_str.rs#L5-L16

adzialocha:
Ah ja, this works as well. You can use the derive macros for that:

```
    #[serde(serialize_with = "serialize_hex", deserialize_with = "deserialize_hex")] Vec<u8>,
```

sandreae:
> 

Ooookay, cool. But this didn't work when I tried it in this commit https://github.com/p2panda/p2panda/pull/495/commits/a0d3205cafc2277c259c94ae2fa0b651bd532676 and you also thought it wasn't correct. Is that because I used the methods in manual implementations of `Serialize` and `Deserialize`? I didn't think that would make a difference...

adzialocha:
> > 
> 
> Ooookay, cool. But this didn't work when I tried it in this commit https://github.com/p2panda/p2panda/pull/495/commits/a0d3205cafc2277c259c94ae2fa0b651bd532676 and you also thought it wasn't correct. Is that because I used the methods in manual implementations of `Serialize` and `Deserialize`? I didn't think that would make a difference...

Ah sorry, I think I didn't look closely and got confused why you didn't just use the derive macros. Then it's all good 👍🏻

adzialocha:
I've reverted everything back to https://github.com/p2panda/p2panda/commit/a0d3205cafc2277c259c94ae2fa0b651bd532676 and will merge now, sorry for the confusion!

PR #507, opened 2023-08-03
`blob_v1` and `blob_piece_v1` schema

sandreae:
Introduces two new system schema:

- `blob_v1`
- `blob_piece_v1`

Each with methods for validating operations against their expected fields.

## 📋 Checklist

### DIFF
--- a/p2panda-rs/src/schema/validate/blob.rs
+++ b/p2panda-rs/src/schema/validate/blob.rs
@@ -6,16 +6,6 @@ use regex::Regex;
 use crate::operation::plain::{PlainFields, PlainValue};
 use crate::schema::validate::error::BlobError;
 
-/// @TODO: Do we want a max and what should it be?
-const MAX_BLOB_LENGTH: i64 = 100000; // 1GB
-
-/// Checks "length" field of operations with "blob_v1" schema id.
-///
-/// 1. It must be less than `MAX_BLOB_LENGTH`
-pub fn validate_length(value: &i64) -> bool {
-    *value <= MAX_BLOB_LENGTH
-}
-
 /// Checks "mime_type" field of operations with "blob_v1" schema id.
 ///
 /// 1. It matches expected mime type format
@@ -37,19 +27,7 @@ pub fn validate_mime_type(value: &str) -> bool {
 /// this should be handled by other validation methods. This method is only checking the
 /// special requirements of this particular system schema.
 pub fn validate_blob_v1_fields(fields: &PlainFields) -> Result<(), BlobError> {
-    // Check "length" field
-    let blob_length = fields.get("length");
-
-    match blob_length {
-        Some(PlainValue::Integer(value)) => {
-            if validate_length(value) {
-                Ok(())
-            } else {
-                Err(BlobError::LengthInvalid)
-            }
-        }
-        _ => Ok(()),
-    }?;
+    // `pieces` and `length` fields don't have any special requirements.
 
     // Check "mime_type" field
     let blob_mime_type = fields.get("mime_type");
@@ -65,8 +43,6 @@ pub fn validate_blob_v1_fields(fields: &PlainFields) -> Result<(), BlobError> {
         _ => Ok(()),
     }?;
 
-    // We don't have anything to validate on the pieces field.
-
     Ok(())
 }
 
@@ -77,7 +53,7 @@ mod test {
     use crate::operation::plain::PlainFields;
     use crate::test_utils::fixtures::random_document_view_id;
 
-    use super::{validate_blob_v1_fields, validate_length, validate_mime_type};
+    use super::{validate_blob_v1_fields, validate_mime_type};
 
     #[rstest]
     #[case(vec![
@@ -138,14 +114,5 @@ mod test {
     #[case("thismimetypealsohasonelongelement/one.element.too.long")]
     fn check_mime_type_field(#[case] name_str: &str) {
         assert!(validate_mime_type(name_str));
-    }
-
-    #[rstest]
-    #[case(100)]
-    #[case(100000)]
-    #[should_panic]
-    #[case(100001)]
-    fn check_length_field(#[case] length_int: i64) {
-        assert!(validate_length(&length_int));
     }
 }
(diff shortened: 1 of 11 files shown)


p2panda-rs/src/schema/validate/blob.rs
```diff
@@ -0,0 +1,151 @@
...
+///
+/// 1. It must be less than `MAX_BLOB_LENGTH`
+pub fn validate_length(value: &i64) -> bool {
+    *value <= MAX_BLOB_LENGTH
+}
+
+/// Checks "mime_type" field of operations with "blob_v1" schema id.
+///
+/// 1. It matches expected mime type format
+pub fn validate_mime_type(value: &str) -> bool {
```
adzialocha, line 22: We could use this crate to test against existing mime types: https://docs.rs/mime/latest/mime/

p2panda-rs/src/schema/validate/blob.rs
```diff
@@ -0,0 +1,151 @@
+// SPDX-License-Identifier: AGPL-&3.0-or-later
+
+use once_cell::sync::Lazy;
+use regex::Regex;
+
+use crate::operation::plain::{PlainFields, PlainValue};
+use crate::schema::validate::error::BlobError;
+
+/// @TODO: Do we want a max and what should it be?
+const MAX_BLOB_LENGTH: i64 = 100000; // 1GB
```
adzialocha, line 10: This probably also relates to this restriction here, right? https://github.com/p2panda/aquadoggo/compare/development#diff-42ddd3729325393c2a3a0f2ae68495cfaca656de0ec6b29a171669fbbce26b30R19

If we set the max. pieces number to 10.000 then we gonna have a maximum length of 10.000 * 256 bytes = 2.56 Megabyte.

Probably we want to adjust these values on both ends and make them reflect the same maximum.

p2panda-rs/src/schema/validate/blob_piece.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::operation::plain::{PlainFields, PlainValue};
+use crate::schema::validate::error::BlobPieceError;
+
+const MAX_BLOB_PIECE_LENGTH: usize = 256;
+
+/// Checks "data" field of operations with "blob_piece_v1" schema id.
+///
+/// 1. It must be less than `MAX_BLOB_PIECE_LENGTH`
+pub fn validate_data(value: &String) -> bool {
+    value.len() <= MAX_BLOB_PIECE_LENGTH
```
adzialocha, line 12: Nice, good reminder that `String::len` actually returns the bytes of the string, not the number of characters 😆

p2panda-rs/src/schema/validate/blob_piece.rs
```diff
@@ -0,0 +1,64 @@
...
+        ("data", "aGVsbG8gbXkgbmFtZSBpcyBzYW0=".into()),
+     ].into())]
+    #[should_panic]
+    #[case(vec![
+        ("data", "aGVsbG8gbXkgbmFtZSBpcyBzYW1oZWxsbyBteSBuYW1lIGlzIHNhbWhlbGxvIG15IG5hbW \
+                  UgaXMgc2FtaGVsbG8gbXkgbmFtZSBpcyBzYW1oZWxsbyBteSBuYW1lIGlzIHNhbWhlbGxv \
+                  G15IG5hbWUgaXMgc2FtaGVsbG8gbXkgbmFtZSBpcyBzYW1oZWxsbyBteSBuYW1lIGlzIHN \
+                  hbWhlbGxvIG15IG5hbWUgaXMgc2FtaGVsbG8gbXkgbmFtZSBpcyBzYW1oZWxsbyBteSBuY \
+                  W1lIGlzIHNhbWhlbGxvIG15IG5hbWUgaXMgc2FtaGVsbG8gbXkgbmFtZSBpcyBzYW1oZWx \
+                  sbyBteSBuYW1lIGlzIHNhbWhlbGxvIG15IG5hbWUgaXMgc2FtaGVsbG8gbXkgbmFtZS".into()),
```
adzialocha, line 59: What secret is hidden here .. ? 🕵🏻 😄

p2panda-rs/src/schema/validate/blob.rs
sandreae: I tried this crate for mime type validation, but it didn't catch many invalid mime types (which my regex did catch :rofl: ). It hasn't been updated in 4 years so maybe it's just out of date...

p2panda-rs/src/schema/validate/blob.rs
sandreae: Yeh, `MAX_BLOB_LENGTH` relates to the size of the complete blob file in bytes (as stated in the `length` field), and then on the other end we have `MAX_BLOB_PIECES` which relates to the maximum number of pieces a blob can contain. 

Reflecting on it though, I don't think we need a limit here in `p2panda-rs`. It's something we configure in `aquadoggo`.

p2panda-rs/src/schema/validate/blob_piece.rs
sandreae: I actually can't remember :rofl: it was definitely something....

p2panda-rs/src/schema/validate/blob.rs
adzialocha: Did you try others?

sandreae: Not yet, there's this one which looks more up-to-date https://github.com/picoHz/mediatype

sandreae: Hmmm, tried that one too, it seems to not care too much if something is an existing mime type, so things like this:

`"this/mime.type.hasoneelementwhichiswaytoolong"`

parse into a valid media type.

sandreae: What behaviour are we expecting here? If we want to be strict we could hard code a list of known mime types and validate against this.

adzialocha: Yeah I guess we shouldn't control that, let's stick to the regex 🖖🏻

Issue #506, opened 2023-08-03
Introduce `Bytes` operation value


sandreae:
`OperationValue::Bytes([u8])`

sandreae:
This would be used in the `data` field of `blob_pieces_v1`.

sandreae:
Thinking on this a little and realised we again have to deal with deserializing ambiguity. A `Bytes` value containing an empty vec would look the same as an empty `PinnedRelationList`. This is why `PlainOperation` has the `AmbiguousRelation` type which can only be typed when a schema is present.

sandreae:
We can expand `AmbiguousRelation` to include bytes, renaming it to something like `AmbiguousVec`, or an alternative approach I thought of would be to specify that `Bytes` can't be empty... don't know if that's a strange solution though.

adzialocha:
I would suspect CBOR encoding byte sequences differently than arrays (or almost anything else, really), but probably that doesn't work if your byte sequence is exactly a _CBOR encoded_ array? I assume you stumbled upon this because you used the cbor macros?

If that's not the case it's fine to go down the ambiguous route, we can't deserialize operations without a schema and maybe byte types are part of this.

adzialocha:
I would suspect CBOR encoding byte sequences differently than arrays (or almost anything else, really), but probably that doesn't work if your byte sequence is exactly a _CBOR encoded_ array? I assume you stumbled upon this because you used the cbor macros?

If that's not the case it's fine to go down the ambiguous route, we can't deserialize operations without a schema and maybe byte types are part of this.

sandreae:
Yeh, i tried with both `Vec<u8>` and `ByteBuf` from the `serde_bytes` crate (thin wrapper around the former which apparently encourages optimised byte encoding) and both got confused when serialising and then deserializing plain operations. Feels like there should be a way for bytes to be treaded differently from other arrays (even when empty), maybe I've just not found the trick yet.

sandreae:
Update: when i specify that the vec should be treated as bytes (through serde macros) then `String` values are interpreted as `Bytes`. Still wrong, but in a different way. Progress? :rofl:

sandreae:
Ok, so my suggestion is to replace the `StringOrRelation(String)` enum variant with `ByteString(&[u8])` which will be encoded as a cbor byte string value. As I understand it, cbor encodes strings and bytes in exactly the same way. 

[...]

PR #513, opened 2023-09-01
`Bytes` operation value

sandreae:
Introduces a new operation value for defining a field which contains a `ByteBuf` array.

- add `Bytes` to `OperationValue` and `PlainValue`
- implement custom `Deserialize` trait for `PlainValue`
- update wasm modules
- update `p2panda-js`, use `Uint8Array` to represent byte fields 

closes: #506 

## 📋 Checklist


p2panda-rs/src/operation/plain/plain_fields.rs
```diff
@@ -191,7 +196,7 @@ mod tests {
             ("it_works", PlainValue::Boolean(true)),
             (
                 "it_works",
-                PlainValue::StringOrRelation("... and ignores duplicates".into()),
+                PlainValue::ByteString(ByteBuf::from("... and ignores duplicates")),
```
adzialocha, line 199: I wonder if the name isn't confusing, I read it as "it is a utf8 string in its byte representation" which is half true? Maybe just calling it "Bytes" as well or "AmbiguousBytes"? 🥴

sandreae:
Finally ready for review @adzialocha :pray:

p2panda-rs/src/operation/encode.rs
```diff
@@ -60,6 +61,9 @@ mod tests {
         let from_operation = encode_operation(&operation).unwrap();
         let from_plain_operation = encode_plain_operation(&plain_operation).unwrap();
 
+        println!("{from_operation}");
+        println!("{from_plain_operation}");
```
adzialocha, line 65: Remove println

p2panda-rs/src/operation/error.rs
```diff
@@ -2,6 +2,7 @@
 
 //! Error types for encoding, decoding and validating operations with schemas and regarding data
 //! types like operation fields, relations or plain operations.
+
```
adzialocha, line 5: Without is actually right 😅

p2panda-rs/src/operation/plain/plain_value.rs
```diff
@@ -1,8 +1,15 @@
...
 
+use std::convert::TryInto;
+
+use ciborium::Value;
 use serde::{Deserialize, Serialize};
+use serde_bytes::ByteBuf;
 
-use crate::document::{DocumentId, DocumentViewId};
+use crate::{
+    document::{DocumentId, DocumentViewId},
```
adzialocha, line 10: Nesting

p2panda-rs/src/wasm/serde.rs
```diff
@@ -14,7 +16,13 @@ use wasm_bindgen::JsValue;
...
 /// BigInt instances.
 pub fn serialize_to_js<T: Serialize + ?Sized>(value: &T) -> Result<JsValue, Error> {
-    let serializer = Serializer::new().serialize_large_number_types_as_bigints(true);
+    let serializer = Serializer::new()
+        .serialize_large_number_types_as_bigints(true)
+        // @TODO: It would be better to serialize into Uint8Array but I couldn't get this to
+        // work during deserialization.
+        //
+        // https://github.com/p2panda/p2panda/issues/516
+        .serialize_bytes_as_arrays(false);
```
adzialocha, line 25: This (`false`) value should serialize to Uint8Array I think?

p2panda-rs/Cargo.toml
```diff
@@ -47,6 +47,7 @@ rstest = { version = "0.16.0", optional = true }
 rstest_reuse = { version = "0.5.0", optional = true }
 serde = { version = "1.0.152", features = ["derive"] }
 serde_bytes = "0.11.12"
+serde_with = "3.3.0"
```
adzialocha, line 50: Is this dependency used anywhere?

p2panda-rs/src/operation/plain/plain_value.rs
```diff
@@ -24,7 +31,10 @@ pub enum PlainValue {
     Float(f64),
 
     /// String value which can be either a text or relation (document id).
-    StringOrRelation(String),
+    String(String),
```
adzialocha, line 34: Isn't this still sometimes a relation?

p2panda-rs/src/operation/plain/plain_value.rs
```diff
@@ -124,10 +160,63 @@ impl From<Vec<DocumentViewId>> for PlainValue {
...
+        Value::Bool(bool) => Ok(bool.into()),
+        Value::Array(array) => cbor_array_to_plain_value_list(array),
+        _ => return Err(PlainValueError::UnsupportedValue),
+    };
+
+    result
+}
+
+/// Helper for converting a cbor array into a plain operation list value.
+fn cbor_array_to_plain_value_list(array: Vec<Value>) -> Result<PlainValue, PlainValueError> {
```
adzialocha, line 182: Comments could be nice here, this code is a bit low level and abstract.

adzialocha (changes requested):

adzialocha (changes requested):

p2panda-rs/src/operation/operation_value.rs
```diff
@@ -9,6 +11,9 @@ pub enum OperationValue {
     /// Boolean value.
     Boolean(bool),
 
+    /// Bytes value.
+    Bytes(ByteBuf),
```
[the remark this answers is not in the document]
sandreae: Yeh, this would work! I had it as a `Vec<u8>` before (for the reason you mention) but I wanted bytes to be `Uint8Array` everywhere in `p2panda-js` which meant this value being a ByteBuf. But with that macro we can have both :+1:

p2panda-rs/src/operation/plain/plain_value.rs
sandreae: Yeh, it is, but I thought it was ok to just call it `String` as this is the concrete CBOR type it represents.

adzialocha: I see! But if we would go by that naming scheme then we would need to also rename ambiguous relation and relation lists because these are no cbor types? I'm fine with either, just consistent

p2panda-rs/src/operation/plain/plain_value.rs
```diff
@@ -124,10 +164,71 @@ impl From<Vec<DocumentViewId>> for PlainValue {
...
+        .iter()
+        .map(|value| match value.as_text() {
+            Some(text) => Ok(text.to_string()),
+            None => Err(PlainValueError::UnsupportedValue),
+        })
+        .collect();
+
+    // If this was successful we stop here and return the value.
+    if let Ok(strings) = ambiguous_relation {
+        return Ok(strings.into());
```
adzialocha, line 202: Great, this is much nicer! Thank you!!

A little detail, but how about writing tgis out to `AmbigiousRelation:: From`, then it gets very explicit what we're dealing with ✊🏻

p2panda-rs/src/operation/plain/plain_value.rs
sandreae: Cool, then I change it back cos I don't wanna change the others everywhere...

sandreae:
> > Check if descriptions and terminology match handbook content (and visa-versa)
> 
> Do we need a handbook update?

I raised an issue for now :+1: https://github.com/p2panda/handbook/issues/292

Issue #518, opened 2023-09-05
Serialize `Hash` derived ids as bytes


sandreae:
For `EncodedOperation` and `EncodedEntry` we `serialize_hex` and `deserialize_hex` methods to conditionally de/serialize into human readable hex or (more compact) machine readable bytes. However this also needs to be implemented on `Hash` in order to make the most use of this functionality when  serializing id values in operations and entries. 

Right now when we serialize an operation we get the following 662 bytes of CBOR. We can see that the id values have been encoded into bytes (displayed as hex string) from there already hex encoded strings.

```
84                                      # array(4)
   01                                   # unsigned(1)
   00                                   # unsigned(0)
   78 4A                                # text(74)
      76656E75655F3030323063363535363761653337656665613239336533346139633764313366386632626632336462646333623563376239616234363239333131316334386663373862 # "venue_0020c65567ae37efea293e34a9c7d13f8f2bf23dbdc3b5c7b9ab46293111c48fc78b"
   A8                                   # map(8)
      63                                # text(3)
         616765                         # "age"
      18 1C                             # unsigned(28)
      68                                # text(8)
         636F6D6D656E7473               # "comments"
      82                                # array(2)
         81                             # array(1)
            78 44                       # text(68)
               3030323034663064643361316238323035623664346365336664346331353862623931633965313331626438343265373237313634656132323062356236643039333436 # "00204f0dd3a1b8205b6d4ce3fd4c158bb91c9e131bd842e727164ea220b5b6d09346"
         82                             # array(2)
            78 44                       # text(68)
               3030323031396564336539623339636431376631646263306636653331613665376239633961623765333439333332653731306339343661343431623764333038656235 # "002019ed3e9b39cd17f1dbc0f6e31a6e7b9c9ab7e349332e710c946a441b7d308eb5"
            78 44                       # text(68)
               3030323039393564353366343630323933633536383663343230333762373237383765643238363638616438623664313865396435663032633564333330313136316630 # "0020995d53f460293c5686c42037b72787ed28668ad8b6d18e9d5f02c5d3301161f0"
      66                                # text(6)
         686569676874                   # "height"
      F9 4300                           # primitive(17152)
      68                                # text(8)
         69735F61646D696E               # "is_admin"
      F4                                # primitive(20)
      6A                                # text(10)
         6D795F667269656E6473           # "my_friends"
      81                                # array(1)
         78 44                          # text(68)
[...]

PR #525, opened 2023-09-17
Serialize `Hash` from bytes

sandreae:
closes: #518 

- impl `Serialize` for `Hash` which serializes hex encoded bytes for human readable formats and raw hash bytes for machine formats
- impl `Deserialize` for `Hash` which deserializes for both above mentioned cases
- derive `Serialize` and `Deserialize` for all hash based types (operation id, document id, etc...) 
- All `PlainValue` variants which represent hash based ids are represented by `Hash`
- `StringOrRelation` is swapped for `BytesOrRelation`

## 📋 Checklist


p2panda-rs/src/document/document_id.rs
```diff
@@ -76,23 +83,6 @@ impl FromStr for DocumentId {
     }
 }
 
-impl<'de> Deserialize<'de> for DocumentId {
```
adzialocha, line 79: Oh nice, this works because deriving Deserialize calls our implementation on Hash which validates everything (hex encoding, length, multihash format), right?

p2panda-rs/src/serde/hex_str.rs
```diff
@@ -5,7 +5,7 @@ use serde_bytes::{ByteBuf as SerdeByteBuf, Bytes as SerdeBytes};
 
 /// Helper method for `serde` to serialize bytes into a hex string when using a human readable
 /// encoding (JSON, GraphQL), otherwise it serializes the bytes directly (CBOR).
-pub fn serialize_hex<S>(value: &[u8], serializer: S) -> Result<S::Ok, S::Error>
+pub fn serialize_hex_bytes<S>(value: &[u8], serializer: S) -> Result<S::Ok, S::Error>
```
adzialocha, line 8: Should we rename the deserialize method as well?

p2panda-rs/src/operation/operation_id.rs
```diff
@@ -146,24 +132,24 @@ mod tests {
             )
             .unwrap(),
         );
+        println!("{}", hex::encode(&bytes));
```
adzialocha, line 135: 🍃

adzialocha (changes requested):
Coooooooool ✊🏻✨

Changelog is missing and I have some understanding questions but otherwise happy to merge!

p2panda-rs/src/serde/hex_str.rs
sandreae: Ah, I actually reverted this name change, thought I already pushed it but obviously not. Earlier I'd added a new util method and the names needed distinction. But I removed that and so this naming change isn't required anymore.

p2panda-rs/src/hash/hash.rs
```diff
@@ -113,17 +115,36 @@ impl Human for Hash {
...
+        };
+
+        let hash_str = hex::encode(hash_bytes);
 
         // Convert and validate format
-        hash.try_into()
-            .map_err(|err: HashError| serde::de::Error::custom(err.to_string()))
+        Hash::new(&hash_str).map_err(|err: HashError| serde::de::Error::custom(err.to_string()))
     }
 }
```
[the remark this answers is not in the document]
sandreae: Ah yeh, we can! Nice.

p2panda-rs/src/hash/traits.rs
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+//! Interfaces for interactions for hash-like structs.
+use crate::hash::Hash;
+
+/// Trait implemented on types which are derived from a hash.
+pub trait HashId {
```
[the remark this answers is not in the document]
sandreae: Took some head scratching to get there :rofl:

p2panda-rs/src/document/document_id.rs
sandreae: Yeh, exactly! It's very neat :+1:

p2panda-rs/src/hash/hash.rs
```diff
@@ -113,17 +116,30 @@ impl Human for Hash {
...
+    where
+        S: serde::Serializer,
+    {
+        if serializer.is_human_readable() {
+            self.as_str().serialize(serializer)
+        } else {
+            ByteBuf::from(self.to_bytes()).serialize(serializer)
+        }
+    }
+}
```
adzialocha, line 130: Here we could maybe also use the utility method or even serde with attribute?

sandreae: You mean the existing util or a new one? The existing one doesn't do what we want, and I thought maybe there wasn't need to create a new util for this one case. Don't feel very strongly about it though.

sandreae: The reason I didn't just make `Hash` contain bytes (which was another valid approach) is that all our APIs around `Hash`, `OperationId`, `DocumentId` etc.. assume strings are the core data type. And I think this is actually nice and fine to keep. Serializing now takes account of the knowledge that this `String` is actually a hex representation of bytes.

adzialocha: > The reason I didn't just make `Hash` contain bytes (which was another valid approach) is that all our APIs around `Hash`, `OperationId`, `DocumentId` etc.. assume strings are the core data type. And I think this is actually nice and fine to keep. 
> 
> The change is that Serializing now takes account of the knowledge that this `String` is actually a hex representation of bytes.

Ah, that was my thought, thanks for clarifying. Hmn, I'm fine with leaving it as is for now but wonder if we want to keep it like that. Guess it's performance optimization related and too much guessing at this stage.

sandreae: Yeh, have a look. I think it means a lot of changes to our current APIs, which I thought was worth avoiding. But happy for a second opinion.

adzialocha (approved):

PR #527, opened 2023-09-25
Optimise document building

sandreae:
Introduce `Reducer` trait and `DocumentReducer` struct which is an accumulator passed into the graph walking method. It constructs the document according to it's implementation of the `combine` method.

Makes our document building logic more efficient and opens the door to custom `Reducer`s in the future :fist_raised: 

## 📋 Checklist


p2panda-rs/src/graph/traits.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+/// Trait to be defined on a reducer which can be passed into `Graph` in order to perform an
+/// operation on all visited nodes.
+///
+/// It has one method `combine` which takes a generic value and mutable self.
+pub trait Reducer<V> {
+    /// Takes a generic value and presumably combines it with some contained state.
+    fn combine(&mut self, value: &V);
```
adzialocha, line 9: I'd like to explore if this should be able to return exceptions, maybe we make an issue for this?

adzialocha (approved):

p2panda-rs/src/graph/traits.rs
sandreae: Ah yeh, could actually just implement that in this PR. Sure there are cases where an error is never expected, but that's no problem. The current pattern when reducing a document and checking the reducer for errors at the end is a bit weird.

sandreae: How's that look now, with `combine` returning an error?

adzialocha: Looking super :+1:

Issue #528, opened 2023-11-18
Deserialization of `DocumentViewId` hex strings broken: `invalid type: string, expected byte buffer`


adzialocha:
I could reproduce the behaviour in this branch: https://github.com/p2panda/p2panda/pull/529

adzialocha:
Ah, after some digging it looks like it so far concerns only `DocumentViewId`

sandreae:
Just had a quick look. Shouldn't the test value for `deserialize_human_readable` be an array of hex encoded hashes? Like so:

```rust
let json = format!(
    r#"
    {{
        "document_view_id": ["{hash_str}"]
    }}
"#
);
```

adzialocha:
> Just had a quick look. Shouldn't the test value for `deserialize_human_readable` be an array of hex encoded hashes? Like so:
> 
> ```rust
> let json = format!(
>     r#"
>     {{
>         "document_view_id": ["{hash_str}"]
>     }}
> "#
> );
> ```

There's no JSON we have like that (GraphQL responses etc.), only in this form: `<id>_<id>`. I had a go at a new `Visitor` which fixes the problem for view ids.

sandreae:
> > Just had a quick look. Shouldn't the test value for `deserialize_human_readable` be an array of hex encoded hashes? Like so:
> > ```rust
> > let json = format!(
> >     r#"
> >     {{
> >         "document_view_id": ["{hash_str}"]
> >     }}
> > "#
> > );
> > ```
> 
> There's no JSON we have like that (GraphQL responses etc.), only in this form: `"<id>_<id>"`. I had a go at a new `Visitor` which fixes the problem for view ids, can push later.

Ah yeh, of course!

adzialocha:
> > > Just had a quick look. Shouldn't the test value for `deserialize_human_readable` be an array of hex encoded hashes? Like so:
> > > ```rust
> > > let json = format!(
> > >     r#"
> > >     {{
> > >         "document_view_id": ["{hash_str}"]
> > >     }}
> > > "#
> > > );
> > > ```
> > 
> > There's no JSON we have like that (GraphQL responses etc.), only in this form: `"<id>_<id>"`. I had a go at a new `Visitor` which fixes the problem for view ids, can push later.
> 
> Ah yeh, of course!

Still a good answer to why it fails 🤣: arrays are not strings (at least not on that level)!

PR #529, opened 2023-11-18
Fix de-/serialization for human-readable encodings

adzialocha:
This PR fixes two issues concerning deserialization and serialization for human-readable encodings (for example JSON):

1. Deserialization of `PlainOperation` was not possible when coming from a human-readable format. That's an edge case as operations should _never_ be encoded in anything else than CBOR (as per specification), but we have exceptions when it is a nice user-interface, for example in `send-to-node`
2. `DocumentViewId` can be represented as a string in human-facing scenarios, but we didn't account for that during deserialization

Closes: https://github.com/p2panda/p2panda/issues/528

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,17 +9,13 @@ Highlights are marked with a pancake 🥞
 
 ## [Unreleased]
 
-### Added
-
-### Changed
-
 ### Fixed
+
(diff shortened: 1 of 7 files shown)


sandreae (approved):
This was a real head scratcher, well done figuring it out 🎉🙏

PR #531, opened 2023-12-03
WIP: Add C API and generate gobject introspection

lw64:


lw64:
You can build this with `cargo cinstall --prefix=path/to/prefix`. Then you can run `g-ir-scanner --header-only --warn-all -n P2panda --identifier-prefix p2panda_ -o P2panda-1.0.gir path/to/prefix/include/p2panda_rs/p2panda_rs.h` to generate the gir file.

adzialocha:
This work is pretty amazing and definitely something we'll do (considering that we're about to start development on our first Gtk application).

Still, after talking briefly to @lw64, learning that there's a better approach to bringing in GObject plus heavily re-modularizing our stack (see `v2` branch for a sneak-peek, we're preparing the first official release towards end November) I'll close this PR now. Let's look at this freshly soon, it will even be easier now with the core data types being nicely separated.

PR #532, opened 2023-12-19
Remove Bamboo entries and YASMF hash

adzialocha:
Up until now the core data-type used by p2panda has been the [Bamboo](https://github.com/AljoschaMeyer/bamboo) append only log. This gave us signed data, cryptographic security and a simple replication mechanism. Each bamboo entry contains a payload which can be deleted, for us this is an `Operation`, which is what we use to carry application data and group it into logical collections (by multiple authors). Although we still want to keep many of the properties bamboo gives us, there have been a few pain points which have lead us to the decision to migrate to something new. Building on the shoulders of giants of course :pray: a summary of these reasons are: 

* our in-built conflict resolution mechanism is based on a DAG where the nodes are operations. We want to be able to delete application data (the payloads) while retaining this DAG (for system concerns like capabilities), to do this we need to move the some fields from the operation to the entry (in particular `previous`)
* we don't expect to use multiple hashing functions in p2panda
* we don't make use of skiplinks for several reasons:
  *  data an author publishes is split between many logs which we expect to remain quite small except for in special cases, and so partial replication of individual logs becomes less important
  * in cases where you want to keep all history of a document you want to keep all operations
  * for cases where logs are expected to get very long, we have an alternate approach to sparse replication in our new data model, where we optionally allow only replicating the latest operations in a log (old operations can be deleted). This is only done when a complete cryptographically verifiable history is not required
    *  skiplinks would have retained this history, but in the cases we're imagining this is an acceptable trade-off

All of this is not a criticism of bamboo, which is designed for a specific purpose and does it very well, rather how we ended up using it doesn't play to it's strengths, and something custom will allow us to keep the properties we really need while adding required p2panda specific elements, simplifying our specification and code base.

## Remove `Entry`

Replacing it with new types `Header` and `Body` (see below).

## Remove `YASMF` multi-hash

Replacing it with `BLAKE3` hashing function.

## Introduce operation `Header` and `Body`

The `Header` contains _always_ the following fields:
* `version`: `u64`
  * Version of the operation format, currently `1`
* `signature`: `Signature`
  * Ed25519 signature over the header
* `public_key`: `PublicKey`
  * Ed25519 public key of the author of this operation
* `payload_hash`: `Hash`
  * BLAKE3 hash of the body of this operation, can be omitted if no body is given
* `payload_size`: `u64`
  * Number of bytes of the body of this operation, can be omitted if no body is given
* `seq_num`: `u64`
  * Number of operations this author has published to this document
  * Begins with `0` and is always incremented by `1` with each new operation by the same author
* `backlink`: `Hash`
  * only required for operations where `seq_num` > 0
  * Hash of the previous operation of the same author
* `timestamp`: `u64`
  * A UNIX timestamp
  * If not the first operation in the document it needs to be larger or equal than the timestamp the operation points at via it's `backlink` or `previous` link (see below)
* `schema_id`: `Hash`
  * id of the schema document which describes the expected shape of this operations body

The following fields are only required when appending operations to an existing document.

* `document_id`: `Hash`
    * Hash of the CREATE operation aka document Id we refer to with this update
    * This is how we know which document an operation is part of
* `previous`: `Hash[]`
    * An optional list of hashes of the operations we want to refer to as the "previous" ones, these are operations from other authors
* `tombstone`: `true`
  * optional field for signifying a document should be deleted 

The `Body` contains only the `OperationFields` which are optional and can itself be deleted (same as payload on a bamboo entry):
* in the future we want to have body only contain arbitrary bytes, we may remove operation fields completely from this crate
* TODO: `Body` needs a little more thought as it has now become only a wrapper around the optional fields. As body itself is an optional attachment to the header, we could make fields no longer optional and then omit the body entirely for `DELETE` operations, and indeed for operations where the body has been deleted. I'd be happy to leave it in a slightly strange state for this PR though.

## Progress

TODO

## Next Steps

TODO

## 📋 Checklist


adzialocha:
Small thought on the change to make timestamp mandatory:

Could be actually nice to still not have `timestamp` as an argument on the Operation Builder constructor, and set it by default to the current time if not explicitly set.

sandreae:
> Small thought on the change to make timestamp mandatory:
> 
> Could be actually nice to still not have `timestamp` as an argument on the `OperationBuilder` constructor, and set it by default to the current time if not explicitly set.
> 
> @sandreae

Yeh, I like this suggestion. 

Not totally trivial though. We can do something like this to get the current UNIX timestamp in operation builder constructor: 

```
let timestamp: u64 = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
```

However `SystemTime::now()` doesn't guarantee monotonically increasing values, so relying on this to construct operations programmatically will sometimes result in a earlier timestamp.

https://doc.rust-lang.org/std/time/struct.SystemTime.html

This might be unavoidable though, you need some kind of app state to be able to guarantee incrementing timestamps are provided. I guess the question is whether it's a big enough foot-gun that we want to keep it mandatory + explicit over having a cleaner api.

adzialocha:
> This might be unavoidable though, you need some kind of app state to be able to guarantee incrementing timestamps are provided. I guess the question is whether it's a big enough foot-gun that we want to keep it mandatory + explicit over having a cleaner api.

Right, I didn't think that thru. Thanks for clarifying! Looking at it now, I think we leave this as-is, it smells like this is a higher level concern which needs state and we don't want to have that here.

sandreae:
> > This might be unavoidable though, you need some kind of app state to be able to guarantee incrementing timestamps are provided. I guess the question is whether it's a big enough foot-gun that we want to keep it mandatory + explicit over having a cleaner api.
> 
> Right, I didn't think that thru. Thanks for clarifying! Looking at it now, I think we leave this as-is, it smells like this is a higher level concern which needs state and we don't want to have that here.

I did implement it, makes the api very nice of course, but agree that we should leave it to a higher level. I'll revert it back.

sandreae:
@adzialocha our current spec for header has moved away from using extensions for any of the fields we need for this PR. In that case I suggest removing `HeaderExtension` entirely from this PR and making `Header` look just as it does in the PR description now.

sandreae:
> @adzialocha our current spec for header has moved away from using extensions for any of the fields we need for this PR. In that case I suggest removing `HeaderExtension` entirely from this PR and making `Header` look just as it does in the PR description now.

Ah, that's not quite right actually. `schema_id` is still in the extensions in the new specification. I suggest adding it to the header directly for now.

adzialocha:
> > @adzialocha our current spec for header has moved away from using extensions for any of the fields we need for this PR. In that case I suggest removing `HeaderExtension` entirely from this PR and making `Header` look just as it does in the PR description now.
> 
> Ah, that's not quite right actually. `schema_id` is still in the extensions in the new specification. I suggest adding it to the header directly for now (description updated to reflect this).

We gonna have more extensions at one point, maybe we can think of a trait to express them?

sandreae:
> > > @adzialocha our current spec for header has moved away from using extensions for any of the fields we need for this PR. In that case I suggest removing `HeaderExtension` entirely from this PR and making `Header` look just as it does in the PR description now.
> > 
> > 
> > Ah, that's not quite right actually. `schema_id` is still in the extensions in the new specification. I suggest adding it to the header directly for now (description updated to reflect this).
> 
> We gonna have more extensions at one point, maybe we can think of a trait to express them?

You mean to express the yet-to-be-add extensions? Like capabilities and secret groups?

PR #535, opened 2024-07-04
Core types

adzialocha:
Introducing p2panda core types based on the new [namakemono specification](https://p2panda.org/specifications/namakemono/). 🦥 

* BLAKE3 `Hash`
* Ed25519 `PrivateKey`, `PublicKey` and `Signature`
* p2panda `Operation`, `Header`, `Body` and some validation methods
* CBOR based encoding with `serde` and `ciborium`

## 📋 Checklist


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
 resolver = "2"
-members = [
-  "p2panda-core",
-]
+members = ["p2panda-core", "p2panda-core/fuzz"]
```
adzialocha, line 3: Should we move the `fuzz` folder into the root of this repository (potentially holding more fuzz tests for other crates?)

p2panda-core/src/serde.rs
```diff
@@ -0,0 +1,603 @@
...
+    E: Clone + Serialize + DeserializeOwned,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut seq = serializer.serialize_seq(Some(2))?;
+
+        seq.serialize_element(&self.0)?;
+        seq.serialize_element(&self.1)?;
```
adzialocha, line 288: I'm afraid that this changes the operation format now with the signature being parsed first instead of the version.

I think it doesn't matter too much _where_ the signature is parsed, probably it's good early on as it's a mandatory field, but the version should always be first to make sure we support backwards compatibility. The parser should be able to adjust itself right from the beginning, according to the operation version.

A solution could be here to combine the serializer methods of `Header` and `UnsignedHeader` with common logic and something along the lines of `fn serialize_header<S: Blubs>(serializer: S, signature: Option<Signature>)`.

sandreae: Oh damn, yeh of course. I'll have a go at that now.

p2panda-core/src/operation.rs
```diff
@@ -0,0 +1,440 @@
...
+where
+    E: Serialize + DeserializeOwned,
+{
+    /// Operation format version, allowing backwards compatibility when specification changes.
+    pub version: u64,
+
+    /// Author of this operation.
+    pub public_key: PublicKey,
+
+    /// Number of bytes of the body of this operation, can be omitted if no body is given.
```
sandreae, line 61: Should read `must be zero if no body is given`

p2panda-core/src/operation.rs
```diff
@@ -0,0 +1,440 @@
...
+    pub payload_size: u64,
+
+    /// Hash of the body of this operation, can be omitted if no body is given.
```
sandreae, line 64: Should read `...must be omitted if payload_size is zero`

p2panda-core/src/operation.rs
```diff
@@ -0,0 +1,440 @@
...
+    pub fn sign(&self, private_key: &PrivateKey) -> Header<E> {
+        let bytes = self.to_bytes();
+        let signature = private_key.sign(&bytes);
+        Header(signature, self.clone())
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+pub struct Header<E>(pub Signature, pub UnsignedHeader<E>)
```
sandreae, line 114: Needs doc strings.

p2panda-core/src/serde.rs
adzialocha: Sounds great! Yeah, `Deserialize` on `UnsignedHeader` is not necessary, makes it easier now ✌️

mycognosist (approved):

PR #536, opened 2024-07-09
`p2panda-store`: basic storage traits plus `MemoryStore` implementation

adzialocha:
Storage traits outlining basic API for inserting, getting and deleting operations and their payloads. This is a low-level API intended to be used when validating and persisting operations which arrive on a node, retrieving all operations from a log during replication, or pruning operations and/or their payloads.

- [x] `OperationStore` trait describing API for handling individual operations
- [x] `LogStore` trait describing API for performing actions involving logs 
- [x] `MemoryStore` implementation of above traits   

## 📋 Checklist

### DIFF
--- a/p2panda-store/Cargo.toml
+++ b/p2panda-store/Cargo.toml
@@ -7,3 +7,8 @@ edition = "2021"
 workspace = true
 
 [dependencies]
+serde = { version = "1.0.203" }
+thiserror = "1.0.61"
+
+[dependencies.p2panda-core]
+path = "../p2panda-core"
--- a/p2panda-store/Cargo.toml
+++ b/p2panda-store/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "p2panda-store"
+version = "0.1.0"
+edition = "2021"
+
+[lints]
+workspace = true
+
+[dependencies]
--- a/p2panda-store/Cargo.toml
+++ b/p2panda-store/Cargo.toml
@@ -7,8 +7,10 @@ edition = "2021"
 workspace = true
 
 [dependencies]
-serde = { version = "1.0.203" }
 thiserror = "1.0.61"
+
+[dev-dependencies]
+serde = { version = "1.0.203" }
 
 [dependencies.p2panda-core]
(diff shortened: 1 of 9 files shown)


p2panda-store/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "p2panda-store"
+version = "0.1.0"
+edition = "2021"
+
+[lints]
+workspace = true
+
+[dependencies]
+serde = { version = "1.0.203" }
```
adzialocha, line 10: Can we do this without serde? And if not, can we put it behind a feature flag for the in-memory variant?

p2panda-store/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+
+pub mod memory_store;
+pub mod traits;
+
+pub use memory_store::MemoryStore;
+pub use traits::{LogStore, OperationStore, StoreError};
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, PartialEq, Eq, std::hash::Hash, Deserialize, Serialize, Default)]
```
mycognosist, line 11: Could you explain why you've included this `LogId` `struct`? Is it strictly necessary?

mycognosist (approved):
Looks good to me. I just had the one question about the `LogId` tuple struct.

When we looked at the code previously you had a `get_logs()` method. Am I remembering that correctly? Did you choose to leave that out for the time being?

p2panda-store/Cargo.toml
sandreae: Ah, good spot, thanks :pray: this is only needed in tests, so I've moved it to `dev-dependencies` where it should have been.

p2panda-store/src/lib.rs
sandreae: It's actually an artefact left over from a previous design where the log id was required to be present in the operation extensions (`Extension<LogId>` was a required trait) but now I removed that requirement by adding `log_id` as an additional argument to `insert_operation()`.

TLDR; it isn't required, I deleted it :wink:

PR #537, opened 2024-08-06
Fix cbor deserialization for correctly ordered `Header` and `Body`

mycognosist:
```rust
#[derive(Debug, PartialEq, Serialize, Deserialize)]
struct Message {
    header: Header<()>,
    body: Body,
}
```

Previously, a `struct` would fail to deserialize from CBOR-encoded bytes if `Header` was the first field (such as in `Message`, seen above). The failure would produce the following error:

```
`Err` value: Semantic(None, "invalid type: break, expected str or bytes")` 
```

Strangely, the deserialization attempt would not fail if `Header` was the second field (e.g. appearing beneath / after `Body` in the above `Message` example).

During investigation it was noted that explicitly specifying the number of elements in the header during serialization resulted in expected behaviour (as opposed to specifying `None`).

This PR introduces a `len()` method on `Header` which counts the number of fields, excluding any `None` values. The method is then used to explicitly specify the number of elements when serializing a header. When the number of elements is explicitly specified, the encoded byte value is one byte shorter; an `FF` byte is not included.

You'll notice that the `serde_header_with_other_types()` test now passes (it was failing before) but the `fixtures()` test fails (because the encoded byte value for each test encoding differs by a single byte). The fixtures will need to be corrected.

## 📋 Checklist


p2panda-core/src/operation.rs
```diff
@@ -128,6 +128,32 @@ where
     pub fn hash(&self) -> Hash {
         Hash::new(self.to_bytes())
     }
+
+    /// Number of fields included in the header.
+    ///
+    /// Fields instantiated with `None` values are excluded from the count.
+    pub fn len(&self) -> usize {
```
mycognosist, line 135: Assuming we go with this change: is `len()` the best name?

Could also be `count_elements()` or similar.

sandreae (approved):
All looks good to me, thanks for the fix :pray:

p2panda-core/src/operation.rs
```diff
@@ -128,6 +128,32 @@ where
     pub fn hash(&self) -> Hash {
         Hash::new(self.to_bytes())
     }
+
+    /// Number of fields included in the header.
+    ///
+    /// Fields instantiated with `None` values are excluded from the count.
+    pub(crate) fn len(&self) -> usize {
```
sandreae, line 135: Ah, sorry, missed this. I thought we decided to change this to `field_count()`, I think that works well.

sandreae (changes requested):
I missed something actually :smile:

p2panda-core/src/operation.rs
mycognosist: All sorted.

PR #540, opened 2024-08-13
Networking

mycognosist:
- [x] Introduce v2 network crate to the workspace
- [x] Introduce gossip overlay connection test
- [x] Document methods and types in `network.rs`
  - [x] `NetworkBuilder`
  - [x] `Network`

Doc comments need a lot more work (describing the API and offering usage recommendations etc.) but I think it's a good foundation for future refinement.

## 📋 Checklist


p2panda-net/src/network.rs
```diff
@@ -0,0 +1,647 @@
...
+}
+
+/// Creates an overlay network for peers grouped under the same network identifier.
+///
+/// All peers can subscribe to multiple topics in this overlay and hook into a data stream per
+/// topic where they'll send and receive data.
+#[derive(Debug)]
+pub struct NetworkBuilder {
+    bind_port: Option<u16>,
+    direct_node_addresses: Vec<NodeAddr>,
```
sandreae, line 52: We talked about renaming this field (in many other places as well) to something like `peer_addresses` or `known_peer_addresses` or `direct_peer_addresses` to avoid confusion with _our own_ direct addresses. Doesn't need changing here in this PR, leaving a comment so we can consider it further though.

sandreae (approved):
Looks good to me! Considering we understand this code to still be in an experimental stage and there is a dedicated design phase on the horizon I don't have so many comments. As a faithful `p2panda-net` user I can say I'm already enjoying the current functionality :wink: 

Thanks @adzialocha for all your work on this and @mycognosist for tracking down and fixing a couple of tricky bugs and making doc improvements already.

sandreae (changes requested):
Oh yeh, just needs a `CHANGELOG` entry.

sandreae (approved):

adzialocha (approved):

PR #550, opened 2024-09-03
Make all storage traits async (and use interior mutability)

sandreae:
- changes all store methods to be `async fn`
- moves internal store data types into an `InnerStore` which is then wrapped in `Arc<RwLock<InnerStore>>` to enable shared read and write access to the store.

BREAKING: this is a breaking change anywhere where we currently need the store would need updating.

## 📋 Checklist

### DIFF
--- a/p2panda-store/Cargo.toml
+++ b/p2panda-store/Cargo.toml
@@ -11,8 +11,8 @@ thiserror = "1.0.61"
 trait-variant = "0.1.2"
 
 [dev-dependencies]
-serde = { version = "1.0.203" }
-tokio = { version = "1.38.0", features = ["full"] }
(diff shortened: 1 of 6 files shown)


p2panda-store/Cargo.toml
```diff
@@ -8,9 +8,11 @@ workspace = true
 
 [dependencies]
 thiserror = "1.0.61"
+trait-variant = "0.1.2"
 
 [dev-dependencies]
 serde = { version = "1.0.203" }
+tokio = { version = "1.38.0", features = ["full"] }
```
adzialocha, line 15: Full might lead to longer compile times for unused functionality during testing

adzialocha (approved):

PR #555, opened 2024-09-16
Stream "engine" for validation, decoding, out-of-order buffering and other extensions

adzialocha:
First steps towards a p2panda "engine" with basic methods implementing the p2panda specification. This gives developers an easy "plug and play" streaming pipeline they can use in their stacks.

For now the engine provides two methods: `decode` and `ingest`.

`decode` simply decodes any bytes to p2panda operations while `ingest` checks operations against their log-integrity, handles out-of-order messages in an internal buffer (with a maximum of attempts), prune flags and persistance.

With our nice `Extension` traits the prune flag implementation is up to the user of this API. Also cool: everything is runtime-agnostic!


adzialocha:
I'd wish for a better error handling as an engine user, but I think this is part of another PR where both `decode` and `ingest` get wrapped into one single pipeline with one common wrapper error type: `EngineError`.

On top we can optimize for the internal o-o-o buffer even more as it's FIFO nature is not optimal, a sorted list (by seq num and maybe even public key) might be more efficient, though I'm not sure about optimal implementations yet, so benchmarks might make sense before we attempt any of this.

mycognosist (approved):

p2panda-engine/src/operation.rs
```diff
@@ -0,0 +1,260 @@
...
+    validate_backlink, validate_operation, Body, Extension, Header, Operation, OperationError,
+};
+use p2panda_store::{LogStore, OperationStore, StoreError};
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+use thiserror::Error;
+
+use crate::extensions::PruneFlag;
+
+/// Encoded bytes of an operation header and optional body.
```
sandreae, line 14: It would be nice to have this in p2panda-core eventually, maybe behind a `cbor` flag

p2panda-engine/src/operation.rs
```diff
@@ -0,0 +1,260 @@
...
+/// The trait bounds requires the operation header to contain a prune flag as specified by the core
+/// p2panda specification.
+pub async fn ingest_operation<S, L, E>(
+    store: &mut S,
+    header: Header<E>,
+    body: Option<Body>,
+) -> Result<IngestResult<E>, IngestError>
+where
+    S: OperationStore<L, E> + LogStore<L, E>,
+    E: Clone + Serialize + DeserializeOwned + Extension<L> + Extension<PruneFlag>,
```
sandreae, line 103: Love our extensions :wink:

sandreae: Although.... we could also move at least `log_id` into the method signature, and not require that `Extension<L>` be present. This gives users more flexibility in how they define log ids, it doesn't need to be something encoded in the operation extensions. The same could be said for prune flag even... but there it gets a bit fuzzy, as we specify where pruning should happen at a protocol level, and to respect that we say a published prune flag is required.

p2panda-engine/src/macros.rs
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+macro_rules! delegate_sink {
```
sandreae, line 3: I'd appreciate a brief comment on what these macros are used for :pray:

p2panda-engine/src/stream/ingest.rs
```diff
@@ -0,0 +1,258 @@
...
+                }
+            };
+            let Some(IngestAttempt(header, body, counter)) = res else {
+                // Both external stream and buffer stream has ended, so we stop here as well
+                return Poll::Ready(None);
+            };
+
+            // 2. Validate and check the log-integrity of the incoming operation. If it is valid it
+            //    get's persisted and the log optionally pruned.
+            let ingest_fut = async { ingest_operation::<S, L, E>(&mut store, header, body).await };
```
sandreae, line 129: Aha, so if we allowed passing in log id as a parameter, then it would actually make it not possible to use generic stream patterns like this. Interesting! So our extensions are indeed very cool :wink: 

Am I right saying this? Had you considered this as well @adzialocha

sandreae (commented):
All fantastic work, well done! 

I had a few comments, most of them I expect are for a wider conversation and will be extracted elsewhere. I am happy to approve and merge once we touch base on these things.

p2panda-engine/src/stream/ingest.rs
adzialocha: > Aha, so if we allowed passing in log id as a parameter, then it would actually make it not possible to use generic stream patterns like this. Interesting! So our extensions are indeed very cool 😉

Can you clarify what you mean with "parameter"? Do you mean an argument on the `ingest` method?

adzialocha: It would be possible to not require the log id to be encoded in the header extension in the standalone `ingest_operation` method, which is cool, but yeah, for this stream we need to get it from each operation from somewhere. We can extract it before we call the method and make the extension mandatory when using streams.

adzialocha: Thinking about it more: its possible to implement an extension trait without encoding anything. You could return a constant value instead in an extension impl, etc.

PR #559, opened 2024-09-18
Connection manager and actor

mycognosist:
This is a first-pass at writing a connection manager and actor for the purpose of facilitating sync sessions.

Connection-related logic is kept in the manager, with the actor primarily serving as a means of responding to connection events by invoking manager methods.

Super minimal functionality for first-pass:

- [x] Keep track of discovered peer-topic combinations
- [x] Connect to new peers
- [x] Handle inbound peer connections
- [x] Invoke sync sessions
- [x] Log successful sync sessions

Existing tests are passing.

## 📋 Checklist


p2panda-net/src/engine/engine.rs
```diff
@@ -410,7 +566,16 @@ impl EngineActor {
...
+        self.peers.on_announcement(topics.clone(), delivered_from);
+
+        // Inform the connection manager about the peer topics
+        if let Some(tx) = &self.connection_actor_tx {
+            tx.send(ToConnectionActor::UpdatePeerTopics {
+                peer: delivered_from,
+                topics,
+            })
+            .await?;
+        }
```
sandreae, line 578: @adzialocha do you agree that this is the correct place to notify the connection manager of a new peer and/or topics? This is the point where the connection manager discovers peers, internally it decides whether to initiate sync based on 1) is an outbound sync session in progress already for this peer+topic? 2) did we already complete a sync session with this peer+topic? If the answer to either of those is "yes" then we decide not to initiate sync.

adzialocha: Yes! Looks good.

This is the place where "discovery" finally brought the peer and its topics together.

sandreae: Great! Just wanted to check :smile:

sandreae (approved):
This all looks good to me. I think we can try to improve separation between manager and actor, reduce similar actor message types, and maybe combine the sync actor and connection manager with a general refactor later. But this is a great initial version.

p2panda-net/src/connection/mod.rs
```diff
@@ -0,0 +1,6 @@
+mod actor;
```
adzialocha, line 1: License header missing

adzialocha (approved):

PR #563, opened 2024-09-26
Refactor connection manager

mycognosist:
This PR aims to simplify our handling of connections for the purpose of running a sync protocol over novel peer-topic pairings.

- Gets rid of the connection and sync actors
- Renames `connection` module to `sync`
- Renames `ConnectionManager` to `SyncManager`
- Moves sync logic into sync module root
- Moves sync initiation into sync manager
- Moves sync acceptance into sync connection handler

Tests are passing. No more erroneously closed connections.

## 📋 Checklist


p2panda-net/src/network.rs
```diff
@@ -657,10 +657,8 @@ mod sync_protocols {
             }
 
             sink.flush().await?;
-            sink.close().await?;
```
mycognosist, line 660: This turned out to be the cause of at least some of the closed connection errors.

p2panda-net/src/sync/manager.rs
```diff
@@ -0,0 +1,154 @@
...
+use p2panda_sync::SyncError;
+use tokio::sync::mpsc::Sender;
+use tracing::{debug, warn};
+
+use crate::engine::ToEngineActor;
+use crate::sync::{self, SYNC_CONNECTION_ALPN};
+
+#[derive(Debug)]
+pub struct SyncManager {
+    known_peer_topics: HashMap<NodeId, HashSet<TopicId>>,
```
mycognosist, line 19: When we get into our upcoming cleanup phase, I'd like to take a close look at our address book story and see if there's potential for consolidation. We currently have a peer map in the engine actor and a topic map here...they record different information but it might be nice to have a single location for look-up. Just flagging that here for now.

p2panda-net/src/sync/manager.rs
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    /// Respond to newly discovered peer topics by initiating a new connection if one is not
+    /// currently underway and a successful sync session has not already been completed.
+    pub async fn update_peer_topics(&mut self, peer: NodeId, topics: Vec<TopicId>) -> Result<()> {
+        debug!("updating peer topics in connection manager");
+
+        // Create a list of (previously) unknown topics that we might like to sync over.
+        let mut new_topics = Vec::new();
```
mycognosist, line 49: This isn't the most elegant solution (compared to what @sandreae had before) but I needed to ensure we didn't have two mutable borrows.

p2panda-net/src/sync/manager.rs
```diff
@@ -0,0 +1,154 @@
...
+            let sync_complete = self
+                .completed_sync_sessions
+                .entry(topic)
+                .or_default()
+                .contains(&peer);
+
+            // Attempt connection in order to initiate a sync session.
+            if !active_peers.contains(&peer) && !sync_complete {
+                active_peers.insert(peer);
+                // @TODO: We might want to disentangle this call from the `update_peer_topics`
```
mycognosist, line 75: One way to do this would be to make `connect()` a standalone function (not a method) and spawn it in a task. I'm not overly concerned about it for now but that could be a target for cleanup phase. I'd probably also want to add a sync connection duration timer so we can get a feeling for how long sync takes in various scenarios.

p2panda-net/src/sync/manager.rs
```diff
@@ -0,0 +1,154 @@
...
+        } else {
+            debug!("sync failure: initiate");
+            self.complete_failed_sync(peer, topic)
+        }
+
+        Ok(())
+    }
+
+    /// Remove the given topic from the set of active sync sessions for the given peer.
+    pub fn complete_failed_sync(&mut self, peer: NodeId, topic: TopicId) {
```
mycognosist, line 132: A future task would be to provide a failure reason.

mycognosist (commented):
Just left a few notes in the review, mostly about next steps.

adzialocha (approved):
😍

PR #565, opened 2024-10-14
Validation method for prunable logs

adzialocha:
This PR introduces a separate validation method to check against backlinks in the context of prunable logs.

If a log is prunable we can look out for "prune flags", as soon as one was set for an operation we don't expect the header of a backlink, otherwise we go on with validation as usual.

### DIFF
--- a/p2panda-core/src/prune.rs
+++ b/p2panda-core/src/prune.rs
@@ -4,6 +4,15 @@ use serde::{Deserialize, Serialize};
 
 use crate::{validate_backlink, Header, OperationError};
 
+/// Alternative backlink validation method for logs which allow pruning.
+///
+/// When a "prune flag" is set in an operation, an author signals to others that all operations can
+/// be deleted (including payloads) in that log _before_ it.
+///
+/// As soon as a prune flag was set for an operation we don't expect the header of a backlink,
+/// otherwise we go on with validation as usual.
+///
+/// Use this method instead of [`validate_backlink`] if you want to support prunable logs.
 pub fn validate_prunable_backlink<E>(
     past_header: Option<&Header<E>>,
     header: &Header<E>,
--- a/p2panda-core/src/prune.rs
+++ b/p2panda-core/src/prune.rs
@@ -1,24 +1,69 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-use serde::de::DeserializeOwned;
-use serde::Serialize;
-use thiserror::Error;
+use serde::{Deserialize, Serialize};
 
-use crate::Header;
-
-#[derive(Error, Debug)]
-pub enum PruneError {
-    // @TODO
-    #[error("operation version {0} is not supported, needs to be <= {1}")]
-    UnsupportedVersion(u64, u64),
-}
+use crate::{validate_backlink, Header, OperationError};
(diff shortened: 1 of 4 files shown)


p2panda-core/src/prune.rs
```diff
@@ -0,0 +1,78 @@
...
+/// Alternative backlink validation method for logs which allow pruning.
+///
+/// When a "prune flag" is set in an operation, an author signals to others that all operations can
+/// be deleted (including payloads) in that log _before_ it.
+///
+/// As soon as a prune flag was set for an operation we don't expect the header of a backlink,
+/// otherwise we go on with validation as usual.
+///
+/// Use this method instead of [`validate_backlink`] if you want to support prunable logs.
+pub fn validate_prunable_backlink<E>(
```
sandreae, line 16: With this method being such a thin wrapper around `validate_backlink`, I almost wonder if we need it at all.... I could imagine an if statement handling this case quite nicely. Maybe that would actually result in clearer code and allow for slightly different behaviour around prune points (maybe a user wants to still validate backlinks when present, even on prune points). 

```rust
let prune_flag: PruneFlag = header.extract();
if prune_flag.is_true() {
  // get backlink...
  validate_backlink(past_header, header)
} else {
  Ok(())
}
```

I think the plus points for of having the specialised method is that we get nice docs which explain what the hell pruning is :smile: and this is a big plus.... but it could also appear somewhere else. 

I'm really not sure though, just floating my initial reaction right now.

p2panda-core/src/prune.rs
```diff
@@ -0,0 +1,78 @@
...
+    past_header: Option<&Header<E>>,
+    header: &Header<E>,
+    prune_flag: bool,
+) -> Result<(), OperationError>
+where
+    E: Clone + Serialize + for<'a> Deserialize<'a>,
+{
+    assert!(
```
mycognosist, line 24: As unlikely as this scenario is, it's still going to panic if it happens. I'm wondering if we should rather return an `OperationError`, as is the case with all other validation failures.

mycognosist (changes requested):
One small suggested change around error handling, otherwise it looks great!

p2panda-core/src/prune.rs
adzialocha: For me it would be legit to panic here since it shows that you have a logical bug in your code. Handling it in the errors would hide that bug?

p2panda-core/src/prune.rs
adzialocha: > I think the plus points for of having the specialised method is that we get nice docs which explain what the hell pruning is 😄 and this is a big plus.... but it could also appear somewhere else.

Thats also what think! I don't mind it's so simple, people can ignore it and implement it themselves (as many other things), but it's nice to explain it here and have a clear reference. It's simple code but a complicated concept with big implications.

p2panda-core/src/prune.rs
adzialocha: Ah ja, good point. If the node implementation just takes out any operation as soon as there's a backlink then this can be exploited indeed, Ill change it to an error then.

sandreae (approved):
Love the docs ❤️

PR #566, opened 2024-10-14
CBOR encode and decode methods for operation headers

adzialocha:
This PR introduces generic methods to encode and decode values into and from CBOR.

This is then internally used to implement conversion traits from bytes to `Header` and vice-versa.

### DIFF
--- a/p2panda-core/src/operation.rs
+++ b/p2panda-core/src/operation.rs
@@ -4,6 +4,7 @@ use serde::de::DeserializeOwned;
 use serde::Serialize;
 use thiserror::Error;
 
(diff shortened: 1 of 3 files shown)


sandreae (approved):

p2panda-core/src/operation.rs
```diff
@@ -86,7 +87,7 @@ impl<E> Default for Header<E> {
             payload_size: 0,
             payload_hash: None,
             timestamp: 0,
-            seq_num: 1,
+            seq_num: 0,
```
sandreae, line 90: :eyes:

PR #568, opened 2024-10-14
`0.1.0`: changes to `p2panda-sync`

sandreae:
* [x] `TopicMap` trait fn should be async and return a vec of log ids
* [x] Move `TopicMap` from `p2panda-store` into `p2panda-sync`
* [x] Many small re-naming
    * [x] `protocols` module name can be more descriptive (maybe separate `log_height` etc. into own modules on root level)
    * [x] `RawOperation` => `Operation`
    * [x] `SyncProtocol::open()` -> `SyncProtocl::initiate()`
* [x] Move sync trait into `lib.rs`
* [x] Can we avoid packing header and body together into a tuple when we send bytes on `app_tx` 
    * Make `FromSync::Bytes()` => `FromSync::Data(Vec<u8>, Option<Vec<u8>>)`
* [x] Move `cbor_codec.rs` and `utils.rs` into `cbor.rs` in root 
* [x] Move `log_height.rs` into root and rename `log_sync.rs` 
* [x] Rename `into_stream` and `into_sink` to better explain the functionality (it also encodes/decodes and frames) 

## 📋 Checklist


p2panda-net/src/engine/engine.rs
```diff
@@ -641,23 +649,43 @@ impl TopicMap {
...
+    /// This method forwards messages to the subscribers for the given topic.
+    pub async fn on_sync_message(
+        &self,
+        topic: TopicId,
+        header: Vec<u8>,
+        payload: Option<Vec<u8>>,
+        delivered_from: PublicKey,
+    ) -> Result<()> {
+        let inner = self.inner.read().await;
+        let out_tx = inner.earmarked.get(&topic).context("on_sync_message")?;
```
mycognosist, line 681: Can we rename `out_tx` to `from_network_tx` to keep parity with naming conventions in `on_gossip_message()`?

p2panda-sync/src/log_sync.rs
```diff
@@ -38,16 +37,16 @@ where
     }
 }
 
-static LOG_HEIGHT_PROTOCOL_NAME: &str = "p2panda/log_height";
+static LOG_HEIGHT_PROTOCOL_NAME: &str = "p2panda/log_sync";
```
mycognosist, line 40: `LOG_HEIGHT_PROTOCOL_NAME` could be changed to `LOG_SYNC_PROTOCOL_NAME`.

mycognosist (changes requested):

p2panda-net/src/engine/mod.rs
```diff
@@ -14,7 +14,7 @@ use futures_util::future::{MapErr, Shared};
 use futures_util::{FutureExt, TryFutureExt};
 use iroh_gossip::net::Gossip;
 use iroh_net::{Endpoint, NodeAddr};
-use p2panda_sync::traits::SyncProtocol;
+use p2panda_sync::SyncProtocol;
```
adzialocha, line 17: Small thing: How about naming this to `Sync`? Then it's similar to the `Discovery` trait (which is not `DiscoveryProtocol`)

p2panda-sync/src/lib.rs
```diff
@@ -1,10 +1,44 @@
...
+use async_trait::async_trait;
+use futures::{AsyncRead, AsyncWrite, Sink};
 use thiserror::Error;
 
+pub type TopicId = [u8; 32];
+
+/// Trait used for mapping a generic topic to a single or collection of logs
+pub trait TopicMap<K, V> {
+    fn get(&self, topic: &K) -> Option<V>;
+}
```
adzialocha, line 20: > "TopicMap trait fn should be async and return a vec of log ids"

Ah, I think this was forgotten here

adzialocha (changes requested):
Nice improvement with the `SyncMessage` / `GossipMessage` naming!

p2panda-net/src/engine/mod.rs
sandreae: The only thing about that is that it conflicts with `Sync` from rust `std`, not sure if that's a problem though... What do you think?

p2panda-sync/src/lib.rs
sandreae: Aha, true, I'll take that in a later PR.

mycognosist (approved):

adzialocha (approved):

PR #569, opened 2024-10-14
Raw store trait methods and in-memory implementation

adzialocha:


p2panda-store/src/lib.rs
```diff
@@ -1,7 +1,112 @@
...
+        body: Option<&Body>,
+        header_bytes: &[u8],
+        log_id: &LogId,
+    ) -> Result<bool, Self::Error>;
+
+    /// Get an operation.
+    async fn get_operation(
+        &self,
+        hash: Hash,
+    ) -> Result<Option<(Header<Extensions>, Option<Body>)>, Self::Error>;
```
sandreae, line 32: Maybe it would be nice to still get `Operation`s back from the store, even though the insert method now takes it's constituent parts?

sandreae: Pros:
- minor (but often occurring) optimisation that the hash doesn't need to be re-generated
- we will likely use `Operation` in other APIs so why not provide that already from the store

Cons:
- if point 2 above isn't actually true :sweat_smile:

adzialocha: I've refactored all APIs which touch the store and realized we're so far fine with returning `(Header, Option<Body>)` instead of `Operation`. The only method where we are hashing and assuming `Operation` (for example in `validate_operation`), is `ingest_operation` (in `p2panda-engine`), but everything fits nicely together there as it seems and no step is redundant ✨

p2panda-sync/src/log_sync.rs
```diff
@@ -260,20 +261,16 @@ async fn remote_needs<T, E>(
 where
     E: Clone + Serialize,
 {
-    let mut log = store
-        .get_log(public_key, log_id)
+    let log = store
+        .get_raw_log(public_key, log_id)
         .await
-        .map_err(|e| SyncError::Protocol(e.to_string()))?;
+        .map_err(|err| SyncError::Custom(format!("could not retrieve log from store: {err}")))?;
```
adzialocha, line 267: This is strictly speaking no `Protocol` error but an critical storage failure. `Custom` feels too generic though as we know quite well what the category of this error is. Should I try to wrap it in an `Io` type?

p2panda-sync/src/log_sync.rs
```diff
@@ -38,6 +38,7 @@ where
     }
 }
 
+// @TODO(adz): Does this need a version number?
```
mycognosist, line 41: Hmm...good question. I'm currently leaning towards no. I suppose the other option for versioning is to include it in the sync protocol handshake (if necessary). Feels fine to leave out for now.

p2panda-sync/src/log_sync.rs
mycognosist: I'd go with the `Io` wrapper here.

mycognosist (approved):
Great work! Seems like some really nice efficiency improvements and I'm sure hands-on time with the external APIs will nudge us toward any further refinements in the future.

p2panda-store/src/lib.rs
```diff
@@ -1,7 +1,114 @@
...
+
+    /// Get "raw" header and body bytes of operation from store.
+    async fn get_raw_operation(&self, hash: Hash) -> Result<Option<RawOperation>, Self::Error>;
+
+    /// Returns `true` if operation exists in store.
+    async fn has_operation(&self, hash: Hash) -> Result<bool, Self::Error>;
```
sandreae, line 40: Handy new method :+1:

p2panda-sync/src/log_sync.rs
sandreae: Yeh, good question! We don't actually do anything with this yet, if we want to check versions on sync then we would need to add a message to send over the version during the handshake. We should definitely do this once things stabilise. For now I'm happy with it in or out.

sandreae (approved):
Really like these improvements to the store API :clap: all looks good to me!

p2panda-sync/src/log_sync.rs
adzialocha: Hmn, I looked a little bit into Input/Output APIs and I'm not sure databases really fit in there. I'll gather some thoughts, this feels like a separate PR, will write about it there!

adzialocha: See https://github.com/p2panda/p2panda/pull/575

PR #570, opened 2024-10-15
Replace `Ready` network message with oneshot receiver on subscribe

mycognosist:
- [x] Remove `Ready` variant from `FromNetwork` enum
- [x] Return a oneshot receiver from `subscribe()`
- [x] Send `()` on the oneshot channel when the gossip overlay has been joined for a particular topic

## 📋 Checklist

### DIFF
--- a/p2panda-net/examples/chat.rs
+++ b/p2panda-net/examples/chat.rs
@@ -30,15 +30,11 @@ async fn main() -> Result<()> {
         .build()
         .await?;
 
-    let (tx, mut rx) = network.subscribe(topic_id).await?;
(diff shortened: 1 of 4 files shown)


p2panda-net/examples/chat.rs
```diff
@@ -55,7 +51,7 @@ async fn main() -> Result<()> {
     });
 
     println!(".. waiting for peers to join ..");
-    let _ = ready_rx.recv().await;
+    let _ = ready.await;
```
mycognosist, line 54: The `tokio` oneshot receiver doesn't have a `recv()` method; it's a future that we can `await` directly.

adzialocha (approved):

PR #574, opened 2024-10-22
Introduce a scheduler to the sync manager

mycognosist:
A simple connection / sync scheduler:

- Incorporated into the existing sync manager
- Receives new peer-topic combinations from the engine (gossip announcements)
- Adds attempts to a scheduler queue
  - Connection errors result in a reattempt
  - Sync errors result in the attempt being dropped
- Keeps track of pending, active and successfully completed attempts

## 📋 Checklist

### DIFF
--- a/p2panda-net/src/engine/engine.rs
+++ b/p2panda-net/src/engine/engine.rs
@@ -166,12 +166,13 @@ impl EngineActor {
         mut gossip_actor: GossipActor,
         sync_manager: Option<SyncManager>,
     ) -> Result<()> {
-        let sync_manager_token = CancellationToken::new();
-        let cloned_sync_manager_token = sync_manager_token.clone();
+        // Used to shutdown the sync manager.
+        let shutdown_token = CancellationToken::new();
+        let cloned_shutdown_token = shutdown_token.clone();
 
         if let Some(sync_manager) = sync_manager {
             tokio::task::spawn(async move {
-                if let Err(err) = sync_manager.run(cloned_sync_manager_token).await {
+                if let Err(err) = sync_manager.run(cloned_shutdown_token).await {
                     error!("sync manager failed to run: {err:?}");
                 }
             });
@@ -190,7 +191,7 @@ impl EngineActor {
             error!(?err, "error during shutdown");
         }
 
-        sync_manager_token.cancel();
+        shutdown_token.cancel();
         gossip_handle.await?;
         drop(self);
 
--- a/p2panda-net/src/engine/engine.rs
+++ b/p2panda-net/src/engine/engine.rs
@@ -17,7 +17,7 @@ use tracing::{debug, error, warn};
 use crate::engine::gossip::{GossipActor, ToGossipActor};
 use crate::engine::message::NetworkMessage;
 use crate::network::{FromNetwork, ToNetwork};
-use crate::sync::{SyncManager, ToSyncManager};
+use crate::sync::manager::{SyncManager, ToSyncManager};
 use crate::{FromBytes, ToBytes};
 
 /// Maximum size of random sample set when choosing peers to join gossip overlay.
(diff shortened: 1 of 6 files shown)


p2panda-net/src/sync/manager.rs
```diff
@@ -3,45 +3,154 @@
...
+            sync_queue_rx,
+        };
+
+        (sync_manager, sync_manager_tx)
+    }
+
+    /// Add a peer and topic combination to the sync connection queue.
+    async fn schedule_attempt(&mut self, peer: NodeId, topic: TopicId) -> Result<()> {
+        let sync_attempt = SyncAttempt::new(peer, topic);
+        self.sync_queue_tx.send(sync_attempt).await?;
```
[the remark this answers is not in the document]
adzialocha: > This will block when the queue is full; I don't think that's a problem for now but something to bear in mind

Is it possible to never make it too full? If we know we only can handle max. 10 sync sessions at a time then we should a) never push to the queue when it's full and b) the queue can not be larger than 10?

It could be otherwise a potential bug source which might be hard to find in the future (unexpected blocking behaviour).

p2panda-net/src/engine/engine.rs
```diff
@@ -545,13 +562,9 @@ impl EngineActor {
...
-        //
-        // NOTE: This will only return once sync has been attempted with all novel peer-topic
-        // combinations.
-        if let Some(ref mut sync_manager) = self.sync_manager {
-            sync_manager
-                .update_peer_topics(delivered_from, topics)
-                .await?;
+        if let Some(sync_manager_tx) = &self.sync_manager_tx {
+            let peer_topics = ToSyncManager::new(delivered_from, topics);
+            sync_manager_tx.send(peer_topics).await?
```
adzialocha, line 567: 👍🏻

p2panda-net/src/sync/manager.rs
```diff
@@ -68,86 +177,90 @@ impl SyncManager {
...
-            // Attempt connection in order to initiate a sync session.
+            // Schedule a sync attempt.
             if !active_peers.contains(&peer) && !sync_complete {
-                active_peers.insert(peer);
-                // @TODO: We might want to disentangle this call from the `update_peer_topics`
-                // method so that it returns immediately and the `connect` method has its own life.
-                if let Err(err) = self.connect(peer, topic).await {
-                    warn!("outbound connection attempt failed: {}", err)
-                }
+                self.schedule_attempt(peer, topic).await?;
```
adzialocha, line 182: If I see it correctly this ties the scheduler to the receipt of announcement messages. We could detach the logics a little bit more, by just learning about the topics here and then checking if we have free "sync slots", if yes, start an attempt. If not, that's fine.

If a sync session ended (`complete_successful_sync`) or all attempts failed plus the queue has some space for us (`complete_failed_sync` + `sync_attempt.attempts > self.max_retry_attempts` + `queue.len() < MAX_CONCURRENT_SYNC_SESSIONS`)  we can look into our topic table again and schedule a new session.

I think this a) unties the logic nicely from "announcements" b) prevents the queue to ever be full (see other comment somewhere else)

adzialocha (changes requested):
Great PR! I have some suggestions for the scheduling logic / queue handling but they are not critical.

Otherwise some smaller change requests over documentation.

p2panda-net/src/engine/engine.rs
mycognosist: Good call on avoiding having to debug unexpected blocking behaviour in the future! I've added a capacity check: if the queue is full then we call `send_timeout()` which will only block for the specified duration before returning an error. Are you happy with this behaviour for now? I've set the timeout to 100 ms (feels quite generous).

p2panda-net/src/sync/manager.rs
mycognosist: Cool idea! Looking into this now.

adzialocha (approved):

PR #575, opened 2024-10-22
Improve `Sync` error types for implementers and API users

adzialocha:
This PR is a draft for a different take on `p2panda-sync` error types. This hopefully gives them more semantic meaning and clearer guidance on how they can be understood when writing and using a `Sync` implementation.

See comments below for details and some open questions!

### DIFF
--- a/p2panda-net/src/sync/handler.rs
+++ b/p2panda-net/src/sync/handler.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use anyhow::Result;
 use futures_lite::future::Boxed as BoxedFuture;
 use iroh_net::endpoint::{self, Connecting, Connection};
-use p2panda_sync::{SyncError, SyncProtocol};
+use p2panda_sync::SyncProtocol;
 use tokio::sync::mpsc;
 use tracing::{debug, debug_span};
 
@@ -42,10 +42,7 @@ impl SyncConnection {
         let _span = debug_span!("connection", connection_id, %remote_addr);
 
         // Create a bidirectional stream on the connection.
-        let (mut send, mut recv) = connection
-            .accept_bi()
-            .await
-            .map_err(|e| SyncError::Protocol(e.to_string()))?;
+        let (mut send, mut recv) = connection.accept_bi().await?;
 
         let sync_protocol = self.sync_protocol.clone();
         let engine_actor_tx = self.engine_actor_tx.clone();
--- a/p2panda-net/src/sync/mod.rs
+++ b/p2panda-net/src/sync/mod.rs
@@ -35,7 +35,7 @@ pub async fn initiate_sync<S: AsyncWrite + Send + Unpin, R: AsyncRead + Send + U
 
     // Set up a channel for receiving new application messages.
     let (tx, mut rx) = mpsc::channel(128);
-    let mut sink = PollSender::new(tx).sink_map_err(|e| SyncError::Protocol(e.to_string()));
+    let mut sink = PollSender::new(tx).sink_map_err(|e| SyncError::Critical(e.to_string()));
 
     // Spawn a task which picks up any new application messages and sends them on to the engine
     // for handling.
@@ -115,7 +115,7 @@ pub async fn accept_sync<S: AsyncWrite + Send + Unpin, R: AsyncRead + Send + Unp
 
     // Set up a channel for receiving new application messages.
     let (tx, mut rx) = mpsc::channel(128);
-    let mut sink = PollSender::new(tx).sink_map_err(|e| SyncError::Protocol(e.to_string()));
+    let mut sink = PollSender::new(tx).sink_map_err(|e| SyncError::Critical(e.to_string()));
 
     // Spawn a task which picks up any new application messages and sends them on to the engine
     // for handling.
--- a/p2panda-sync/Cargo.toml
+++ b/p2panda-sync/Cargo.toml
@@ -5,7 +5,6 @@ edition = "2021"
 
 [features]
 cbor = [
-    "dep:ciborium",
     "dep:serde",
     "dep:tokio",
     "dep:tokio-stream",
@@ -15,7 +14,6 @@ log-sync = ["dep:p2panda-core", "dep:p2panda-store", "cbor"]
 
 [dependencies]
 async-trait = "0.1.82"
-ciborium = { version = "0.2.2", optional = true }
 futures = "0.3.30"
 p2panda-core = { path = "../p2panda-core", optional = true }
 p2panda-store = { path = "../p2panda-store", optional = true }
--- a/p2panda-sync/src/cbor.rs
+++ b/p2panda-sync/src/cbor.rs
@@ -66,9 +66,7 @@ where
                         // @TODO
                         panic!("{}", err);
                     }
-                    err => Err(SyncError::RemoteUnexpectedBehaviour(format!(
-                        "CBOR codec failed decoding message from remote peer, {err}"
-                    ))),
+                    err => Err(SyncError::InvalidEncoding(err.to_string())),
                 }
             }
         }
--- a/p2panda-sync/src/cbor.rs
+++ b/p2panda-sync/src/cbor.rs
@@ -58,17 +58,13 @@ where
         let reader = src.reader();
         let result: Result<Self::Item, _> = decode_cbor(reader);
         match result {
-            // If we read the item, we also need to advance the underlying buffer.
             Ok(item) => Ok(Some(item)),
-            Err(ref error) => {
-                match error {
-                    DecodeError::Io(err) => {
-                        // @TODO
-                        panic!("{}", err);
-                    }
-                    err => Err(SyncError::InvalidEncoding(err.to_string())),
-                }
-            }
+            Err(ref error) => match error {
+                DecodeError::Io(err) => Err(SyncError::Critical(format!(
+                    "CBOR codec failed decoding message due to i/o error, {err}"
+                ))),
+                err => Err(SyncError::InvalidEncoding(err.to_string())),
+            },
         }
     }
 }
--- a/p2panda-sync/src/cbor.rs
+++ b/p2panda-sync/src/cbor.rs
@@ -3,9 +3,10 @@
 use std::marker::PhantomData;
 
 use futures::{AsyncRead, AsyncWrite, Sink, Stream};
+use p2panda_core::cbor::{decode_cbor, encode_cbor, DecodeError};
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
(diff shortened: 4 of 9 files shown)


p2panda-net/src/sync/handler.rs
```diff
@@ -42,10 +42,7 @@ impl SyncConnection {
         let _span = debug_span!("connection", connection_id, %remote_addr);
 
         // Create a bidirectional stream on the connection.
-        let (mut send, mut recv) = connection
-            .accept_bi()
-            .await
-            .map_err(|e| SyncError::Protocol(e.to_string()))?;
+        let (mut send, mut recv) = connection.accept_bi().await?;
```
adzialocha, line 45: This used `SyncError` before but I'm not sure if we this is an actual error coming from the sync protocol itself. Connectivity issues should maybe be considered a separate issue and deserve their own error type?

p2panda-net/src/sync/manager.rs
```diff
@@ -92,10 +92,7 @@ impl SyncManager {
             .await?;
 
         // Create a bidirectional stream on the connection.
-        let (mut send, mut recv) = connection
-            .open_bi()
-            .await
-            .map_err(|e| SyncError::Protocol(e.to_string()))?;
+        let (mut send, mut recv) = connection.open_bi().await?;
```
adzialocha, line 95: .. same here

p2panda-net/src/sync/mod.rs
```diff
@@ -36,7 +36,7 @@ pub async fn initiate_sync<S: AsyncWrite + Send + Unpin, R: AsyncRead + Send + U
 
     // Set up a channel for receiving new application messages.
     let (tx, mut rx) = mpsc::channel(128);
-    let mut sink = PollSender::new(tx).sink_map_err(|e| SyncError::Protocol(e.to_string()));
+    let mut sink = PollSender::new(tx).sink_map_err(|e| SyncError::Critical(e.to_string()));
```
adzialocha, line 39: If I understand this correctly this is mapping _all_ error types coming from the regarding `Sync` implementation and converting them to one type (previously `Protocol`)?

That would defy the whole reason of having custom error types in sync. Do I missunderstand this code or should we look for a way of passing the actual sync error type further?

p2panda-sync/src/cbor.rs
```diff
@@ -37,13 +38,10 @@ where
...
-        &mut self,
-        item: T,
-        dst: &mut tokio_util::bytes::BytesMut,
-    ) -> Result<(), Self::Error> {
-        let mut bytes = Vec::new();
-        ciborium::into_writer(&item, &mut bytes).map_err(|e| SyncError::Codec(e.to_string()))?;
+    fn encode(&mut self, item: T, dst: &mut BytesMut) -> Result<(), Self::Error> {
+        let bytes = encode_cbor(&item).map_err(|err| {
+            SyncError::Critical(format!("CBOR codec failed encoding message, {err}"))
+        })?;
```
adzialocha, line 44: If we failed to encode our _own_ messages this would be a critical system error

p2panda-sync/src/cbor.rs
```diff
@@ -56,20 +54,23 @@ where
...
-            },
+            Err(ref error) => {
+                match error {
+                    DecodeError::Io(err) => {
+                        // @TODO
+                        panic!("{}", err);
+                    }
+                    err => Err(SyncError::RemoteUnexpectedBehaviour(format!(
+                        "CBOR codec failed decoding message from remote peer, {err}"
+                    ))),
```
adzialocha, line 71: If we failed decoding a remote peer's messages then we can consider this a protocol violation

p2panda-sync/src/cbor.rs
```diff
@@ -56,20 +54,23 @@ where
...
-                // Sometimes the EOF is signalled as IO error
-                ciborium::de::Error::Io(_) => Ok(None),
-                e => Err(SyncError::Codec(e.to_string())),
-            },
+            Err(ref error) => {
+                match error {
+                    DecodeError::Io(err) => {
+                        // @TODO
+                        panic!("{}", err);
+                    }
```
adzialocha, line 68: I'm a little bit unsure about this part. Ideally we would just convert any Io error into a critical system error (because Io will probably not be related to bad behaviour from a remote peer), but I've stumbled over this "EOF" note. Is this still the case or is this a remaining of us troubleshooting the connection bug we had in the past?

p2panda-sync/src/log_sync.rs
```diff
@@ -78,44 +76,41 @@ where
         let mut sink = into_cbor_sink(tx);
         let mut stream = into_cbor_stream(rx);
 
-        // Get the log ids which are associated with this topic.
+        // Get the log ids which are associated with this topic
         let Some(log_ids) = self.topic_map.get(topic).await else {
-            return Err(SyncError::Protocol("Unknown topic id".to_string()));
+            return Err(SyncError::Critical(format!("unknown {topic:?} topic")));
```
adzialocha, line 81: If we tried to initate a sync session over a topic we didn't know, this smells like a bug

p2panda-sync/src/log_sync.rs
```diff
@@ -78,44 +76,41 @@ where
...
         for log_id in log_ids {
-            let log_heights = self
-                .store
-                .get_log_heights(&log_id)
-                .await
-                .expect("memory store error");
-
+            let log_heights = self.store.get_log_heights(&log_id).await.map_err(|err| {
+                SyncError::Critical(format!("can't retreive log heights from store, {err}"))
+            })?;
```
adzialocha, line 91: This shouldn't panic

p2panda-sync/src/log_sync.rs
```diff
@@ -263,7 +261,7 @@ where
     let log = store
         .get_raw_log(public_key, log_id)
         .await
-        .map_err(|err| SyncError::Custom(format!("could not retrieve log from store: {err}")))?;
+        .map_err(|err| SyncError::Critical(format!("could not retrieve log from store, {err}")))?;
```
adzialocha, line 264: Follow up on the discussion in PR https://github.com/p2panda/p2panda/pull/569#discussion_r1810248002:

Storage errors are _not_ I/O errors, as potentially nothing went wrong with the communication between us and the database but with the database itself. On top semantically it's more interesting to know if it's an critical system error and we should maybe worry about it

p2panda-sync/src/cbor.rs
adzialocha: @sandreae

p2panda-sync/src/lib.rs
```diff
@@ -42,21 +42,39 @@ pub trait SyncProtocol<'a>: Send + Sync + Debug {
...
 
-    /// Error which occurs when encoding or decoding protocol messages.
-    #[error("codec error: {0}")]
-    Codec(String),
+    /// Critical error due to system failure on our end.
+    ///
+    /// This indicates that our system is running out of resources (storage layer failure etc.) or
+    /// we have a buggy implementation.
+    #[error("sync session failed due critical system error: {0}")]
+    Critical(String),
```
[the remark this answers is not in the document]
adzialocha: Since `RemoteTimeout` is more a connection-layer concern, I'm wondering if we can have that on a higher-level in the connection manager?

Not saying we need this already, but more if it's possible.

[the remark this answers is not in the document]
sandreae: Thanks for looking into these error types, really like the changes you made so far, definitely more expressive for users + implementers. 

- `RemoteUnexpectedBehaviour`: I wonder if this could be split into two distinct types representing: `1)` the remote sent badly formed data; `2)` the remote sent a valid message but it violated the protocol in some way. `2)` is still very broad (could be unexpected message order, or access-control violation) and it might be that we add useful variants later, but maybe to start that's enough. Naming wise, maybe `RemoteInvalidMessage` & `RemoteUnexpectedBehaviour`?  
- `RemoteTimeout`: I agree this could be moved into a higher layer, currently `-sync` doesn't really know about "connections" exactly... but I also wouldn't mind leaving it here. If it stayed here, maybe `RemoteUnresponsive` could be good?

p2panda-sync/src/cbor.rs
sandreae: This is an artefact which came across from a cbor codec implementation I found [here](https://github.com/vorner/tokio-serde-cbor/blob/fdd3b348f9c215466ce2785e2be9d3073c3ba19a/src/lib.rs#L135). Right now I can't remember if I observed this behaviour myself, but it wasn't anything to do with the connection bug de-bugging. If all tests pass with this removed I'd say we don't need it.

p2panda-sync/src/lib.rs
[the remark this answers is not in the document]
mycognosist: I'm happy with these changes. Regarding the `RemoteTimeout` variant, I can see it living in either place but probably lean towards moving it into `SyncAttemptError` in the sync manager. Then we would just be concerned whether the `connect_and_sync()` attempt completed in the given time limit or not.

mycognosist (commented):
Looks good to me overall. Seems like we just need to figure out the alternative solution for `sink_map_err` and move some variants around (such as timeout to sync manager).

p2panda-sync/src/lib.rs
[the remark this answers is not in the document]
adzialocha: > Regarding the RemoteTimeout variant, I can see it living in either place but probably lean towards moving it into SyncAttemptError in the sync manager. Then we would just be concerned whether the connect_and_sync() attempt completed in the given time limit or not.

I was wondering that as well. I think the `RemoteTimeout` makes sense if you can actually measure the timing _between_ messages (which I'm not sure the connection manager can see), but I think that's already too much entering "nice to have" territory. I'll leave it out for now and we can reason about timeout handling in another PR, maybe it's better to have on connection handler level.

p2panda-sync/src/cbor.rs
adzialocha: Tests are passing and it's still mysterious to me that whole part, but that's for another PR. I removed it for now and then we tackle it next.

p2panda-net/src/sync/mod.rs
adzialocha: Cool, I think this can be looked at in a separate PR as part of the "improved error handling in the sync manager" task

p2panda-sync/Cargo.toml
```diff
@@ -5,7 +5,6 @@ edition = "2021"
 
 [features]
 cbor = [
-    "dep:ciborium",
```
sandreae, line 8: Nice we can drop a dependency here

sandreae (approved):
Thanks for the greatly improved error types :clap:

PR #577, opened 2024-10-25
Handle generic topics in `-net` and `-sync`

sandreae:
We want users to be able to introduce their own topic types which can be used to derive the collection of data gossip overlays and sync sessions operate over. What constitutes a topic is left entirely up to the user. When discussing this change we arrived at several requirements/properties which need to be met in this implementation:

- `p2panda-net` requires that a topic `T` can be uniquely identified by a `[u8; 32]` byte array. This is so that it can be used as a gossip overlay identifier in `iroh_gossip`.
- `p2panda-net` does not need to know about the actual "gossip scope" (the collection of data a `T` refers to) of a topic `T`, this is rather an application concern, it is expected that handling which messages to send/receive on which topic `T` is handled by users. 
- `p2panda_sync::SyncProtocol` implementations must be able to map a topic `T` to a collection of data which the sync session is concerned with, aka the "sync scope". This can be different from the "gossip scope".

## `TopicId` trait in `p2panda-net`
This trait is only required in `p2panda-net` so that topics can be used as gossip overlay ids.

```rust
/// Topics are identified by a network-unique id encoded as a `[u8; 32]`. Topic ids are announced
/// on the network and used to identify peers with overlapping interests. Once identified, peers
/// join a gossip overlay and, if a sync protocol has been provided, attempt to synchronize past
/// state before entering "live mode".
pub trait TopicId {
    fn id(&self) -> [u8; 32];
}
```

## Keep a map of `[u8; 32] => T` relations in `Network`
Peers only announce topics by their `[u8; 32]` ids, when joining a gossip overlay the concrete topic type is never used in `-net` (as explained above the application layer should take care of sending/receiving correct gossip messages), however when we initiate a sync session, we need to be able to map a topic id `[u8; 32]` back to it's concrete topic `T` in order to pass this into `SyncProtocol::initiate(..)`. For this reason we keep a map of all topics which were registered via the `subscribe()` method.

## `Topic` super-trait in `p2panda-sync`
Used in both `-net` and `-sync` to place requirements on topic objects.

```rust
/// A topic is used to identify the particular data-set a peer is interested in syncing. Exactly how
/// this is expressed is left up to the user to decide. During sync the "initiator" sends their topic to
/// a remote peer where it is be mapped to their local data-set and access-control checks can be
/// performed. Once this "handshake" is complete both peers will proceed with the designated sync
/// protocol.
pub trait Topic:
    Clone + Debug + Eq + Hash + Send + Sync + Serialize + for<'a> Deserialize<'a>
{
}
```

## `TopicMap` trait in `p2panda-sync`
Strictly speaking, this trait doesn't need to be provided, as an implementation of `SyncProtocol` can take any approach they like to deriving scope from a topic `T`. I like that we have a place where we can explain this relationship, maybe it could be made clear in the docs that this is only a _suggested_ approach. Alternatively this can move into `log_sync` and become a little more opinionated.

```rust
/// Each `SyncProtocol` implementation defines the type of data it is expecting to sync and how
/// the scope for a particular session should be identified. Sync protocol users can provide an
/// implementation of `TopicMap` so that a scope `S` can be retrieved for a specific topic `T`
/// when a peer initiates or accepts a sync session.
#[async_trait]
pub trait TopicMap<T, S> {
    async fn get(&self, topic: &T) -> Option<S>;
}
```
## 📋 Checklist


p2panda-net/src/lib.rs
```diff
@@ -8,15 +8,36 @@ pub mod network;
...
 #[cfg(feature = "log-sync")]
 pub use p2panda_sync::log_sync::LogSyncProtocol;
 
 pub type NetworkId = [u8; 32];
-pub type TopicId = [u8; 32];
+
+/// Topics are identified by a network-unique id encoded as a `[u8; 32]`.
+/// Topic ids are announced on the network and used to identify peers with
+/// similar interests. Once identified, peers join a gossip overlay and, if
+/// a sync protocol has been provided, attempt to synchronize past state.
```
mycognosist, line 29: Not sure if we've settled on "live mode" or if we want to refer to this in some other way.

```suggestion
/// Topic ids are announced on the network and used to identify peers with
/// overlapping interests. Once identified, peers join a gossip overlay and, if
/// a sync protocol has been provided, attempt to synchronize past state before entering "live mode".
```

p2panda-net/src/sync/manager.rs
```diff
@@ -248,12 +258,19 @@ impl SyncManager {
         let sync_protocol = self.sync_protocol.clone();
         let engine_actor_tx = self.engine_actor_tx.clone();
 
+        let topic = self
+            .topic_map
+            .get(&topic)
+            // @TODO: I'm not sure if this case can ever occur, if it can, it would be best to
```
mycognosist, line 264: I believe the returned value will always be `Some`; `connect_and_sync()` will only be called if a sync attempt has been scheduled (and that requires a `ToSyncManager` message with topic).

p2panda-net/src/sync/mod.rs
```diff
@@ -9,41 +9,52 @@ use std::sync::Arc;
...
 
     // Spawn a task which picks up any new application messages and sends them on to the engine
     // for handling.
     let mut sync_handshake_success = false;
+    let topic_id = topic.id();
     tokio::spawn(async move {
         while let Some(message) = rx.recv().await {
             // We expect the first message to be a topic id
-            if let FromSync::Topic(id) = &message {
+            // @TODO: We don't need to receive the topic from within the sync session as we
```
mycognosist, line 51: Seems worthwhile to keep this in place for now but I'll defer to @adzialocha .

mycognosist (approved):
Super nice work! You made the change look pretty seamless. I've left a few small comments but overall I'm happy.

p2panda-net/src/lib.rs
```diff
@@ -8,15 +8,36 @@ pub mod network;
...
+}
+
+impl<T> Topic for T
+where
+    T: Clone + Debug + Display + Send + Sync,
+{
+    fn id(&self) -> [u8; 32] {
+        Hash::new(self.to_string().as_bytes()).as_bytes().to_owned()
+    }
+}
```
adzialocha, line 43: This auto implementation might already do to much "magic" maybe. I think it's good to expose users to this concept and make them take a concious decision by themselves? What do you think?

p2panda-sync/src/log_sync.rs
```diff
@@ -149,76 +166,87 @@ where
...
+                        for log_id in log_ids {
+                            let latest_operation = self
+                                .store
+                                .latest_operation(&public_key, &log_id)
+                                .await
+                                .map_err(|err| {
+                                    SyncError::Critical(format!(
+                                        "can't retreive log heights from store, {err}"
+                                    ))
+                                })?;
```
adzialocha, line 213: We probably want a method for this in the store (`latest_operations(log_ids: Vec<_>)`) for optimal store queries, just noting that here and maybe we can make a ticket?

p2panda-sync/src/log_sync.rs
```diff
@@ -319,25 +347,24 @@ mod tests {
...
+    impl LogHeightTopicMap {
         pub fn new() -> Self {
-            Self(HashMap::new())
+            LogHeightTopicMap(HashMap::new())
         }
 
-        fn insert(&mut self, topic: TopicId, log_ids: Vec<String>) -> Option<Vec<String>> {
-            self.0.insert(topic, log_ids)
+        fn insert(&mut self, topic: &str, logs: Logs<u64>) -> Option<Logs<u64>> {
+            self.0.insert(topic.to_string(), logs)
```
adzialocha, line 358: Rather than encoding every struct as a `String` (using `Display`) we could make `Topic: StdHash` and let the (fast) hashing algorithm do the work

p2panda-sync/src/log_sync.rs
```diff
@@ -319,25 +347,24 @@ mod tests {
     }
 
     #[derive(Clone, Debug)]
-    struct LogIdTopicMap(HashMap<TopicId, Vec<String>>);
+    struct LogHeightTopicMap(HashMap<String, Logs<u64>>);
```
adzialocha, line 350: Very cool, that concept is so nice now and that simple mapper comes in handy for many use cases already ✨

p2panda-net/src/sync/mod.rs
adzialocha: Oh, haven't realized that this was also in the "initiator" logic, it shouldn't need to be there indeed as far as I can see it!

p2panda-sync/src/log_sync.rs
```diff
@@ -39,20 +41,21 @@ where
...
+    pub store: MemoryStore<L, E>,
 }
 
 #[async_trait]
-impl<'a, S, T, E> SyncProtocol<'a> for LogSyncProtocol<S, T, E>
+impl<'a, T, TM, L, E> SyncProtocol<T, 'a> for LogSyncProtocol<TM, L, E>
 where
-    S: Debug + TopicMap<TopicId, T> + Send + Sync,
-    T: Clone
+    T: Clone + Debug + Send + Sync + Serialize + for<'de> Deserialize<'de>,
```
adzialocha, line 52: I think this should be the newly introduced topic trait here?

p2panda-net/src/lib.rs
```diff
@@ -8,15 +8,36 @@ pub mod network;
...
 pub type NetworkId = [u8; 32];
-pub type TopicId = [u8; 32];
+
+/// Topics are identified by a network-unique id encoded as a `[u8; 32]`.
+/// Topic ids are announced on the network and used to identify peers with
+/// similar interests. Once identified, peers join a gossip overlay and, if
+/// a sync protocol has been provided, attempt to synchronize past state.
+///
+/// The `Topic` trait must be implemented on any user defined topic types.
+pub trait Topic: Clone + Debug + Display + Send + Sync {
```
adzialocha, line 32: Do we need this to be `StdHash` (see other comment) and `Serialize` and `Deserialize` as well? If it's sent over the wire we want it to be serializable, and hashable to insert it into all sorts of hash maps etc.

p2panda-net/src/sync/mod.rs
adzialocha: > It is worth considering that the sync scope portion of a topic may have change since the network subscription occurred, if that is the correct behavior then this is actually a good way to receive the most current topic state.

I'm not sure I understand this / haven't considered this. How and where can the topic change? Do you mean between calling "subscribe" and the actual sync session happening?

p2panda-net/src/sync/mod.rs
```diff
@@ -72,23 +83,28 @@ pub async fn initiate_sync<S: AsyncWrite + Send + Unpin, R: AsyncRead + Send + U
...
             .await
             .expect("engine channel closed");
     });
 
     // Run the sync protocol.
     let result = sync_protocol
         .initiate(
-            topic.as_bytes(),
+            // @TODO: When generic topic is implemented in `p2panda-sync` we can pass the topic
+            // directly here.
```
adzialocha, line 106: Can this comment be removed? Not sure I understand it, we're passing in the topic and it is generic?

adzialocha (commented):
I like your PR description, very nice description of the new "topic" concept, should definitely make us of that in documentation.

I have some comments / questions for now for understanding, but generally it all looks good.

adzialocha:
Maybe it's nice to rename all variables of type `TopicId` from `topic` -> `topic_id`, just for better readability (until our naming changes, who knows)

adzialocha:
What happens if I'm subscribed to two different topics who result in the same topic id? Would our logic consider them as separate?

p2panda-net/src/sync/mod.rs
sandreae: I was thinking about the fact that a topic `T` is stored in multiple locations, both internally inside network, and externally by the user. The internal version is passed into `subscribe` and from that point is never changed, the user stored copy is exposed to any sync protocol via a `TopicMap` and this _may_ be mutated without the networks knowledge. The point I'm making is that it's possible for a topic to get "out of sync" with itself... in a way... maybe this is more of an API documentation point: "if you change a topic in any way, even if the id remains the same, it's advised that `subscribe()` be re-called".

p2panda-net/src/lib.rs
sandreae: Haha, yeh, I almost just added this for fun cos I thought it was cool :rofl: I actually agree that it could be removed though :+1:

p2panda-sync/src/log_sync.rs
sandreae: The `Topic` trait is defined in `p2panda-net` and primarily used for guaranteeing that a `[u8; 32]` id is available for gossip overlay identification. This id isn't required in `p2panda-sync` for anything, so it might be strange to add this trait as a bound here.

p2panda-net/src/lib.rs
sandreae: A `Topic` is actually never sent over the wire in `p2panda-net`, and per my other comment there is a separation between bounds required on topics in `-net` and `-sync`. Maybe this is confusing though? I thought the separation was good to make it clear that the id requirement was about gossip only, but if it causes confusion in other directions I'm happy to consider using the trait in `-sync` as well and adding more bounds.

sandreae:
> What happens if I'm subscribed to two different topics who result in the same topic id? Would our logic consider them as separate?
> 
> I'm looking at https://github.com/p2panda/p2panda/blob/generic-topic/p2panda-net/src/engine/engine.rs#L444-L446 and similar parts, probably we need to distinct subscriptions differently? This might have been an issue before actually.

Yeh, good point, need to inspect the behaviour here more closely. I don't believe it's possible to subscribe to two topics with the same id though, in `p2panda-net` they would be considered the same. Need to check what exactly happens, the desired behaviour would be that the second call to `subscribe` triggers the existing topic to be "unsubscribed" and all state cleaned up before then actually subscribing to it again.

This actually relates to my comment about a topic getting "out-of-sync" with itself.

adzialocha:
> > What happens if I'm subscribed to two different topics who result in the same topic id? Would our logic consider them as separate?
> > 
> > I'm looking at https://github.com/p2panda/p2panda/blob/generic-topic/p2panda-net/src/engine/engine.rs#L444-L446 and similar parts, probably we need to distinct subscriptions differently? This might have been an issue before actually.
> 
> Yeh, good point, need to inspect the behaviour here more closely. I don't believe it's possible to subscribe to two topics with the same id though, in `p2panda-net` they would be considered the same. Need to check what exactly happens, the desired behaviour would be that the second call to `subscribe` triggers the existing topic to be "unsubscribed" and all state cleaned up before then actually subscribing to it again.
> 
> This actually relates to my comment about a topic getting "out-of-sync" with itself.

It would be possible if we decouple the key to store these sessions from the topic id and make it unique (random hash or counting up a number by one).

In gossip we can keep the topic id mapping and just multiplex all operations into multiple subscriptions which use it.

p2panda-net/src/lib.rs
adzialocha: I'd try first without and then see if something is missing after some time?

p2panda-sync/src/log_sync.rs
adzialocha: Thanks for clarifying, I think this doesn't fully get across looking at the APIs, both concepts are called topic right now (for gossip and sync) and the trait bounds do not communicate themselves well (I think we need to combine the concepts or introduce a second super trait).

Ill write more thoughts in a comment with some ideas!

p2panda-net/src/lib.rs
adzialocha: Ja, I got confused here, might be easier if they are somehow consolidated, will add a comment with some thoughts

p2panda-net/src/sync/mod.rs
adzialocha: > I was thinking about the fact that a topic `T` is stored in multiple locations, both internally inside network, and externally by the user. The internal version is passed into `subscribe` and from that point is never changed, the user stored copy is exposed to any sync protocol via a `TopicMap` and this _may_ be mutated without the networks knowledge. The point I'm making is that it's possible for a topic to get "out of sync" with itself... in a way... maybe this is more of an API documentation point: "if you change a topic in any way, even if the id remains the same, it's advised that `subscribe()` be re-called".

Interesting, I'd consider Topic almost like a messaging API a user defines themselves for telling other peers what they would like to have, and similar to an API like that the message doesn't change once its sent (or you have to re-send) if sth. changed.

I can't see in the code how that change might happen, can you walk me through an example in the code? Maybe there's a way to prevent it.

sandreae: It isn't exactly visible in our code, as it's more about use of the API. The main point is that calls to `subscribe` on network need to be coordinated with updates to a `TopicMap` (the newly subscribed topic needs to be present as a key here).

It somehow relates to our discussion around "what happens if I subscribe to two topics which have the same id" as defining the expected behaviour here is relevent.

sandreae:
> > > What happens if I'm subscribed to two different topics who result in the same topic id? Would our logic consider them as separate?
> > > 
> > > I'm looking at https://github.com/p2panda/p2panda/blob/generic-topic/p2panda-net/src/engine/engine.rs#L444-L446 and similar parts, probably we need to distinct subscriptions differently? This might have been an issue before actually.
> > 
> > Yeh, good point, need to inspect the behaviour here more closely. I don't believe it's possible to subscribe to two topics with the same id though, in `p2panda-net` they would be considered the same. Need to check what exactly happens, the desired behaviour would be that the second call to `subscribe` triggers the existing topic to be "unsubscribed" and all state cleaned up before then actually subscribing to it again.
> > 
> > This actually relates to my comment about a topic getting "out-of-sync" with itself.
> 
> It would be possible if we decouple the key to store these sessions from the topic id and make it unique (random hash or counting up a number by one).
> 
> In gossip we can keep the topic id mapping and just multiplex all operations into multiple subscriptions which use it.

Ah yeh, interesting, hadn't considered multiplexing different topics which have the same gossip id. The reason for this would be mainly for having sync sessions running over different data sets, correct? As topic subscriptions which have the same gossip id will receive identical (all) gossip messages.

p2panda-net/src/sync/mod.rs
sandreae: Thinking about this again, in many cases `TopicMap` will wrap calls to some store, and it will be able to interpret whatever topic it gets, and dynamically understand the scope. So this "out-of-sync" behaviour is actually a side-effect of only some (probably naive and niche) implementations.

adzialocha (changes requested):
Generally the functionality is already doing everything we need. I suggest tackling "refactoring" issues like better stream building APIs and gossip multiplexing in another PR. My points ore mostly around how to communicate the types well across the stack (low level and high level)?

For readability and ergonomics I'd make use if `StdHash` and change the variable names accordingly as commented above, next to the other small changes.

Also, I wonder if we can make the APIs more explicit. This is an idea:

1. `Topic` as a super-trait should define the bounds required for the **object** which gets sent to another peer over sync. An object which implements that trait has the guarantee it can do it
2. `id` as a method of that trait returns 32 bytes, the topic id. This **result type** is the type which is used for gossip, *not* the object which implements the trait itself
3. `Topic` should require the object to be hashable (for data types), serializable and deserializable (for transmission during sync)

Then I think we can use `Topic` as a type in function signatures like sync initiation etc. and `[u8; 32]` everywhere else where we do gossipping.

I'm not sure if the names translate these concepts well but I'm fine finding better names later (for example "gossip topic id" and "sync / topic scope / query" etc., use `type` and so on.

I'm imagining writing something like this one day:

```rust
#[derive(Serialize, Deserialize, Hash)]
enum Query {
  AllEvents,
  Events(from: Date, to: Date),
  AllResources,
  EventResources(event: Hash),
}

impl Topic for Query {
  fn id(&self) -> [u8; 32] {
    match self {
      Self::AllEvents => EVENTS_TOPIC_ID,
      Self::Events(...) => EVENTS_TOPIC_ID,
      Self::Resources => RESOURCES_TOPIC_ID,
      Self::EventResources(event) => Hash::new(event + RESOURCES_TOPIC_ID).to_bytes(),
    }
  }
}

struct QueryMap(CalendarStore);

impl TopicMap for QueryMap {
  async fn topic_logs(&self, topic: Query) -> Vec<(PublicKey, u64)> {
    match topic {
      Query::AllEvents => store.events_logs().await,
      // ...
    }
  }
}

let (tx, tx) = network.subscribe(Query::EventResources(klangkeller)).await?;
```

Maybe it's weird to strictly tie "topic" and "topic id" together in only one trait, we could alternatively try something like `network.subscribe(topic: T, topic_id: [u8; 32])` but that gets ugly as soon as we introduce options.

.. Or have a `StreamBuilder` with methods to add a "sync query" and a "gossip overlay id" for now? This would then return your stream sender and receiver as usual but offer an extensible, explicit and not too ugly interface?

p2panda-net/src/sync/mod.rs
sandreae: Turns out we actually want to wait in `initiate_sync`, but the message variant naming was confusing. I changed the name from `FromSync::Topic` to `FromSync::HandshakeSuccess` to better explain what we are waiting for. The contained topic is required in `accept_sync` but not here in `initiate_sync`. However in both cases we want to wait for the handshake to complete.

p2panda-net/src/engine/engine.rs
```diff
@@ -426,20 +429,20 @@ impl EngineActor {
...
         self.topics
-            .earmark(topic, from_network_tx, gossip_ready_tx)
+            .earmark(topic.clone(), from_network_tx, gossip_ready_tx)
             .await;
 
         // If we haven't joined a gossip overlay for this topic yet, optimistically try to do it
         // now. If this fails we will retry later in our main loop
-        if !self.topics.has_joined(&topic).await {
-            self.join_topic(topic).await?;
+        if !self.topics.has_joined(&topic.id().into()).await {
```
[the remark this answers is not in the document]
sandreae: I changed `TopicId` to `[u8; 32]` everywhere now :+1:

p2panda-sync/src/log_sync.rs
sandreae: https://github.com/p2panda/p2panda/issues/579

mycognosist (approved):

p2panda-net/src/engine/sync.rs
```diff
@@ -1,291 +0,0 @@
-// SPDX-License-Identifier: AGPL-3.0-or-later
```
adzialocha, line 1: Is this deleted by accident? @sandreae

adzialocha (approved):

PR #578, opened 2024-10-25
Remove stream name and improve prune flag extensions

adzialocha:
* Move `PruneFlag` to `p2panda-core` and improve it's API
* Remove `StreamName` for now


p2panda-core/src/prune.rs
```diff
@@ -1,9 +1,43 @@
...
+        Self(flag)
+    }
+
+    pub fn is_set(&self) -> bool {
+        self.0
+    }
+
+    pub fn is_not_set(&self) -> bool {
+        !self.0
+    }
```
adzialocha, line 24: This method is a bit funny, but helps to have a nice `serde` usage, not sure how to do it better

p2panda-core/src/prune.rs
```diff
@@ -82,4 +119,47 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn prune_flag_can_be_optional() {
+        #[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+        struct Extensions {
+            #[serde(
+                rename = "p",
+                skip_serializing_if = "PruneFlag::is_not_set",
```
mycognosist, line 144: I don't mind the `is_not_set` method (I like how clear `PruneFlag::is_not_set` reads) but you could also use `std::ops::Not::not` instead.

Two options:

```rust
use std::ops::Not;

#[serde(skip_serializing_if = "Not::not")]

#[serde(skip_serializing_if = "<&bool>::not")]
```

p2panda-core/src/prune.rs
mycognosist: I left a suggestion in the test code.

mycognosist (approved):

p2panda-core/src/prune.rs
adzialocha: Oh, also very nice! Thank you!

PR #580, opened 2024-10-29
Refactor `p2panda-net` engine, separating "topic discovery" and "topic stream" handling

adzialocha:
This PR is an attempt to de-clutter the engine implementation in `p2panda-net` a little:

- Move gossip buffer into own module
- Move all logic around the "network-wide gossip overlay" into `TopicDiscovery`
- Move `PeerMap` into own module, now called `AddressBook`
- Move `TopicMap` into own module, now called `TopicStreams`
- `TopicStreams` is the main orchestrator of all gossip, sync and gossip buffer handling now
- Changed `SyncActor` to not use topic id's as keys but topics instead
- Multiplexer logic to account for subscriptions using the same topic or topic ids
- General doc-string and comment improvements (hopefully)


p2panda-net/Cargo.toml
```diff
@@ -4,8 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-default = ["cbor", "mdns-discovery"]
-cbor = ["dep:ciborium"]
```
adzialocha, line 8: CBOR is _always_ used to encode announcement messages for topic discovery

p2panda-net/src/engine/address_book.rs
```diff
@@ -0,0 +1,108 @@
...
+        inner
+            .known_peer_addresses
+            .entry(node_id)
+            .and_modify(|addrs| {
+                addrs.push(node_addr.clone());
+            })
+            .or_insert(vec![node_addr]);
+    }
+
+    /// Connect peer with a topic id they are interested in.
```
sandreae, line 59:
```suggestion
    /// Associate peer with a topic id they are interested in.
```

sandreae: "Connect" has too many networking connotations :sweat_smile:

p2panda-net/src/engine/mod.rs
```diff
@@ -89,31 +95,27 @@ where
         }
     }
 
+    /// Adds a peer to the address book.
+    ///
+    /// This method can be manually called to register known peers or automatically, for example by
+    /// an background "peer discovery" process.
```
sandreae, line 101:
```suggestion
    /// a background "peer discovery" process.
```

p2panda-net/src/engine/topic_discovery.rs
```diff
@@ -0,0 +1,133 @@
...
+/// Within a network peers can be interested in different topics. How topics are defined is up to
+/// the applications. Within a network there can be even multiple applications co-existing.
+///
+/// To find out which peer is interested in what topic we need a process called "topic discovery".
+/// Currently this is (rather naively) implemented as a network-wide gossip overlay where peers
+/// frequently broadcast their interests. Later we might look into other approaches, for example
+/// applying a random-walk algorithm which traverses the network and learning about it over time.
+// @TODO(adz): Would be great to already express this interface as traits so it's easier to swap
+// out the strategies with something else. The API could even look similar to our current
+// `Discovery` trait (for peer discovery), adjusted to work with topics.
```
sandreae, line 34: Yeh, agree this would be good to do. Maybe a "two onion" task ;-)

p2panda-net/src/engine/topic_streams.rs
```diff
@@ -0,0 +1,466 @@
...
+/// A stream has quite a bit of state to deal with, this includes:
+/// 1. Try to enter a gossip overlay for sending messages in "live mode" over a topic id.
+/// 2. Help the sync manager with learning about topics of interest and guide it to connect to
+///    peers for syncing up state with them.
+/// 3. Intercept and temporarily buffer incoming gossip messages of a peer when we're currently in
+///    a sync session with them. As soon as this sync session has finished we can re-play the
+///    messages. This helps reducing the number of out-of-order messages.
+/// 4. Applications can subscribe to topics multiple times, or to different topics but with the
+///    same topic ids. This stream handler multiplexes messages to the right place, even when
+///    there's duplicates.
```
sandreae, line 37: :sparkles:

p2panda-net/src/engine/topic_streams.rs
```diff
@@ -0,0 +1,466 @@
...
+                                .await
+                        }
+                    };
+
+                    if let Err(err) = result {
+                        // @TODO(adz): This fails silently right now, shouldn't this be propagated
+                        // further to the user?
+                        error!("failed broadcasting message to gossip for topic {topic:?}: {err}");
+                        break;
+                    }
```
sandreae, line 153: And the reason this would fail is if there was an error sending on the gossip_actor_tx channel?

p2panda-net/src/engine/topic_streams.rs
```diff
@@ -0,0 +1,466 @@
...
+            })?;
+        }
+
+        Ok(())
+    }
+
+    /// Process sync session finishing.
+    pub async fn on_sync_done(&mut self, topic: T, node_id: NodeId) -> Result<()> {
+        let topic_id = topic.id();
+        let counter = self.gossip_buffer.unlock(node_id, topic_id);
```
sandreae, line 340: Just to check I'm understanding the behaviour here:
- a peer subscribes to two topics `A`+`B` which both have the same topic id `1`
- two independent sync sessions kick-in, one for each topic
- one buffer is created and used for both sync sessions (buffer counter == `2`)
- sync session for topic `A` finishes (buffer counter == `1`)
- sync session for topic `B` finishes (buffer counter == `0`)
- buffer counter is `0` so all messages are drained and forwarded to the two existing subscription channels for topic `A`+`B` (because both have topic id `1`)

p2panda-net/src/network.rs
```diff
@@ -166,8 +175,8 @@ where
 
     /// Sets the sync protocol for this network.
     ///
-    /// If a sync protocol is provided, a sync session will be completed before entering gossip
-    /// mode with any known peers with whom we share topics of interest.
+    /// If a sync protocol is provided, a sync session will be completed with any known peers with
+    /// whom we share topics of interest.
     pub fn sync(mut self, protocol: impl for<'a> SyncProtocol<'a, T> + 'static) -> Self {
```
sandreae, line 180: I could imagine that we soon want to be able to set sync protocol for a certain set of topics, that would be cool for supporting multiple applications sharing the same network. Just a reflection now....

sandreae (approved):
This looks great to me, very pleasing to have the code arrive at these logical modules. Particularly nice to have `TopicStreams` isolated as it has a lot of state handling.

I only had a clarification request and some typo spots so I'll approve already now. Thanks for the good work.

p2panda-net/src/engine/topic_streams.rs
adzialocha: Sounds right!

We don't always need to necessarily block for incoming sync requests when there's no bi-directional data exchange, but since we don't control the actual sync implementation this is the most straight-forward.

p2panda-net/src/engine/topic_streams.rs
adzialocha: Right, this should only happen if the receiver got dropped and we didn't close everything properly (which is the case right now), but it's otherwise fine?

adzialocha: It doesn't propagate but would it be useful to handle this error? https://github.com/p2panda/p2panda/blob/net-cleanup/p2panda-net%2Fsrc%2Fengine%2Fgossip.rs#L107

It means that your message was not broadcasted which maybe is good to know?

p2panda-net/src/engine/engine.rs
```diff
@@ -255,180 +200,112 @@ where
...
-                .await?;
+        self.address_book.add_peer(node_addr).await;
 
-            // Do not attempt peer sync if the topic_id is the network id.
-            if topic_id == self.network_id {
-                return Ok(());
-            }
-        }
+        // Hot path: Attempt starting topic discovery as soon as we've learned about at least one
+        // peer. If this fails we'll try again soon again in our internal loop.
```
mycognosist, line 274:
```suggestion
        // peer. If this fails we'll try again soon in our internal loop.
```

p2panda-net/src/engine/engine.rs
```diff
@@ -439,410 +316,71 @@ where
...
 
         Ok(())
     }
 
-    /// Process an inbound message from the network.
+    /// Process an inbound message from one of the gossip overlays.
+    ///
+    /// If the network comes from the "network-wide" gossip overlay (determined by the "network
+    /// id"), then it gets handled by the "topic discovery" mechanism. Otherwise it is a regular,
+    /// custom application-related gossip overlay around a "topic id".
```
mycognosist, line 343:
```suggestion
    /// If the message comes from the "network-wide" gossip overlay (determined by the "network
    /// id"), then it gets handled by the "topic discovery" mechanism. Otherwise it is from a regular,
    /// custom application-related gossip overlay around a "topic id".
```

p2panda-net/src/engine/topic_discovery.rs
```diff
@@ -0,0 +1,133 @@
...
+    #[default]
+    Idle,
+    Pending,
+    Active,
+}
+
+/// Manages the "topic discovery" background process.
+///
+/// Within a network peers can be interested in different topics. How topics are defined is up to
+/// the applications. Within a network there can be even multiple applications co-existing.
```
mycognosist, line 26: These changes will still need a `cargo fmt` run (I'm making the suggestion in the GH UI to save some time).

```suggestion
/// Peers can be interested in different topics within a single network. How topics are defined is up to
/// the applications. Multiple applications can even co-exist within the same network.
```

p2panda-net/src/engine/topic_discovery.rs
```diff
@@ -0,0 +1,133 @@
...
+///
+/// To find out which peer is interested in what topic we need a process called "topic discovery".
+/// Currently this is (rather naively) implemented as a network-wide gossip overlay where peers
+/// frequently broadcast their interests. Later we might look into other approaches, for example
+/// applying a random-walk algorithm which traverses the network and learning about it over time.
```
mycognosist, line 31:
```suggestion
/// applying a random-walk algorithm which traverses the network and learns about it over time.
```

p2panda-net/src/engine/topic_discovery.rs
```diff
@@ -0,0 +1,133 @@
...
+// @TODO(adz): Would be great to already express this interface as traits so it's easier to swap
+// out the strategies with something else. The API could even look similar to our current
+// `Discovery` trait (for peer discovery), adjusted to work with topics.
+pub struct TopicDiscovery {
+    address_book: AddressBook,
+    gossip_actor_tx: mpsc::Sender<ToGossipActor>,
+    network_id: NetworkId,
```
mycognosist, line 38: I wonder if we could get rid of the `network_id` field here, since `address_book` already includes the network id internally? You could add a `network_id()` method to `AddressBook` for convenient look-up.

p2panda-net/src/engine/topic_streams.rs
```diff
@@ -0,0 +1,466 @@
...
+                        // here.
+                        //
+                        // For now this is fine as the user has two options:
+                        //
+                        // 1. They're combining sync with gossip. If the user stores all messages
+                        //    before sending them (which they probably always should if they care
+                        //    about consistency) sync will make sure that peers will catch up with
+                        //    this data as soon as they connect to somebody.
+                        // 2. They're don't care about consistency, but they are waiting for the
+                        //    "gossip ready" signal before they send any messages.
```
mycognosist, line 133:
```suggestion
                        // 2. They don't care about consistency, but are waiting for the
                        //    "gossip ready" signal before sending any messages.
```

mycognosist (approved):
Woah, this is amazing work! I love the separation you've achieved and the doc comments offer useful context.

I've left a few suggestions regarding comments (mostly grammar changes and typo fixes). Then there's also one comment about the `network_id` field. I'm happy for this to be merged whenever you wish.

p2panda-net/src/engine/topic_streams.rs
adzialocha: I'll add it to the roadmap :+1:

p2panda-net/src/engine/topic_discovery.rs
adzialocha: I think intuitively I wouldn't understand why the address book is the source for knowing about the network id (it _also_ needs to know it but I wouldn't see it as the solely source)

mycognosist: Agreed. That makes sense to me!

p2panda-net/src/network.rs
sandreae: I meant that the you could register a sync protocol per topic (or collection of topic). The method signature might look like this:

```rust
pub fn sync(mut self, topics: Vec<T>, protocol: impl for<'a> SyncProtocol<'a, T> + 'static) -> Self {...}
```

sandreae: Nothing for this PR though of course :pray:

PR #581, opened 2024-11-05
Propagate sync errors to the manager and inform the engine actor

mycognosist:
WIP.

## 📋 Checklist

### DIFF
--- a/p2panda-net/src/sync/manager.rs
+++ b/p2panda-net/src/sync/manager.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 
 use anyhow::{Context, Error, Result};
 use iroh_net::{Endpoint, NodeId};
-use p2panda_sync::{SyncProtocol, Topic};
+use p2panda_sync::{SyncError, SyncProtocol, Topic};
 use thiserror::Error;
 use tokio::sync::mpsc::{self, Receiver, Sender};
 use tokio::time::Duration;
@@ -61,9 +61,9 @@ enum SyncAttemptError {
     #[error("sync attempt failed due to connection or stream error")]
     Connection,
 
-    /// Error occurred while initiating a sync session.
-    #[error("sync attempt failed due to sync protocol error")]
-    Sync,
+    /// Error occurred while initiating or accepting a sync session.
+    #[error(transparent)]
+    Sync(#[from] SyncError),
 }
 
 /// An API for scheduling outbound connections and sync attempts.
@@ -246,8 +246,7 @@ where
             sync_protocol,
             engine_actor_tx,
         )
-        .await
-        .map_err(|_| SyncAttemptError::Sync)?;
+        .await?;
 
         // Clean-up the streams.
         send.finish()?;
@@ -257,9 +256,10 @@ where
         Ok(())
     }
 
-    /// Remove the given topic from the set of active sync sessions for the given peer. Reschedule
-    /// a sync attempt if the failure was caused by a connection error. Otherwise, drop the attempt
-    /// and schedule the next pending attempt.
+    /// Remove the given topic from the set of active sync sessions for the given peer and notify
+    /// the engine actor of the failed sync attempt. Reschedule a sync attempt if the failure was
+    /// caused by a connection error. Otherwise, drop the attempt and schedule the next pending
+    /// attempt.
     async fn complete_failed_sync(
         &mut self,
         sync_attempt: SyncAttempt<T>,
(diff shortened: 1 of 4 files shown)


p2panda-net/src/sync/mod.rs
```diff
@@ -105,18 +113,14 @@ where
...
             Box::new(&mut send),
             Box::new(&mut recv),
             Box::new(&mut sink),
         )
-        .await;
-
-    if let Err(err) = result {
-        error!("sync protocol initiation failed: {err}");
-    }
+        .await?;
```
mycognosist, line 123: There's a slight asymmetry here, in that `initiate_sync()` returns the error to the caller (`connect_and_sync()` in the manager) whereas `accept_sync()` uses the oneshot channel to inform the engine actor more directly.

I'm not unhappy about the slight difference, since returning the error in `initiate` allows us to match on connection vs. sync errors and take appropriate action (rescheduling an attempt or informing the engine actor).

adzialocha: An comment in this place might be nice!

p2panda-net/src/sync/manager.rs
```diff
@@ -257,9 +256,15 @@ where
...
 
-    /// Remove the given topic from the set of active sync sessions for the given peer. Reschedule
-    /// a sync attempt if the failure was caused by a connection error. Otherwise, drop the attempt
-    /// and schedule the next pending attempt.
+    /// Clean up after a failed sync attempt.
+    ///
+    /// First remove the given topic from the set of active sync sessions for the given peer. Then
+    /// reschedule a sync attempt if the failure was caused by a connection error and the maximum
+    /// number of retry attempts has not already been exceeded. Otherwise, notify the engine actor,
+    /// drop the attempt and schedule the next pending attempt.
```
adzialocha, line 264: This part kinda repeats what the code describes, I wonder if it's just too much information which will get outdated fast as soon as someone touches the code. I'd try to only write things in comments the code _can't_ convey.

p2panda-net/src/sync/mod.rs
```diff
@@ -105,18 +113,24 @@ where
...
+    //
+    // Any errors occuring during the sync protocol session are returned to the caller, which in
+    // this case is the `connect_and_sync()` method in the sync manager. Returning the error allows
+    // the sync manager to handle the response and take different paths based on the type of error
+    // which occurred (either a connection or sync error). This allows, for example, retry attempts
+    // to be made for the same topic and peer.
+    //
+    // When an error happens while _accepting_ a sync session (as in `accept_sync()` below) we
+    // simply notify the engine actor directly, since the acceptor does not need to track
+    // reattempts.
```
adzialocha, line 125: Similar here, I think the second paragraph already describes perfectly everything to prevent a person from wondering why we don't have a oneshot channel here, the rest should be part of the code.

p2panda-net/src/sync/mod.rs
```diff
@@ -141,59 +155,87 @@ where
...
+
+                        continue;
+                    }
+
+                    // 2. Data Sync Phase.
+                    // ~~~~~~~~~~~~~~~~~~~
+                    //
+                    // The topic must be known at this point in order to process further messages.
+                    let Some(topic) = &topic else {
+                        error!("topic not received");
```
adzialocha, line 212: This silently fails when the handshake phase was not fullfilled, I assume we should catch that as well as this indicates that the other peer didn't follow the protocol?

p2panda-net/src/sync/mod.rs
```diff
@@ -53,12 +53,18 @@ where
...
             while let Some(message) = rx.recv().await {
-                // We expect the first message to be a topic.
+                // 1. Handshake Phase.
+                // ~~~~~~~~~~~~~~~~~~~
+                //
+                // At the beginning of every sync session the "initiating" peer needs to send over
+                // the topic to the "accepting" peer during the handshake phase. This is the first
+                // message we're expecting:
                 if let FromSync::HandshakeSuccess(_) = &message {
                     if sync_handshake_success {
```
adzialocha, line 69: Here we also end up with a protocol violation which gets silently caught

adzialocha (commented):
Next to some comment-style nitpicks I'm not sure if this PR covers _all_ potential error cases.

I wonder if errors above the sync implementation get handled correctly, see my comments.

Additionally I'm not sure how to deal with all of these channel errors, we should at least look at each of them and try to understand if they can break other parts of the application as they are all panics currently - closing channels can also occur during a graceful shutdown.

I'm fine though to tackle that in another PR.

adzialocha:
> I wonder if errors above the sync implementation get handled correctly, see my comments.

Thinking about this a little bit more I assume we have different layers of errors here which need different handling:

1. Something went wrong during my `Sync` implementation -> `SyncError`
2. The `Sync` implementation is incorrect / doesn't fullfil our "2. Phase` (Handshake + Data) flow -> ???, is this worth a panic? Feels like a buggy implementation
3. Channels are failing because other parts of the application are shutting down gracefully or crashed -> Not much we need to do maybe, but just making sure it doesn't crash uncontrolled when it shouldn't?

p2panda-net/src/sync/mod.rs
mycognosist: It's definitely a protocol violation. The question for me is: what do we do with that information? From the perspective of the engine, no sync session has actually started in this case; it only "starts" when the accepting peer receives the topic and then sends the `ToEngineActor::SyncHandshakeSuccess` event.

I can imagine potentially having a "naughty list" for peers who repeatedly try to run sync sessions with faulty protocols. In that case, we might want to be reporting this peer to some higher-level (the sync connection handler maybe, or the engine). But we would still only know the peer (not the topic).

p2panda-net/src/sync/mod.rs
mycognosist: Good catch. In this case we should send a `ToEngineActor::SyncFailed` event so the buffers can be unlocked and drained.

p2panda-net/src/sync/mod.rs
adzialocha: > It's definitely a protocol violation. The question for me is: what do we do with that information?

I'd think that it's more about "how can I avoid future me to deal with bugs and footguns" and a clear API surface with errors is one step towards that, logging usually only helps if you know what you're looking for.

I don't want to suggest this needs urgent fixing but lets maybe do a further ticket for it?

mycognosist:
> Thinking about this a little bit more I assume we have different layers of errors here which need different handling:

That's a useful separation of error types / handling scenarios.

>     2. The `Sync` implementation is incorrect / doesn't fullfil our "Two Phase" (Handshake + Data) flow -> ???, is this worth a panic? Feels like a buggy implementation ❓
>
> What makes looking at these errors so tricky is that some of them happen inside a tokio task, so extra work would be required if we somehow want to propagate this up, but maybe it's not always necessary (for example on a shutdown).

I think it could be worth the extra work to return `SyncError::UnexpectedBehaviour` errors from the tokio task when protocol violations occur, then match on that and return the error if it occurs.

adzialocha:
> I think it could be worth the extra work to return `SyncError::UnexpectedBehaviour` errors from the tokio task when protocol violations occur, then match on that and return the error if it occurs.

If the sync protocol implementation was correct but a remote peer did something wrong it should be `UnexpectedBehavior` but here this failed and wasn't caught by the implementer, so the implementation is faulty, I'd assume it's a `Critical` error.

adzialocha (approved):

PR #584, opened 2024-11-14
Introduce supertraits for `LogId` and `Extensions`

mycognosist:
Aims to improve API readability and ergonomics by grouping trait bounds into supertraits for `LogId` and `Extensions`. 

The new supertraits are used in `p2panda-store`, `p2panda-sync` and `p2panda-engine`.

### DIFF
--- a/p2panda-core/src/extensions.rs
+++ b/p2panda-core/src/extensions.rs
@@ -6,9 +6,15 @@ use serde::{Deserialize, Serialize};
 
 use crate::Header;
 
-pub trait Extensions: Clone + Debug + Default + Send + Sync {}
+pub trait Extensions:
+    Clone + Debug + Default + for<'de> Deserialize<'de> + Serialize + Send + Sync
+{
+}
 
-impl<T> Extensions for T where T: Clone + Debug + Default + Send + Sync {}
+impl<T> Extensions for T where
+    T: Clone + Debug + Default + for<'de> Deserialize<'de> + Serialize + Send + Sync
+{
+}
 
 #[derive(Clone, Default, Debug, Serialize, Deserialize)]
 pub struct DefaultExtensions {}
--- a/p2panda-core/src/extensions.rs
+++ b/p2panda-core/src/extensions.rs
@@ -6,6 +6,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::Header;
 
+/// Extensions can be used to define custom fields in the operation header.
 pub trait Extensions:
     Clone + Debug + Default + for<'de> Deserialize<'de> + Serialize + Send + Sync
 {
--- a/p2panda-core/src/extensions.rs
+++ b/p2panda-core/src/extensions.rs
@@ -1,8 +1,14 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+use std::fmt::Debug;
+
 use serde::{Deserialize, Serialize};
 
 use crate::Header;
+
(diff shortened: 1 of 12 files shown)


p2panda-sync/src/log_sync.rs
```diff
@@ -50,18 +49,9 @@ pub struct LogSyncProtocol<TM, L, E> {
...
-        + Eq
-        + StdHash
-        + Send
-        + Sync
-        + for<'de> Deserialize<'de>
-        + Serialize
-        + 'a,
-    E: Clone + Debug + Default + Send + Sync + for<'de> Deserialize<'de> + Serialize + 'a,
+    TM: TopicMap<T, Logs<L>>,
+    L: LogId + for<'de> Deserialize<'de> + Serialize + 'a,
```
mycognosist, line 53: We run into some complexity if we add the `for<'de> Deserialize<'de>` bound directly to the `LogId` trait (specifically, an "implementation of `Deserialize` is not general enough" error). I imagine there's a work-around but it'll be quite technical and I don't want to pour time into understanding the solution right now.

`Serialize` could be included as an implicit bound but I think it's clearer if `Deserialize` and `Serialize` are both explicitly added as bounds.

This applies to `LogId` and `Extensions`.

p2panda-core/src/extensions.rs
```diff
@@ -1,9 +1,15 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+use std::fmt::Debug;
+
 use serde::{Deserialize, Serialize};
 
 use crate::Header;
 
+pub trait Extensions: Clone + Debug + Default + Send + Sync {}
```
mycognosist, line 9: I'd welcome any ideas / input regarding a doc comment for `Extensions`.

p2panda-sync/src/log_sync.rs
adzialocha: Oh dammit! How did we make it work for `Topic`? There we have the serde traits defined, I wonder what makes it problematic, is it the defaults?

p2panda-core/src/extensions.rs
adzialocha: Something like: "Extensions can be used to define custom fields in the operation header"?

adzialocha (commented):
Are there other places where we can update the new trait bounds, for example in `p2panda-engine`?

p2panda-sync/src/log_sync.rs
mycognosist: So, thankfully this issue does not actually apply to `Extensions`; it works fine for us to include the `Deserialize` and `Serialize` bounds there.

It's just an issue for `LogId`. If we add the bounds to the supertrait, the memory store breaks but log sync is fine. If we remove the bounds then we have to include them explicitly for log sync. This is just in a single place so maybe it's not a terrible outcome?

adzialocha (approved):

p2panda-sync/src/log_sync.rs
adzialocha: Makes sense! Also, `LogId` is quite a special thing, the core data types do not require it, but it's used as soon as data is persisted or synced, for the latter we need serde as it is sent over the wire (at least for log-height sync, maybe it's not required for set reconciliation).

sandreae: Yeh, I agree that `LogId` is a bit different from `Extensions`, it's not an inherently "on the wire" data type, whereas `Extensions` are. So having to add serde traits where needed maybe isn't so bad.

PR #588, opened 2024-11-22
Fix multi-session sync bug, introduce resync logic and improve sync config

mycognosist:
- Fix bugs in sync attempt state-tracking which resulted in multiple concurrent attempts
- Introduce `SyncConfiguration` type (with internal `ResyncConfiguration`) for fine-grained control over sync sessions
- Add resync ability

### DIFF
--- a/p2panda-net/src/network.rs
+++ b/p2panda-net/src/network.rs
@@ -17,7 +17,7 @@ use iroh_net::relay::{RelayMap, RelayNode};
 use iroh_net::{Endpoint, NodeAddr, NodeId};
 use p2panda_core::{PrivateKey, PublicKey};
 use p2panda_discovery::{Discovery, DiscoveryMap};
-use p2panda_sync::{SyncProtocol, Topic};
+use p2panda_sync::Topic;
 use tokio::sync::{broadcast, mpsc, oneshot};
 use tokio::task::{JoinError, JoinSet};
 use tokio_util::sync::CancellationToken;
@@ -28,7 +28,7 @@ use crate::addrs::DEFAULT_STUN_PORT;
 use crate::config::{Config, DEFAULT_BIND_PORT};
 use crate::engine::Engine;
 use crate::protocols::{ProtocolHandler, ProtocolMap};
-use crate::sync::SYNC_CONNECTION_ALPN;
+use crate::sync::{SyncConfiguration, SYNC_CONNECTION_ALPN};
 use crate::{NetworkId, RelayUrl, TopicId};
 
 /// Maximum number of streams accepted on a QUIC connection.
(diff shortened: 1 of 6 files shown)


p2panda-net/src/sync/manager.rs
```diff
@@ -176,9 +185,29 @@ where
...
+                    if let Err(err) = self.schedule_attempt(sync_attempt).await {
+                        // The attempt will fail if the sync queue is full, indicating that a high
+                        // volume of sync sessions are underway. In that case, we drop the attempt
+                        // completely. Another attempt will be scheduled when the next announcement of
+                        // this peer-topic combination is received from the network-wide gossip
+                        // overlay.
+                        error!("failed to schedule sync attempt: {}", err)
+                    }
+                }
+                _ = attempt_resync_interval.tick() => {
```
mycognosist, line 199: I wonder if there's a better solution for this resync behaviour...

Right now we're using a simple `VecDeque`: every 5 seconds we pull an attempt from the queue and schedule it if the previous attempt was completed more than 60 seconds ago. If not, we push it to the back of the queue.

Is a 5 second interval frequent enough? What if we end up with 50 or 100 attempts in the queue? Do we need a solution that offers reattempt timing guarantees (e.g. reattempt will be made more than 60 seconds after the previous successful attempt, but not more than 70 seconds after)?

p2panda-net/src/network.rs
```diff
@@ -176,9 +178,15 @@ where
...
     ///
     /// If a sync protocol is provided, a sync session will be completed with any known peers with
-    /// whom we share topics of interest.
-    pub fn sync(mut self, protocol: impl for<'a> SyncProtocol<'a, T> + 'static) -> Self {
+    /// whom we share topics of interest. The resync flag determines whether sync should only be
+    /// completed once or repeated at regular intervals.
+    pub fn sync(
+        mut self,
+        protocol: impl for<'a> SyncProtocol<'a, T> + 'static,
+        resync: bool,
```
adzialocha, line 186: Would be cool to pass a whole `SyncConfiguration` struct here, so we can add further options in the future. There's already a few others which are currently hard-coded which could live here.

mycognosist: Ya, good idea!

p2panda-net/src/sync/manager.rs
adzialocha: The logic looks fine as-is to me right now.

"First ever sync attempts" should always have higher priority I believe and "> 1th attempts" lower priority, but I think this is currently given with the two separate "priority" channels (`inbox` vs. `resync_queue`).

It gets hard to reason about timings when there's a higher priority channel involved with unknown number of items, so I think it's fine to say we can't define the upper timing limit, only the minimum.

adzialocha (approved):

PR #589, opened 2024-11-25
Sign topic discovery messages

sandreae:
- [x] add signature and public key to `TopicDiscoveryMessage` 
- [x] attempt to join gossip overlay and sync with the _signing_ peer (rather than the peer which delivered the message)
- [x] don't notify engine of sync error on connection failures (after x retry attempts)

### DIFF
--- a/p2panda-net/src/network.rs
+++ b/p2panda-net/src/network.rs
@@ -886,7 +886,7 @@ mod tests {
     use crate::addrs::DEFAULT_STUN_PORT;
     use crate::bytes::ToBytes;
     use crate::config::Config;
-    use crate::network::sync_protocols::{DummyProtocol, PingPongProtocol};
+    use crate::network::sync_protocols::PingPongProtocol;
     use crate::sync::SyncConfiguration;
     use crate::{NetworkBuilder, RelayMode, RelayUrl, TopicId};
 
--- a/p2panda-net/src/network.rs
+++ b/p2panda-net/src/network.rs
@@ -1243,17 +1243,17 @@ mod tests {
         node_2.add_peer(node_1_addr).await.unwrap();
         node_3.add_peer(node_2_addr).await.unwrap();
 
-        // Subscribe to the same topic from both nodes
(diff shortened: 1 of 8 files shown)


p2panda-net/src/sync/manager.rs
```diff
@@ -308,18 +308,14 @@ where
...
             match err {
+                // If the sync attempt failed because of a connection error we want to retry up to
+                // `max_retry_attempts`. If error occurs after this we simply stop trying without
+                // informing the engine as it never knew the attempts were occurring.
                 SyncAttemptError::Connection => {
                     warn!("sync attempt failed due to connection error");
                     if sync_attempt.attempts <= self.config.max_retry_attempts {
                         self.reschedule_attempt(sync_attempt).await?;
                         return Ok(());
-                    } else {
```
sandreae, line 316: Notifying the engine of `SyncFailed` here was causing a panic when the gossip buffer was unlocked as it was never locked in the first place. @mycognosist @adzialocha do you mind confirming that this doesn't have any other repercussions I'm not aware of?

adzialocha: Yes! I stumbled upon it as well but wasn't sure.

On a connection fail the session never starts, but a `SyncFailed` is sent anyhow, so indeed it should be removed (or treated differently by the engine).

adzialocha:
I'd give the panic a more meaningful message maybe

p2panda-net/src/network.rs
```diff
@@ -804,6 +805,10 @@ mod sync_protocols {
                     }
                     Message::Pong => {
                         debug!("pong message received");
+                        app_tx
+                            .send(FromSync::Data("PONG".as_bytes().to_owned(), None))
+                            .await
+                            .unwrap();
```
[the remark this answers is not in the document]
sandreae: Haha, actually, yes :sweat_smile: it's updating the `PingPong` test protocol to send messages to the app layer, which we use in the new tests.

PR #592, opened 2024-11-27
Fix CBOR decoder when buffer contains less than a full frame

adzialocha:
This PR fixes an bug in the CBOR decoder where it always advanced the buffer, even when it only contained an incomplete message frame.

### DIFF
--- a/p2panda-sync/src/cbor.rs
+++ b/p2panda-sync/src/cbor.rs
@@ -115,16 +115,15 @@ where
     }
 }
 
-/// Returns a reader for your data type, receiving CBOR encoded byte-streams of it and handling the
-/// framing of them automatically.
+/// Returns a reader for your data type, automatically decoding CBOR byte-streams and handling the
+/// message framing.
 ///
 /// This can be used in various sync protocol implementations where we need to receive data via a
 /// wire protocol between two peers.
 ///
 /// This is a convenience method if you want to use CBOR encoding and serde to handle your wire
 /// protocol message encoding and framing without implementing it yourself. If you're interested in
-/// your own approach you can either implement your own `FramedRead` or `Sink` (if you're not
-/// interested in the tokio codec traits).
+/// your own approach you can either implement your own `FramedRead` or `Sink`.
 pub fn into_cbor_stream<'a, M>(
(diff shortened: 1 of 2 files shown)


adzialocha:
Oupsi, the https://github.com/p2panda/p2panda/pull/592/commits/a098b56c1997ec01a50e91a13ad18a2d2fa19289 commit sneaked into this PR and I'm too lazy to remove it now, hope that's okay 😆

p2panda-sync/src/cbor.rs
```diff
@@ -38,10 +46,14 @@ where
 {
     type Error = SyncError;
 
+    /// Encodes an serializable item into CBOR bytes and adds them to the buffer.
```
mycognosist, line 49:
```suggestion
    /// Encodes a serializable item into CBOR bytes and adds them to the buffer.
```

p2panda-sync/src/cbor.rs
```diff
@@ -54,17 +66,44 @@ where
     type Item = T;
     type Error = SyncError;
 
+    /// CBOR decoder method taking as an argument the bytes that has been read so far, and when it
+    /// is called, it will be in one of the following situations:
```
mycognosist, line 70:
```suggestion
    /// CBOR decoder method taking as an argument the bytes that have been read so far; when called,
    /// it will be in one of the following situations:
```

p2panda-sync/src/cbor.rs
```diff
@@ -54,17 +66,44 @@ where
...
+        // Attempt decoding the buffer and remember how many bytes we've advanced it doing that.
+        //
+        // This will succeed in case 2. and 3.
+        let result: Result<Self::Item, _> = decode_cbor(&mut bytes);
+        let ending = bytes.len();
+
         match result {
-            Ok(item) => Ok(Some(item)),
+            Ok(item) => {
+                // We've successfully could read one full frame from the buffer. We're finally
```
mycognosist, line 88:
```suggestion
                // We've successfully read one full frame from the buffer. We're finally
```

mycognosist (approved):
Amazing work! You slayed the final boss 🙌🏻 

Beautiful tests; super clear. I added some small suggested changes relating to grammar in the comments but they're not blocking.

PR #593, opened 2024-11-27
Improve sync error handling

adzialocha:
General improvements of error handling in the sync module of `p2panda-net` which can be summarized as:

* Handle _all_ errors
* Consistent `SyncStart` events for both acceptors and initiators
* Do not send `SyncDone` events on errors in initiators
* Consider broken pipes an "unexpected behaviour" error and not something critical as it might have happened because the _other_ peer crashed


p2panda-net/src/engine/engine.rs
```diff
@@ -47,7 +47,7 @@ pub enum ToEngineActor<T> {
         topic_id: [u8; 32],
     },
     SyncStart {
-        topic: T,
+        topic: Option<T>,
```
adzialocha, line 50: With making `topic` optional we can now also send `SyncStart` events for acceptors. This is not very much used, but the symmetry helps to understand what is going on (and might prevent future bugs when we start using the `on_sync_start` method).

p2panda-net/src/sync/accept.rs
```diff
@@ -0,0 +1,221 @@
...
+    R: AsyncRead + Send + Unpin,
+{
+    debug!("accept sync session with peer {}", peer);
+
+    engine_actor_tx
+        .send(ToEngineActor::SyncStart { topic: None, peer })
+        .await
+        .map_err(|err| {
+            SyncError::Critical(format!("engine_actor_tx failed sending sync start: {err}"))
+        })?;
```
adzialocha, line 38: Acceptors send the `SyncStart` event now as well at the beginning.

p2panda-net/src/sync/accept.rs
```diff
@@ -0,0 +1,221 @@
...
+    //
+    // Additionally, the task forwards any synced application data straight to the engine.
+    let glue_task_handle: JoinHandle<Result<(), SyncError>> = tokio::spawn(async move {
+        let mut topic = None;
+
+        loop {
+            tokio::select! {
+                biased;
+
+                Ok(err) = &mut sync_error_rx => {
```
adzialocha, line 65: This was moved up again in the `select!` priority. We need to catch errors on the oneshot channel, otherwise we can never propagate the error further.

p2panda-net/src/sync/accept.rs
```diff
@@ -0,0 +1,221 @@
...
+                        .send(ToEngineActor::SyncFailed {
+                            peer,
+                            topic: topic.clone(),
+                        })
+                        .await
+                        .map_err(|err| {
+                            SyncError::Critical(
+                                format!("engine_actor_tx failed sending sync failed: {err}")
+                            )
+                        })?;
```
adzialocha, line 76: All channels propagate their send errors up now

p2panda-net/src/sync/accept.rs
```diff
@@ -0,0 +1,221 @@
...
+                    // the topic of the "initiating" peer during the handshake phase. This is
+                    // _always_ the first message we're expecting:
+                    if let FromSync::HandshakeSuccess(handshake_topic) = message {
+                        // It should only be sent once so topic should be `None` now.
+                        if topic.is_some() {
+                            return Err(
+                                SyncError::Critical(
+                                    "received topic twice from sync session in handshake phase"
+                                    .into()
+                                )
```
adzialocha, line 100: All violations of the "2 Phase Protocol Flow" are considered critical errors (as we can assume they are due to buggy `Sync` implementations.

p2panda-net/src/sync/accept.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    // We're expecting the task to exit with a result soon, we're awaiting it here ..
+    let glue_task_result = glue_task_handle
+        .await
+        .map_err(|err| SyncError::Critical(format!("glue task handle failed: {err}")))?;
+
+    // .. to inform some brave developer who will read the error logs ..
+    if let Err(SyncError::Critical(err)) = &glue_task_result {
+        error!("critical error in sync protocol: {err}");
```
adzialocha, line 216: There is other places where we print logs on errors, I hope this is not getting too much?

p2panda-net/src/sync/accept.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    // .. and forward it further!
+    glue_task_result
```
adzialocha, line 220: All errors, including from the task but also from the sync session itself get forwarded

p2panda-net/src/sync/initiate.rs
```diff
@@ -0,0 +1,173 @@
...
+        }
+        return Err(err);
+    }
+
+    engine_actor_tx
+        .send(ToEngineActor::SyncDone { peer, topic })
+        .await
+        .map_err(|err| {
+            SyncError::Critical(format!("engine_actor_tx failed sending sync done: {err}"))
+        })?;
```
adzialocha, line 170: We should only send a `SyncDone` when _no_ error occurred in the initiator.

adzialocha: `SyncFailed` is handled in the sync manager, that makes the code a little bit hard to read though?

p2panda-net/src/sync/initiate.rs
```diff
@@ -0,0 +1,173 @@
...
+    // The same we're doing with errors coming from the sync protocol implementation itself.
+    if let Err(err) = result {
+        match &err {
+            SyncError::Critical(err) => {
+                error!("critical error in sync protocol: {err}");
+            }
+            _ => {
+                warn!("error in sync protocol: {err}");
+            }
+        }
```
adzialocha, line 161: Similarly to above, I'm not sure if this is not too much logging (we might have other places which already log on sync errors).

p2panda-net/src/sync/tests.rs
```diff
@@ -0,0 +1,456 @@
...
+            if let FailingProtocol::InitiatorSendsTopicTwice = *self {
+                sink.send(ProtocolMessage::Topic(topic.clone())).await?;
+            }
+
+            // Simulate some critical error which occurred inside the sync session.
+            if let FailingProtocol::InitiatorFailsCritical = *self {
+                return Err(SyncError::Critical(
+                    "something really bad happened in the initiator".to_string(),
+                ));
+            }
```
adzialocha, line 73: I've moved the critical failure higher in the protocol so we can test what happens if the connection suddenly disappears.

p2panda-net/src/sync/tests.rs
```diff
@@ -0,0 +1,456 @@
...
+        Some(ToEngineActor::SyncStart { .. })
+    ));
+
+    assert!(matches!(
+        rx_acceptor.recv().await,
+        Some(ToEngineActor::SyncFailed { .. })
+    ));
+
+    // Expected handler results.
+    assert_eq!(initiator_handle.await.unwrap(), Ok(()));
```
adzialocha, line 362: Interestingly we can't detect a closed pipe if the acceptor "crashed" .. I'm not sure why that is.

p2panda-net/src/engine/engine.rs
mycognosist: Nice, I like the symmetry. This also gives us the option to record sync session duration (and other stats) from the "accepting" side in the future.

p2panda-net/src/sync/accept.rs
mycognosist: In this case the error would be logged twice, right? Once here and once in `src/sync/handler.rs` L66:

`error!("sync failure as acceptor: {}", err)`

I'd prefer just to keep one of them but it's not a blocker for me.

p2panda-net/src/sync/accept.rs
mycognosist: This is great! Really nice to condense everything into a single result.

p2panda-net/src/sync/initiate.rs
mycognosist: It does make it a little confusing (to have `SyncDone` here but not `SyncFailed`), though I'm not sure if there's a desirable alternative; it needs to be handled by the manager because we don't interpret it as a failure in all cases (such as a connection error when we still have more reattempts to go through).

Perhaps the second best option is to include a comment in `initiate.rs` (maybe L162) which explains that `SyncFailed` is sent by the sync manager according to the current state for this specific peer-topic attempt.

p2panda-net/src/sync/initiate.rs
mycognosist: In this case we also log in `src/sync/manager.rs` L277 or L291 (depending on error type) but we don't differentiate between sync error types (ie. we don't specifically match on a critical error).

Again, would be nice to collapse these into a single log (just because it might be confusing when debugging in the future: "is this one thing going wrong or two?").

p2panda-net/src/sync/tests.rs
mycognosist: I really like the way you separated out the helper function for running each protocol impl; it makes the actual tests very easy to reason about.

mycognosist (approved):
I've given some feedback about logs but overall this looks great! Feels good to be making these incremental improvements.

p2panda-net/src/sync/accept.rs
adzialocha: Okay, let me remove one of them then 👍🏻

p2panda-net/src/sync/initiate.rs
adzialocha: Thank you! That makes sense. I agree, let's keep it as is and I'll add a comment.

p2panda-net/src/sync/initiate.rs
adzialocha: Okay, let me remove one of them then 👍🏻

adzialocha:
@mycognosist I've added more doc-strings which hopefully explain the difference between Acceptors and Initiators in regards to the `SyncFailed` handling.

On top I've reduced the logging on failures (to-be-continued I assume when we have more first-hand experiences).

mycognosist (approved):
Awesome, looks good! 🟢

p2panda-net/src/sync/accept.rs
```diff
@@ -0,0 +1,243 @@
...
+    // It picks up any messages from the sync session, making sure that the "Two-Phase Sync Flow"
+    // is followed (I. "Handshake" Phase & II. "Data Sync" Phase), and informs the engine
+    // accordingly.
+    //
+    // If the task detects any invalid behaviour from the sync flow it fails critically, indicating
+    // that the sync protocol implementation does not behave correctly and is not compatible with
+    // the engine.
+    //
+    // Additionally, the task forwards any synced application data straight to the engine.
+    let glue_task_handle: JoinHandle<Result<(), SyncError>> = tokio::spawn(async move {
```
sandreae, line 80: Really great well documented code here, we've come a long way with understanding these sync state transitions. Feels like it would even benefit from having it's own struct (`SyncDriver` or something) to make the state transitions even more explicit. Need to think about it more, but it could even be something we expose from the `p2panda-sync` crate :thinking: not anything for now, but will keep it in mind.

p2panda-net/src/sync/tests.rs
sandreae, line 1: AMAZING TESTS :star2: :star2: :star2: :star2: :star2:

sandreae (approved):
Fantastic work! This is really the point several complex systems meet, well done unpicking all the flows and expected behaviours.

PR #609, opened 2024-12-04
Documentation for `p2panda-store`

mycognosist:
Documentation for all public modules, traits, structs and methods in `p2panda-store`.

### DIFF
--- a/p2panda-store/src/lib.rs
+++ b/p2panda-store/src/lib.rs
@@ -4,7 +4,7 @@
 //!
 //! The provided APIs allow for efficient implementations of `Operation` and `Log` stores. These
 //! persistence and query APIs are utilised by higher-level components of the `p2panda` stack, such
-//! as `p2panda-sync` and `p2panda-engine`. For detailed information concerning the `Operation` and
+//! as `p2panda-sync` and `p2panda-stream`. For detailed information concerning the `Operation` and
 //! `Log` types, please consult the documentation for the `p2panda-core` crate.
 //!
 //! The traits provided here are not intended to offer generic storage solutions for non-`p2panda`
--- a/p2panda-store/src/lib.rs
+++ b/p2panda-store/src/lib.rs
@@ -4,11 +4,18 @@
 //!
 //! The provided APIs allow for efficient implementations of `Operation` and log stores. These
 //! persistence and query APIs are utilised by higher-level components of the p2panda stack, such
-//! as `p2panda-sync` and `p2panda-stream`. For detailed information concerning the `Operation` and
-//! log types, please consult the documentation for the `p2panda-core` crate.
+//! as `p2panda-sync` and `p2panda-stream`. For detailed information concerning the `Operation`
+//! type, please consult the documentation for the `p2panda-core` crate.
 //!
-//! The traits provided here are not intended to offer generic storage solutions for non-p2panda
-//! data types, nor are they intended to solve application-layer storage concerns.
+//! Logs in the context of `p2panda-store` are simply a collection of operations grouped under a
+//! common identifier. The precise type for which `LogId` is implemented is left up to the
+//! developer to decide according to their needs. With this in mind, the traits and implementations
+//! provided by `p2panda-store` do not perform any validation of log integrity. Developers using
+//! this crate must take steps to ensure their log design is fit for purpose and that all operations
+//! have been thoroughly validated before being persisted.
+//!
+//! Also note that the traits provided here are not intended to offer generic storage solutions for
+//! non-p2panda data types, nor are they intended to solve application-layer storage concerns.
 //!
 //! An in-memory storage solution is provided in the form of a `MemoryStore` which implements both
 //! `OperationStore` and `LogStore`. The store is gated by the `memory` feature flag and is enabled
@@ -24,6 +31,8 @@ pub use memory_store::MemoryStore;
 use p2panda_core::{Body, Hash, Header, PublicKey, RawOperation};
 
 /// Uniquely identify a single-author log.
+///
+/// The `LogId` exists purely to group a set of operations.
 ///
 /// A blanket implementation is provided for any type meeting the required trait bounds.
 pub trait LogId: Clone + Default + Debug + Eq + Send + Sync + std::hash::Hash {}
--- a/p2panda-store/src/lib.rs
+++ b/p2panda-store/src/lib.rs
@@ -32,9 +32,29 @@ use p2panda_core::{Body, Hash, Header, PublicKey, RawOperation};
 
 /// Uniquely identify a single-author log.
 ///
-/// The `LogId` exists purely to group a set of operations.
+/// The `LogId` exists purely to group a set of operations and is intended to be implemented for
+/// any type which meets the design requirements of a particular application.
 ///
 /// A blanket implementation is provided for any type meeting the required trait bounds.
+///
+/// Here we briefly outline several implementation scenarios:
+///
+/// An application relying on a one-log-per-author design might choose to implement `LogId` for a thin
+/// wrapper around an Ed25519 public key; this effectively ties the log to the public key of the
+/// author. Secure Scuttlebutt (SSB) is an example of a protocol which relies on this model.
+///
+/// In an application where one author may produce operations grouped into multiple logs,
+/// `LogId` might be implemented for a `struct` which includes both the public key of the author
+/// and a unique number for each log instance.
+///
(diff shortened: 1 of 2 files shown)


p2panda-store/src/memory_store.rs
```diff
@@ -17,18 +19,25 @@ type RawHeader = Vec<u8>;
 type LogMeta = (SeqNum, Timestamp, Hash);
 type StoredOperation<L, E> = (L, Header<E>, Option<Body>, RawHeader);
 
+/// An in-memory store for core p2panda data types: `Operation` and `Log`.
 #[derive(Clone, Debug)]
 pub struct InnerMemoryStore<L, E> {
```
mycognosist, line 24: It feels a bit strange to have this as part of our public API, but maybe we want to offer that level of access?

The alternative would be to make `InnerMemoryStore`, `read_store()` and `write_store()` private and only allowing interaction via the API of `MemoryStore` itself.

adzialocha: I'm undecided at this point. It feels not very useful at this point to expose it, on the other hand I can see how people would like to re-use the store implementation and add custom methods on it. And wouldn't it be possible to also add additional methods for example for extensions?

p2panda-store/src/lib.rs
```diff
@@ -1,5 +1,19 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+//! Interfaces and implementations of persistence layers for core p2panda data types.
+//!
+//! The provided APIs allow for efficient implementations of `Operation` and `Log` stores. These
```
adzialocha, line 5:
```suggestion
//! The provided APIs allow for efficient implementations of `Operation` and log stores. These
```

Don't think we have a code-equivalent called `Log`.

p2panda-store/src/lib.rs
```diff
@@ -1,5 +1,19 @@
...
+//! persistence and query APIs are utilised by higher-level components of the `p2panda` stack, such
```
adzialocha, line 6:
```suggestion
//! persistence and query APIs are utilised by higher-level components of the p2panda stack, such
```

p2panda-store/src/lib.rs
```diff
@@ -1,5 +1,19 @@
...
+//! as `p2panda-sync` and `p2panda-stream`. For detailed information concerning the `Operation` and
+//! `Log` types, please consult the documentation for the `p2panda-core` crate.
```
adzialocha, line 8:
```suggestion
//! log types, please consult the documentation for the `p2panda-core` crate.
```

p2panda-store/src/lib.rs
```diff
@@ -1,5 +1,19 @@
...
+//!
+//! The traits provided here are not intended to offer generic storage solutions for non-`p2panda`
```
adzialocha, line 10:
```suggestion
//! The traits provided here are not intended to offer generic storage solutions for non-p2panda
```

p2panda-store/src/lib.rs
adzialocha: Similar here, there's no crate named `p2panda`

p2panda-store/src/lib.rs
```diff
@@ -1,5 +1,19 @@
...
+//! data types, nor are they intended to solve application-layer storage concerns.
+//!
+//! An in-memory storage solution is provided in the form of a `MemoryStore` which implements both
+//! `OperationStore` and `LogStore`. The store is gated by the `memory` feature flag and is enabled
+//! by default.
+
```
adzialocha, line 16: Remove new-line here (we don't do them for both lib-level or mod-level doc-strings)

p2panda-store/src/lib.rs
```diff
@@ -10,11 +24,17 @@ pub use memory_store::MemoryStore;
...
 
-/// The unique identifier of a single-author log.
+/// Uniquely identify a single-author log.
+///
+/// A blanket implementation is provided for any type meeting the required trait bounds.
 pub trait LogId: Clone + Default + Debug + Eq + Send + Sync + std::hash::Hash {}
 
 impl<T> LogId for T where T: Clone + Default + Debug + Eq + Send + Sync + std::hash::Hash {}
 
+/// Interface for storing, deleting and querying `Operations`.
```
adzialocha, line 34:
```suggestion
/// Interface for storing, deleting and querying operations.
```

adzialocha: .. or `Operation`s but that always makes it harder to read?

p2panda-store/src/lib.rs
```diff
@@ -57,6 +79,10 @@ pub trait LocalOperationStore<LogId, Extensions>: Clone {
     async fn delete_payload(&mut self, hash: Hash) -> Result<bool, Self::Error>;
 }
 
+/// Interface for storing, deleting and querying `Log` entries.
```
adzialocha, line 82:
```suggestion
/// Interface for storing, deleting and querying logs.
```

p2panda-store/src/memory_store.rs
```diff
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+//! In-memory persistence for p2panda operations and logs.
+
```
adzialocha, line 4: Remove newline here

p2panda-store/src/memory_store.rs
```diff
@@ -17,18 +19,25 @@ type RawHeader = Vec<u8>;
...
 type StoredOperation<L, E> = (L, Header<E>, Option<Body>, RawHeader);
 
+/// An in-memory store for core p2panda data types: `Operation` and `Log`.
 #[derive(Clone, Debug)]
 pub struct InnerMemoryStore<L, E> {
     operations: HashMap<Hash, StoredOperation<L, E>>,
     logs: HashMap<(PublicKey, L), BTreeSet<LogMeta>>,
 }
 
+/// An in-memory store for core p2panda data types: `Operation` and `Log`.
```
adzialocha, line 29:
```suggestion
/// An in-memory store for core p2panda data types: `Operation` and log.
```

p2panda-store/src/memory_store.rs
```diff
@@ -17,18 +19,25 @@ type RawHeader = Vec<u8>;
...
+///
+/// `MemoryStore` supports usage in asynchronous and multi-threaded contexts by wrapping an
+/// `InnerMemoryStore` with ai `RwLock` and `Arc`. Convenience methods are provided to obtain a
```
adzialocha, line 32:
```suggestion
/// `InnerMemoryStore` with an `RwLock` and `Arc`. Convenience methods are provided to obtain a
```

adzialocha (changes requested):
Looking good! I only have minor formatting changes and a "conceptual" note.

Re: Exposing the lower-level `RwLock` access I'm open for any approach, I think we need to learn still how these APIs are used and I can see how it can go in either direction.

Describing "Log" as a core-data type is a little bit tricky. While it kinda is an important concept in our specification it is not a "Rust Type" and nowhere implemented as such. This is why I removed all mentions of `Log` as that implies we have a struct somewhere waiting.

The interesting thing with how we treat logs here is that we're simply just grouping operations under a common identifier defined by `LogId` which is up to the developer to decide what that is.

If `LogId` is equals the public key of an author then we're in a SSB world where only one log exists per author, if we're setting it to `u64` we're entering Bamboo territory (many logs per author, identified by a number), if we're giving them a string or something similar we can give them a semantic grouping based on our application data.

On top we're not doing any validation of the integrity of logs here, the storage does not care about that. So developers should make sure their regarding "log approach" is sane and all operations have been checked against integrity before being persisted.

mycognosist:
@adzialocha Thanks for the suggested changes and great feedback!

Since we're understandably still unsure about API usage at this point, what if we remove `InnerMemoryStore`, `read_store()` and `write_store()` from the public API for now? We can easily export them publicly if there ends up being demand. The flip-side is that we leave them in and then have to introduce a breaking change if we want to remove them later. I don't see anywhere in our codebase where we currently use these types (outside of `p2panda-store` itself).

@sandreae might also have some thoughts to add.

adzialocha (approved):

adzialocha:
> Since we're understandably still unsure about API usage at this point, what if we remove `InnerMemoryStore`, `read_store()` and `write_store()` from the public API for now?

I'm up for that!

adzialocha:
Merging this now, we can always change things :-)

Issue #541, opened 2024-08-13
Run all code examples in `README.md` when running `cargo test`


sandreae:
Using: `#[doc = include_str!("../README.md")]`

mycognosist:
Cool, this would be a welcome change. We'd just need to ensure that the `README` incorporates all the existing documentation from the crate-level docs.

PR #622, opened 2024-12-06
Introduce a RELEASE doc with step-by-step instructions

mycognosist:
Here we add a `RELEASE.md` doc with step-by-step instructions to release a new p2panda version.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,39 @@
+# Releasing p2panda
+
+_This example assumes we are publising version `1.2.0`._
+
+_Remember to update the version number accordingly!_
+
+## Checks and preparations
+
+1. Check that the CI has passed on the p2panda project's
+  [Github page](https://github.com/p2panda/p2panda).
+2. Make sure you are on the `main` branch.
+3. Run the test suite and make sure all tests pass:
+    - `cargo test --all-features`
+4. Make sure that all examples in each crate `README.md` (including the ones in the
+   sub-folders) are still up-to-date with the latest API changes.
+
+## Changelog time!
+
+5. Check the git history for any commits on main that have not been mentioned
+   in the _Unreleased_ section of `CHANGELOG.md` but should be.
+6. Add an entry in `CHANGELOG.md` for this new release and move over all the
+   _Unreleased_ stuff. Follow the formatting given by previous entries.
+7. Remember to update the links to your release and the unreleased git log at
+   the bottom of `CHANGELOG.md`.
+
+## Tagging and versioning
+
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,39 @@
...
+9. Commit the version changes with a commit message `1.2.0`.
+10. Run `git tag v1.2.0` and push (including your tags) using `git push origin
+    main --tags`.
+
+## Publishing releases
+
+11. Copy the changelog entry you authored into Github's [new release
+    page](https://github.com/p2panda/p2panda/releases/new)'s description field.
+    Title it with your version `v1.2.0`.
+12. Run `cargo publish` from the workspace root.
```
adzialocha, line 38: Ah, amazing! This allows us to release all crates at the same time? Maybe we should add a note what to do if we want to release only one crate? Or is cargo doing this for us automatically?

RELEASE.md
```diff
@@ -0,0 +1,39 @@
...
+_Remember to update the version number accordingly!_
+
+## Checks and preparations
+
+1. Check that the CI has passed on the p2panda project's
+  [Github page](https://github.com/p2panda/p2panda).
+2. Make sure you are on the `main` branch.
+3. Run the test suite and make sure all tests pass:
+    - `cargo test --all-features`
+4. Make sure that all examples in each crate `README.md` (including the ones in the
```
adzialocha, line 14: This seems to be a thing: https://github.com/p2panda/p2panda/issues/541 but nothing important for this PR, just FYI

adzialocha (approved):

RELEASE.md
mycognosist: Ah no, I just didn't update that part properly (kept the text from the v1 document). For now we'll need to do them one by one.

In the future we may want to have a look at https://lib.rs/crates/cargo-workspaces for more efficient workflows.

mycognosist: It occurs to me now that we'll seldom be publishing all crates in one go (that's just a bit of an outlier for the initial release). Mostly we'll be publishing a single crate at a time, or maybe a handful. I'll update the doc accordingly.

mycognosist: So we'll also want to create git tags which are prefixed by the crate name, e.g. `p2panda-sync-v1.2.0`.

Don't mind me, just thinking out loud here 😅

adzialocha: > So we'll also want to create git tags which are prefixed by the crate name, e.g. `p2panda-sync-v1.2.0`.

Oh, good point. We could also decide to have everything being always bumped together (this is how iroh does it). I like both approaches.

Issue #630, opened 2024-12-16
Reset sync state after disconnect


mycognosist:
Peers should start syncing again after reconnecting to allow catching up on what happened when they've been offline.

Scenario:

Alice and Bob are both interested in "cats". They find one another via the gossip-wide overlay, sync their data-sets and move into "live (gossip) mode". Alice then loses her connection for five minutes before coming back online. During that time, Bob posted a lot of cat photos that Alice needs to catch up on. The problem is that Alice has the Bob-cats sync state marked as `Complete`; this means she won't sync state again.

Solution:

We want to detect the disconnection and reset sync state for all peer-topic combinations we know about. Then, initiate a sync session after the connection returns. We may also want to alter the sync protocol to exchange data in both directions.

mycognosist:
This is really quite an interesting problem; specifically: how to detect the disconnection?

If we do not have any active sync session at the moment when the disconnection happens then we don't receive any connection error. As far as I can tell, `iroh-gossip` does not return any event which alerts us to the disconnection.

One option is to reach for the lower-level [`Monitor`](https://docs.rs/iroh-net/0.25.0/iroh_net/net/netmon/struct.Monitor.html) in `iroh-net` which would allow us to subscribe to network changes. We could listen for "major" changes and use that to trigger sync state resets.

https://docs.rs/iroh-net/0.25.0/src/iroh_net/net/netmon.rs.html#96-105

PR #648, opened 2024-12-18
Reset sync and gossip state on major network interface change

mycognosist:
## Expectation

Peers should resync and reenter network-wide and topic gossip overlays following an extended loss of connectivity.

## Problem

Network disconnections > 30 seconds result in gossip connections dropping. This means that we "fall out" of the network-wide and topic gossip overlays. In addition to gossip issues, peers who had previously synced will not sync again if the resync flag is not enabled.

## Solution

We listen for major network interface changes and use them as a trigger to reset sync and gossip state. This takes place on the node that lost connection.

This alone is not enough for a full recovery of the system once a connection is reestablished, since sync is currently unidirectional and the peer that stayed online will not necessary attempt to resync. Bidirectional sync is introduced in https://github.com/p2panda/p2panda/pull/657

Closes https://github.com/p2panda/p2panda/issues/630

## 📋 Checklist

- [ ] ~~New files contain a SPDX license header~~

### DIFF
--- a/p2panda-blobs/Cargo.toml
+++ b/p2panda-blobs/Cargo.toml
@@ -20,10 +20,10 @@ bytes = "1.7.1"
 futures-buffered = "0.2.8"
 futures-lite = "2.3.0"
 futures-util = "0.3.30"
-iroh-base = "0.25.0"
-iroh-blobs = { version = "0.25.0", features = ["downloader", "fs-store"], default-features = false }
+iroh-base = "0.26.0"
+iroh-blobs = { version = "0.26.0", features = ["downloader", "fs-store"], default-features = false }
 iroh-io = "0.6.1"
-iroh-net = "0.25.0"
+iroh-net = "0.26.0"
 p2panda-core = { path = "../p2panda-core", version = "0.1.0" }
 p2panda-net = { path = "../p2panda-net", version = "0.1.0" }
 p2panda-sync = { path = "../p2panda-sync", version = "0.1.0" }
--- a/p2panda-net/src/engine/engine.rs
+++ b/p2panda-net/src/engine/engine.rs
@@ -209,15 +209,16 @@ where
                 // Inform the topic discovery process and sync actor about a major network
                 // interface change.
                 Some(is_major) = interface_change_rx.recv() => {
-                    // In the event of a disconnection, we will drop out of the network-wide gossip
-                    // overlay and may fall out of sync with peers with whom we had previously
-                    // synced. Here we inform the sync actor (if one exists) of the interface
-                    // change and reset the state of the topic discovery process. This should result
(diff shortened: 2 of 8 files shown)


mycognosist:
I still need to update the `CHANGELOG` etc. before this is ready for review.

p2panda-blobs/Cargo.toml
```diff
@@ -20,10 +20,10 @@ bytes = "1.7.1"
 futures-buffered = "0.2.8"
 futures-lite = "2.3.0"
 futures-util = "0.3.30"
-iroh-base = "0.25.0"
-iroh-blobs = { version = "0.25.0", features = ["downloader", "fs-store"], default-features = false }
+iroh-base = "0.26.0"
```
mycognosist, line 23: I bumped these because the inclusion of the `netwatch` dependency resulted in a conflict with `derive_more`.

p2panda-net/src/engine/engine.rs
```diff
@@ -185,6 +206,24 @@ where
...
+                Some(is_major) = interface_change_rx.recv() => {
+                    // In the event of a disconnection, we will drop out of the topic and
+                    // network-wide gossip overlays and may fall out of sync with peers with whom we
+                    // had previously synced. Here we inform the sync actor (if one exists) of the
+                    // interface change and reset the state of the topic discovery process. This
+                    // should result in us reentering the network-wide gossip overlay and resyncing
+                    // with our peers before entering "live mode" again.
+                    if is_major {
+                        debug!("detected major network interface change");
+                        self.topic_discovery.reset_status().await;
```
mycognosist, line 220: This is where we trigger all the state resets. The existing plumbing makes it very simple!

p2panda-net/src/engine/engine.rs
```diff
@@ -224,6 +266,11 @@ where
             ToEngineActor::GossipJoined { topic_id } => {
                 self.on_gossip_joined(topic_id).await;
             }
+            ToEngineActor::GossipNeighborDown { peer } => {
```
mycognosist, line 269: `GossipNeighborDown` indicates that we've lost connection with a direct neighbour in the HyParView cluster. We use that event to initiate a reset of the completed sync state for the peer with whom we lost a connection. This means that we'll be all-clear to initiate sync with that peer when we receive their next topic announcement via the network-wide gossip overlay (ie. whenever they come back online).

Thanks to @sandreae for suggesting this addition!

adzialocha: Neighbor down events do not necessarily imply that the peer went offline, so it might mean we initiate sync without any new outcomes, on top we might do that more often as there's potentially multiple gossip overlays. Of course it _can_ sometimes help if that peer went actually offline and _maybe_ came up again in our strong ebt links, but that's a lot of maybes.

Im undecided if this is really what we want?

PR #653, opened 2024-12-27
Decentralised group management

adzialocha:
This PR introduces the `p2panda-groups` crate which is the starting point of a full DCGKA implementation based on this [paper](https://dl.acm.org/doi/10.1145/3460120.3484542). This implementation roughly follows the Java version [here](https://github.com/trvedata/key-agreement) with additional optimizations which will be documented while this PR progresses.

The first methods for this crate will be a distributed `Group` struct which helps with managing membership in an offline-first network over generic member identity and message id parameters. Members can be added and removed from a group, all existing members explicitly acknowledge these changes until everyone converges to the same "view" again.

Some ordering solution needs to exist to make this reliably work, but this is rather part of `p2panda-core` as we're re-using the hash-based vector clocks ("previous" links) of our append-only log data type for it.

The idea is to make this implementation be useful for all sorts of applications, also outside of the p2panda universe, hence the generic approach over member ids, message ids and ordering algorithms.

To design groups with roles, for example where some members have additional "moderator" or "admin" rights, groups can be stacked on top of each other. For example one can manage an "admin" group whose state again is checked before a new user is added to a "chat" group. All of this is already higher level usage, but it gives an idea!

## Todos

- [ ] Extensive doc-strings to explain this whole thing to myself
- [ ] A whole bunch of unit tests
- [ ] Fuzz tests for random adds and removals
- [ ] Methods to import / export whole group state (required for DCGKA protocol)
- [ ] Store traits to persist group states (with default in-memory implementation)
- [ ] Clear removed members as soon as removal has been acked by everyone (this is an improvement which doesn't exist in the Java implementation, neither the paper)

## 📋 Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "p2panda-blobs",
     "p2panda-core",
     "p2panda-discovery",
+    "p2panda-groups",
     "p2panda-net",
     "p2panda-store",
     "p2panda-stream",
--- a/p2panda-groups/Cargo.toml
+++ b/p2panda-groups/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "p2panda-groups"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+thiserror = "2.0.9"
+
+[lints]
+workspace = true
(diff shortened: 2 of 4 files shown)


adzialocha:
Closing PR for now, will re-open as soon as we're getting back to it again :-)

Issue #654, opened 2024-12-31
mDNS should not error when no interface to bind socket is given


adzialocha:
This is a fairly common case in an offline-first world, we should rather just park the process and come back to it when an interface was detected

mycognosist:
**Problem**

Calling `LocalDiscovery::new()` returns an error when no socket is available to be bound (for example, when the device network is disabled). The error occurs in `socket_v4()` (`src/mdns/socket.rs`) and originates with `socket.join_multicast_v4()`.

```
local discovery service: join_multicast_v4

Caused by:
    No such device (os error 19)
```

PR #660, opened 2025-01-10
Introduce offline-first capability for `LocalDiscovery` mDNS service

mycognosist:
This PR introduces offline-first capability for our `LocalDiscovery` mDNS service.

If the initial attempt to fully configure and bind the UDP socket fails, we create an unbound socket as a fallback and periodically attempt to configure and bind a new socket. The socket is also recreated when a major network interface change is detected.

A minor API change is introduced: the `LocalDiscovery::new()` no longer returns a `Result`.

Addresses: https://github.com/p2panda/p2panda/issues/654

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Highlights are marked with a pancake 🥞
 
 ### Changed
 
+- Introduce offline-first capability for `LocalDiscovery` mDNS service [#660](https://github.com/p2panda/p2panda/pull/660)
 - Update to iroh `v0.28.1` [#661](https://github.com/p2panda/p2panda/pull/661)
 - Make log sync protocol bidirectional [#657](https://github.com/p2panda/p2panda/pull/657)
 - `TopicMap` replaced by `TopicLogMap` [#650](https://github.com/p2panda/p2panda/pull/650)
--- a/p2panda-discovery/src/mdns/mod.rs
+++ b/p2panda-discovery/src/mdns/mod.rs
@@ -5,19 +5,23 @@ mod dns;
 mod socket;
 
 use std::collections::HashMap;
+use std::net::Ipv4Addr;
 use std::str::FromStr;
 use std::time::Duration;
 
 use anyhow::Result;
 use flume::Sender;
-use futures_lite::StreamExt;
+use futures_lite::{FutureExt, StreamExt};
 use hickory_proto::rr::Name;
 use iroh_base::base32;
 use iroh_net::NodeAddr;
+use netwatch::netmon::{CallbackToken, Monitor};
+use tokio::sync::mpsc::{self, Receiver};
 use tokio_util::task::AbortOnDropHandle;
+use tracing::debug;
 
 use crate::mdns::dns::{make_query, make_response, parse_message, MulticastDNSMessage};
-use crate::mdns::socket::{send, socket_v4};
+use crate::mdns::socket::{send, socket_v4, MDNS_IPV4};
 use crate::{BoxedStream, Discovery, DiscoveryEvent};
 
 const MDNS_PROVENANCE: &str = "mdns";
@@ -39,6 +43,24 @@ pub struct LocalDiscovery {
     tx: Sender<Message>,
 }
 
+/// Create a new network monitor and subscribe to major interface changes.
+async fn network_monitor() -> Result<(Monitor, CallbackToken, Receiver<bool>)> {
+    let network_monitor = Monitor::new().await?;
+    let (interface_change_tx, interface_change_rx) = mpsc::channel(8);
+    let token = network_monitor
+        .subscribe(move |is_major| {
+            debug!("detected major network interface change");
+            let interface_change_tx = interface_change_tx.clone();
+            async move {
+                interface_change_tx.send(is_major).await.ok();
+            }
+            .boxed()
+        })
+        .await?;
+
+    Ok((network_monitor, token, interface_change_rx))
+}
+
 impl LocalDiscovery {
     pub fn new() -> Result<Self> {
         let (tx, rx) = flume::bounded(64);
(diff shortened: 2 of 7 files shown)


mycognosist:
Not yet working properly.

Logs from aardvark:

```
2025-01-10T10:52:21.939975Z DEBUG p2panda_discovery::mdns: failed to join ipv4 multicast for mdns discovery; waiting for major network interface change
2025-01-10T10:52:51.991612Z DEBUG p2panda_discovery::mdns: detected major network interface change
2025-01-10T10:52:51.991646Z DEBUG p2panda_discovery::mdns: detected major network interface change
2025-01-10T10:52:51.991661Z DEBUG p2panda_discovery::mdns: failed to join ipv4 multicast for mdns discovery; waiting for major network interface change
2025-01-10T10:52:52.174229Z DEBUG p2panda_discovery::mdns: detected major network interface change
2025-01-10T10:52:52.174257Z DEBUG p2panda_discovery::mdns: detected major network interface change
2025-01-10T10:52:52.176191Z ERROR p2panda_discovery::mdns::socket: failed sending mdns message on udp socket: Operation not permitted (os error 1)
2025-01-10T10:52:53.175609Z ERROR p2panda_discovery::mdns::socket: failed sending mdns message on udp socket: Operation not permitted (os error 1)
```

p2panda-discovery/src/mdns/mod.rs
```diff
@@ -10,18 +10,22 @@ use std::time::Duration;
...
+use tracing::{debug, warn};
 
 use crate::mdns::dns::{make_query, make_response, parse_message, MulticastDNSMessage};
-use crate::mdns::socket::{send, socket_v4};
+use crate::mdns::socket::{send, socket_v4, socket_v4_unbound};
 use crate::{BoxedStream, Discovery, DiscoveryEvent};
 
 const MDNS_PROVENANCE: &str = "mdns";
 const MDNS_QUERY_INTERVAL: Duration = Duration::from_millis(1000);
+const SOCKET_REBIND_INTERVAL: Duration = Duration::from_millis(5000);
```
mycognosist, line 28: We might consider adjusting this interval. Five seconds seems pretty reasonable for an ambient discovery service.

p2panda-discovery/src/mdns/mod.rs
```diff
@@ -39,23 +43,63 @@ pub struct LocalDiscovery {
...
+                interface_change_tx.send(is_major).await.ok();
+            }
+            .boxed()
+        })
+        .await?;
+
+    Ok(interface_change_rx)
+}
+
+impl Default for LocalDiscovery {
```
mycognosist, line 64: Added this to satisfy clippy.

p2panda-discovery/src/mdns/mod.rs
```diff
@@ -39,23 +43,63 @@ pub struct LocalDiscovery {
...
-        let socket = socket_v4()?;
+        let mut socket_is_bound = false;
+        let mut socket = match socket_v4() {
+            Ok(socket) => {
+                socket_is_bound = true;
+                socket
+            }
+            Err(err) => {
+                warn!("failed to create udp socket for mdns discovery: {}", err);
+                socket_v4_unbound().expect("create udp socket")
```
mycognosist, line 82: This is the tradeoff we have for being able to remove the `Result` return type. The "unbound" socket creation involves a low-level system call and doesn't bind to any interface or address. The likelihood of an error here seems very low and would likely represent a critical issue.

p2panda-discovery/src/mdns/mod.rs
```diff
@@ -98,12 +142,12 @@ impl LocalDiscovery {
...
                         }
                     },
-                    _ = interval.tick() => {
+                    _ = interval.tick(), if socket_is_bound => {
                         for service_name in subscribers.keys() {
                             send(&socket, make_query(service_name)).await;
                         }
                     },
-                    Ok(msg) = rx.recv_async() => {
+                    Ok(msg) = rx.recv_async(), if socket_is_bound => {
```
mycognosist, line 150: We only try to `send` and `recv` if the socket is fully bound.

p2panda-net/src/engine/engine.rs
```diff
@@ -208,20 +208,18 @@ where
                 },
                 // Inform the topic discovery process and sync actor about a major network
                 // interface change.
-                Some(is_major) = interface_change_rx.recv() => {
+                Some(true) = interface_change_rx.recv() => {
```
mycognosist, line 211: This change isn't directly related to the PR but I noticed it while working on the network monitor code.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Highlights are marked with a pancake 🥞
 
 ### Changed
 
+- Introduce offline-first capability for `LocalDiscovery` mDNS service [#660](https://github.com/p2panda/p2panda/pull/660)
```
adzialocha, line 14: I think that's not a "Change" but a "Feature" 🥳

adzialocha (approved):
Amazing! This is so useful!

Do I see it right that because we need to have more "low-level" access to the sockets we actually can't use a crate in the future like `swarm-discovery`?

mycognosist:
> Amazing! This is so useful!
> 
> Do I see it right that because we need to have more "low-level" access to the sockets we actually can't use a crate in the future like `swarm-discovery`?

Yeah exactly. iroh now integrates `swarm-discovery` into `iroh-net` (https://docs.rs/iroh-net/0.28.1/iroh_net/endpoint/struct.Builder.html#method.discovery_local_network) which makes it super convenient to add via the `Endpoint` `Builder`. I tried this on a branch of `p2panda-net` and it resulted in a major decrease in code but we lose out on low-level access.

If the endpoint had some kind of "restart discovery service" API then that might work in our case. The only other option I could think of was recreating the entire endpoint on major network interface changes, which seemed like a drastic measure to take.

Issue #652, opened 2024-12-19
Report stale peers after some duration


adzialocha:
We can keep a table where we track "activity" of an peer, if no activity has been seen after a set duration we consider them as "stale", that event can be propagated further to applications (they can interpret these events as "peer disconnected" etc.)

adzialocha:
Probably we want to send these events via this interface: https://github.com/p2panda/p2panda/issues/636

Issue #639, opened 2024-12-17
Automatically unsubscribe from topic if both `tx` and `rx` gets dropped


adzialocha:

mycognosist:
https://github.com/n0-computer/iroh/issues/2962

> When dropping the subscription stream, neither connections nor subscriptions are fully cleaned up.

Issue #637, opened 2024-12-17
Diagnostics for sync session events


adzialocha:
When a sync session ended we want to report diagnostics like:

* How many bytes have been transmitted
* How long did the sync session last

Issue #636, opened 2024-12-17
API to receive system events, for example when peer connected or sync started / ended


adzialocha:
Ideas for events:

* New peer discovered
* Sync started
* Sync ended
* Sync failed
* Connection attempt failed

adzialocha:
We can also report these events here:

- https://github.com/p2panda/p2panda/issues/652

.. and report diagnostics on "Sync ended" events:

- https://github.com/p2panda/p2panda/issues/637

adzialocha:
This can be closed, right?

@mycognosist

PR #669, opened 2025-01-21
Introduce network system events API

mycognosist:
Here we introduce the ability to subscribe to network system events.

```rust
let mut event_rx = network.events().await?;
```

I've only implemented a limited set of system events for now. I mostly wanted to get the plumbing in place and ensure we have a solid foundation for future additions.

Note that the `GossipLeft` event requires `unsubscribe()` to be implemented (see https://github.com/p2panda/p2panda/issues/639).

Partially addresses https://github.com/p2panda/p2panda/issues/636 and paves the way for reporting https://github.com/p2panda/p2panda/issues/652 and https://github.com/p2panda/p2panda/issues/637

## 📋 Checklist


p2panda-net/src/engine/engine.rs
```diff
@@ -326,20 +344,37 @@ where
         Ok(())
     }
 
+    /// Return a receiver for system events.
+    fn events(&mut self) -> broadcast::Receiver<SystemEvent<T>> {
```
mycognosist, line 348: We're using a broadcast channel to allow for multiple receivers to be created (if needed). This means that all receivers will receive a sent event (not just one of them).

p2panda-net/src/engine/gossip.rs
```diff
@@ -205,13 +207,17 @@ where
     async fn on_joined(&mut self, topic_id: [u8; 32], stream: GossipTopic) -> Result<()> {
         self.joined.insert(topic_id);
 
-        // Split the gossip stream and insert handles to the receiver and sender
+        // Split the gossip stream and insert handles to the receiver and sender.
         let (stream_tx, stream_rx) = stream.split();
+
+        // Collect all our current direct neighbors for this gossip topic.
+        let peers: Vec<PublicKey> = stream_rx.neighbors().collect();
```
mycognosist, line 214: This is how we learn who we have just connected to (ie. the peer / peers who served as our entry-point into the gossip overlay).

p2panda-net/src/network.rs
```diff
@@ -1526,4 +1537,100 @@ pub(crate) mod tests {
...
+        assert!(ready_1.await.is_ok());
+
+        // Start a third node.
+        let node_3 = NetworkBuilder::new(network_id).build().await.unwrap();
+        let node_3_id = node_3.endpoint().node_id();
+        node_3.add_peer(node_1_addr).await.unwrap();
+        let (_tx_3, _rx_3, ready_3) = node_3.subscribe(chat_topic.clone()).await.unwrap();
+        assert!(ready_3.await.is_ok());
+
+        // Events we expect to receive on node 1.
```
sandreae, line 1587: So cool to be able to test the life-cycle of a node like this :star_struck:

sandreae (approved):
This looks perfect to me :pray: already a lot of useful and interesting node life-cycle information given from just these first few events.

p2panda-net/src/events.rs
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! System events API.
+
```
adzialocha, line 4: Styleguide Nitpick: No newlines after mod-level docs.

p2panda-net/src/events.rs
```diff
@@ -0,0 +1,27 @@
...
+use iroh_net::key::PublicKey;
+
+/// Network system events.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum SystemEvent<T> {
+    /// Joined a gossip topic via a connection to the given peer(s).
+    GossipJoined {
+        topic_id: [u8; 32],
+        peers: Vec<PublicKey>,
+    },
```
adzialocha, line 14: Styleguide Nitpick: Newlines between enum variants

p2panda-net/src/events.rs
```diff
@@ -0,0 +1,27 @@
...
+    // https://github.com/p2panda/p2panda/issues/639
+    GossipLeft { topic_id: [u8; 32] },
+    /// Established a connection with a neighbor.
+    GossipNeighborUp { topic_id: [u8; 32], peer: PublicKey },
+    /// Lost a connection to a neighbor.
+    ///
+    /// This event will be emitted approximately 30 seconds after the connection is lost.
+    GossipNeighborDown { topic_id: [u8; 32], peer: PublicKey },
+    /// Completed a sync session.
+    SyncDone { topic: T, peer: PublicKey },
```
adzialocha, line 26: Is there a reason why we can't have `SyncStart` and `SyncFailed` yet?

Also I'd be interested in peer discovery events, that would help to estimate how many peers are online.

mycognosist: No reason other than pragmatism; they're not immediately required and `SyncDone` already gives a good level of introspection into sync activity. Eventually it will also include the sync session duration and number of bytes synced.

That being said, I don't mind adding `SyncStart` and `SyncFailed` if you want them in this PR.

I can add the discovery event now. Do you want `SystemEvent::PeerDiscovered { peer: PublicKey }` or `SystemEvent::TopicsDiscovered { topic_ids: Vec<[u8; 32]>, peer: PublicKey }`?

adzialocha: If I think from an UI perspective, it makes sense to indicate if I'm currently syncing (show a spinner etc.), so I think it would be quite good to have an `SyncStart` and `SyncFailed` event as well to do that.

Gossipping doesn't tell too much about the number of active peers, so discovery would be cool as well as an "more precise" metric (I guess one needs a timeout on app layer as well and measure activities for each peer etc.)

mycognosist: Cool! I've added `SyncStarted`, `SyncFailed` and `PeerDiscovered`.

adzialocha (approved):
Woho! This is a great feature!

Also, thank you for adding the extra events. <3

PR #670, opened 2025-01-21
Allow `extract` method to have access to header when building extensions

sandreae:
Give access to `Header` in the method `Extension::extract`. This allows for cases where the extension value needs to be derived from header material. Additionally add a helper method to `Header` for accessing extension values.
 
## 📋 Checklist

- [x] ~~Add tests that cover your changes~~
- [x] ~~Link this PR to any issues it closes~~


p2panda-stream/src/stream/ingest.rs
```diff
@@ -153,7 +153,7 @@ where
             // 2. Validate and check the log-integrity of the incoming operation. If it is valid it
             //    get's persisted and the log optionally pruned.
             let ingest_fut = async {
-                let log_id: L = header
```
adzialocha, line 156: Is this a small clean up (because the compiler can derive that `log_id` is `L`) or is this doing something specific?

p2panda-core/src/extensions.rs
```diff
@@ -61,10 +61,19 @@ use crate::Header;
...
 /// Trait definition of a single header extension type.
 pub trait Extension<T>: Extensions {
-    /// Extract the value of an extension based on it's type.
+    /// Extract the raw extension value of an extension based on it's type.
     fn extract(&self) -> Option<T> {
         None
     }
+
+    /// Extract the extension value with the option to derive it from material contained in the
+    /// passed header.
```
adzialocha, line 70: I can imagine it's fairly abstract for someone to look at all of that functionality. Maybe we can provide some examples (not in Rust code but in "use cases") in the docs?

adzialocha:
Do we have a reason to keep the `extract` method next to the newly introduced `with_header` method, other than not breaking the API for existing implementations?

If that's the only reason, I'd vote for removing the old `extract` and renaming `with_header` to it as we can and will break APIs for a while.

adzialocha (changes requested):
CHANGELOG entry seems to be missing.

Re: PR, looks all fine and is a good solution! Would maybe consider breaking the API instead if there's no good reason to keep both methods.

p2panda-stream/src/stream/ingest.rs
sandreae: Yeh, this is just a clean up :+1:

sandreae:
> Do we have a reason to keep the `extract` method next to the newly introduced `with_header` method, other than not breaking the API for existing implementations?
> 
> If that's the only reason, I'd vote for removing the old `extract` and renaming `with_header` to it as we can and will break APIs for a while.

Yeh, that's a valid point. I felt that only having `with_header` would make for an ugly API when `Extension` is implemented on the `Header` itself, but I'm a bit on the fence about it.

It's between `let log_id: L = header.extract()` and `let log_id: L = Extensions::with_header(&header)`. If we go for removing `extract()` we could find a better name for `with_header`.

adzialocha:
Maybe it's the definition of "Extensions" which needs changing: So far we've only considered "Extensions" as a way to add extra fields to the header, maybe we can re-define them as "enhance your header with stuff you need for your application or this method!" and that can be through interpreting parts of the header for your needs and / or adding extra fields?

sandreae:
> Maybe it's the definition of "Extensions" which needs changing: So far we've only considered "Extensions" as a way to add extra fields to the header, maybe we can re-define them as "enhance your header with stuff you need for your application or this method!" and that can be through interpreting parts of the header for your needs and / or adding extra fields?

Oh my, that is very cool! I like it :-)

sandreae:
I can try to adjust the API (with breaking changes) to reflect that understanding.

p2panda-core/src/extensions.rs
```diff
@@ -61,8 +28,8 @@ use crate::Header;
 
 /// Trait definition of a single header extension type.
 pub trait Extension<T>: Extensions {
-    /// Extract the value of an extension based on it's type.
-    fn extract(&self) -> Option<T> {
+    /// Extract the extension value from a header.
+    fn extract(_header: &Header<Self>) -> Option<T> {
         None
     }
 }
```
sandreae, line 35: We're back to one `extract` method on `Extension` which takes a header as argument.

sandreae: I wonder if we want to actually return `T` directly now, without wrapping in an `Option`. If a field is actually optional, this could be taken care of in your custom extension type definition.

sandreae: Hmm, although I'm not sure, the fact that extensions are optional on header themselves effects this return value too.

adzialocha: > Hmm, although I'm not sure, the fact that extensions are optional on header themselves effects this return value too.

Yeah right, it's a good way to check if a value was given in the header or not (even if mandatory), so the application can accordingly consider an operation invalid.

p2panda-core/src/extensions.rs
```diff
@@ -20,39 +20,6 @@
...
-//! }
-//!
-//! // Create a custom extension instance, this can be added to an operation's header.
-//! let extensions = CustomExtensions {
-//!     expiry: Expiry(1733170246),
-//! };
-//!
-//! // Extract the extension we are interested in.
-//! let expiry: Expiry = extensions.extract().expect("expiry field should be set");
-//! ```
```
adzialocha, line 55: 😢 can we have an example here?

adzialocha: The one from the `examples` folder is nice, maybe just paste it in here? Or you think in one place is enough?

adzialocha (changes requested):
Cool, this looks great! I like the helper method as well ✨ 

Minor thing: Changelog is missing.

Otherwise I was wondering if we need a pass over the documentation to see if we can clarify / update the new concept of extensions.

p2panda-core/src/extensions.rs
sandreae: Yeh, I'll paste the one from `examples` here too :+1:

sandreae:
I updated docs, added an doc-string example and CHANGELOG entry. Should be ready now :+1:

adzialocha (approved):

Issue #643, opened 2024-12-17
SQLite store


adzialocha:
Prior art:

https://github.com/p2panda/aquadoggo/blob/main/aquadoggo/src/db/mod.rs
https://github.com/p2panda/aquadoggo/blob/main/aquadoggo/src/db/stores/operation.rs
https://github.com/p2panda/aquadoggo/blob/main/aquadoggo/src/db/stores/log.rs

mycognosist:
Lots of useful SQLx examples in the SQLite test code: 

https://github.com/launchbadge/sqlx/blob/6fa0458ff3a2921a5b2f3757b7ed39cc29d51c45/tests/sqlite/sqlite.rs

PR #680, opened 2025-01-27
Implement SQLite `OperationStore` & `LogStore`

mycognosist:
SQLite store.

- [x] `OperationStore`
  - [x] `insert_operation()`
  - [x] `get_operation()`
  - [x] `get_raw_operation()`
  - [x] `hash_operation()`
  - [x] `delete_operation()`
  - [x] `delete_payload()`
- [x] `LogStore`
  - [x] `get_log()`
  - [x] `get_raw_log()`
  - [x] `latest_operation()`
  - [x] `delete_operations()`
  - [x] `delete_payloads()`
  - [x] `get_log_heights()`

Addresses https://github.com/p2panda/p2panda/issues/643

## 📋 Checklist


p2panda-store/src/lib.rs
```diff
@@ -22,6 +22,8 @@
 //! by default.
 #[cfg(feature = "memory")]
 pub mod memory_store;
```
mycognosist, line 24: I'd suggest renaming this from `memory_store` to `memory`.

Then we have `p2panda_store::{memory, sqlite};`

Or we rather go with `memory_store` and `sqlite_store`.

What do you think?

sandreae: I like this suggesting :+1:

p2panda-store/src/sqlite/store.rs
```diff
@@ -0,0 +1,525 @@
...
+        header: &Header<E>,
+        body: Option<&Body>,
+        header_bytes: &[u8],
+        log_id: &L,
+    ) -> Result<bool, Self::Error> {
+        // Start a transaction.
+        //
+        // Any insertions after this point, and before `execute()`, will be rolled back in the
+        // event of an error.
+        let mut tx = self.pool.begin().await?;
```
sandreae, line 117: Is this transaction required? I understood that if only one query was being made then rollback is not an issue (as it either succeeds or fails). Maybe it's just required for the sqlx api....

sandreae: Yeh, the pool itself implements [`Executor`](https://docs.rs/sqlx/latest/sqlx/trait.Executor.html) so I believe you can pass it directly into the `execute()` method.

p2panda-store/migrations/20250125065817_create-operations.sql
```diff
@@ -0,0 +1,18 @@
+-- SPDX-License-Identifier: MIT OR Apache-2.0
+
+CREATE TABLE IF NOT EXISTS operations_v1 (
+    hash                    TEXT            NOT NULL    PRIMARY KEY,
+    log_id                  TEXT            NOT NULL,
+    version                 TEXT            NOT NULL,
+    public_key              TEXT            NOT NULL,
+    signature               TEXT            NULL,
```
sandreae, line 8: On our `Header` struct signature is optional, but this is only a side-effect of us having a two step approach to constructing the header (eg. add the header fields, then call `sign` to add the signature). I would advocate for our store implementations expecting that a signature will always be present, and if it isn't we return an error on insertion.

p2panda-store/src/sqlite/models.rs
```diff
@@ -0,0 +1,73 @@
...
+            let (hex, rest) = row_previous.split_at(32);
+            // We assume database values are valid and therefore we're safe to unwrap.
+            previous.push(Hash::from_str(hex).unwrap());
+            row_previous = rest.to_string();
+        }
+
+        Header {
+            version: row.version.parse::<u64>().unwrap(),
+            public_key: PublicKey::from_str(&row.public_key).unwrap(),
+            signature: row.signature.map(|hex| Signature::from_str(&hex).unwrap()),
```
sandreae, line 55: Following my previous comment, maybe we could match on the signature here and error if it is none.

sandreae (changes requested):
Looking great :+1: reminded me how much I like working on database implementations :smile:  

I had a few suggestions which I'd like to discuss before approving, but nothing major.

p2panda-store/src/sqlite/models.rs
sandreae: I guess this will require adding our own error type...

adzialocha: Its fine to panic here, as itll be a bug I assume

p2panda-store/src/sqlite/store.rs
mycognosist: Ooh yeah, nice one! Much better.

p2panda-store/src/sqlite/models.rs
mycognosist: Good call!

p2panda-store/migrations/20250125065817_create-operations.sql
mycognosist: Good suggestion. I've added a `header.signature.is_none()` check on `insert_operation()` in both store implementations, returning an error if `true`.

mycognosist:
@sandreae This is ready for a final review 🙏🏻

sandreae (approved):
This all looks great to me, nice work!

I had one comment about signatures again, it's not a big deal so already happy to approve this PR. Thanks!

PR #688, opened 2025-02-13
Allow bootstrap peer to already enter topic-discovery overlay without any known peers

### LINKED ISSUE #685: Resolving & Establishing connection to peer fails in gossip overlay when direct address is unknown

adzialocha:
This PR fixes a bug where peers couldn't learn about each other when not being on the same network (not using mDNS for discovery).

The underlying issue was that we didn't allow a peer to enter the topic-discovery overlay when they didn't know about any peers yet. We still need to have at least one peer using a "boostrap node id" to join the overlay, but the bootstrap node was never reacting because it never subscribed itself.

## Buggy version:

```
Node A ("Bootstrap Node")
1. Attempt Subscription to "Topic Discovery" Overlay
2. Reject because no peers are known

Node B w. Node A as Bootstrap
1. Subscribe to "Topic Discovery" Overlay w. Node A as Bootstrap
2. .. wait forever because we never receive a response from Node A
```

## Fixed version:

```
Node A ("Bootstrap Node")
1. Subscribe to "Topic Discovery" Overlay
2. .. wait until someone joins
3. Node B joins: Establish gossip overlay

Node B w. Node A as Bootstrap
1. Subscribe to "Topic Discovery" Overlay w. Node A as Bootstrap
2. .. wait until connection got established and join!
```

With mDNS discovery this bug never surfaced because we learned about at least one external node id and then could pass the check.

Closes #685 

## 📋 Checklist


p2panda-net/examples/chat.rs
```diff
@@ -45,6 +45,7 @@ async fn main() -> Result<()> {
     let private_key = PrivateKey::new();
 
     let network = NetworkBuilder::new(network_id)
+        .private_key(private_key.clone())
```
adzialocha, line 48: Unrelated to this fix but helpful to have for future debugging

mycognosist:
Hmm... `network::tests::multi_hop_topic_discovery_and_sync` is blocking in CI, though it passes just fine on my machine. That's exciting.

PR #698, opened 2025-02-19
Introduce concrete error type for `SqliteStore`

mycognosist:
Replaces the current `anyhow` errors and result type with a custom error type: `SqliteStoreError`.

The sqlite error has variants to account for extension encoding, extension decoding and all other errors stemming from `sqlx` (basically anything relating to SQL or the SQLite database).  

Addresses https://github.com/p2panda/p2panda/issues/692

## 📋 Checklist

### DIFF
--- a/p2panda-store/src/memory.rs
+++ b/p2panda-store/src/memory.rs
@@ -413,31 +413,6 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn insert_operation_with_unsigned_header() {
-        let mut store = MemoryStore::default();
-        let private_key = PrivateKey::new();
-
-        // Create the first operation.
-        let body = Body::new("hello!".as_bytes());
-        let (hash, mut header, header_bytes) = create_operation(&private_key, &body, 0, 0, None);
-
-        // Set signature to `None` for the sake of the test.
-        header.signature = None;
-
-        // Only insert the first operation into the store.
-        let inserted = store
-            .insert_operation(hash, &header, Some(&body), &header_bytes, &0)
-            .await;
-
-        // Ensure that the lack of a header signature returns an error.
-        assert!(inserted.is_err());
-        assert_eq!(
-            format!("{}", inserted.unwrap_err()),
-            "operation header must be signed prior to insertion"
-        );
-    }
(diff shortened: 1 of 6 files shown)


adzialocha (approved):
Looking good! I think we can remove `anyhow` now as a dependency, otherwise its ready to merge ✨

mycognosist:
> Looking good! I think we can remove `anyhow` now as a dependency, otherwise its ready to merge ✨

Oh yeah, good catch! To do that I needed to also introduce a custom error for the memory store.

p2panda-store/src/memory.rs
```diff
@@ -87,9 +93,7 @@ where
         log_id: &L,
     ) -> Result<bool, Self::Error> {
         if header.signature.is_none() {
-            return Err(anyhow!(
-                "store error: operation header must be signed prior to insertion"
-            ));
+            return Err(MemoryStoreError::MissingHeaderSignature);
         }
```
adzialocha, line 97: This is caught in validation methods before, I believe this is not a concern of the storage itself

adzialocha (changes requested):

p2panda-store/src/memory.rs
mycognosist: Hmm...I hear what you're saying about the separation of concerns (validation and storage).

If we remove the check here then I'd expect we do the same for `insert_operation()` in `SqliteStore` (to keep the behaviour the same). If a header without a signature makes it into the `SqliteStore`, the corresponding `get_operation()` call is going to panic when the `from(row: OperationRow)` function tries to build the `Signature`. Then one would need to delete that operation.

If we're OK with that possibility then I'm happy to remove the `MissingHeaderSignature` variant from both error types.

sandreae: I'm also happy with panics occuring in the memory store if a header without a signature is inserted 👍

mycognosist: Great, done! It turns out we get the validation for free anyways; SQLx returns an error on the insertion attempt because `signature` is defined as `NOT NULL`.

Issue #701, opened 2025-02-24
Sync errors should also lead to sync re-attempts


sandreae:
We schedule a sync session for a retry attempt after a connection error occurs, but so far not if any other kind of error occurs during sync. This is a valid approach, if we take the critical sync error to mean that a peer is breaking the protocol, then dropping the session makes sense. However, sync errors can occur for other reasons, in one reported case this is occurring because of a peer not being able to connect reliably with another service. In this case a retry would be the prefered behaviour. 

https://github.com/p2panda/p2panda/blob/1c5264e105883c50a5b79c99d5ed38529a8b19dd/p2panda-net/src/sync/manager.rs#L335-L356

adzialocha:
Some more context why this is happening:

If peer B abruptly crashes / goes offline, peer A will get an abruptly closed I/O stream (usually CBOR bytes). We handle that case here: https://github.com/p2panda/p2panda/blob/main/p2panda-sync/src/lib.rs#L243

Since we're only re-attempting on connection failures this case is never handled by the sync manager.

PR #702, opened 2025-02-25
Schedule sync re-attempt after any error occurred

sandreae:
Currently we only schedule a sync attempt retry if a session failed because of a connection error. We want to retry after any error occurred.

This PR also does some cleanup by moving our test protocols into `p2panda_sync`.

Issue details: https://github.com/p2panda/p2panda/issues/701

## 📋 Checklist

### DIFF
--- a/p2panda-net/src/sync/manager.rs
+++ b/p2panda-net/src/sync/manager.rs
@@ -386,14 +386,14 @@ mod tests {
     use iroh::{Endpoint, RelayMode};
     use iroh_quinn::TransportConfig;
     use p2panda_core::PublicKey;
+    use p2panda_sync::test_protocols::{PingPongProtocol, SyncTestTopic as TestTopic};
+    use p2panda_sync::SyncProtocol;
     use tokio::sync::mpsc;
     use tokio::time::{Duration, sleep};
     use tokio_util::sync::CancellationToken;
     use tracing::warn;
 
     use crate::engine::ToEngineActor;
-    use crate::network::sync_protocols::PingPongProtocol;
-    use crate::network::tests::TestTopic;
     use crate::protocols::ProtocolMap;
     use crate::sync::{SYNC_CONNECTION_ALPN, SyncConnection};
     use crate::{ResyncConfiguration, SyncConfiguration, to_public_key};
@@ -421,7 +421,8 @@ mod tests {
 
     // Sync actor creation, along with all prerequisite setup steps, to reduce boilerplate
     // duplication in the tests which follow.
-    async fn prepare_for_sync(
+    async fn prepare_for_sync<P>(
+        protocol: P,
         resync: bool,
     ) -> (
         TestTopic,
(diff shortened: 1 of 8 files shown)


p2panda-sync/src/test_protocols.rs
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+        // Simulate some critical error which occurred inside the sync session.
+        if let FailingProtocol::InitiatorFailsCritical = *self {
+            return Err(SyncError::Critical(
+                "something really bad happened in the initiator".to_string(),
+            ));
+        }
+
+        // Simulate unexpected behaviour (such as a broken pipe due to disconnection).
```
mycognosist, line 302: This is the newly-introduced test protocol variant that we use to test the resync-after-any-error behaviour.

p2panda-net/src/sync/manager.rs
```diff
@@ -332,26 +332,23 @@ where
...
-                        .await?;
-                }
+        if let Some(err) = err.downcast_ref::<SyncAttemptError>() {
+            // If the sync attempt failed for any reason we want to retry up to
+            // `max_retry_attempts`.
+            warn!("sync attempt failed: {err}");
+
+            // We need to inform the engine of the failed attempt so that the gossip buffer counter
+            // can be decremented.
+            self.engine_actor_tx
```
mycognosist, line 342: I invite careful consideration of this behaviour.

The context here is a failed sync session from the perspective of the initiator (we are in the sync manager where connections and sessions are initiated). Since the topic is known by the initiator, a `SyncStart` event will have been sent to the engine (this happens in `p2panda-net/src/sync/initiate.rs`). This results in a locking of the gossip buffer (thereby incrementing the counter). As a result, we need to send `SyncFailed` here to ensure the gossip buffer counter is correctly decremented.

mycognosist (commented):
The diff is quite noisy due to the housekeeping (moving test protocols to a better home) but the actual changes of interest are minimal.

p2panda-sync/src/lib.rs
```diff
@@ -20,6 +20,8 @@
 pub mod cbor;
 #[cfg(feature = "log-sync")]
 pub mod log_sync;
+#[cfg(feature = "test-protocols")]
```
sandreae, line 23: Like this a lot!

p2panda-net/src/sync/manager.rs
```diff
@@ -332,26 +332,23 @@ where
...
+                .send(ToEngineActor::SyncFailed {
+                    topic: Some(sync_attempt.topic.clone()),
+                    peer: sync_attempt.peer,
+                })
+                .await?;
+
+            if sync_attempt.attempts <= self.config.max_retry_attempts {
+                self.reschedule_attempt(sync_attempt).await?;
```
sandreae, line 350: @mycognosist do you mind clarifying for me what the interval for sync re-attempts are? I couldn't quite grok it looking at the configurable sync polling / resync interval values.

mycognosist: Good question. The current logic pushes the failed attempt back into the regular sync queue. This means that the next sync attempt will occur as soon as it's pulled from the queue (if it's the only attempt in the queue it will be pulled immediately). In other words, there is currently no enforced interval. We only enforce a default interval of > 60 seconds for a _resync_ attempt (where the previous session was successful).

In answering your question, I see that this situation is sub-optimal. We'd really rather not be firing off five attempts in rapid succession after the initial failure.

sandreae: Yeh, that's what I was thinking about. The question came up when talking to Konstantin about the behaviour he was observing. It wasn't clear if sync sessions were being re-attempted on fail, but so quickly that they all failed and the max limit was reached before whatever caused the failure had a chance to recover.

sandreae: Would the polling interval already rate limit re-attempts to max one per second?

mycognosist: The polling interval (of one second) only applies to the resync queue. The main queue is pulled from whenever an item is available.

Let me work on a solution.

mycognosist: @sandreae I've opened a second PR with an overall refactor of the sync manager. It also introduces logic for the retry queue (with configurable intervals between attempts). 

My suggestion would be to merge this PR as-is and then review and merge [PR#714](https://github.com/p2panda/p2panda/pull/714).

PR #710, opened 2025-03-03
Partial ordering algorithm for operations based on `previous` links 

sandreae:
- generic `PartialOrder` struct for non-deterministically sorting items in a DAG by applying a partial order
- dependency store traits and memory implementation used by `PartialOrder`
- specialised wrapper around `PartialOrder` which sorts p2panda operations based on dependencies contained in their `previous` field

Closes: https://github.com/p2panda/p2panda/issues/644

## Next Steps
- fuzz testing
 
## 📋 Checklist


adzialocha (changes requested):
Haven't checked the PR properly yet, but have some more general change requests:

* Think this can have more documentation. This will be a core functionality and we should be able to explain it's logic to people who don't know what this is
* The `stream` module is for `Stream` implementations (_not_ our namakemono specification "Stream" which probably will not be our future name for it anyhow), maybe you can think of another folder for that code?
* License headers :-)

sandreae:
> The stream module is for Stream implementations (not our namakemono specification "Stream" which probably will not be our future name for it anyhow), maybe you can think of another folder for that code?

I couldn't think of a great place for this code to live... moving it to `p2panda-core` makes the most sense but this also didn't feel quite right. Do you have any ideas @adzialocha

adzialocha:
> I couldn't think of a great place for this code to live... moving it to `p2panda-core` makes the most sense but this also didn't feel quite right. Do you have any ideas @adzialocha ?

Haven't checked closely now what this code does, but from first look it just seems to wrap what you have in `partial/mod.rs`? Maybe move the code just into the same file?

p2panda-stream/src/ordering/partial/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+        }
+    }
+
+    /// Pop the next item from the ready queue.
+    pub fn next(&mut self) -> Option<K> {
+        self.ready_queue.pop_front()
+    }
+
+    /// Process a new item which may be in a "ready" or "pending" state.
+    pub async fn process(&mut self, key: K, dependencies: Vec<K>) -> Result<(), PartialOrderError> {
```
adzialocha, line 88: `dependencies` can be a `&[K]`, with `K` being `Copy` it should be easy to handle.

p2panda-stream/src/ordering/partial/store.rs
```diff
@@ -0,0 +1,101 @@
...
+/// An implementation of this store trait provides the following functionality:
+/// - maintain a list of all items which have all their dependencies met
+/// - maintain a list of items which don't have their dependencies met
+/// - return all pending items which depend on a given item
+pub trait PartialOrderStore<K>
+where
+    K: Clone + Copy + StdHash + PartialEq + Eq,
+{
+    /// Add an item to the store which has all it's dependencies met already.
+    async fn add_ready(&mut self, key: K) -> Result<bool, PartialOrderError>;
```
adzialocha, line 20: Just a thought: How about `mark_ready` and `mark_pending`?

p2panda-stream/src/ordering/partial/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+///
+/// A <-- ?? <-- C
+///   \-- B1 <--/
+///
+/// C is not processed yet as we are missing one of its dependencies:
+///
+/// [A, B1]
+///
+/// Note that no checks are made for cycles occurring in the graph, this should be validated on
+/// another layer.
```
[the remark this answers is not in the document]
adzialocha: The doc-string might need another round of edits maybe, I'm not sure if I understand it reading it for the first time, it might get harder for people who don't know the concept of "ordering", "partial ordering", "dependency checking" or "Directed acyclic graph".

Maybe you can try to frame it in a way where this introduces a user to _what_ they are doing here (from the perspective of someone who is rather new to the concept)?

> "Struct for establishing partial order over items in a Directed-Acyclic-Graph."

The DAG structure is correct but it seems to me that this is a side-effect of creating a structure which represents dependencies (a "dependency graph"). I think the latter is easier to understand because it explains better what the interesting information is we use in this algorithm. 

Also there is no explanation of why we want to get a partial order of this and what a partial order is. You say:

> "This approach establishes a partial order, meaning not all items in the graph are comparable, and is non-deterministic"

.. I'm not sure what "comparable" means in this context and _maybe_ I know what you mean with "non-deterministic" but compared to what and how?

> "The main requirement is that all dependencies of an item are sorted "before" the item itself, the exact order is not a concern."

Not sure if I understand that sentence 😅

p2panda-stream/src/ordering/partial/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    /// Pop the next item from the ready queue.
+    pub fn next(&mut self) -> Option<K> {
+        self.ready_queue.pop_front()
+    }
+
+    /// Process a new item which may be in a "ready" or "pending" state.
+    pub async fn process(&mut self, key: K, dependencies: Vec<K>) -> Result<(), PartialOrderError> {
+        if !self.store.ready(&dependencies).await? {
```
adzialocha, line 89: Looks very efficient ✨

p2panda-stream/src/ordering/partial/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+    pub fn new(store: S) -> Self {
+        Self {
+            store,
+            ready_queue: VecDeque::new(),
+        }
+    }
+
+    /// Pop the next item from the ready queue.
+    pub fn next(&mut self) -> Option<K> {
+        self.ready_queue.pop_front()
```
adzialocha, line 84: Is it possible to mark an item as "ready" _after_ it got popped from the queue?

I'm thinking about crash resiliance here as we might get intro trouble when items got "freed" by being marked as ready, _then_ moved into the ready queue. IF the process crashes we would loose them and that information as the ready queue only lives in memory.

I can see that it would be easy to just mark them as ready as soon as they get returned, but maybe that messes with the logic in other parts?

p2panda-stream/src/ordering/partial/operations.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// A store containing p2panda operations.
+    operation_store: OS,
+
+    /// The inner PartialOrder struct which sorts the operation DAG into a partial order.
+    inner: InnerPartialOrder<Hash, POS>,
+
+    /// In memory cache of operations we have processed, before retrieving "ready" operations from
+    /// the store they will be taken from here if present. The store can be cleared at any time
+    /// in order to reduce memory use.
+    operation_cache: HashMap<Hash, Operation<E>>,
```
adzialocha, line 28: A ring buffer would be cool here, with a configurable size on start up. I think `VecDeque` can be used as a ring buffer here.

p2panda-stream/src/ordering/partial/operations.rs
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+        if let Some((header, body)) = self
+            .operation_store
+            .get_operation(hash)
+            .await
+            .map_err(|err| OperationDependencyCheckerError::StoreError(err.to_string()))?
+        {
+            let operation = Operation {
+                hash: header.hash(),
```
adzialocha, line 78: Is it possible to avoid the `hash()` operation here? It would be more costly than just using the already computed hash.

p2panda-stream/src/ordering/partial/store.rs
```diff
@@ -0,0 +1,101 @@
...
+use std::collections::{HashMap, HashSet};
+use std::fmt::Debug;
+use std::hash::Hash as StdHash;
+
+use super::PartialOrderError;
+
+/// Trait defining a store API for handling ready and pending dependencies.
+///
+/// An implementation of this store trait provides the following functionality:
+/// - maintain a list of all items which have all their dependencies met
```
adzialocha, line 12: Is it possible to only keep track of "pending" items and as soon as they are _not_ in the list anymore they are considered "ready"? That would allow to not grow the size of storage, though it would also not be much more than the operation itself already.

p2panda-stream/src/ordering/partial/operations.rs
```diff
@@ -0,0 +1,376 @@
...
+        let Some(hash) = self.inner.next() else {
+            return Ok(None);
+        };
+
+        if let Some(operation) = self.operation_cache.remove(&hash) {
+            return Ok(Some(operation));
+        }
+
+        if let Some((header, body)) = self
+            .operation_store
```
adzialocha, line 72: Interesting that this assumes that the operations are already stored, maybe this needs to be mentioned somewhere?

If we ingest operations _after_ the ordering we might need to change that in the future.

adzialocha (changes requested):
Very nice and well readable code, especially the tests are great.

Next to the obvious CHANGELOG entry I think this could still need a bit of doc-string love and then I have some more detailled points.

p2panda-stream/src/ordering/partial/mod.rs
adzialocha: In a "robust" implementation which would use this I can see how one would have a single SQL transaction for whatever the application does next and marking this as "ready", to not loose any data in that transition.

p2panda-stream/src/ordering/partial/store.rs
sandreae: Yeh, I like that!

p2panda-stream/src/ordering/partial/store.rs
sandreae: I considered this approach, as we discussed it before, however I can't see a way to make it work. If the "pending" items set is empty and we receive a new item with dependencies `[x, y]` we wouldn't know if these were items which have been processed and ready (therefore removed from the pending set), or that we have never seen (and therefore never added to the pending set).

sandreae: Maybe I'm misunderstanding your suggestion though!

p2panda-stream/src/ordering/partial/operations.rs
sandreae: It's quite optimal that `HashMap` is fast for look-up, we'd lose this property with a `VecDeque` though. A capacity to the cache would be nice, it's of course possible to have your own logic to decide when to call `clear_cache()`.

adzialocha: I'm sure there is ways to build a cache which is both performant to look up and has an automatic (configurable) limit

p2panda-stream/src/ordering/partial/store.rs
adzialocha: Ah, yes! Let's not do it then!

adzialocha (changes requested):
Thanks for tackling the transaction part!

I was wondering, maybe we can remove the caching part for now? Next to the API, I think it's opening up too many questions, we don't do it anywhere else and seems like an optimization we don't need yet. On top caching _is_ hard (what happens if an operation gets removed, the cache will not learn about that, is this causing any bugs etc.).

I'd be up for skipping that part for now and moving on with what we have (next to documentation improvements, that would be good to have before merging it). What do you think @sandreae?

p2panda-stream/src/ordering/partial/mod.rs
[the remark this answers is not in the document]
sandreae: Thanks for the feedback! I've made some (hopefully) improvements to the docs, let me know what you think.

adzialocha (approved):
Great! The docs are really nice now! Thank you!

I've added a small intro what "partial ordering" means in the context of causality and fixed a minor broken sentence, I hope that was correct.

PR #714, opened 2025-03-07
Refactor sync manager to reduce complexity

mycognosist:
Significant reduction in complexity, mostly due to more elegant state tracking for sync attempts.

We also introduce a separate retry queue for failed sync attempts.

Related to https://github.com/p2panda/p2panda/issues/701

## 📋 Checklist

### DIFF
--- a/p2panda-net/src/sync/manager.rs
+++ b/p2panda-net/src/sync/manager.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use std::collections::{HashMap, HashSet, VecDeque};
+use std::collections::hash_map::Entry as HashMapEntry;
+use std::collections::{HashMap, VecDeque};
 
 use anyhow::{Context, Error, Result};
 use iroh::Endpoint;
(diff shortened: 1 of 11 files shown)


p2panda-net/src/sync/manager.rs
```diff
@@ -322,64 +303,53 @@ where
...
-        if self.config.is_resync() {
-            trace!("schedule re-sync attempt");
-            sync_attempt.completed = Some(Instant::now());
-            self.resync_queue.push_back(sync_attempt);
+    /// The attempt is pushed to the back of the retry queue if the maximum number of retry
+    /// attempts has not been exceeded.
+    async fn complete_failed_sync(&mut self, scope: Scope<T>, err: Error) -> Result<()> {
+        warn!("sync attempt failed for scope {:?}: {}", scope, err);
+
+        // @TODO(glyph): There is a bug in this flow. If the sync initiation attempt fails before
```
mycognosist, line 329: Here I describe the gossip buffer issue. We should be sending `SyncFailed` but it causes the `retry_after_error` test to fail because the gossip buffer unlocking code panics. I propose we rather not panic and simply log a warning.

p2panda-net/src/sync/manager.rs
```diff
@@ -35,23 +33,47 @@ impl<T> ToSyncActor<T> {
     }
 }
 
-#[derive(Debug)]
-struct SyncAttempt<T> {
+/// Sync session status.
```
sandreae, line 36: This is nice, I see how state for all sessions is handled more clearly now.

p2panda-net/src/sync/manager.rs
sandreae: Just to check I understand: if a sync session is initiated but then fails before `HandshakeSuccess` message is received, we never locked our gossip buffer, and so when we handle the `SyncFailed` event and decrement the buffer a panic occurs in the engine because the buffer counter can go below zero. It could be that there are other sync sessions occurring, and so it wouldn't panic yet, but it _would_ panic later when a different sync session finished and tried to decrement the same buffer counter. 

I can't think of any downside to removing the panic in the engine. In all cases it seems that there are no sync sessions running and so the buffer should be unlocked.

sandreae (commented):
This looks good to me! Only question is around what to do about the issue you take note of where the gossip buffer can be decremented below `0` and panic. Do you suggest resolving that in this PR or a follow-up?

p2panda-net/src/sync/manager.rs
mycognosist: Your understanding is exactly correct.

mycognosist:
> This looks good to me! Only question is around what to do about the issue you take note of where the gossip buffer can be decremented below `0` and panic. Do you suggest resolving that in this PR or a follow-up?

I'm leaning towards resolving the issue in this PR. It would be a very small change, probably just replacing the panic with a logging statement. Then I can uncomment the `SyncFailed` send code in the manager and everything will be in a working state upon merging this PR.

sandreae (approved):

Issue #717, opened 2025-03-15
"consistent topic to stream id mapping" panic in topic streams


adzialocha:
p2panda version `0.3.0` in Aardvark

```
thread 'tokio-runtime-worker' panicked at /home/adz/dev/.gnome-builder/projects/aardvark/builds/org.p2panda.aardvark.json-flatpak-org.gnome.Platform-47-x86_64-adz-is-a-relay/cargo-home/registry/src/index.crates.io-1949cf8c6b5b557f/p2panda-net-0.3.0/src/engine/topic_streams.rs:339:14:
consistent topic to stream id mapping
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_display
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:261:5
   3: core::option::expect_failed
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/option.rs:2024:5
   4: core::option::Option<T>::expect
             at /usr/lib/sdk/rust-stable/lib/rustlib/src/rust/library/core/src/option.rs:933:21
   5: p2panda_net::engine::topic_streams::TopicStreams<T>::on_sync_message::{{closure}}
             at ./dev/.gnome-builder/projects/aardvark/builds/org.p2panda.aardvark.json-flatpak-org.gnome.Platform-47-x86_64-adz-is-a-relay/cargo-home/registry/src/index.crates.io-1949cf8c6b5b557f/p2panda-net-0.3.0/src/engine/topic_streams.rs:336:26
   6: p2panda_net::engine::engine::EngineActor<T>::on_actor_message::{{closure}}
             at ./dev/.gnome-builder/projects/aardvark/builds/org.p2panda.aardvark.json-flatpak-org.gnome.Platform-47-x86_64-adz-is-a-relay/cargo-home/registry/src/index.crates.io-1949cf8c6b5b557f/p2panda-net-0.3.0/src/engine/engine.rs:314:22
   7: p2panda_net::engine::engine::EngineActor<T>::run_inner::{{closure}}
[...]

PR #719, opened 2025-03-16
Only proceed with sync if handshake topic is of interest

mycognosist:
## Problem

In an application context with many topics, a peer may attempt to initiate a sync session with one who is not interested in that topic. In other words: the session acceptor is not subscribed to the topic. A panic then occurs when the accepting peer tries to retrieve the stream for the received topic (because the stream doesn't exist; it was never created because no topic subscription occurred).

## Solution

We need a way for the sync session acceptor to query whether the topic is of interest. If not: return an error. If so, continue to phase II and begin processing data messages from the initiator.

## Implementation

After receiving the handshake message from the initiator (which includes the topic), we create a oneshot channel and include the sender as part of the `SyncHandshakeSuccess` message. The message is sent to the actor, allowing the `topic_streams` module to report-back concerning whether or not it is subscribed to the given topic. We listen on the oneshot receiver and exit the sync session before any data messages are exchanged (in "phase II" of the sync protocol) if the topic is not known.

A new error variant has been introduced to signal this failure scenario: `SyncError::UnknownTopic`.

Addresses: https://github.com/p2panda/p2panda/issues/717

-----

## Open Questions

Do we want to introduce an error wire message in `LogSyncProtocol`? Right now we only have `Message::Have(_, _)`, `Message::Data(_, _)` and `Message::Done`. What if there was a `Message::Denied(_)` variant (or similar) which is used to communicate: "nothing went wrong but I don't want to sync with you". This would be useful in the case of two peers not sharing mutual topic interests; the acceptor can then clearly communicate that they're ending the session early because they're not interested in the topic of the initiator (and not simply end the session without any context). 

Right now we simply return an `UnexpectedBehaviour` error when the topic query sent by the initiator is not in the topic map of the acceptor: https://github.com/p2panda/p2panda/blob/main/p2panda-sync/src/log_sync.rs#L253-L258

## 📋 Checklist

### DIFF
--- a/p2panda-net/src/engine/engine.rs
+++ b/p2panda-net/src/engine/engine.rs
@@ -65,7 +65,7 @@ pub enum ToEngineActor<T> {
     SyncHandshakeSuccess {
         topic: T,
         peer: PublicKey,
-        topic_is_known_tx: oneshot::Sender<bool>,
+        topic_is_known_tx: Option<oneshot::Sender<bool>>,
     },
     SyncMessage {
         topic: T,
--- a/p2panda-net/src/engine/engine.rs
+++ b/p2panda-net/src/engine/engine.rs
@@ -65,6 +65,7 @@ pub enum ToEngineActor<T> {
     SyncHandshakeSuccess {
         topic: T,
         peer: PublicKey,
+        topic_is_known_tx: oneshot::Sender<bool>,
     },
     SyncMessage {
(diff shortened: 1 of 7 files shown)


mycognosist:
Closing this, as we already return an error in the acceptor if the topic is unknown. Not entirely sure how the error was thrown in the first place.

PR #734, opened 2025-03-27
DCGKA for message encryption scheme & identity- and pre-key registry

adzialocha:
> This PR is based on https://github.com/p2panda/p2panda/pull/733 and introduces the last part of the DCGKA paper. The message ratchet and `Group` wrapper APIs are introduced in later PRs.

### DCGKA

A decentralized continuous group key agreement protocol (DCGKA) for p2panda's "message encryption" scheme.

The implementation follows the DCGKA protocol specified in the paper: "Key Agreement for Decentralized Secure Group Messaging with Strong Security Guarantees" by Matthew Weidner, Martin Kleppmann, Daniel Hugenroth, Alastair R. Beresford (2020).

### Identity & Pre-Key Registry

`PreKeyRegistry` and `IdentityRegistry` Traits and reference-implementation to maintain public pre-key bundles and identity keys of other members in the network.

## 📋 Checklist

### DIFF
--- a/p2panda-group/src/message_scheme/acked_dgm.rs
+++ b/p2panda-group/src/message_scheme/acked_dgm.rs
@@ -1,0 +1,438 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+#[cfg(any(test, feature = "test_utils"))]
+pub mod test_utils {
+    use std::collections::{HashMap, HashSet};
+    use std::fmt::Debug;
+    use std::marker::PhantomData;
+
+    use serde::{Deserialize, Serialize};
+    use thiserror::Error;
+
+    use crate::traits::{AckedGroupMembership, IdentityHandle, OperationId};
+
+    /// Non-optimal "Acked" Decentralised Group Membership CRDT implementation for p2panda's
+    /// message encryption scheme.
+    ///
+    /// This does not support removing and re-adding a member.
+    #[derive(Clone, Debug)]
+    pub struct AckedTestDGM<ID, OP> {
+        _marker: PhantomData<(ID, OP)>,
+    }
+
+    impl<ID, OP> AckedTestDGM<ID, OP>
+    where
+        ID: IdentityHandle + Serialize + for<'a> Deserialize<'a>,
+        OP: OperationId + Serialize + for<'a> Deserialize<'a>,
+    {
+        pub fn init(my_id: ID) -> State<ID, OP> {
+            State {
+                my_id,
+                members: HashSet::new(),
+                removed_members: HashSet::new(),
+                infos: HashMap::new(),
+                remove_infos: HashMap::new(),
+                adds_by_msg: HashMap::new(),
+                removes_by_msg: HashSet::new(),
+            }
+        }
+    }
+
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    pub struct State<ID, OP>
+    where
+        ID: IdentityHandle,
+        OP: OperationId,
+    {
+        my_id: ID,
+        members: HashSet<ID>,
+        removed_members: HashSet<ID>,
+        infos: HashMap<ID, MemberInfo<ID, OP>>,
+        remove_infos: HashMap<OP, RemoveInfo<ID>>,
+        adds_by_msg: HashMap<OP, ID>,
+        removes_by_msg: HashSet<OP>,
+    }
+
+    impl<ID, OP> AckedGroupMembership<ID, OP> for AckedTestDGM<ID, OP>
+    where
+        ID: IdentityHandle + Serialize + for<'a> Deserialize<'a>,
+        OP: OperationId + Serialize + for<'a> Deserialize<'a>,
+    {
+        type State = State<ID, OP>;
+
+        type Error = TestAckedGroupError<ID, OP>;
+
+        fn from_welcome(my_id: ID, mut y: Self::State) -> Result<Self::State, Self::Error> {
+            y.my_id = my_id;
+            Ok(y)
+        }
+
+        fn create(my_id: ID, initial_members: &[ID]) -> Result<Self::State, Self::Error> {
+            let mut initial_members = initial_members.to_vec();
+            if !initial_members.contains(&my_id) {
+                initial_members.push(my_id);
+            }
+
+            let mut infos = HashMap::with_capacity(initial_members.len());
+            let mut members = HashSet::with_capacity(initial_members.len());
+
+            for member in &initial_members {
+                infos.insert(*member, MemberInfo::new(*member, None, &initial_members));
+                members.insert(*member);
+            }
+
+            Ok(Self::State {
+                my_id,
+                members,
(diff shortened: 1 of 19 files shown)


adzialocha:
I believe there's two things missing in this PR:

1. Support for re-adding members: This needs a better DGM implementation (length-prefixed CRDT instead of insert-and-remove-only-set) and some additional adjustments in the DCGKA protocol (removing ratchet secrets after `process_remove` etc.). I would tackle that as part of the DGM work + add dedicated tests
2. More tests, more test tooling (especially around processing messages in a network), since the tests here already take a lot of space I suggest continuing with adding more tests in subsequent PRs, also looking at this [`dcgka-fuzz`](https://github.com/p2panda/p2panda/compare/dcgka...dcgka-fuzz)

p2panda-group/src/message_scheme/dcgka.rs
```diff
@@ -0,0 +1,1587 @@
...
+        rng: &Rng,
+    ) -> DcgkaProcessResult<ID, OP, PKI, DGM, KMG> {
+        y.dgm =
+            DGM::create(y.my_id, &initial_members).map_err(|err| DcgkaError::DgmOperation(err))?;
+        Self::process_seed(y, &sender, seq, direct_message, rng)
+    }
+
+    /// Called by group members when they receive the "ack" message.
+    ///
+    /// In this function, `ackID` and `ackOP` are the user id and id of the acknowledged message.
```
mycognosist, line 346: Nitpick: in some places we write `id` and in `ID` in others. Would be great to stick with one approach throughout.

p2panda-group/src/message_scheme/dcgka.rs
```diff
@@ -0,0 +1,1587 @@
...
+    /// Adds a new group member.
+    ///
+    /// The added group member will receive a direct "welcome" message, every member will process
+    /// an "add" control message.
+    pub fn add(
+        y: DcgkaState<ID, OP, PKI, DGM, KMG>,
+        added: ID,
+        rng: &Rng,
+    ) -> DcgkaOperationResult<ID, OP, PKI, DGM, KMG> {
+        // Construct a control message of type "add" to broadcast to the group
```
mycognosist, line 503:
```suggestion
        // Construct a control message of type "add" to broadcast to the group.
```

p2panda-group/src/message_scheme/dcgka.rs
```diff
@@ -0,0 +1,1587 @@
...
+        // concurrently C adds D to the group. When A distributes a new seed secret through
+        // 2SM-encrypted direct messages, D will not be a recipient of one of those direct
+        // messages, since A did not know about D's addition at the time of sending. D cannot
+        // derive any of the member secrets for this update. When B updates its KDF ratchet using
+        // A's seed secret, it will compute an update secret that D does not know, and D will not
+        // be able to decrypt B's subsequent application messages.
+        //
+        // In this example, B may receive the add and the update in either order. If B processes
+        // A's update first, the seed secret from A is already incorporated into B's ratchet state
+        // at time time of adding D; since B sends this ratchet state to D along with its "add-ack"
```
mycognosist, line 1033:
```suggestion
        // at the time of adding D; since B sends this ratchet state to D along with its "add-ack"
```

p2panda-group/src/message_scheme/dcgka.rs
```diff
@@ -0,0 +1,1587 @@
...
+        // message, no further action is needed. On the other hand, if B processes the addition of
+        // D first, then when B subsequently processes A's update, B must take the member secret it
+        // derives from A's seed secret and forward it to D, so that D can compute B's update
+        // secret for A's update.
+        //
+        // Recall that first we set recipients to be the set of group members at the time the
+        // update/remove was sent, except for the sender. We then compute the current set of
+        // members according to the local node. The set difference thus computes the set of users
+        // whose additions have been processed by the local user, but who were not yet known to
+        // sender of the update.
```
mycognosist, line 1043:
```suggestion
        // the sender of the update.
```

p2panda-group/src/message_scheme/dcgka.rs
```diff
@@ -0,0 +1,1587 @@
...
+        };
+        let (y_2sm_i, y_my_keys_i, plaintext) =
+            TwoParty::<KMG, OneTimeKeyBundle>::receive(y_2sm, y.my_keys, ciphertext)?;
+        y.my_keys = y_my_keys_i;
+        y.two_party.insert(*sender, y_2sm_i);
+        Ok((y, plaintext))
+    }
+
+    /// Generates the next update secret for group member ID. It implements the outer KDF of the
+    /// ratchet. The ratchet state is stored in γ.ratchet[ID]; we use a HMAC-based key derivation
```
mycognosist, line 1146:
```suggestion
    /// ratchet. The ratchet state is stored in γ.ratchet[ID]; we use an HMAC-based key derivation
```

mycognosist (approved):
🌈 🌱 

Left a few suggested changes (for docs) and a comment or two.

p2panda-group/src/message_scheme/acked_dgm.rs
```diff
@@ -0,0 +1,562 @@
...
+        /// Please note that a user can only be added to a group once.
+        fn add(
+            mut y: Self::State,
+            adder: ID,
+            added: ID,
+            message_id: OP,
+        ) -> Result<Self::State, Self::Error> {
+            let mut added_info = MemberInfo::new(added, Some(adder), &[]);
+            added_info.acks.insert(adder);
+            added_info.acks.insert(added);
```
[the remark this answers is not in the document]
adzialocha: This seems to come from the Java implementation where added members do not need to ack their own adds, see comment here: https://github.com/trvedata/key-agreement/blob/main/group_protocol_library/src/main/java/org/trvedata/sgm/StrongRemoveDgm.java#L159

The paper does not talk about this implementation detail. I assume it's because ratchets use hard-coded constants during the adding phase which are already known by everyone?

Since this is an test implementation I don't think this will stay long anyhow, actually, we _are_ creating acks when being added currently and it works as well. I assume this is just an optimization.

p2panda-group/src/message_scheme/acked_dgm.rs
```diff
@@ -0,0 +1,562 @@
...
+
+            // If `actor` acknowledged adding or removing a member in the past, then we can be sure
+            // that `added` also acknowledges it as they must have been made aware of this history
+            // by receiving the welcome message from `actor`.
+            for member in &y.members {
+                let member_info = y
+                    .infos
+                    .get_mut(member)
+                    .expect("infos values should be consistent with members keys");
+                if member_info.acks.contains(&adder) {
```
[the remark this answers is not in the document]
adzialocha: I haven't looked close into this implementation (as it will be removed eventually), but similar to the point above I assume it's an optimization.

sandreae (approved):

PR #742, opened 2025-04-16
"Data Encryption" Group API & orderer

adzialocha:
## 📋 Checklist


adzialocha:
This definitely needs more tests, I will do this with follow up PRs. One will be a fuzz tests, going through all combinations of group operations. More tests will follow as part of the "integration" work in the next weeks, next to improved documentation.

sandreae (approved):

PR #749, opened 2025-05-12
`p2panda-auth`: group management protocol

sandreae:
This PR introduces the core protocol and data primitives which will be used in `p2panda-auth` for managing and representing groups of identities in distributed systems. This crate itself is agnostic with regards to how an identity is represented, it is only required that any identifier be globally unique. Systems for verifying identities and messages are abstracted away behind trait interfaces in order to support integration into different systems. 

## Design goals

Groups will be used in `p2panda-auth` and in more "opinionated" p2panda tools to understand _who_ has permission to read and write data to a topic. This information will inform what data peers sync, who data is encrypted towards, and at an application layer, if an author has permission to perform a write action. Individual identities should be added to a group and assigned an access level, the group can then be associated with some data topic, certain group members should be able to manage the group itself.

Our approach to group management is influenced on our research of the problem area, our experiences of what application developers need when implementing user management patterns and the fact that we will be integrating this system with `p2panda-encryption` in order to encrypt application data based on group members.

## Requirements

- generic over the precise primitives used to represent an individual or group.
- support all common group management actions (create, add/remove member, re-add member).
- group members can be assigned different access levels including being elevated to "admin" position allowing them to perform actions on the group itself.
- concurrent conflicting actions are detected and resolved based on custom logic (we don't bake in one approach to conflict resolution).
- as well as the protocol defined access levels (see below) application developers can include additional "conditions" which restrict access further based on domain specific logic.
- groups can be nested and are re-usable, making it possible to model hierarchical access patterns, roles, or multi-device identities. 

## Approach

A group contains a set of members which are either "individuals" or themselves a group. Individuals can be understood to be "stateless" as they are represent a single immutable identity, whereas groups can be understood to be "stateful", as they contain a mutable set of members. Each group member is assigned one of the following access levels: `Pull`, `Read`, `Write` & `Manage`, these levels are hierarchical with any greater level containing all lower access levels. An additional generic `C` condition can be attached to an access level in order to place additional conditions on a given access level based on application domain requirements. The `PartialOrd` bound is required of any `C`. Although not introduced in this PR, groups will eventually be used to understand who has permission to perform certain actions with regards to an encrypted application topic/message stream. The access levels confer the following meaning:
- `Pull` permission to sync encrypted data from a topic
- `Read` permission to sync and read (+decrypt) data from a topic
- `Write` permission to sync, read (+decrypt) and write (+encrypt) data to/from a topic
- `Manage` all previous plus permission to manage the members of a group (aka admin)

Group encryption for application messages will be introduced with the integration of `p2panda-auth` and `p2panda-encryption`.

The core data structure used for group management is a Directed Acyclic Graph (DAG) of operations containing group actions. Possible actions are `Create`, `Add`, `Remove`, `Promote` & `Demote`. Operations in the graph refer to the previous state of the graph (from the actors perspective) as well as any sub-group graph heads. This allows a system level orderer (dependency checker) to ensure all peers process all operations only once their dependencies have been met. Member state at each point in the graph is represented by a [state-based CRDT](https://github.com/p2panda/p2panda/pull/748) object, although this state is only used locally, it allows us to not require total ordering of operations and to merge states from any branches in the graph without requiring a graph re-build (no graph traversal).

### Nested groups example
groups_2

### Concurrent operations example
groups

A concrete resolver implementation is implemented in this follow-up PR https://github.com/p2panda/p2panda/pull/753 following a "strong remove" strategy based on the approach described in this [blog post](https://p2panda.org/2025/02/24/group-encryption.html).

image

In this crate, as well as in other p2panda crates, we provide several traits which represent core interfaces in a generic manner, allowing for integration into a variety of systems, and to remain agnostic over the core primitives used. There is then an implementation of the group management protocol (still agnostic over identity primitives) which provides all core types and functionality. Importantly, even with this opinionated implementation, it is still possible to customise certain aspects of group management which are very dependent on the specific requirements of the application domain. 

### Trait interfaces

- `AuthGroup` generic interface with basic methods needed for implementing a the group management messaging protocol. The `prepare` method takes a generic action and a state object, and should return an operation which other instances can pass into there `process` method. This is the trait which is implemented on `Group`.
- `Resolver` trait bound on the generic "resolver" parameter provided to an AuthGroup implementation. Allows users of a concrete AuthGroup implementation to still customise which operations should be included in the final CRDT state. For example, in the case of group management, one could use a "strong remove" or a "seniority wins" resolver to customise how concurrent member removals are handled. 
- `Operation` trait bound required on the associated generic message type of the `Ordering` trait expressing that messages must have a sender id, message id, list of dependencies and a payload.
- `Ordering` abstract interface for ordering messages/operations in a system. In the case of `auth` operations form a graph of dependencies, and so this can be leveraged to linearize operations into a partial-order and expect that they are only processed in this order.
- `GroupStore` interface for providing shared access to all groups in the current context (needs refactoring to account for new transactional approach to persistent)

### Implementation

- `Group` the core group protocol implementation. Implements `AuthGroup` and processes `GroupControlMessage`'s and `GroupState`
- `GroupState` representation of a single group's state, contains references to shared group and orderer state.
- `GroupAction` enum of actions which can be applied to a group state
- `GroupControlMessage` protocol message enum
- a display method for printing auth graphs in DOT format for graph visualization (see above diagrams) 

### Testing

For testing there is an `Ordering` and `GroupStore` implementation provided, as well as a `Network` abstraction for modelling group actions issued by multiple actors (possibly concurrently).

### Next steps
- fuzz tests
- adjust state methods to account for promoting more than one step
- allow C conditions to be attached to any access level
- introduce Access trait with validation method in order to allow C conditions to effect system access levels
  - eg. introducing "super" admins which are the only ones who can remove users, or designating specific read scopes 

## 📋 Checklist


p2panda-auth/Cargo.toml
```diff
@@ -16,6 +16,9 @@ keywords = ["auth", "access-control", "groups"]
 all-features = true
 
 [dependencies]
+petgraph = { version = "0.8.1", features = ["serde-1"] }
+rand = { version = "0.9.1", features = ["alloc"] }
```
adzialocha, line 20: `rand` is used in the test utils. I suggest you introduce a `test_utils` feature flag and restrict these features to that flag and `test` environments, then you can also make this dependency optional

p2panda-auth/src/traits/group_store.rs
```diff
@@ -0,0 +1,22 @@
...
+    type State: Clone + Debug;
+
+    type Error: Error;
+
+    /// Insert a group state into the store.
+    fn insert(y: Self::State, id: &ID, group: &G) -> Result<Self::State, Self::Error>;
+
+    /// Get a group's state from the store.
+    fn get(y: &Self::State, id: &ID) -> Result<Option<G>, Self::Error>;
+}
```
adzialocha, line 22: In our future storage / transaction API design we will not have any write-methods on concrete trait interfaces (only read), otherwise we can't assure atomicity.

Writes will be expressed with implementing an `WriteToStore` trait (name pending) on the `GroupStore::State` object against a concrete store implementation.

You can see an example in our playground!

p2panda-auth/src/group/resolver.rs
```diff
@@ -0,0 +1,76 @@
...
+use thiserror::Error;
+
+use crate::group::{GroupControlMessage, GroupState};
+use crate::traits::{GroupStore, IdentityHandle, Operation, OperationId, Ordering, Resolver};
+
+use super::GroupStateInner;
+
+// TODO: introduce all error types.
+#[derive(Debug, Error)]
+pub enum GroupResolverError {}
```
adzialocha, line 16: Nitpick: I found code easier to read which starts with the most concrete implementation at the top, error types seem less important

p2panda-auth/src/group/resolver.rs
```diff
@@ -0,0 +1,76 @@
...
+use std::collections::HashSet;
+use std::fmt::Display;
+use std::{fmt::Debug, marker::PhantomData};
+
+use thiserror::Error;
+
+use crate::group::{GroupControlMessage, GroupState};
+use crate::traits::{GroupStore, IdentityHandle, Operation, OperationId, Ordering, Resolver};
+
+use super::GroupStateInner;
```
adzialocha, line 12: Replace with `crate` import

p2panda-auth/Cargo.toml
adzialocha: Alternatively you can move `rand` just into dev-dependencies, but I assume you want to fuzz your code at one point, then it makes sense to already expose the `test_utils`

p2panda-auth/src/group/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+            Ok(operation) => operation,
+            Err(_) => panic!(),
+        };
+
+        // Queue the operation in the orderer.
+        //
+        // Even though we know the operation is ready for processing (ordering dependencies are
+        // met), we need to queue it so that the orderer progresses to the correct state.
+        //
+        // TODO: maybe this should rather happen inside ORD::next_message?
```
adzialocha, line 610: I agree, we usually don't need to make the orderer aware of our own messages (they are always already processed), or at least we shouldn't need to enforce this already on this level

p2panda-auth/src/traits/group_store.rs
adzialocha: Some possible approach could be:

1. Remove `insert` from `GroupStore`
2. Keep `GroupStore` otherwise around with `get` method, it allows different auth groups to have access to a "global, shareable" store to read information from
3. Remove `State` from `GroupStore` (a type like `SqliteStore` would implement `GroupStore`, so a state object doesn't need to be known outside of the database)
4. I see only one spot where `insert` is called, that is inside of a `AuthGraph::process` implementation. I would suggest to move that inserted group into the `AuthGraph::State`
5. You add a method on `GroupState` which has a `get_group` (etc.) method. This method looks up if a group exists in memory (in `GroupState`), if not, it'll ask the store which implements `GroupStore` (probably async in the future)
6. All methods using the concrete `GroupState` type call `get` on the `GroupState` instead of `GroupStore`
7. Later, when have transactions we will implement an write trait on top of `GroupState` which will do the actual insertion into SqliteStore

Or, even simpler (for now, as we don't have the trait interfaces yet for transactions):

1. Remove `GroupStore`
2. Have everything inside the concrete type `GroupState`
3. When `GroupState` gets initialised, the API asks for a reference-countable object which has the HashMap inside
4. Every instance learns about that object, this is how you get the "shared in-memory" database

adzialocha (changes requested):

p2panda-auth/src/group/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+/// state required for GroupStore. In that case we don't need this separation and the inner and
+/// outer states could be combined again.  
+///
+/// TODO: One other reason these are separate structs is because I got into "recursive generic
+/// parameter" hell when trying to define a generic type (GroupStoreState) which returned another
+/// type (GroupState) which itself contained the container generic type (GroupStoreState)....
+/// Returning an "inner" state from the store (which doesn't know about the store) got around this
+/// issue.
+#[derive(Clone, Debug)]
+pub struct GroupStateInner<ID, OP, MSG>
```
adzialocha, line 177: The handling of `GroupStateInner` is rather confusing and makes implementing the APIs hard, would be good to get rid of it or at least not surface it to the outside

adzialocha:
Next to the `GroupError` infinite cycle issue I also observed another "cycle" (I believe) when using the API:

```
error[E0275]: overflow evaluating the requirement `GroupResolver<MemberId, MessageId, DocumentMessage>: Resolver<GroupState<MemberId, MessageId, ..., ..., ...>, ...>`
   --> p2panda-document/src/lib.rs:163:24
    |
163 |         let universe = Universe::<GroupResolver<MemberId, MessageId, DocumentMessage>, SqliteStore> {
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: required by a bound in `Universe`
   --> p2panda-document/src/lib.rs:108:9
    |
106 | pub struct Universe<RS, GS>
    |            -------- required by a bound in this struct
107 | where
108 |     RS: Resolver<AuthGroupState<RS, GS>, DocumentMessage> + fmt::Debug,
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Universe`

For more information about this error, try `rustc --explain E0275`.
error: could not compile `p2panda-document` (lib test) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

This happens when keeping the resolver generic and handling trait bounds like that:

`RS: Resolver<AuthGroupState<RS, GS>, DocumentMessage>`

`RS` is defined in `AuthGroupState` which again requires `AuthGroupState`, so one ends up with something like this:

`RS: Resolver<AuthGroupState<Resolver<AuthGroupState<Resolver<AuthGroupState<Resolver<AuthGroupState<RS, GS>, DocumentMessage>, GS>, DocumentMessage>, GS>, DocumentMessage>, GS>, DocumentMessage>` etc.

p2panda-auth/src/traits/operation.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn sender(&self) -> ID;
+
+    /// Other operation dependencies.
+    fn dependencies(&self) -> &Vec<OP>;
+
+    /// The last graph tips in the group this operation belongs to.
+    fn previous(&self) -> &Vec<OP>;
+
+    /// Payload of this operation.
+    fn payload(&self) -> &P;
```
adzialocha, line 22: I have trouble using the borrowed P type as conversions are required in the `payload` method and the returned type would not live long enough. I'd change `P` to an owned type instead

p2panda-auth/src/group/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+    #[error("operation for group {0} processed in group {1}")]
+    IncorrectGroupId(ID, ID),
+}
+
+/// A group member which can be a single stateless individual, or a stateful group. In both cases
+/// the member identifier is the same generic ID.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
+pub enum GroupMember<ID> {
+    Individual(ID),
+    Group { id: ID },
```
adzialocha, line 66: `Group { id: ID }` -> `Group(ID)` (for consistency)

p2panda-auth/src/traits/auth_graph.rs
```diff
@@ -0,0 +1,45 @@
...
+///
+/// Generic parameter RS (resolver) allows for introducing custom logic which decides if
+/// operations should be included in any state-deriving process. This can include the handling of
+/// concurrent operations which would cause conflicting state changes.
+pub trait AuthGraph<ID, OP, RS, ORD>
+where
+    ID: IdentityHandle,
+    OP: OperationId + Ord,
+    RS: Clone + Resolver<Self::State, ORD::Message>,
+    ORD: Clone + Ordering<ID, OP, Self::Action>,
```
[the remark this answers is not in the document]
sandreae: I've managed to remove Clone trait bounds for all states except for in one place where GroupStoreState needs to be cloned, I'm working on removing this too though.

p2panda-auth/src/traits/group_store.rs
sandreae: Thanks for the suggestions :+1: 

I've had a go at a variant of the simpler option 2. I've kept the trait definition, but removed the state associated type, so it now represents a "classic" group store. My thinking is that we keep it like this until the transaction APIs are ready, and then migrate directly to that pattern. Let me know what you think.

p2panda-auth/src/group/mod.rs
```diff
@@ -0,0 +1,821 @@
...
+    },
+    Promote {
+        member: GroupMember<ID>,
+        access: Access<()>,
+    },
+    Demote {
+        member: GroupMember<ID>,
+        access: Access<()>,
+    },
+}
```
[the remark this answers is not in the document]
sandreae: Haven't dealt with this yet, but will do!

p2panda-auth/src/group/mod.rs
```diff
@@ -0,0 +1,752 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#![allow(clippy::type_complexity)]
+#![allow(dead_code)]
```
adzialocha, line 3: Clippy attributes should be local and maybe even with a comment to avoid overseeing linting errors.

I suggest taking them all away first, checking if they're still necessary, most stuff is maybe even used already, otherwise consider exporting on crate level. If that still doesn't work, it's fine to apply the rule, but better on local level.

p2panda-auth/src/group/test_utils/group_store.rs
```diff
@@ -0,0 +1,49 @@
...
+where
+    ID: IdentityHandle,
+{
+    fn default() -> Self {
+        Self(Rc::new(RefCell::new(HashMap::new())))
+    }
+}
+
+impl GroupStore<char> for TestGroupStore<char> {
+    type Group = TestGroupState;
```
adzialocha, line 32: Not sure if this needs to be generic. This will almost always be a `GroupState` or not?

p2panda-auth/src/traits/group_store.rs
```diff
@@ -0,0 +1,20 @@
...
+use std::error::Error;
+
+use super::IdentityHandle;
+
+/// API for global group store.
+pub trait GroupStore<ID>
+where
+    ID: IdentityHandle,
+{
+    type Group;
```
adzialocha, line 12: Not sure if this needs to be generic. This will almost always be a `GroupState` or not?

p2panda-auth/src/traits/ordering.rs
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+#![allow(dead_code)]
+
+use std::error::Error;
+
+use super::Operation;
```
adzialocha, line 6: `super` -> `crate`

p2panda-auth/Cargo.toml
```diff
@@ -3,20 +3,30 @@ name = "p2panda-auth"
...
+    "glyph <glyph@mycelial.technology>",
 ]
 repository = "https://github.com/p2panda/p2panda"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
 keywords = ["auth", "access-control", "groups"]
 
+[features]
+default = []
+test-utils = ["rand"]
```
adzialocha, line 17: Minor thing for consistency, but we use `test_utils` (with underscore) across all crates so far

p2panda-auth/src/group/mod.rs
```diff
@@ -0,0 +1,752 @@
...
+            group_id,
+            states: Default::default(),
+            operations: Default::default(),
+            ignore: Default::default(),
+            graph: Default::default(),
+            group_store,
+            orderer_y,
+            _phantom: PhantomData,
+        }
+    }
```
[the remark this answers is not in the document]
sandreae: Nope, I'll make it public.

p2panda-auth/src/traits/group_store.rs
sandreae: Yes, you're right. If we had a new AuthGroup implementation it would likely have it's own state store. Nice.

p2panda-auth/src/group/resolver.rs
```diff
@@ -0,0 +1,99 @@
...
+
+// TODO: introduce all error types.
+#[derive(Debug, Error)]
+pub enum GroupResolverError {}
+
+/// Resolver for group membership auth graph.
+#[derive(Clone, Debug, Default)]
+pub struct GroupResolver<ID, OP, C, ORD, GS> {
+    _phantom: PhantomData<(ID, OP, C, ORD, GS)>,
+}
```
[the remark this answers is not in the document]
sandreae: Yeh, I see that this might be an issue... In order to avoid code collisions with the `strong-remove`  PR which is open I'll deal with this in a follow-up step.

PR #753, opened 2025-05-23
`p2panda-auth` strong-remove resolver

sandreae:
This PR introduces a group membership resolver implementation which follows a "strong remove" ruleset. Concurrent operations are identified and processed, any which should be invalidated are added to the operation filter and not applied to the group state. Once an operation has been filtered, any operations which depended on any resulting state will not be applied to group state either.

## Strong remove behaviors
image

## Ruleset for Concurrent Operations

The following ruleset is applied when choosing which operations to "filter" when concurrent operations are processed. It can be assumed that the behavior is equivalent for an admin member being removed, or demoted from admin to a lower access level.

### 1) Removals

If a removal has occurred, filter any concurrent operations by the removed member, as long as it's 1) not a predecessor of the remove operation, and 2) not a mutual removal (removal of the remover by the removed member).

### 2) Mutual removals

Mutual removals result in both members being removed from the group, and any dependent concurrent branches are not applied to group state. We imagine further implementations taking different approaches, like resolving by seniority, hash id, quorum or some other parameter.

If a mutual removal has occurred, we want to retain the removal operations but filter all concurrent operations performed by the removed members (keeping predecessors of the remove).

### 3) Re-adding member concurrently

If Alice removes Charlie and Bob removes then adds Charlie concurrently, Charlie is still in the group. However, if Charlie performed any concurrent actions, these will be filtered along with any dependent operations.

### 4) Filtering of dependent operations

When an operation is "explicitly" filtered it may cause dependent operations to become invalid, these operations will not be applied to the group state. 

### Example 1
- A and B concurrently remove each other, both are removed from the group
graphviz(53)

### Example 2
- removal of B causes concurrent operations to be filtered
- B is re-added
- after "merge" B is back in the group and can perform actions
graphviz(51)

### Example 3
- removal of B causes concurrent operations to be filtered
- operations from C are also filtered as they depended on B's operation 
graphviz(52)

## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/group/graph.rs
+++ b/p2panda-auth/src/group/graph.rs
@@ -1,5 +1,5 @@
 use petgraph::graphmap::DiGraphMap;
-use petgraph::visit::{Dfs, DfsPostOrder, Reversed};
+use petgraph::visit::{Dfs, Reversed};
 use std::collections::HashSet;
 
 use crate::traits::OperationId;
@@ -53,7 +53,7 @@ where
 {
     // Get all successors.
     let mut successors = HashSet::new();
-    let mut dfs = DfsPostOrder::new(&graph, target);
+    let mut dfs = Dfs::new(&graph, target);
     while let Some(nx) = dfs.next(&graph) {
         successors.insert(nx);
     }
@@ -61,7 +61,7 @@ where
     // Get all predecessors.
     let mut predecessors = HashSet::new();
     let reversed = Reversed(graph);
-    let mut dfs_rev = DfsPostOrder::new(&reversed, target);
+    let mut dfs_rev = Dfs::new(&reversed, target);
     while let Some(nx) = dfs_rev.next(&reversed) {
         predecessors.insert(nx);
     }
--- a/p2panda-auth/src/group/graph.rs
+++ b/p2panda-auth/src/group/graph.rs
@@ -1,10 +1,13 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::collections::HashSet;
+
 use petgraph::graphmap::DiGraphMap;
 use petgraph::visit::{Dfs, Reversed};
-use std::collections::HashSet;
 
 use crate::traits::OperationId;
 
-pub fn concurrent_bubble<OP>(
+fn concurrent_bubble<OP>(
     graph: &DiGraphMap<OP, ()>,
     target: OP,
     processed: &mut HashSet<OP>,
--- a/p2panda-auth/src/group/graph.rs
+++ b/p2panda-auth/src/group/graph.rs
@@ -1,0 +1,200 @@
+use petgraph::graphmap::DiGraphMap;
+use petgraph::visit::{Dfs, DfsPostOrder, Reversed};
+use std::collections::HashSet;
+
+use crate::traits::OperationId;
+
+pub fn concurrent_bubble<OP>(
+    graph: &DiGraphMap<OP, ()>,
+    target: OP,
+    processed: &mut HashSet<OP>,
+) -> HashSet<OP>
+where
+    OP: OperationId + Ord,
+{
+    let mut bubble = HashSet::new();
+    bubble.insert(target);
+
+    concurrent_operations(graph, target)
+        .into_iter()
+        .for_each(|op| {
+            if processed.insert(op) {
+                bubble.extend(concurrent_bubble(graph, op, processed).iter())
+            }
+        });
+
+    bubble
+}
+
+/// Walk the graph and identify all sets of concurrent operations.
+pub fn concurrent_bubbles<OP>(graph: &DiGraphMap<OP, ()>) -> Vec<HashSet<OP>>
+where
+    OP: OperationId + Ord,
+{
+    let mut processed: HashSet<OP> = HashSet::new();
+    let mut bubbles = Vec::new();
+
+    graph.nodes().for_each(|target| {
+        if processed.insert(target) {
+            let bubble = concurrent_bubble(graph, target, &mut processed);
+            if bubble.len() > 1 {
+                bubbles.push(bubble)
+            }
+        }
+    });
+
+    bubbles
+}
+
+/// Return any operations concurrent with the given target operation.
+fn concurrent_operations<OP>(graph: &DiGraphMap<OP, ()>, target: OP) -> HashSet<OP>
+where
+    OP: OperationId + Ord,
+{
+    // Get all successors.
+    let mut successors = HashSet::new();
+    let mut dfs = DfsPostOrder::new(&graph, target);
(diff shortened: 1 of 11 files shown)


p2panda-auth/src/group/resolver.rs
```diff
@@ -30,70 +42,703 @@ where
...
+                    // as long as it's 1) not a predecessor of the remove operation, and 2) not a
+                    // mutual removal (removal of the remover by the removed member).
+                    if let GroupAction::Remove {
+                        member: removed_member,
+                    } = action
+                    {
+                        println!("processing removal of {:?}", removed_member);
+
+                        if concurrent_operation.sender() == removed_member.id()
+                            && !operation.previous().contains(concurrent_operation_id)
```
sandreae: I believe we need to do some graph traversal here to understand if an operation is a predecessor, rather than just looking in `previous`. If I'm understanding what should happen here correctly!

p2panda-auth/src/group/resolver.rs
```diff
@@ -30,70 +42,703 @@ where
...
-        // Get all current tip operations.
-        //
-        // TODO: should be checking against transitive heads here.
-        let heads = y.heads();
+        let is_concurrent = !get_concurrent_operations(&y.graph, operation.id()).is_empty();
 
-        // Detect concurrent operations by comparing the current heads with the new operations'
-        // dependencies.
-        let is_concurrent = heads != HashSet::from_iter(operation.dependencies().clone());
+        // @TODO: Think we need to check here if there is a concurrent operation which means
```
sandreae: As I say in the comment, I believe we also need to check here for the reverse case, where the concurrent remove/demote already exists, and we need to react to it because of concurrent operations arriving afterwards.

p2panda-auth/src/group/mod.rs
```diff
@@ -534,66 +546,66 @@ where
...
-                    &previous_operations,
-                    &action,
-                )?;
-            }
-            // No action required as revokes would have triggered a rebuild in the previous step.
-            //
+        //
+        // Do not update the state if the attempt to apply the action fails.
+        let mut y_i = match control_message {
+            // TODO: either here or inside apply_action we need to account for expected errors
```
sandreae: I believe this is the cause of one of the test failures, as once some actions get filtered out, then later ones may trigger an error (eg. applying a now "ignored" member's action)

p2panda-auth/src/group/graph.rs
```diff
@@ -0,0 +1,200 @@
+use petgraph::graphmap::DiGraphMap;
```
adzialocha: License headers missing

p2panda-auth/src/group/graph.rs
```diff
@@ -0,0 +1,200 @@
...
+use petgraph::visit::{Dfs, DfsPostOrder, Reversed};
+use std::collections::HashSet;
```
adzialocha, line 3: Nit: Import grouping

p2panda-auth/src/group/graph.rs
```diff
@@ -0,0 +1,200 @@
...
+
+use crate::traits::OperationId;
+
+pub fn concurrent_bubble<OP>(
```
adzialocha: Does this need to be public?

p2panda-auth/src/group/graph.rs
```diff
@@ -0,0 +1,200 @@
...
+        successors.insert(nx);
+    }
+
+    // Get all predecessors.
+    let mut predecessors = HashSet::new();
+    let reversed = Reversed(graph);
+    let mut dfs_rev = DfsPostOrder::new(&reversed, target);
+    while let Some(nx) = dfs_rev.next(&reversed) {
+        predecessors.insert(nx);
+    }
```
adzialocha: More of an understanding question:

Why ist post-order sorting required here? I can see that we're collecting all items into an unordered set, is there a special reason though? A comment would be nice maybe to understand the reasoning better.

Doesn't need changing if post-order isn't required and doesn't have any impact on performance etc.

p2panda-auth/src/group/resolver.rs
```diff
@@ -1,99 +1,966 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use std::collections::HashSet;
+//! Strong remove group resolver implementation.
+
```
adzialocha: Nit: no newline after mod-level doc string

p2panda-auth/src/group/resolver.rs
```diff
@@ -1,99 +1,966 @@
...
+///
+/// If a mutual removal has occurred, we want to retain the removal operations but filter all
+/// concurrent operations performed by the removed members (keeping predecessors of the
+/// remove).
+///
+/// 3) Re-adding member concurrently
+///
+/// If Alice removes Charlie and Bob removes then adds Charlie concurrently, Charlie is still
+/// in the group. However, if Charlie performed any concurrent actions, these will be filtered
+/// along with any dependent operations.
```
adzialocha, line 48: I believe this would weaken the "strong" removal guarantee as a removal can be made ineffectual. I know we're assuming a lot of trust anyhow but allowing this will make removals attackable (as long as the second, third, .. malicious peer wasn't detected and also removed).

I don't suggest changing this behavior, however we maybe should weaken the "strong removal" guarantee in the docs and mention the exception explicitly?

adzialocha: Another question:

IF we would not allow anyone but Alice to revoke their removal of Charlie (by re-adding them inside the same concurrency bubble), then I'd expect the removal to "never happen" so why remove all concurrent group changes of Charlie?

p2panda-auth/src/group/graph.rs
sandreae: Good observation, we don't require post order graph traversal in this case and there may be a performance benefit to using `Dfs` so I swapped it now :+1:

p2panda-auth/src/group/resolver.rs
sandreae: This is definitely the least intuitive of the behaviours we implemented towards a strong remove resolver strategy. I think it definitely deserves some more thought and discussion, maybe it can be improved from the current approach, in a way which we feel retains the strong remove behaviour users will expect. 

The current approach was chosen during implementation because it was quite elegant technically (due to how membership state merges are is resolved using grow only counters) and we didn't have a strong feeling that there was a more intuitive approach. It will be possible to take a different technical approach in order to realise different behaviour.

I need to think about this as an attack vector a little more, if we come to the conclusion that it is then definitely let's find a better solution.

In any case, I'd like to move this conversation into a dedicated issue if that's ok and make a decision there. Thanks for raising it!

sandreae: Issue here: https://github.com/p2panda/p2panda/issues/760

PR #758, opened 2025-06-26
Introduce high-level group membership API

mycognosist:
This PR introduces `GroupManager` which exposes the high-level API for creating and updating groups.

Groups provide a means for restricting access to application data and resources. They are comprised of members, which may be individuals or groups, and are assigned a user-chosen identity (`group_id`). Each member is assigned a unique user-chosen identifier and access level. Access levels are used to enforce restrictions over access to data and the mutation of that data. They are also used to grant permissions which allow for mutating the group state by adding, removing and modifying the access level of other members.

The `GroupManager` API is a thin wrapper around the `GroupState` and `Group` APIs, with the addition of validation checks to ensure that invalid control message operations are not created and appended to the group state DAG. For example, it should not be possible to remove a member who is not currently part of the group. This would result in a `GroupManagerError` being returned.

## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/group/dgm.rs
+++ b/p2panda-auth/src/group/dgm.rs
@@ -99,7 +99,11 @@ where
     ID: IdentityHandle + Display,
     OP: OperationId + Ord + Display,
     C: Clone + Debug + PartialEq + PartialOrd,
-    RS: Resolver<ORD::Message, State = GroupState<ID, OP, C, RS, ORD, GS>> + Debug,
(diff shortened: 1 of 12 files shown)


sandreae (approved):
Great work, this all looks fine to me. We can improve naming maybe, as this higher-level user facing struct `GroupManager` should feel like the first thing people should look at. But changes like that can come when looking at the whole of the -auth module in next steps.

Issue #760, opened 2025-07-02
Expected behaviour when re-adding a concurrently removed member


sandreae:
A conversation was opened during implementation of a Strong Remove strategy for group management [here](https://github.com/p2panda/p2panda/pull/753#discussion_r2171884774) regarding the current behaviour when a member (C) is removed by valid member A, but concurrently removed and then re-added by valid member B. In the current implementation this would result in the member C being in the group after all branches are merged but any concurrent changes they made will have been filtered out. As discussed in the linked conversation, we need to consider if this behaviour can be exploited as an attack vector (malicious members colluding to over-write removes). Additionally, the fact that the member remains in the group, but their concurrent changes are not applied is counter intuitive.

PR #765, opened 2025-07-04
Improve auth docs

mycognosist:
This PR increases documentation coverage throughout the `p2panda-auth` crate and renames several modules and types.


p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -900,6 +885,33 @@ where
...
+where
+    ID: IdentityHandle + Display,
+    OP: OperationId + Ord + Display,
+    C: Clone + Debug + PartialEq + PartialOrd,
+    RS: Resolver<ID, OP, C, ORD, GS> + Debug,
+    ORD: Orderer<ID, OP, GroupControlMessage<ID, OP, C>> + Debug,
+    ORD::Operation: Clone,
+    GS: GroupStore<ID, OP, C, RS, ORD> + Clone + Debug,
+{
+    /// Action was applied an no error occurred.
```
adzialocha: Typo "an"

p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -900,6 +885,33 @@ where
...
+    Ok {
+        state: GroupCrdtState<ID, OP, C, RS, ORD, GS>,
+    },
+    /// Action was not applied because it failed internal validation.
```
adzialocha, line 904: Nit: We usually have newlines between complex enum types

p2panda-auth/src/group/message.rs
```diff
@@ -2,39 +2,41 @@
...
-    /// conflicts (for example with "strong removal") strategy. By issuing a revoke message
-    /// revoking the message which first added a member into the group, it's possible to
-    /// completely erase that member from the group history. There can be an implicit "seniority"
-    /// rule in play, where it's only possible for an admin to revoke messages that they
-    /// published, or from when they were in the group.
+    /// Publishing a revocation message allows for invalidation of other group control messages
+    /// which are already included in the group graph. This action is agnostic to any, probably more
+    /// nuanced, resolving logic which reacts to group actions.
+    //
+    // TODO: revoking messages is not implemented yet. I'm still considering if it is required in
```
adzialocha, line 24: Maybe give a name to the TODO as it's written in "I" form

p2panda-auth/src/group/mod.rs
```diff
@@ -31,21 +23,23 @@ use thiserror::Error;
...
 use crate::traits::{
-    AuthGroup, GroupMembership, GroupMembershipQuery, GroupStore, IdentityHandle, Operation,
-    OperationId, Ordering, Resolver,
+    Group as GroupTrait, GroupMembership, GroupStore, IdentityHandle, Operation, OperationId,
+    Orderer, Resolver,
 };
 
 #[derive(Debug, Error)]
-pub enum GroupManagerError<ID, OP, C, RS, ORD, GS>
+// TODO: Rename to `GroupError`.
```
adzialocha: Remove TODO

p2panda-auth/src/group/mod.rs
```diff
@@ -65,30 +59,31 @@ where
...
-/// Each `GroupManager` method performs internal validation to ensure that the desired group
-/// action is valid in light of the current group state. Attempting to perform an invalid action
-/// results in a `GroupManagerError`. For example, attempting to remove a member who is not
-/// currently part of the group.
-pub struct GroupManager<ID, OP, C, RS, ORD, GS>
+/// Each `Group` method performs internal validation to ensure that the desired group action is
+/// valid in light of the current group state. Attempting to perform an invalid action results in a
+/// `GroupError`. For example, attempting to remove a member who is not currently part of the
+/// group.
+// TODO: Rename to `Group`.
```
adzialocha: Remove TODO

p2panda-auth/src/group/mod.rs
```diff
@@ -65,30 +59,31 @@ where
 
 /// Decentralised Group Management (DGM).
```
adzialocha, line 59: DGM in the DCGKA paper stands for "Decentralised Group Membership". Not sure if we want to be consistent here, up to you!

p2panda-auth/src/group/resolver.rs
```diff
@@ -42,13 +44,13 @@ use crate::traits::{GroupStore, IdentityHandle, Operation, OperationId, Ordering
 /// concurrent operations performed by the removed members (keeping predecessors of the
 /// remove).
 ///
-/// 3) Re-adding member concurrently
+/// ### Re-adding member concurrently
 ///
 /// If Alice removes Charlie and Bob removes then adds Charlie concurrently, Charlie is still
 /// in the group. However, if Charlie performed any concurrent actions, these will be filtered
 /// along with any dependent operations.
```
adzialocha, line 51: What happens if Charlie removes and re-adds themselves?

p2panda-auth/src/group/resolver.rs
```diff
@@ -25,13 +25,15 @@ use crate::traits::{GroupStore, IdentityHandle, Operation, OperationId, Ordering
 /// operations are processed. It can be assumed that the behavior is equivalent for an admin
 /// member being removed, or demoted from admin to a lower access level.
 ///
-/// 1) Removals
+/// ## Strong Remove Concurrency Rules
```
adzialocha, line 28: Struggling a bit to call it "strong" with the rules around re-adding (see https://github.com/p2panda/p2panda/issues/760), but that maybe gets resolved in the future

p2panda-auth/src/lib.rs
```diff
@@ -1,5 +1,74 @@
...
+//! 1) Removal or demotion of a manager causes any concurrent actions by that member to be
+//!    invalidated
+//! 2) Mutual removals, where two managers remove or demote one another concurrently, are not
+//!    invalidated; both removals are applied to the group state but any other concurrent actions
+//!    by those members are invalidated
+//! 3) Re-adds are allowed; if Alice removes Charlie then re-adds them, they are still a member of
+//!    the group but all of their concurrent actions are invalidated
+//! 4) Invalidation of transitive operations; invalidation of an operation due to the application
+//!    of the aforementioned rules results in all dependent operations being invalidated
+
```
adzialocha: Nit: Remove newline

p2panda-auth/src/lib.rs
```diff
@@ -1,5 +1,74 @@
...
+//! has an associated access level which can be used to determine their permissions. The access
+//! levels are `Pull`, `Read`, `Write` and `Manage`. Each access level is a superset of the lower
+//! levels and can be assigned an associated set of conditions; this allows fine-grained
+//! partitioning of each access level. For example, `Read` conditions could be assigned with a path
+//! to restrict access to areas of a dataset. Finally, only members with `Manage` access are allowed
+//! to modify the group state by adding, removing, promoting or demoting other members.
+//!
+//! The access levels defined by `p2panda-auth` may prove useful when controlling replication of
+//! datasets. Custom sync protocols can be defined which rely on group membership and access levels
+//! to determine who to sync with and over which subsets of data.
```
adzialocha: Could be nice to mention also how it helps building application-layer stuff (for example by establishing moderation rules around conditions and write access)

p2panda-auth/src/group/message.rs
sandreae: I've removed revocation message in a different branch meaning this TODO is gone, so won't fix it here.

p2panda-auth/src/group/resolver.rs
sandreae: Charlie's remove of themselves would be applied, and then as they are no longer a member, they wouldn't be able to re-add themselves.

p2panda-auth/src/group/resolver.rs
sandreae: If feel feel strongly we should rename this to just "Concurrency Rules" and not refer to "Strong" I wouldn't mind doing that. 

My position right now is that we made an attempt to arrive at "Strong Remove" logic, but there is a question around how we handle re-adds which needs resolving. We should find a solution we're happy with which we feel provides the strong remove property we're after. For that reason (that we will make changes until we're happy) I'm also ok with leaving the description as it is (keeping the "Strong").

p2panda-auth/src/group/mod.rs
sandreae: I'll change that here for consistency  as we definitely got that acronym from DGM in DCGKA

sandreae: Or maybe up to you @mycognosist ?

p2panda-auth/src/group/resolver.rs
adzialocha: My worry is that "strong removal" might indicate a false security around "full guarantee around removals in a byzantine setting" while it doesn't fully provide that. I understand that almost all other guarantees are given, but for a more "activist context" it's still not enough.

adzialocha: I don't think we need to guarantee that on this level though, I'm fine with the current scenario. Just that in other contexts it might be understood differently?

sandreae: I still need to consider the current approach (re-adds by valid members respected) as an attack vector, not saying it isn't, but I haven't fully understood it yet.

adzialocha: > I still need to consider the current approach (re-adds by valid members respected) as an attack vector, not saying it isn't, but I haven't fully understood it yet.

Yes, no rush here! I think the issue consists of two questions: 1. Attacks 2. Intuition / expectations around what it means to "remove" someone from a group (especially when removals are always "strong")

sandreae: Cool, I suggest we keep the name then and explore the open questions to assure we fulfil the expectations of that name.

adzialocha: Okay, that sounds alright to me

Issue #779, opened 2025-08-08
Detect and handle mutual remove cycles in strong remove resolver


sandreae:
Mutual remove cycles should be detected and handled according to strong remove rules (all members removed).

Example 1: `A remove B -> B remove C -> C remove A`
Example 2: `A remove B -> B add C -> C remove A`

PR #781, opened 2025-08-12
`p2panda-auth`: one global state object and operation graph for all groups 

sandreae:
The motivation for this PR is issue #782  related to resolving nested groups during state re-build. The exact issue came from the fact that during group graph traversal, when we build concurrent bubbles, we weren't of any concurrent operations which may have occurred in sub-groups. This is due to those operations not being directly attached to the same graph, but rather being retrieved and from the store. We could request the state at an exact position (from the perspective of say an Add operation), but this doesn't account for any concurrent changes which may have occurred. The solution is to include all operations for the root group and sub-groups in the same graph, preserving their causal relationships, and when we identify concurrency bubbles, we can do so looking at all related groups. When we go in this direction though, it makes much more sense to just have one auth groups state, which contains the entire global graph of operations for all groups. Doing this actually simplifies state handling and reduces potentially a lot of duplication when managing state for many groups. Although the core changes to `GroupCrdt` and `GroupCrdtState` were fairly straight forward, updating all the unit tests, fuzz tests, test tooling, displaying graphs and high-level api resulted in A LOT of lines of code changes :bow: . I recommend looking mostly at `group::crdt::mod` and `group::resolver` to get an idea of what has changed, everything else is updating due to these changes and refactoring.

## Highlights
- group store removed :partying_face: 
- only one state object handled for all groups
- all auth control messages go into one graph
- `previous` field removed from operation (we only need one field denoting the last global state now)
- reduced generic trait parameter hell in some places
- test utilities somewhat simplified due to reduced complexity around group state handling
- drop support for adding a sub-group with manager access (until #779) is resolved and adapted to handle sub-groups
- notable changes to the higher-level dgm trait interface

## What do the graphs look like now?

graphviz(7)

closes: #782 

## 📋 Checklist

### DIFF
--- a/fuzz/fuzz_targets/auth_group.rs
+++ b/fuzz/fuzz_targets/auth_group.rs
@@ -226,8 +226,7 @@ impl Member {
         }
 
         if member.is_group() {
-            // Group create a sub-group for themselves incase they should be added as a
-            // sub-group.
+            // Members which are acting as a group create a group for themselves.
             member.create_group(
                 member.id(),
                 vec![(GroupMember::Individual(member.my_id), Access::manage())],
--- a/p2panda-auth/src/group/crdt/mod.rs
+++ b/p2panda-auth/src/group/crdt/mod.rs
@@ -424,7 +424,7 @@ where
 
         groups_y = match result {
             StateChangeResult::Ok { state } => state,
-            StateChangeResult::Noop { error, .. } => {
+            StateChangeResult::Error { error, .. } => {
                 // Noop shouldn't happen when processing new operations as the
                 // rebuild logic should have occurred instead.
                 return Err(GroupCrdtError::StateChangeError(operation_id, error));
@@ -506,7 +506,7 @@ where
 
         match result {
             StateChangeResult::Ok { state } => state,
-            StateChangeResult::Noop { error, .. } => {
+            StateChangeResult::Error { error, .. } => {
                 // Noop shouldn't happen when processing new operations as the
                 // rebuild logic should have occurred instead.
                 return Err(GroupCrdtError::StateChangeError(operation.id(), error));
@@ -598,7 +598,7 @@ where
             // change, however we do want to accept them into our graph so as to ensure
             // consistency consistency across peers.
             groups_y.insert(group_id, members_y);
-            StateChangeResult::Noop {
+            StateChangeResult::Error {
                 state: groups_y,
                 error: err,
             }
@@ -616,7 +616,7 @@ where
     Ok { state: GroupStates<ID, C> },
 
     /// Action was not applied because it failed internal validation.
-    Noop {
+    Error {
         state: GroupStates<ID, C>,
         #[allow(unused)]
         error: GroupMembershipError<GroupMember<ID>>,
@@ -634,7 +634,7 @@ where
     pub fn state(&self) -> &GroupStates<ID, C> {
         match self {
             StateChangeResult::Ok { state }
-            | StateChangeResult::Noop { state, .. }
+            | StateChangeResult::Error { state, .. }
             | StateChangeResult::Filtered { state } => state,
         }
     }
--- a/p2panda-auth/src/group/mod.rs
+++ b/p2panda-auth/src/group/mod.rs
@@ -13,7 +13,10 @@ pub mod resolver;
 pub use action::GroupAction;
 pub(crate) use crdt::apply_action;
 pub use crdt::state::{GroupMembersState, GroupMembershipError, MemberState};
-pub use crdt::{GroupCrdt, GroupCrdtError, GroupCrdtInnerState, GroupCrdtState, StateChangeResult};
+pub use crdt::{
+    GroupCrdt, GroupCrdtError, GroupCrdtInnerError, GroupCrdtInnerState, GroupCrdtState,
+    StateChangeResult,
+};
 pub use member::GroupMember;
 pub use message::GroupControlMessage;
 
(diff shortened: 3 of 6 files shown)


fuzz/fuzz_targets/auth_group.rs
```diff
@@ -218,30 +223,17 @@ impl Member {
...
-        // All members create a sub-group for themselves incase they should be added as a
-        // sub-group.
-        member.create_group(
-            member.id(),
-            vec![(GroupMember::Individual(member.my_id), Access::manage())],
-            operations,
-        );
+        if member.is_group() {
+            // Group create a sub-group for themselves incase they should be added as a
+            // sub-group.
```
adzialocha: nit: This sentence misses something

p2panda-auth/src/group/mod.rs
```diff
@@ -54,77 +55,78 @@ where
...
-
-    /// Store for all known groups.
-    store: GS,
-
-    /// Message orderer state.
-    orderer: ORD::State,
-
-    _phantom: PhantomData<(ID, OP, C, RS, ORD, GS)>,
+    y: Option<GroupCrdtState<ID, OP, C, ORD>>,
+    _phantom: PhantomData<(ID, OP, C, RS, ORD)>,
```
adzialocha: Only `RS` should be necessary here

p2panda-auth/src/group/mod.rs
```diff
@@ -133,103 +135,81 @@ where
...
+            group_id,
             action: GroupAction::Create { initial_members },
         };
 
-        let (y, operation) = GroupCrdt::prepare(y, &action)?;
-        let y = GroupCrdt::process(y, &operation)?;
+        let y = self.y.take().expect("state object present");
+        let (y_i, operation) = GroupCrdt::prepare(y, &action)?;
+        let y_ii = GroupCrdt::process(y_i, &operation)?;
+        let _ = self.y.insert(y_ii);
```
adzialocha, line 160: Redundant `let` assignment?

p2panda-auth/src/group/mod.rs
```diff
@@ -133,103 +135,81 @@ where
...
-    fn receive_from_remote(
-        y: Self::State,
-        remote_operation: ORD::Operation,
-    ) -> Result<Self::State, Self::Error> {
+    fn receive_from_remote(&mut self, remote_operation: ORD::Operation) -> Result<(), Self::Error> {
         // Validation is performed internally by `process()`.
-        let y = GroupCrdt::process(y, &remote_operation)?;
+        let y = self.y.take().expect("state object present");
+        let y_i = GroupCrdt::process(y, &remote_operation)?;
+        let _ = self.y.insert(y_i);
```
adzialocha, line 173: Same here

p2panda-auth/src/group/mod.rs
```diff
@@ -133,103 +135,81 @@ where
...
                 access,
             },
         };
 
-        let (y, operation) = GroupCrdt::prepare(y, &action)?;
-        let y = GroupCrdt::process(y, &operation)?;
+        let y = self.y.take().expect("state object present");
+        let (y_i, operation) = GroupCrdt::prepare(y, &action)?;
+        let y_ii = GroupCrdt::process(y_i, &operation)?;
+        let _ = self.y.insert(y_ii);
```
adzialocha, line 213: And here

p2panda-auth/src/group/mod.rs
```diff
@@ -241,34 +221,37 @@ where
...
                 member: GroupMember::Individual(removed),
             },
         };
 
-        let (y, operation) = GroupCrdt::prepare(y, &action)?;
-        let y = GroupCrdt::process(y, &operation)?;
+        let y = self.y.take().expect("state object present");
+        let (y_i, operation) = GroupCrdt::prepare(y, &action)?;
+        let y_ii = GroupCrdt::process(y_i, &operation)?;
+        let _ = self.y.insert(y_ii);
```
adzialocha, line 255: Aaaand here

p2panda-auth/src/group/resolver.rs
```diff
@@ -1,22 +1,28 @@
...
+
+/// Error types for GroupCrdt.
+#[derive(Debug, Error)]
+pub enum StrongRemoveError<OP>
+where
+    OP: OperationId,
+{
+    #[error(transparent)]
+    Groups(#[from] GroupCrdtInnerError<OP>),
+}
```
adzialocha: If there's no other error types we can probably remove this wrapper

p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -147,2062 +113,1399 @@ where
...
-    },
+    Ok { state: GroupStates<ID, C> },
 
     /// Action was not applied because it failed internal validation.
     Noop {
-        state: GroupCrdtState<ID, OP, C, RS, ORD, GS>,
+        state: GroupStates<ID, C>,
         #[allow(unused)]
         error: GroupMembershipError<GroupMember<ID>>,
     },
```
adzialocha: "Noop" doesn't seem to fit the meaning of this variant anymore as it actually describes a "failure" case now. Maybe calling it "Err" (to reflect the "Ok") or similar is better?

p2panda-auth/src/group/mod.rs
sandreae: You mean line 157? I get clippy warnings without the let assignment....

PR #784, opened 2025-08-14
`p2panda-spaces`: introduce `AuthStore` trait

sandreae:
There is now only one auth state used for all groups (and therefore spaces). This should be globally accessible on the manager, so that any spaces can retrieve it, and set it again once any changes have been applied. For this we can use a new "auth store" interface (trait) with methods for getting and setting the auth state. This is implemented on the generic S store and as Manager is accessible from within a Space we can also access and set the auth state when required.

## 📋 Checklist


adzialocha (approved):
Looking good and the improvements are really nice!

I noticed that a few TODOs have been removed even though they were not related (orderer state, code repetition etc.), do you track them somewhere else now? Otherwise I'd bring them back.

PR #785, opened 2025-08-16
`p2panda-spaces`: introduce `space.add` method

sandreae:
- [x] add dependencies graph to both auth and encryption orderer state and attach the tips as dependencies when next_message is called
- [x] update dependencies graphs when new operations are forged or processed
- [x] pre-compute next space membership state by processing a fake auth control operation  
- [x] introduce `space.add` method
  - [x] exclude members with only "pull" access from the encryption group 
- [x] tests

## 📋 Checklist

### DIFF
--- a/p2panda-spaces/src/space.rs
+++ b/p2panda-spaces/src/space.rs
@@ -136,7 +136,7 @@ where
 
         // 7. Persist new state.
 
-        Self::persist_space(manager_ref.clone(), y, auth_y).await?;
+        Self::set_state(manager_ref.clone(), y, auth_y).await?;
 
         Ok((
             Self {
@@ -204,7 +204,7 @@ where
 
         // 6. Persist new state.
 
-        Self::persist_space(self.manager.clone(), y, auth_y).await?;
+        Self::set_state(self.manager.clone(), y, auth_y).await?;
 
         Ok(message)
     }
@@ -270,7 +270,7 @@ where
 
         // 4. Persist new state.
 
-        Self::persist_space(self.manager.clone(), y, auth_y).await?;
+        Self::set_state(self.manager.clone(), y, auth_y).await?;
 
         Ok(vec![])
     }
@@ -431,28 +431,6 @@ where
         (encryption_y, auth_y)
     }
 
-    /// Persist both auth and space state.
-    async fn persist_space(
-        manager_ref: Manager<S, F, M, C, RS>,
-        space: SpaceState<M>,
-        auth: AuthGroupState<C>,
-    ) -> Result<(), SpaceError<S, F, M, C, RS>> {
-        let mut manager = manager_ref.inner.write().await;
-        manager
-            .store
-            .set_auth(&auth)
-            .await
-            .map_err(SpaceError::AuthStore)?;
-        let space_id = space.space_id;
-        manager
-            .store
-            .set_space(&space_id, space)
-            .await
-            .map_err(SpaceError::SpaceStore)?;
-
-        Ok(())
-    }
-
     /// Forge a space message from an auth and optional encryption message. This produces a signed
     /// message which can be hashed to compute the final operation id.
     async fn forge(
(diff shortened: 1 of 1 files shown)


p2panda-spaces/src/message.rs
```diff
@@ -139,20 +170,29 @@ where
...
         let control_message = match (auth_action, encryption_control_message) {
             (
                 AuthGroupAction::Create { initial_members },
                 EncryptionControlMessage::Create { .. },
             ) => ControlMessage::Create { initial_members },
+            (AuthGroupAction::Add { member, access }, EncryptionControlMessage::Add { .. })
+                if !access.is_pull() =>
+            {
+                ControlMessage::Add { member, access }
+            }
```
adzialocha: We would panic here if we're adding a member with only "pull" access, right? Wouldn't the correct behaviour be to _always_ return the `Add` control message? On the level of space messages we don't distinct between encryption and auth.

p2panda-spaces/src/tests.rs
```diff
@@ -262,3 +279,196 @@ async fn send_and_receive() {
...
+//     );
+//
+//     // Orderer states have been updated.
+//     let manager_ref = bob_manager.inner.read().await;
+//     let y = manager_ref.store.space(&space_id).await.unwrap().unwrap();
+//     assert_eq!(vec![message_02.id()], y.encryption_y.orderer.heads);
+//
+//     let auth_y = manager_ref.store.auth().await.unwrap();
+//     assert_eq!(vec![message_02.id()], auth_y.orderer_y.heads);
+// }
```
adzialocha: Could also be removed for this PR as it's just dead code (it's in the history if you need it for the future)

p2panda-spaces/src/space.rs
```diff
@@ -195,6 +192,132 @@ where
...
+    ) -> Result<M, SpaceError<S, F, M, C, RS>> {
+        let mut y = {
+            let manager = self.manager.inner.read().await;
+            manager
+                .store
+                .space(&self.id)
+                .await
+                .map_err(SpaceError::SpaceStore)?
+                .expect("space exists")
+        };
```
adzialocha: You can also use the `space.state` method here

p2panda-spaces/src/space.rs
```diff
@@ -195,6 +192,132 @@ where
...
+                }
+                GroupMember::Group(_) => {
+                    // @TODO: in order to support adding a group to a space we need to add _all_
+                    // new transitive members which have read access. We already compute the new
+                    // to-be-added secret members above and have it in _secret_members_diff.
+                    unimplemented!();
+                }
+            }
+        };
+        y.encryption_y = encryption_y;
```
adzialocha: This should not be executed if the added member has only "pull" access

p2panda-spaces/src/space.rs
```diff
@@ -195,6 +192,132 @@ where
...
+            };
+
+            SpacesArgs::from_args(self.id, Some(auth_args), Some(encryption_args))
+        };
+
+        let mut manager = self.manager.inner.write().await;
+
+        // @TODO: Can't use ephemeral private key for signing "create" message as this will make
+        // the author / sender different from the person we want to do a key agreement with when
+        // processing it in `p2panda-encryption`.
```
adzialocha: This TODO doesn't seem to fit here

p2panda-spaces/src/space.rs
```diff
@@ -374,6 +490,89 @@ where
...
+        // us.
+        let auth_y_i = AuthGroup::process(auth_y, &fake_op).map_err(SpaceError::AuthGroup)?;
+        let next_members = secret_members(auth_y_i.members(space_id));
+        let diff_members = next_members
+            .iter()
+            .cloned()
+            .filter(|actor| previous_members.contains(actor))
+            .collect::<Vec<_>>();
+        Ok((next_members, diff_members))
+    }
```
adzialocha: Cool that this was so easy to do in the end. For performance we could also later do an even simpler approach where we take the current transitive members and add / remove / promote / demote them manually. Since this is a local operation we don't need to check for concurrency.

p2panda-spaces/src/space.rs
```diff
@@ -374,6 +490,89 @@ where
...
+                encryption_y.add_dependency(message.id(), encryption_dependencies);
+            }
+            // @TODO: also include application messages in auth and encryption dependencies.
+            SpacesArgs::Application { .. } => (),
+        };
+        (encryption_y, auth_y)
+    }
+
+    // Persist both auth and space state.
+    async fn persist_space(
```
adzialocha: How about calling this `set_state`? Then it's nicely fitting the `state` method.

adzialocha: It could also move closer to `state` for readability

p2panda-spaces/src/space.rs
adzialocha: It wouldn't panic, but I think that's okay

p2panda-spaces/src/space.rs
sandreae: I don't think there's a performance consideration here, as if an operation isn't concurrent (it's state points at the current tip) then no heavy processing kicks in (the rebuild logic). I think using the existing API here is nice, it keeps the validation consistent and in one (already well considered) place and adding a nested group is handled already (we'd have to query each added/removed group and resolve transitive members manually).

adzialocha: Ah great! This sounds good indeed!

adzialocha (approved):

PR #788, opened 2025-08-25
`p2panda-auth`: improved strong remove resolver

sandreae:
## 📋 Checklist

### DIFF
--- a/fuzz/fuzz_targets/groups_message.rs
+++ b/fuzz/fuzz_targets/groups_message.rs
@@ -201,17 +201,19 @@ impl StateMachine {
         let mut options = Vec::new();
 
         if try_options.contains(&Options::Add)
-            && let Some(added) = self.values.random_member(rng) {
-                options.push(Operation::Add {
-                    added,
-                    members_in_welcome: self.values.active_members.clone(),
-                });
-            }
+            && let Some(added) = self.values.random_member(rng)
+        {
+            options.push(Operation::Add {
+                added,
+                members_in_welcome: self.values.active_members.clone(),
+            });
+        }
 
         if try_options.contains(&Options::Remove)
-            && let Some(removed) = self.values.random_active_member(rng) {
-                options.push(Operation::Remove { removed });
-            }
+            && let Some(removed) = self.values.random_active_member(rng)
+        {
+            options.push(Operation::Remove { removed });
+        }
 
         if try_options.contains(&Options::Update) {
             options.push(Operation::Update);
--- a/fuzz/fuzz_targets/groups_message.rs
+++ b/fuzz/fuzz_targets/groups_message.rs
@@ -200,20 +200,18 @@ impl StateMachine {
     fn suggest_valid(&self, try_options: &[Options], rng: &Rng) -> Operation {
         let mut options = Vec::new();
 
-        if try_options.contains(&Options::Add) {
-            if let Some(added) = self.values.random_member(rng) {
+        if try_options.contains(&Options::Add)
+            && let Some(added) = self.values.random_member(rng) {
                 options.push(Operation::Add {
                     added,
                     members_in_welcome: self.values.active_members.clone(),
                 });
             }
-        }
 
-        if try_options.contains(&Options::Remove) {
-            if let Some(removed) = self.values.random_active_member(rng) {
+        if try_options.contains(&Options::Remove)
+            && let Some(removed) = self.values.random_active_member(rng) {
                 options.push(Operation::Remove { removed });
             }
-        }
 
         if try_options.contains(&Options::Update) {
             options.push(Operation::Update);
--- a/p2panda-auth/src/graph.rs
+++ b/p2panda-auth/src/graph.rs
@@ -4,6 +4,7 @@
 
 use std::collections::HashSet;
 
+use petgraph::algo::has_path_connecting;
 use petgraph::graphmap::DiGraphMap;
 use petgraph::visit::{Dfs, Reversed};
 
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Highlights are marked with a pancake 🥞
 ### Changed
 
 - One global state object and operation graph for all groups [#781](https://github.com/p2panda/p2panda/pull/781)
+- Improvements to strong remove resolver [#788](https://github.com/p2panda/p2panda/pull/788)
```
adzialocha: The CHANGELOG entry (and PR title) can be made more meaningful. You already have a good template in your PR description:

> Handle several new edge cases in strong remove resolver, improve concurrent re-add logic and handle cycles in nested groups.

How about: "Handle strong removal edge cases and cycles in nested groups, improve concurrent re-adds"

fuzz/fuzz_targets/groups_message.rs
```diff
@@ -200,19 +200,19 @@ impl StateMachine {
...
         }
 
-        if try_options.contains(&Options::Remove) {
-            if let Some(removed) = self.values.random_active_member(rng) {
-                options.push(Operation::Remove { removed });
-            }
+        if try_options.contains(&Options::Remove)
+            && let Some(removed) = self.values.random_active_member(rng)
+        {
+            options.push(Operation::Remove { removed });
```
adzialocha: Oh, interesting. What is this change in the `p2panda-encryption` fuzz test doing?

sandreae: Ah, this is a clippy suggested change, I can revert it from this PR.

p2panda-auth/src/graph.rs
```diff
@@ -81,20 +82,54 @@ where
...
+}
+
+/// Return `true` if no linear path exists between `a` and `b` in either direction.
+///
+/// This indicates whether or not the given operations occurred concurrently.
+pub fn is_concurrent<OP>(graph: &DiGraphMap<OP, ()>, a: OP, b: OP) -> bool
+where
+    OP: OperationId + Ord,
+{
+    !has_path(graph, a, b) && !has_path(graph, b, a) && a != b
```
mycognosist: A minor performance-related comment:

The `a != b` check could be made into the first check (ie. moved to the front of the three assertions) as it is the least computationally-expensive.

I can't intuit the performance difference in practice, and the `a == b` case is probably very rare anyway, but I thought I'd mention it since the graph passed into `is_concurrent()` could be very large.

p2panda-auth/src/group/authority_graphs.rs
```diff
@@ -0,0 +1,489 @@
...
+    pub fn add_removal(&mut self, group_id: ID, remover: ID, removed: ID, op: OP) {
+        if remover == removed {
+            return;
+        }
+        self.removals
+            .entry(group_id)
+            .or_default()
+            .push((remover, removed, op));
+
+        // Smash the cache.
```
mycognosist, line 109: "Smash the cache." is an epic code comment 🔥

p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -16,6 +16,13 @@ use crate::group::{
 };
 use crate::traits::{Conditions, IdentityHandle, Operation, OperationId, Orderer, Resolver};
 
+/// Max depth of group nesting allowed.
+///
+/// Depth is checked during group state queries and if the depth is exceeded further additions are
+/// ignored. The main reason for this check is to protect against accidental group nesting cycles
+/// which may occur as a result of concurrent operations.
+const MAX_NESTED_DEPTH: u32 = 1000;
```
mycognosist, line 24: Super cool to see a limit introduced on nesting!

I wonder if this number is too high? Do you have thoughts to share on how you arrived at 1000? I don't have a sense of what an appropriate limit would be...mostly curious to hear your reasoning.

p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -165,7 +180,14 @@ where
         group_id: ID,
         members: &mut HashMap<ID, Access<C>>,
         root_access: Option<Access<C>>,
+        mut depth: u32,
     ) {
+        // If we reached max nesting depth exit from the traversal.
+        if depth == MAX_NESTED_DEPTH {
```
mycognosist, line 186: Is there any point at which we communicate "upstream" that we've hit this limit? Or would you imagine an application author needing to ensure that they didn't allow nesting beyond this point? Currently the code returns silently.

p2panda-auth/src/graph.rs
sandreae: Ahh, thanks, nice spot. I'll make this change.

p2panda-auth/src/group/crdt/mod.rs
sandreae: Yeh, I agree it seems like a high limit, but my reason for this is that nested groups could be used to model a strongly delegation based access control system (group managers unilaterally passing on access further and further), and in this setting the delegation chains could get arbitrarily long (depends on the use case). So although 1000 seems like an unnecessarily high limit for modelling concrete user groups, in the delegation case I have less idea what an upper limit should be. Considering that it is mainly in place to help detect nested cycles then I thought I high number was the safer bet.

p2panda-auth/src/group/crdt/mod.rs
sandreae: This I'm not sure how to handle, I could look into introducing an error type, but intuitively I feel that might introduce errors to the API in a way that makes it a bit ugly. I thought that communicating the limit and behaviour might be enough, again considering that it's mostly a fail-safe for group cycles.

p2panda-auth/src/group/crdt/mod.rs
mycognosist: Ah right, I hadn't thought of the delegation chain application. Cool! Thanks for sharing 🙏🏻

p2panda-auth/src/group/crdt/mod.rs
mycognosist: Makes sense to not return an error if the check is mainly in place as a fail safe. Riffing on your "communicate the limit and behaviour" approach, I wonder if there's any way to signal the current depth to the caller (without adversely affecting the API). I'm not sure how difficult that would be but I figured I'd throw the idea out there. If the caller knows the max depth and the current depth (plus the overall behaviour) then they're well-equipped to act accordingly.

sandreae: I've made this into an error as I'd like to consider it more after merging this PR https://github.com/p2panda/p2panda/issues/799

Thanks for raising this issue!

PR #790, opened 2025-09-01
WIP: Implement supervised gossip actor sub-system

mycognosist:
This PR introduces a new crate: `p2panda-net-next`, which is intended to serve as a clean-slate for refactoring `p2panda-net` into a fault-tolerant system using supervised actors.

The refactored gossip sub-system forms the focus of this PR. It's split into the following modules / actors:

 - Gossip : main interface with userland (manage channels, spawn sessions); also processes overlay events 
   - GossipSession : super a gossip session over a single topic
     - GossipListener : receive messages from userland and forward them to gossip sender
     - GossipSender : broadcast messages into the topic overlay
     - GossipReceiver : receive messages from the topic overlay and forward them to the session

## 📋 Checklist

### DIFF
--- a/p2panda-net-next/Cargo.toml
+++ b/p2panda-net-next/Cargo.toml
@@ -5,11 +5,15 @@ edition = "2024"
 
 [dependencies]
 anyhow = "1.0.97"
+futures-lite = "2.6.1"
 iroh = { version = "0.34.1", default-features = false }
 iroh-base = "0.34.1"
 iroh-gossip = "0.34.1"
 p2panda-core = { path = "../p2panda-core", version = "0.4.0" }
+ractor = { version = "0.15.8", features = ["cluster"] }
 serde = { version = "1.0.219", features = ["derive"] }
+tokio = { version = "1.44.2", features = ["rt", "sync", "time"] }
+tracing = "0.1.41"
 
 [lints]
 workspace = true
--- a/p2panda-net-next/Cargo.toml
+++ b/p2panda-net-next/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "p2panda-net-next"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+
+[lints]
+workspace = true
--- a/p2panda-net-next/Cargo.toml
+++ b/p2panda-net-next/Cargo.toml
@@ -4,6 +4,12 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
+anyhow = "1.0.97"
+iroh = { version = "0.34.1", default-features = false }
+iroh-base = "0.34.1"
+iroh-gossip = "0.34.1"
+p2panda-core = { path = "../p2panda-core", version = "0.4.0" }
+serde = { version = "1.0.219", features = ["derive"] }
 
 [lints]
 workspace = true
--- a/p2panda-net-next/src/actors/gossip/mod.rs
+++ b/p2panda-net-next/src/actors/gossip/mod.rs
@@ -1,0 +1,201 @@
+//! An iroh-specific gossip actor for message broadcast.
+
+mod listener;
+mod receiver;
+mod sender;
+mod session;
+
+use std::collections::HashMap;
+
+use iroh::Endpoint as IrohEndpoint;
+use iroh_gossip::net::Gossip as IrohGossip;
+use iroh_gossip::proto::DeliveryScope as IrohDeliveryScope;
+use p2panda_core::PublicKey;
+use ractor::{Actor, ActorId, ActorProcessingErr, ActorRef, Message, RpcReplyPort};
+use session::ToGossipSession;
+use tokio::sync::mpsc::{self, Receiver, Sender};
+
+use crate::actors::gossip::session::GossipSession;
+use crate::network::{FromNetwork, ToNetwork};
+use crate::{from_public_key, TopicId};
+
+// NOTE: The `GossipSession` actor does not know the topic id for it's session. Instead, the
+// `Gossip` actor holds a mapping of `topic_id` -> `actor_id` for each gossip subscription
+// session.
+
+pub enum ToGossip {
+    /// Return a handle to the iroh gossip actor.
+    ///
+    /// This is required when registering the gossip ALPN with the router.
+    Handle(RpcReplyPort<IrohGossip>),
+
+    /// Join the given topic, using the given peers as gossip bootstrap nodes.
+    Join {
+        topic_id: TopicId,
+        peers: Vec<PublicKey>,
+        reply: RpcReplyPort<(Sender<ToNetwork>, Receiver<FromNetwork>)>,
+    },
+
+    /// Joined a topic by connecting to the given peers.
+    Joined {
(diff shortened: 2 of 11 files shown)


p2panda-net-next/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+[dependencies]
+anyhow = "1.0.97"
+futures-lite = "2.6.1"
+iroh = { version = "0.91.2", default-features = false }
+iroh-base = "0.91.2"
+iroh-gossip = "0.91.0"
+p2panda-core = { path = "../p2panda-core", version = "0.4.0" }
+ractor = { version = "0.15.8", features = ["cluster"] }
+serde = { version = "1.0.219", features = ["derive"] }
+tokio = { version = "1.47.1", features = ["rt", "sync", "time"] }
```
adzialocha: Maybe have a pass on the actually used features / dependencies. I can't see any use of `time`-related code from tokio for example, except in tests.

p2panda-net-next/src/actors/gossip/receiver.rs
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Gossip recevier actor which holds the gossip topic receiver, receives overlay messages and sends
+//! them to the gossip session actor.
+
+use futures_lite::StreamExt;
+use iroh_gossip::api::GossipReceiver as IrohGossipReceiver;
```
adzialocha, line 6: It would be nice if we can already abstract away any concrete iroh types wherever possible, so we only use actual iroh code in one place and have all other gossip-related-interfaces agnostic to the concrete PlumTree/HyParView implementation.

This is a "nice to have", but if you can have a quick check if that's easy to do already that would help with any future work and separate the APIs nicely.

adzialocha: .. that being said: if some of the actors exist _because_ iroh's design requires them to exist, then of course it's fine if they have iroh-types

adzialocha: I can see how `GossipReceiver` is a stream over standard events, this looks straight-forward to abstract over. Maybe the `joined` can be realized by observing the first "neighbor up" event ourselves

p2panda-net-next/src/actors/gossip/joiner.rs
```diff
@@ -0,0 +1,73 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Join a set of peers on a gossip topic.
+
```
adzialocha: Style consistence: No new-lines after mod-level doc-strings.

p2panda-net-next/src/actors/gossip/mod.rs
```diff
@@ -0,0 +1,987 @@
...
+    use tokio::sync::mpsc::error::TryRecvError;
+    use tokio::time::sleep;
+
+    use crate::network::{FromNetwork, ToNetwork};
+    use crate::{from_private_key, from_public_key};
+
+    use super::{Gossip, ToGossip};
+
+    #[tokio::test]
+    async fn two_peer_gossip() {
```
adzialocha, line 588: Great tests! Would be nice one day if we could just test the gossip behaviour without any concrete endpoint we need to bootstrap for proper isolation of test concerns.

p2panda-net-next/src/actors/gossip/mod.rs
```diff
@@ -0,0 +1,987 @@
...
+                    let _ = reply.send(gossip);
+                }
+
+                Ok(())
+            }
+            ToGossip::Subscribe(topic_id, peers, reply) => {
+                // TODO: How do we handle a subscribe for a topic that already has an active
+                // subscription? Either we multiplex over a subscription or we establish an
+                // additional session. Right now I (glyph) am leaning towards establishing
+                // additional sessions.
```
adzialocha, line 178: On that layer I think it's fine to have it like that (create a new session). On a higher level we probably want to multiplex.

We can document that behaviour and just have it like this :+1:

p2panda-net-next/src/actors/gossip/mod.rs
```diff
@@ -0,0 +1,987 @@
...
+
+pub struct GossipState {
+    gossip: IrohGossip,
+    sessions_by_actor_id: HashMap<ActorId, TopicId>,
+    sessions_by_topic_id: HashMap<TopicId, Vec<ActorRef<ToGossipSession>>>,
+    neighours_by_topic_id: HashMap<TopicId, HashSet<PublicKey>>,
+    from_gossip_senders: HashMap<TopicId, Vec<Sender<FromNetwork>>>,
+    gossip_joined_senders: HashMap<ActorId, OneshotSender<u8>>,
+    topic_delivery_scopes: HashMap<TopicId, Vec<IrohDeliveryScope>>,
+}
```
adzialocha, line 88: Maybe good to have some assertions in the tests just to make sure these maps really contain what we expect after state transitions

p2panda-net-next/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// Unique 32 byte identifier for a network.
+///
+/// Peers operating on the same network identifier will eventually discover each other. This is the
+/// most global identifier to group peers into networks. Different applications may choose to share
+/// the same underlying network infrastructure by using the same network identifier.
+///
+/// Please note that the network identifier should _never_ be the same as any other topic
+/// identifier.
+pub type NetworkId = [u8; 32];
```
adzialocha: Maybe we keep that out for now as this is not used by the PR, I believe we need to revisit now with the refactoring what the role of a "network id" is in general.

p2panda-net-next/src/network.rs
```diff
@@ -0,0 +1,24 @@
...
+    GossipMessage {
+        bytes: Vec<u8>,
+        delivered_from: PublicKey,
+    },
+    SyncMessage {
+        header: Vec<u8>,
+        payload: Option<Vec<u8>>,
+        delivered_from: PublicKey,
+    },
+}
```
adzialocha: We will separate gossip ("ephemeral subscription") from sync ("subscription") much more strictly, so these types should probably not be part of the same enum.

I'd be fine with removing this part altogether and look at it in future PRs.

p2panda-net-next/src/lib.rs
mycognosist: Good call. I've removed it.

p2panda-net-next/src/network.rs
mycognosist: I think that's a good idea. Have removed this completely for now (just dealing with raw `Vec<u8>` instead).

p2panda-net-next/src/actors/gossip/mod.rs
mycognosist: I've added a `TestSupervisor` actor which can be used to "catch" the outcome of some actor being tested (it either fails or terminates successfully). Terminated actors return their state, which then allows us to perform these kinds of assertion (e.g. did the gossip actor correctly store the session according to topic id, did it store the correct neighbours by topic id etc.).

PR #792, opened 2025-09-04
`p2panda-spaces`: process encryption messages after welcome

sandreae:
We want to queue up all group encryption control and application messages until we have actually been welcomed into the group. This can happen in a "Create" or an "Add" message. In the case of the latter, many other control messages may have arrived before we actually join the group and should process them.

To do this we implement the set_welcome method on the encryption orderer and only return next messages when the DGM flagged us as "welcomed".

Additionally a check is added which returns an error if `space.publish` is called before we are welcomed to a group.

## 📋 Checklist

### DIFF
--- a/p2panda-spaces/src/space.rs
+++ b/p2panda-spaces/src/space.rs
@@ -272,7 +272,7 @@ where
 
         Self::set_state(self.manager.clone(), y, auth_y).await?;
 
-        Ok(to_events(self.id(), encryption_output))
+        Ok(encryption_output_to_events(self.id(), encryption_output))
     }
 
     /// Process a group membership change on the group encryption state.
@@ -351,7 +351,7 @@ where
             .await
             .map_err(SpaceError::SpaceStore)?;
 
-        Ok(to_events(self.id(), encryption_output))
+        Ok(encryption_output_to_events(self.id(), encryption_output))
     }
 
     /// Process an auth control message before the "authored" version has been forged. This is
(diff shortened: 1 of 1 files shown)


p2panda-spaces/src/encryption/orderer.rs
```diff
@@ -33,11 +35,25 @@ impl<M> EncryptionOrderer<M> {
...
+    /// Queue of operations we have not yet processed.
+    ///
+    /// This should only grow in the case where we are not yet welcomed into the group by an "Add"
+    /// message.
+    queue: VecDeque<OperationId>,
+
+    // @TODO: We keep all messages in memory currently which is bad. We need a persistence
+    // layer where we can fetch messages from.
+    /// In-memory store of all messages.
+    messages: HashMap<OperationId, EncryptionMessage>,
```
adzialocha, line 53: Can we already use the `OperationStore` here? They will already exist in the database and if we have the `queue` with the IDs around we can get the real objects from there when it is time.

sandreae: Nice idea, I expect that we can indeed do that. I'll give it go tomorrow.

sandreae: I think this is the right way to go, but I'd like to take it later in a dedicated PR. Thanks for mentioning it here though!

p2panda-spaces/src/space.rs
```diff
@@ -621,6 +603,24 @@ pub fn added_members(current_members: Vec<ActorId>, next_members: Vec<ActorId>)
         .collect::<Vec<_>>()
 }
 
+fn to_events<M>(space_id: ActorId, encryption_output: Vec<EncryptionGroupOutput<M>>) -> Vec<Event> {
```
adzialocha: I'd change the method name to something more explicit, what is converted "to events" here?

PR #796, opened 2025-09-16
`p2panda-spaces`: global shared auth state and basic groups API

sandreae:
There is one global auth state which is shared by all spaces. This allows for groups to be re-used across multiple spaces. Keeping spaces in sync with the global auth state poses several challenges considering the auth state itself will be mutated concurrently by multiple actors.

Another property we need to uphold is per-space atomicity, meaning that any encryption message must be processed along with it's accompanying auth message, or not at all.

These two requirements can be satisfied if whenever an auth group membership change occurs, an auth message is forged and published towards a global auth state _and_ a space message (containing the encryption messages) is forged and published to all known spaces referencing this message.

With this approach all spaces are effected when the global state changes, and per-space atomicity is retained as encryption state changes will always occur along with their related auth state change.

There are special concurrency cases which need to be handled (auth state changes needing to be retroactively applied to concurrently created spaces), these will be dealt with in further PRs.

image

image
_EDIT: space id is generic now, not the operation hash as stated in the above design._

### Tasks
- [x] share auth state across all spaces
   - [x] update message types
   - [x] update spaces API logic
   - [x] introduce `MessageStore`
   - [x] sync local auth changes with all spaces (we know of)
- [x] implement basic groups API
  - [x] create
  - [x] add
  - [x] remove
- [x] generic space id
- [x] tests

### Next steps
- keep spaces in sync with _concurrent_ shared auth state changes too 

## 📋 Checklist

### DIFF
--- a/p2panda-spaces/src/encryption/message.rs
+++ b/p2panda-spaces/src/encryption/message.rs
@@ -1,14 +1,18 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use p2panda_auth::traits::Conditions;
+use p2panda_auth::traits::{Conditions, Operation};
 use p2panda_encryption::crypto::xchacha20::XAeadNonce;
 use p2panda_encryption::data_scheme::GroupSecretId;
 use p2panda_encryption::traits::{GroupMessage as EncryptionOperation, GroupMessageContent};
 
+use crate::auth::message::AuthMessage;
 use crate::encryption::dgm::EncryptionGroupMembership;
 use crate::message::{AuthoredMessage, SpacesArgs, SpacesMessage};
+use crate::space::removed_members;
 use crate::traits::SpaceId;
-use crate::types::{ActorId, EncryptionControlMessage, EncryptionDirectMessage, OperationId};
+use crate::types::{
+    ActorId, AuthGroupAction, EncryptionControlMessage, EncryptionDirectMessage, OperationId,
+};
 
 #[derive(Clone, Debug)]
 pub enum EncryptionArgs {
@@ -37,39 +41,7 @@ pub enum EncryptionMessage {
 }
 
 impl EncryptionMessage {
-    pub(crate) fn from_membership<ID, M, C>(space_message: &M) -> Vec<Self>
-    where
-        ID: SpaceId,
-        M: AuthoredMessage + SpacesMessage<ID, C>,
-        C: Conditions,
-    {
-        let SpacesArgs::SpaceMembership {
-            space_dependencies,
-            control_messages,
-            ..
-        } = space_message.args()
-        else {
-            panic!("unexpected message type")
-        };
-
-        control_messages
-            .clone()
-            .into_iter()
-            .map(|message| {
-                let encryption_args = EncryptionArgs::System {
-                    dependencies: space_dependencies.clone(),
-                    control_message: message.encryption_control_message(),
-                    direct_messages: message.direct_messages().to_owned(),
-                };
-                EncryptionMessage::Forged {
-                    author: space_message.author(),
-                    operation_id: space_message.id(),
-                    args: encryption_args,
-                }
-            })
-            .collect()
-    }
-
+    /// Construct an encryption message from a space application message.
     pub(crate) fn from_application<ID, M, C>(space_message: &M) -> Self
     where
         ID: SpaceId,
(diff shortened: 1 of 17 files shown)


p2panda-spaces/src/message.rs
```diff
@@ -27,269 +18,70 @@ pub trait AuthoredMessage: Debug {
...
 
-pub trait SpacesMessage<C> {
-    fn args(&self) -> &SpacesArgs<C>;
+pub trait SpacesMessage<ID, C> {
+    fn args(&self) -> &SpacesArgs<ID, C>;
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub enum SpacesArgs<C> {
+pub enum SpacesArgs<ID, C> {
```
adzialocha, line 26: Mini readabiliy-nitpick: Maybe add spaces between each variant, together with the doc-strings it gets very dense (it's an impressive enum! 😅)

p2panda-spaces/src/space.rs
```diff
@@ -5,327 +5,461 @@ use std::convert::Infallible;
...
-                    // every new member. For this we need to attach an array of encryption messages to
-                    // the space control message.
-                    EncryptionGroup::remove(encryption_y, *removed.first().unwrap(), &manager.rng)
-                        .map_err(SpaceError::EncryptionGroup)
+                    for removed in all_removed {
+                        let (encryption_y_inner, message) =
+                            EncryptionGroup::remove(encryption_y, removed, rng)
+                                .map_err(SpaceError::EncryptionGroup)?;
+                        encryption_y = encryption_y_inner;
+                        direct_messages.extend(message.direct_messages());
```
adzialocha, line 423: I believe it's only necessary to extend the array with the direct messages vec from the _last_ iteration.

adzialocha: It's not a problem to have them all, just redundant data / bloated payload size.

Hope to address this upstream with this issue: #806

p2panda-spaces/src/message.rs
```diff
@@ -27,269 +18,70 @@ pub trait AuthoredMessage: Debug {
...
+        ///
+        /// This is a dependency and should be considered when ordering space messages.
+        auth_message_id: OperationId,
 
-        /// Encrypted, direct messages to members in the group, used for key agreement.
+        /// All direct messages that a local peer generated when processing the referenced auth
+        /// message on this space.
         direct_messages: Vec<EncryptionDirectMessage>,
     },
+    SpaceUpdate {
```
adzialocha, line 67: What is a space update message and how is it different to a "membership" one? (Maybe some doc-strings could help)

p2panda-spaces/src/group.rs
```diff
@@ -1,3 +1,289 @@
...
+    #[error("{0}")]
+    Forge(F::Error),
+
+    #[error("{0}")]
+    AuthStore(<S as AuthStore<C>>::Error),
+
+    #[error("{0}")]
+    MessageStore(<S as MessageStore<M>>::Error),
+
+    // @TODO: We lose the concrete error type which caused sync of spaces to fail, ideal we would
```
sandreae, line 285: Need a better solution here, if we coerce directly from a `ManagerError` then we end up with a cyclical error type and stack overflow...

p2panda-spaces/src/encryption/message.rs
```diff
@@ -36,46 +41,134 @@ pub enum EncryptionMessage {
...
+        } = space_message.args()
+        else {
+            panic!("unexpected message type");
+        };
+
+        // Sanity check.
+        assert_eq!(auth_message.id(), *auth_message_id);
+
+        // Check if there are any direct messages for me.
+        let my_direct_message = direct_messages
```
adzialocha: Maybe rename this to has_my_direct_message or similar, as this reads as if its type is a direct message and not a bool

p2panda-spaces/src/encryption/message.rs
```diff
@@ -36,46 +41,134 @@ pub enum EncryptionMessage {
...
+    /// arguments.
+    ///
+    /// This method is required when we receive a space message and associated auth message and we
+    /// want to adjust our local encryption state accordingly. The main requirement is that we
+    /// process our own direct messages (contained in the space message), in many cases the actual
+    /// encryption control message type and content is redundant as the DGM state is always
+    /// manually replaced with the latest membership state provided by p2panda-auth. The only case
+    /// where it does matter is if we ourselves were added or removed from the group, here we
+    /// should make sure that the control message contains our own actor id.
+    pub(crate) fn from_membership<ID, M, C>(
```
adzialocha, line 86: Logic looking good :+1:

p2panda-spaces/src/store.rs
```diff
@@ -6,23 +6,31 @@ use p2panda_auth::traits::Conditions;
...
         &self,
-        id: &ActorId,
-    ) -> impl Future<Output = Result<Option<SpaceState<M>>, Self::Error>>;
+        id: &ID,
+    ) -> impl Future<Output = Result<Option<SpaceState<ID, M, C>>, Self::Error>>;
 
-    fn has_space(&self, id: &ActorId) -> impl Future<Output = Result<bool, Self::Error>>;
+    fn has_space(&self, id: &ID) -> impl Future<Output = Result<bool, Self::Error>>;
+
+    fn spaces(&self) -> impl Future<Output = Result<Vec<ID>, Self::Error>>;
```
adzialocha: Should this be called `spaces_ids` etc. instead as it only is a vec of `ID`?

p2panda-spaces/src/space.rs
```diff
@@ -5,327 +5,461 @@ use std::convert::Infallible;
...
-            space_id,
-            y.encryption_y,
-            &auth_message,
-        )
-        .await?;
-        y.encryption_y = encryption_y;
+    /// Forge a space message from an already existing auth message and apply any resulting group
+    /// membership changes. Any resulting encryption direct messages are included in the space
+    /// message alongside a reference to the auth message.
+    pub(crate) async fn process_auth_message(
```
adzialocha, line 149: Not a biggie: That's maybe my memory from the older code-base but I get a little confused that the naming is around "auth message" for the central control message type? Shouldn't this be a "spaces message" as it contains both arguments for auth _and_ encryption?

p2panda-spaces/src/manager.rs
```diff
@@ -180,54 +221,130 @@ where
...
+                    .process(message, None)
+                    .await
+                    .map_err(ManagerError::Space)?
             }
         };
 
         Ok(events)
     }
+
+    /// Sync all spaces with a shared auth state change.
```
adzialocha: Could be nice to read more about in the doc string what is meant with "sync" here (maybe also confusing word in peer-to-peer contexts) and why it is necessary to keep states "in sync".

p2panda-spaces/src/space.rs
adzialocha: Or does "auth" come from "authored" (the `AuthoredMessage` trait)? That's maybe equally confusing to read 😆

p2panda-spaces/src/group.rs
adzialocha: Not 100% sure, but it feels like you're missing some sort of new layer which handles the glue between spaces and groups, currently this logic is "hidden" in multiple places, in the manager, group and space, but maybe you need something new?

Space handles group changes on "add", "remove" and "create" and Group handles space changes to keep them "in sync" (using a manager method, even though it's only groups who ever call it, but maybe that changes in the future).

```
Space -> Group
Group -> Manager -> Space
```

So rather than spaces doing things with groups and vice versa you might need this extra layer sitting right _after_ manager which keeps these two worlds in sync. It probably requires some untangling of code which I can't tell how complex that might be.

```
Manager -> SyncThingy -> Space / Group
```

Sync-Thingy is maybe the wrong word, it's probably more something sitting around where every local change (create, add, remove, ...) and remote ones (process) are made? Something which "orchestrates" that groups get changed, spaces get changed, and that they are kept in sync.

sandreae: Yeh, I totally agree that the current separation of concerns between the apis isn't great, and as you say, results in some confusing flows. 

Just so I understand your suggestion (I like the idea of adding a dedicated "orchestrator" layer), we would still require method calls on a group to trigger "space sync" and method calls on a space to effect the relevant group, but a new "orchestrator" api/layer on manager would help keep all this logic together in one place and make the code more readable/comprehensible. Is that roughly what you're thinking?

p2panda-spaces/src/space.rs
sandreae: Yeh, naming here could be better, thanks for the spot. What is being processed is a "spaces auth message" `SpacesArgs::Auth` and what comes out is a "spaces space membership message" `SpacesArgs::SpaceMembership`. So maybe `process_spaces_auth_message(..)` would be appropriate?

p2panda-spaces/src/group.rs
```diff
@@ -1,3 +1,289 @@
...
+    ActorId, AuthControlMessage, AuthGroup, AuthGroupAction, AuthGroupError, AuthGroupState,
+    AuthResolver, EncryptionGroupError,
+};
+use crate::utils::{typed_member, typed_members};
+
+#[derive(Debug)]
+pub struct Group<ID, S, F, M, C, RS> {
+    /// Reference to the manager.
+    ///
+    /// This allows us build an API where users can treat "group" instances independently from the
```
mycognosist:
```suggestion
    /// This allows us to build an API where users can treat "group" instances independently from the
```

p2panda-spaces/src/tests.rs
```diff
@@ -16,22 +18,31 @@ use crate::auth::orderer::AuthOrderer;
...
 use crate::test_utils::MemoryStore;
-use crate::types::{ActorId, AuthGroupState, OperationId, StrongRemoveResolver};
+use crate::traits::SpaceId;
+use crate::types::{
+    ActorId, AuthControlMessage, AuthGroupAction, AuthGroupState, OperationId, StrongRemoveResolver,
+};
 
 type SeqNum = u64;
 
+// Implement SpaceId for i32 which is what we use space identifiers in the tests.
```
mycognosist:
```suggestion
// Implement SpaceId for i32 which is what we use as space identifiers in the tests.
```

mycognosist (approved):
🟢 ✨

p2panda-spaces/src/group.rs
adzialocha: > Is that roughly what you're thinking?

Yes! This sounds like what I was thinking about as well. It's hard for me to see _exactly_ right now what logic is required, but I'm happy to walk through it with you if you want.

Of course, this is overkill for this PR, I think you're current "error workaround" is fine and an orchestrator will just be an "cosmetic refactor".

p2panda-spaces/src/space.rs
adzialocha: I can see it also contains encryption / key agreement information (direct messages), so "spaces auth messages" are the most high-level message type and "auth" is understood here as "I authorized you to use that space" (and _not_ related to only `p2panda-auth`), right?

p2panda-spaces/src/group.rs
sandreae: Yeh, cool, I think I'll let it sit for a bit, and also see what additional logic comes in when I implement the logic for handling concurrent group changes which require a space to be "repaired" (this is where the throttling is required). Maybe a nice refactoring path becomes clear then.

adzialocha: Oh great! Yes, that sounds perfect.

p2panda-spaces/src/space.rs
sandreae: It may be as you say, but I grok it slightly differently. "spaces auth messages" `SpacesArgs::Auth` are messages which get applied directly to the shared auth state, these only get applied to a space when a `SpacesArgs::SpaceMembership` message arrives (referencing the earlier auth message), _this_ message contains the encryption direct messages.

So the method `process_auth_message` takes a `SpacesArgs::Auth` message, applies it to a space, and returns the resulting `SpacesArgs::SpaceMembership` message.

PR #802, opened 2025-09-22
`p2panda-spaces`: events for group and space membership changes 

sandreae:
This PR introduces events which are emitted when group and/or space membership changes occur due to processing remote messages.

The two event variants for "groups" and "spaces" have slightly different semantics. Group events contain member changes where the members can be a group or an individual and all members have a corresponding access levels. Space events contain only changes which occurred to the "secret" member set (members with >= "read" access) and access levels are not given. 

Both event variants are enriched with additional state information which may be useful for consumers. Group events contain both `members` (the direct group members, can be groups or individuals) and `transitive_members` (the resolved member set, only contains individuals). Space events contain only `members` which is the set of individuals in the space encryption context (resolved members with >= "read" access). 

```rust

#[derive(Clone, Debug, PartialEq, Eq)]
#[allow(clippy::large_enum_variant)]
pub enum Event<ID, C> {
    Application { space_id: ID, data: Vec<u8> },
    Group(GroupEvent<C>),
    Space(SpaceEvent<ID>),
}

#[derive(Clone, Debug, PartialEq, Eq)]
pub struct GroupContext<C> {
    /// The actor who authored this action.
    pub author: ActorId,

    /// Root group actors, can be individuals or groups.
    pub group_actors: Vec<(GroupActor, Access<C>)>,

    /// Members of this group.
    pub members: Vec<(ActorId, Access<C>)>,
}

#[derive(Clone, Debug, PartialEq, Eq)]
pub struct SpaceContext {
    /// The actor who authored this group change action.
    pub auth_author: ActorId,

    /// The actor who applied this action to the spaces state.
    ///
    /// Note: this can be different to the auth_actor in cases where concurrent auth changes which
    /// effect a space are applied later.
    pub spaces_author: ActorId,

    /// Id of the group associated with this space.
    pub group_id: ActorId,

    /// Members in the spaces' encryption context.
    pub members: Vec<ActorId>,
}

/// Events emitted on auth group membership change.
#[derive(Clone, Debug, PartialEq, Eq)]
pub enum GroupEvent<C> {
    /// A group was created.
    Created {
        /// Group id.
        group_id: ActorId,

        /// Initial group members.
        initial_members: Vec<(GroupActor, Access<C>)>,

        /// Additional event context and group state after any change occurred.
        context: GroupContext<C>,
    },

    /// A member was added to a group.
    Added {
        /// Group id.
        group_id: ActorId,

        /// Group actor that was added, can be individual or group.
        added: GroupActor,

        /// Access level assigned to the added members.
        access: Access<C>,

        /// Additional event context and group state after any change occurred.
        context: GroupContext<C>,
    },

    /// A member was removed from a group.
    Removed {
        /// Group id.
        group_id: ActorId,

        /// Group actor that was removed, can be individual or group.
        removed: GroupActor,

        /// Additional event context and group state after any change occurred.
        context: GroupContext<C>,
    },
}

/// Events emitted on space encryption group membership change.
#[derive(Clone, Debug, PartialEq, Eq)]
pub enum SpaceEvent<ID> {
    /// A space was created.
    Created {
        /// Space id.
        space_id: ID,

        /// Initial members in the encryption context.
        initial_members: Vec<ActorId>,

        /// Additional event context and space state after any change occurred.
        context: SpaceContext,
    },

    /// One or many individuals were added to the space.
    Added {
        /// Space id.
        space_id: ID,

        /// Members added to the encryption context.
        added: Vec<ActorId>,

        /// Additional event context and space state after any change occurred.
        context: SpaceContext,
    },

    /// One or many individuals were removed from the space.
    Removed {
        /// Space id.
        space_id: ID,

        /// Members removed from the encryption context.
        removed: Vec<ActorId>,

        /// Additional event context and space state after any change occurred.
        context: SpaceContext,
    },

    /// Local actor was removed from the space.
    Ejected {
        /// Space id.
        space_id: ID,
    },
}
```

## Next steps
- ~~emit all events from a  channel `rx` which can be subscribed to on manager~~
- also emit events for state changes _we_ triggered by calling create, add, remove etc...

## 📋 Checklist

### DIFF
--- a/p2panda-spaces/src/event.rs
+++ b/p2panda-spaces/src/event.rs
@@ -3,11 +3,12 @@
 use p2panda_auth::Access;
 use p2panda_auth::group::GroupMember;
 use p2panda_auth::traits::Conditions;
-use p2panda_encryption::data_scheme::{ControlMessage, GroupOutput};
+use p2panda_encryption::data_scheme::GroupOutput;
 
 use crate::ActorId;
 use crate::auth::message::AuthMessage;
-use crate::encryption::message::{EncryptionArgs, EncryptionMessage};
+use crate::message::{AuthoredMessage, SpacesArgs, SpacesMessage};
+use crate::space::{added_members, removed_members};
 use crate::traits::SpaceId;
 use crate::types::{AuthGroupAction, AuthGroupState, EncryptionGroupOutput};
 
@@ -46,6 +47,7 @@ impl Member {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[allow(clippy::large_enum_variant)]
 pub enum Event<ID, C> {
     Application { space_id: ID, data: Vec<u8> },
     Group(GroupEvent<C>),
@@ -113,7 +115,7 @@ pub enum SpaceEvent<ID> {
         /// Individual in the spaces' encryption context.
         members: Vec<ActorId>,
     },
-    /// An individual was added to the space.
+    /// One or many individuals were added to the space.
     Added {
         /// Space id.
         space_id: ID,
@@ -121,13 +123,13 @@ pub enum SpaceEvent<ID> {
         /// Id of the group associated with this space.
         group_id: ActorId,
 
-        /// Individual added to the encryption context.
-        added: ActorId,
+        /// Individuals added to the encryption context.
+        added: Vec<ActorId>,
 
-        /// Individual in the spaces' encryption context.
+        /// Individuals in the spaces' encryption context.
         members: Vec<ActorId>,
     },
-    /// An individual was removed from the space.
+    /// One or many individuals were removed from the space.
     Removed {
         /// Space id.
         space_id: ID,
@@ -135,10 +137,10 @@ pub enum SpaceEvent<ID> {
         /// Id of the group associated with this space.
         group_id: ActorId,
 
-        /// Individual removed from the encryption context.
-        removed: ActorId,
+        /// Individuals removed from the encryption context.
+        removed: Vec<ActorId>,
 
-        /// Individual in the spaces' encryption context.
+        /// Individuals in the spaces' encryption context.
         members: Vec<ActorId>,
     },
     /// Local actor was removed from the space.
@@ -165,68 +167,13 @@ where
             }),
             GroupOutput::Control(_control_message) => {
                 // @TODO: when do control messages get emitted in group output?
-                todo!()
+                unimplemented!()
             }
             GroupOutput::Removed => Some(Event::Space(SpaceEvent::Ejected {
                 space_id: *space_id,
             })),
         })
         .collect()
-}
-
-pub(crate) fn encryption_message_to_space_event<ID, C>(
-    space_id: &ID,
-    group_id: ActorId,
-    current_members: &mut Vec<ActorId>,
-    encryption_message: &EncryptionMessage,
-) -> Option<Event<ID, C>>
-where
-    ID: SpaceId,
-    C: Conditions,
-{
-    let args = match encryption_message {
-        EncryptionMessage::Args(auth_args) => auth_args,
-        EncryptionMessage::Forged { args, .. } => args,
-    };
-
-    if let EncryptionArgs::System {
-        control_message, ..
-    } = args
-    {
-        let event = match control_message {
-            ControlMessage::Create { initial_members } => SpaceEvent::Created {
-                space_id: *space_id,
-                group_id,
-                members: initial_members.to_owned(),
-            },
-            ControlMessage::Remove { removed } => SpaceEvent::Removed {
-                space_id: *space_id,
-                group_id,
-                removed: *removed,
-                members: {
-                    let (idx, _) = current_members
-                        .iter()
-                        .enumerate()
-                        .find(|(_, member)| **member == *removed)
-                        .expect("member exists");
-                    current_members.remove(idx);
-                    current_members.clone()
-                },
-            },
-            ControlMessage::Add { added } => SpaceEvent::Added {
-                space_id: *space_id,
-                group_id,
-                added: *added,
-                members: {
-                    current_members.push(*added);
-                    current_members.clone()
-                },
-            },
-            ControlMessage::Update => unimplemented!(),
-        };
-        return Some(Event::Space(event));
-    }
-    None
 }
 
 pub(crate) fn auth_message_to_group_event<ID, C>(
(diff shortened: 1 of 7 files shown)


p2panda-spaces/src/event.rs
```diff
@@ -1,6 +1,282 @@
...
+
+use crate::ActorId;
+use crate::auth::message::AuthMessage;
+use crate::message::{AuthoredMessage, SpacesArgs, SpacesMessage};
+use crate::space::{added_members, removed_members};
+use crate::traits::SpaceId;
+use crate::types::{AuthGroupAction, AuthGroupState, EncryptionGroupOutput};
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Member {
```
adzialocha: We already have a `Member` struct in our API, it could be confusing to have a second one. Either merge them or rename one?

adzialocha: Also here it looks like a "member" can also be a group, while so far we only have "members" who are owners of key material (member -> group -> space).

I still like that naming (and I think it avoids confusion for the API users as nested "members" is much harder to grasp).

p2panda-spaces/src/event.rs
```diff
@@ -1,6 +1,282 @@
...
+    /// A group was created.
+    Created {
+        /// Group id.
+        group_id: ActorId,
+
+        /// Root members, can be individuals or groups.
+        members: Vec<(Member, Access<C>)>,
+
+        /// Transitive members, can only be individuals.
+        transitive_members: Vec<(Member, Access<C>)>,
```
adzialocha: Think we should stick to the `Member` type for the "leaf actors" and not have the same for the general "members" method (which needs a different name).

Maybe members -> group_actors (to make clear that this is a special thing) and transitive_members -> members?

p2panda-spaces/src/event.rs
```diff
@@ -1,6 +1,282 @@
...
+        added: Member,
+
+        /// Access level assigned to the added members.
+        access: Access<C>,
+
+        /// Root members, can be individuals or groups.
+        members: Vec<(Member, Access<C>)>,
+
+        /// Transitive members, can only be individuals.
+        transitive_members: Vec<(Member, Access<C>)>,
```
adzialocha: The "enriched" meta data / context is cool but a bit confusing maybe? Are all of these members added / removed / etc.?

Maybe we prefix the field name or put it in a struct to seperate the event from meta data?

p2panda-spaces/src/event.rs
```diff
@@ -1,6 +1,282 @@
...
+    Created {
+        /// Space id.
+        space_id: ID,
+
+        /// Id of the group associated with this space.
+        group_id: ActorId,
+
+        /// Individual in the spaces' encryption context.
+        members: Vec<ActorId>,
+    },
```
adzialocha, line 143: Nit: Add newlines between complex enums to improve readability

p2panda-spaces/src/event.rs
```diff
@@ -1,6 +1,282 @@
...
+        group_id: ActorId,
+
+        /// Individuals added to the encryption context.
+        added: Vec<ActorId>,
+
+        /// Individuals in the spaces' encryption context.
+        members: Vec<ActorId>,
+    },
+    /// One or many individuals were removed from the space.
+    Removed {
```
adzialocha, line 158: Do we want to mention who was the "remover" and "adder"? Can imagine that this is common UI feature to have

p2panda-spaces/src/event.rs
```diff
@@ -1,6 +1,282 @@
...
+        .into_iter()
+        .filter_map(|event| match event {
+            EncryptionGroupOutput::Application { plaintext } => Some(Event::Application {
+                space_id: *space_id,
+                data: plaintext,
+            }),
+            GroupOutput::Control(_control_message) => {
+                // @TODO: when do control messages get emitted in group output?
+                unimplemented!()
+            }
```
adzialocha, line 193: Good question! Probably only "updates" are coming from here?

p2panda-spaces/src/event.rs
```diff
@@ -1,6 +1,282 @@
...
+        /// Transitive members, can only be individuals.
+        transitive_members: Vec<(Member, Access<C>)>,
+    },
+}
+
+/// Events emitted on space encryption group membership change.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum SpaceEvent<ID> {
+    /// A space was created.
+    Created {
```
adzialocha, line 134: Should we have a flag for added / created if we are affected?

p2panda-spaces/src/event.rs
```diff
@@ -1,6 +1,282 @@
...
+                space_id: *space_id,
+                data: plaintext,
+            }),
+            GroupOutput::Control(_control_message) => {
+                // @TODO: when do control messages get emitted in group output?
+                unimplemented!()
+            }
+            GroupOutput::Removed => Some(Event::Space(SpaceEvent::Ejected {
+                space_id: *space_id,
+            })),
```
adzialocha, line 199: Is there a reason why we use encryption here for DGM stuff? Can we get that info from auth instead?

p2panda-spaces/src/event.rs
sandreae: Yeh, I struggled with this naming and wasn't happy with where it ended up really.

I think `GroupActor` which can be a group or an individual member would be fine.

adzialocha: I also like group actors! It's clear then that it is related to the group API!

p2panda-spaces/src/event.rs
sandreae: I'll add an `author` field to the new context struct :+1:

p2panda-spaces/src/event.rs
sandreae: I won't add this for now, but we can add it later if it's required.

p2panda-spaces/src/event.rs
sandreae: I thought it was maybe good to hook into the encryption output for this message, partly to ensure it is occurring as we expect. We could derive this information from the auth state instead though. Unless there is a strong argument not to I'd leave it as is for now.

adzialocha: Cool, that's fine. I stumbled upon this as it was surprising to see we're _not_ using auth for DGM-related code (it's using it internally of course, but still), so maybe it's good to add a note here for future developers?

adzialocha (approved):

PR #808, opened 2025-09-24
`p2panda-spaces`: key management

sandreae:
This PR introduces an `IdentityManager` struct which exposes an API for managing key bundles (register, retrieve, rotate) and also forging messages signed by the local actors private key. A `Credentials` struct which unifies the two keys (identity secret and private key) that make up an actors identity is introduced. This helps to express that the two keys are connected and should not be rotated independently in our APIs, validation is also introduced to ensure this is the case. A `Config` struct is introduced which configures parameters around key bundle rotation and the local peers credentials. Another change worth noting is that the `KeyStore` and `Forge` traits are now bound to the same generic parameter `K`, this again expresses better that these keys represent a single identity and allows for the IdentityManager to require only this store.

Some reorganising of modules has also occurred, eg. all traits being collected into the same place.

### Done

* [x] Support key bundle "member" system messages
* [x] Make expiry and "renewal window" configurable 
* [x] Method to get own key bundle should automatically rotate when its about to expire 
* [x] Method to learn if we're about to expire 
* [x] Unify management of author and encryption identity keys (make it harder/impossible for users to rotate either independently)

## 📋 Checklist


p2panda-encryption/src/crypto/rng.rs
```diff
@@ -44,6 +35,12 @@ impl Rng {
             .map_err(|_| RngError::NotEnoughRandomness)?;
         Ok(out)
     }
+
+    pub fn from_seed(seed: [u8; 32]) -> Self {
```
sandreae, line 44: I made this public so that I could derive a new `Rng` from an existing one which is required in the IdentityManager which doesn't have a handle to the top level manager. If there's another way to achieve this I'm very open, or if it's not a problem this method being public then we can leave it like this.

p2panda-encryption/src/crypto/x25519.rs
```diff
@@ -23,12 +23,6 @@ pub const SHARED_SECRET_SIZE: usize = 32;
 pub struct SecretKey(Secret<SECRET_KEY_SIZE>);
 
 impl SecretKey {
-    #[cfg(not(feature = "test_utils"))]
```
sandreae, line 26: I made this method public so that a new `SecretKey` could be generated in a `Credentials` constructor. Not sure the reasons it wasn't public before though. If it should indeed be private we can simply pass in a `SecretKey` to the `Credentials` constructor.

p2panda-spaces/src/identity.rs
```diff
@@ -0,0 +1,437 @@
...
+
+    /// Process a key bundle received from the network.
+    pub async fn process_key_bundle(
+        &mut self,
+        author: ActorId,
+        key_bundle: &LongTermKeyBundle,
+    ) -> Result<(), IdentityError<ID, K, M, C>> {
+        key_bundle.verify()?;
+
+        // @TODO: validate that the key bundle was indeed created by the correct author.
```
sandreae: @adzialocha this is a requirement we've discussed but I can't see a way to perform the validation. I could be understanding the requirement wrong though. I believe we want to ensure that the peer `ActorId` who generated the bundle is the same as the one who _published_ it (signed the M message), however I can't find any way to find an `ActorId` which is associated with the bundle we just received (except for the author M which doesn't help us as this is the key we want to verify against).

p2panda-spaces/src/traits/key_store.rs
```diff
@@ -0,0 +1,42 @@
...
+use std::fmt::Debug;
+
+use p2panda_core::{PrivateKey, PublicKey};
+use p2panda_encryption::key_manager::KeyManagerState;
+use p2panda_encryption::key_registry::KeyRegistryState;
+
+use crate::ActorId;
+use crate::message::SpacesArgs;
+
+pub trait Forge<ID, M, C> {
```
adzialocha, line 12: We can remove the `mut` requirement in `forge`, `set_key_manager` etc. on this level. Developers will choose internal mutability when needed.

p2panda-spaces/src/traits/key_store.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    fn public_key(&self) -> PublicKey;
+
+    fn forge(&mut self, args: SpacesArgs<ID, C>) -> impl Future<Output = Result<M, Self::Error>>;
+
+    fn forge_ephemeral(
+        &mut self,
+        private_key: PrivateKey,
+        args: SpacesArgs<ID, C>,
+    ) -> impl Future<Output = Result<M, Self::Error>>;
```
adzialocha: Is this ever used? I think we can remove it now

p2panda-spaces/src/traits/key_store.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+pub trait KeyStore {
```
adzialocha: Nice, this will soon move into `p2panda-store`. Maybe you add a TODO note here about that (or make an GitHub issue)?

p2panda-spaces/src/traits/key_store.rs
```diff
@@ -0,0 +1,42 @@
...
+    type Error: Debug;
+
+    fn key_manager(&self) -> impl Future<Output = Result<KeyManagerState, Self::Error>>;
+
+    fn key_registry(&self) -> impl Future<Output = Result<KeyRegistryState<ActorId>, Self::Error>>;
```
adzialocha: It could make sense to separate key management from key registry into two separate traits as developers might want to treat them differently: Key registry is more like an "address book" and doesn't contain sensitive information (can be stored in SQLite next to app for example), while the key manager contains secret keys and some applications probably want to bring this into a secure vault managed by the OS etc.

p2panda-spaces/src/traits/message.rs
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::fmt::Debug;
+
+use crate::message::SpacesArgs;
+use crate::{ActorId, OperationId};
+
+// @TODO: This could be an interesting trait for `p2panda-core`, next to another one where we
+// declare dependencies.
```
adzialocha: Oh yes! I've just introduced the first raits for `OperationId` and `Identifier` (https://github.com/p2panda/p2panda/blob/a6762e9831ccc8b6c008caf655468d6e75cff408/p2panda-core/src/traits.rs), can definitely start using that after some time (maybe we adjust the TODO note then).

It could make sense to separate `id` (`Identifier`) from `autor` as well, to prepare already the API for the new traits.

p2panda-spaces/src/traits/message.rs
```diff
@@ -0,0 +1,20 @@
...
+pub trait AuthoredMessage: Debug {
+    fn id(&self) -> OperationId;
+
+    fn author(&self) -> ActorId;
```
adzialocha, line 11: This uses `ActorId` and is thus a leaky abstraction. I guess we need to make it generic.

p2panda-spaces/src/traits/message.rs
```diff
@@ -0,0 +1,20 @@
...
+
+    // @TODO: Do we need a method here to check the signature?
```
adzialocha: Good question! I assume if we want to (and I think so) make the DSA scheme generic then yes.

We would then have:

```rust
type Error;

fn author(&self) -> P;

fn signature(&self) -> S;

// Here we can discuss if we want to cover the "tried to verify signature but failed doing so" case or just return false instead.
fn verify_signature(&self) -> Result<bool, Self::Error>
```

p2panda-spaces/src/traits/message.rs
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::fmt::Debug;
+
+use crate::message::SpacesArgs;
+use crate::{ActorId, OperationId};
+
+// @TODO: This could be an interesting trait for `p2panda-core`, next to another one where we
+// declare dependencies.
+pub trait AuthoredMessage: Debug {
```
adzialocha: Is the `Debug` requirement needed here on that level or can that be shifted higher up where it is really needed?

adzialocha: To make that part of the official `p2panda-core` API is maybe weird

p2panda-spaces/src/test_utils/spaces_store.rs
```diff
@@ -5,23 +5,22 @@ use std::convert::Infallible;
...
+use crate::test_utils::{TestConditions, TestMessage};
 use crate::traits::SpaceId;
-use crate::types::{ActorId, AuthGroupState};
+use crate::traits::spaces_store::{AuthStore, MessageStore, SpaceStore};
+use crate::types::AuthGroupState;
+
+pub type TestSpacesStore<ID> = MemoryStore<ID, TestMessage<ID>, TestConditions>;
 
 #[derive(Debug)]
 pub struct MemoryStore<ID, M, C>
```
adzialocha, line 20: This will also move to `p2panda-store` eventually, can leave a TODO note here as well

p2panda-spaces/src/credentials.rs
```diff
@@ -0,0 +1,53 @@
...
+use p2panda_core::{PrivateKey, PublicKey};
+use p2panda_encryption::crypto::x25519::SecretKey;
+use p2panda_encryption::{Rng, RngError};
+use serde::{Deserialize, Serialize};
+
+/// Every peer has two secret keys and _both_ required in order to interact with
+/// p2panda-spaces and neither can be rotated without losing access to all spaces. _If_ key
+/// rotation is required then both keys should be rotated together.
+///  
+/// A peers' identity secret is used for key agreement and encryption of messages in
```
adzialocha: Would remove the "and encryption" part as it's technically only used for key agreement (X3DH) and signing the pre-keys (which is part of key agreement so we can omit that part)

p2panda-spaces/src/credentials.rs
```diff
@@ -0,0 +1,53 @@
...
+/// is used to identify the peer (eg. for access control purposes).  
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Credentials {
+    pub(crate) private_key: PrivateKey,
+    pub(crate) identity_secret: SecretKey,
+}
+
+impl Credentials {
+    pub fn new(rng: &Rng) -> Result<Self, RngError> {
+        let private_key = PrivateKey::from_bytes(&rng.random_array()?);
```
adzialocha, line 23: Ooooooh, this is really good to know that we can re-use the rng here for that as well. I'd like to bring that into `p2panda-core` eventually so we can leverage from the seedable nature for tests and be consistent with the code-base etc.. I'll make a ticket about it.

p2panda-encryption/src/crypto/x25519.rs
adzialocha: I suggest to do this instead:

```rust
pub fn from_rng(rng: &Rng) -> Result<Self, RngError> {
    Ok(Self::from_bytes(rng.random_array()?))
}
```

.. this makes the use a little bit less hazmat.

p2panda-spaces/src/credentials.rs
```diff
@@ -0,0 +1,53 @@
...
+
+/// Every peer has two secret keys and _both_ required in order to interact with
+/// p2panda-spaces and neither can be rotated without losing access to all spaces. _If_ key
+/// rotation is required then both keys should be rotated together.
+///  
+/// A peers' identity secret is used for key agreement and encryption of messages in
+/// p2panda-encryption. Their private key is used to sign messages and the associated public key
+/// is used to identify the peer (eg. for access control purposes).  
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Credentials {
```
adzialocha, line 16: Can imagine we want this later to also be a trait, to allow "interactive" use of it where you only load your secret key material into memory exactly when it is required, or even better: leave it always in a more secured environment of the underlying hardware (like a YubiKey etc.) and only return the signature to us.

This will need more research for a correct API, so something for a GitHub issue for now.

p2panda-spaces/src/credentials.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub(crate) identity_secret: SecretKey,
+}
+
+impl Credentials {
+    pub fn new(rng: &Rng) -> Result<Self, RngError> {
+        let private_key = PrivateKey::from_bytes(&rng.random_array()?);
+
+        // @TODO: needed to make this method public in p2panda-encryption in order to construct a
+        // new SecretKey manually here. If the method indeed shouldn't be public we can adjust
+        // this constructor method to accept the SecretKey from outside as an argument.
```
adzialocha: Yes, let's do something like `SecretKey::from_rng` instead (and have that on `PrivateKey` as well some day, I'll make another ticket).

p2panda-spaces/src/credentials.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub fn from_keys(private_key: PrivateKey, identity_secret: SecretKey) -> Self {
+        Self {
+            private_key,
+            identity_secret,
+        }
+    }
+
+    pub fn private_key(&self) -> PrivateKey {
+        self.private_key.clone()
+    }
```
adzialocha, line 40: More a note, but it's a good place where "it shows": The "private key" naming is very confusing as soon as there's more "private keys" around. I suggest we rename it in the future to "SigningKey" and "VerifyingKey" (public), that's closer to how Ed25519 is meant to be used and how other crates name it.

p2panda-spaces/src/credentials.rs
adzialocha: .. and this will maybe also be in `p2panda-core` (as I can imagine using that for any forge as well in the future)?

p2panda-spaces/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+
+    pub fn credentials(&self) -> &Credentials {
+        &self.credentials
+    }
+
+    pub fn lifetime(&self) -> Lifetime {
+        Lifetime::new(self.pre_key_lifetime.as_secs())
+    }
```
adzialocha: Maybe this can be made `pub(crate)` as it might not be useful for a public interface

p2panda-encryption/src/traits/key_bundle.rs
```diff
@@ -13,5 +13,7 @@ pub trait KeyBundle {
 
     fn onetime_prekey_id(&self) -> Option<OneTimePreKeyId>;
 
+    fn lifetime(&self) -> &Lifetime;
```
adzialocha, line 16: :+1:

p2panda-encryption/src/crypto/rng.rs
adzialocha: I have an idea here to keep it private, let me know if that works (in the other comment)

p2panda-spaces/src/identity.rs
```diff
@@ -0,0 +1,437 @@
...
+    C: Conditions,
+{
+    pub async fn new(
+        key_store: K,
+        config: &Config,
+        rng: &Rng,
+    ) -> Result<Self, IdentityError<ID, K, M, C>> {
+        // @TODO: needed to take Rng::from_seed out from behind the test_utils flag in
+        // p2panda-encryption in order to do this. It's needed so that IdentityManager can have
+        // it's own rng without having a handle to the manager (as we do elsewhere).
```
adzialocha: Is it possible to explicitly pass in `&Rng` as an argument to every method which uses it instead and have it be passed on from the manager?

adzialocha: As far as I can tell there's nothing publicly facing which would require an `Rng` and even if, it could be wrapped by the `Manager`.

p2panda-spaces/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+use p2panda_encryption::key_bundle::Lifetime;
+use serde::{Deserialize, Serialize};
+
+use crate::Credentials;
+
+/// Configuration for a spaces instance.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Config {
+    /// Credentials (private key & identity secret) for local peer.
+    pub(crate) credentials: Credentials,
```
adzialocha: I somehow don't expect secret key material to be living in the config. Is it possible to leave that out?

(Also reading in the `IdentityManager` that this is the place where credentials live, but it's half true, it would also live in the config).

p2panda-spaces/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// our new one and the old one expiring.
+    pub(crate) pre_key_rotate_after: Duration,
+}
+
+impl Config {
+    pub fn new(credentials: &Credentials) -> Self {
+        Self {
+            credentials: credentials.to_owned(),
+            pre_key_lifetime: Duration::from_secs(60 * 60 * 24 * 90), // 90 days
+            pre_key_rotate_after: Duration::from_secs(60 * 60 * 24 * 60), // 60 days
```
adzialocha, line 24: These fields are actually not configurable :-)

p2panda-spaces/src/space.rs
```diff
@@ -691,7 +680,8 @@ where
         };
 
         // Forge message.
-        let message = manager.forge.forge(args).await.map_err(SpaceError::Forge)?;
+        let message = manager.identity.forge(args).await?;
+        // @TODO: persist message.
```
adzialocha, line 681: The forge will persist that message independently. This will be required for atomicity.

adzialocha: (otherwise we end up in races and self-forks where one await point forges a message, then switching to another await point forging another message while the `insert_operation` one has not been polled yet)

p2panda-spaces/src/test_utils/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+impl<ID> TestPeer<ID>
+where
+    ID: SpaceId + StdHash,
+{
+    pub async fn new(peer_id: u8) -> Self {
+        let rng = Rng::from_seed([peer_id; 32]);
+        let credentials = Credentials::new(&rng).unwrap();
+        let config = Config::new(&credentials);
```
adzialocha: As I've written somewhere else I think we want to separate config from credentials here, then one can also nicely write `Config::default` instead.

p2panda-spaces/src/config.rs
adzialocha: There's also a potential hazmat otherwise where people dump their `Config` somewhere (because it's serializable and should be) and later found out it contained their secrets :-P

p2panda-spaces/src/identity.rs
```diff
@@ -0,0 +1,437 @@
...
+/// Neither of a peers keys should be rotated individually, this would result in undefined
+/// behavior. Rotating both keys is possible but will result in the loss of access to existing
+/// spaces.
+#[derive(Debug)]
+pub struct IdentityManager<ID, K, M, C> {
+    key_store: K,
+    credentials: Credentials,
+    pre_key_lifetime: Duration,
+    pre_key_rotate_after: Duration,
+    my_keys_rotated_at: u64,
```
adzialocha, line 41: This only lives in memory, right? So we would be back at `0` whenever the process starts again which is maybe not the desired behaviour?

p2panda-spaces/src/identity.rs
```diff
@@ -0,0 +1,437 @@
...
+        if identity_secret != &self.credentials.identity_secret() {
+            return Err(IdentityError::IdentitySecretRotated);
+        }
+        let public_key = self.key_store.public_key();
+        if public_key != self.credentials.public_key() {
+            return Err(IdentityError::PrivateKeyRotated);
+        }
+
+        Ok(())
+    }
```
adzialocha, line 88: That's helpful!

p2panda-spaces/src/identity.rs
```diff
@@ -0,0 +1,437 @@
...
+
+    /// The public key of the local actor.
+    pub(crate) fn id(&self) -> ActorId {
+        self.credentials.public_key().into()
+    }
```
adzialocha, line 93: Nice, and no async required 😆

p2panda-encryption/src/traits/key_bundle.rs
adzialocha: Though, I can't see where this is used anywhere? Have some ideas though how it'll help you with other issues, so I think we should keep it.

p2panda-spaces/src/identity.rs
```diff
@@ -0,0 +1,437 @@
...
+/// spaces.
+#[derive(Debug)]
+pub struct IdentityManager<ID, K, M, C> {
+    key_store: K,
+    credentials: Credentials,
+    pre_key_lifetime: Duration,
+    pre_key_rotate_after: Duration,
+    my_keys_rotated_at: u64,
+    rng: Rng,
+    _phantom: PhantomData<(ID, K, M, C)>,
```
adzialocha: Mini: `K` is used actually and can be removed here

p2panda-spaces/src/identity.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        let key_manager_y = self
+            .key_store
+            .key_manager()
+            .await
+            .map_err(IdentityError::KeyStore)?;
+
+        // Automatically rotate pre key when it reached critical expiry date.
+        let key_bundle = if now() - self.my_keys_rotated_at > self.pre_key_rotate_after.as_secs() {
+            self.my_keys_rotated_at = now();
```
adzialocha, line 105: This information will be lost because it is only stored in-memory. Can we use the `lifetime()` method you've exposed now on the current key bundle instead to compute it?

p2panda-spaces/src/identity.rs
```diff
@@ -0,0 +1,437 @@
...
+        member.key_bundle().verify()?;
+
+        let y = self
+            .key_store
+            .key_registry()
+            .await
+            .map_err(IdentityError::KeyStore)?;
+
+        // @TODO: Setting longterm bundle should overwrite previous one if this is newer.
+        let y_ii = KeyRegistry::add_longterm_bundle(y, member.id(), member.key_bundle().clone());
```
adzialocha, line 148: I think we should fix that as part of this PR (otherwise it will break fairly fastly), happy to assist here or rewrite that logic

p2panda-spaces/src/identity.rs
adzialocha: The act of publishing is by forging a message and appending it to the log, so we're not looking for the information on that layer anymore (where you already removed the message and only have the "validated" key bundle).

I assume that we will have that validation on the operation-layer already and I assume somewhere we are using the public key of that operation for the actor id, but it's good to keep a note somewhere for people who will "bring their own datatype" to `p2panda-spaces`

p2panda-spaces/src/manager.rs
```diff
@@ -43,38 +43,61 @@ use crate::types::{ActorId, AuthResolver, OperationId};
...
+    /// Instantiate a new manager with custom configuration.
+    #[allow(clippy::result_large_err)]
+    pub async fn new_with_config(
+        spaces_store: S,
+        key_store: K,
+        config: &Config,
+        rng: Rng,
+    ) -> Result<Self, ManagerError<ID, S, K, M, C, RS>> {
+        // @TODO: kick off service which will handle key bundle rotation and space repair.
+        // @TODO: return rx channel where new events/messages will be emitted.
```
adzialocha: As we've discussed this will not be necessary anymore :+1:

p2panda-spaces/src/manager.rs
```diff
@@ -251,6 +233,59 @@ where
         Ok(events)
     }
 
+    // @TODO: Make it work without async
+    /// The public key of the local actor.
+    pub async fn id(&self) -> ActorId {
```
adzialocha: Oh no, async is still here lol

p2panda-spaces/src/manager.rs
```diff
@@ -251,6 +233,59 @@ where
...
+        let manager = self.inner.read().await;
+        manager.identity.id()
+    }
+
+    /// The local actor id and their long-term key bundle.
```
adzialocha, line 244: It's the high-level API and users might want to know here that their long-term bundle will be automatically renewed when it reaches the "key rotation before expiry" window

p2panda-spaces/src/manager.rs
```diff
@@ -251,6 +233,59 @@ where
...
+    ) -> Result<(), ManagerError<ID, S, K, M, C, RS>> {
+        let mut manager = self.inner.write().await;
+        manager
+            .identity
+            .register_member(member)
+            .await
+            .map_err(ManagerError::IdentityManager)
+    }
+
+    pub async fn key_bundle_expired(&self) -> bool {
```
adzialocha, line 273: Here we should mention in docs more what this is used for and what the expected flow is

p2panda-spaces/src/manager.rs
```diff
@@ -251,6 +233,59 @@ where
...
+        let manager = self.inner.read().await;
+        manager.identity.key_bundle_expired().await
+    }
+
+    pub async fn key_bundle(&self) -> Result<M, ManagerError<ID, S, K, M, C, RS>> {
```
adzialocha: Maybe name this `key_bundle_message` or similar as this is what the return value is?

adzialocha: On a higher-level API we don't really expose key bundles but `Member` to the users, so maybe it makes sense to actually call that method `member_message`?

adzialocha: Also, doc-strings will be nice here as well

p2panda-spaces/src/manager.rs
```diff
@@ -251,6 +233,59 @@ where
...
+        let mut manager = self.inner.write().await;
+        manager
+            .identity
+            .key_bundle()
+            .await
+            .map_err(ManagerError::IdentityManager)
+    }
+
+    pub(crate) async fn forge(
```
adzialocha, line 294: Doc-strings will be nice here as well, even when not public

p2panda-encryption/src/crypto/x25519.rs
sandreae: Thanks, this works for generating a new `SecretKey` in `p2panda-spaces`.

p2panda-encryption/src/crypto/rng.rs
sandreae: If we can take the same approach here as you suggested for SecretKey then we're all good. The use case I have is that I want to generate a new `Rng` from an existing one to be owned by the `IdentityManager`. I could add this method to `Rng`:

```rust
pub fn from_rng(rng: &Rng) -> Result<Self, RngError> {
    Ok(Self::from_seed(rng.random_array()?))
}
```

p2panda-spaces/src/config.rs
sandreae: Yeh, I hesitated about putting credentials here for the reasons you mention, I did in the end to keep all configuration info together (both lifetime and credentials are needed to configure the key manager). But in retrospect, it's better to prioritize the security concerns you mention. I'll move credentials out of the config.

p2panda-spaces/src/credentials.rs
sandreae: Yeh, this makes sense. I'll add it to the tracking issue for now, I plan to pull out all remaining tasks into their own issues at the end of the week.

p2panda-spaces/src/credentials.rs
sandreae: Yeh, for now I tried to follow existing naming convention, would be great to agree on the nomenclature and then change it across all crates. I like your naming suggestion :+1:

p2panda-encryption/src/crypto/rng.rs
adzialocha: Would the `from_seed` method be available without turning on the `test_utils` flag? If yes, I'd be fine with this API method.

The reason is that I don't want the seedable rng to be available by default as it's a potential hazmat as well (people want to _always_ use a CSPRNG). So we either move that feature behind an explicit feature flag (like `seedable_rng` or `hazmat`) which I would like to avoid or do the argument passing instead as I've mentioned.

p2panda-spaces/src/config.rs
adzialocha: Great! It's a security concern but also an API design question, I expect configuration to be independent of key handling and would assume that I can never touch configuring a system when not necessary and just stick to the defaults while with keys this is not possible.

p2panda-encryption/src/crypto/rng.rs
sandreae: > Would the `from_seed` method be available without turning on the `test_utils` flag? If yes, I'd be fine with this API method.

The `from_seed` method would stay behind the `test_utils` flag, but the `from_rng` would be public. I made this change in the code now so you can see what I'm suggesting there too.

adzialocha: Looking good!

p2panda-spaces/src/space.rs
sandreae: Aha, ok, this needs changing elsewhere then as we are currently persisting messages returned from the forge in the `Space` and `Group` api. 

So I'm clear, we can assume that the forge will persist all messages it "forges" and so can expect to be able to get them from the `MessageStore` later (but within the same method call).

I'll raise an issue for this and deal with it in another PR.

adzialocha: Let me think about this a little bit more in terms of consistency (this will be important now for the general "publishing" part and the guarantees around it). Having it as a separate issue is super for now!

Generally we can say though that we can guarantee that you get the messages from the store afterwards (either through dirty reads within a transaction or the guarantee of the data model).

p2panda-spaces/src/traits/message.rs
sandreae: I'll add this to the tracking issue.

p2panda-spaces/src/traits/message.rs
sandreae: I'll add this to the tracking issue.

p2panda-spaces/src/traits/message.rs
sandreae: Updated TODO and added to tracking issue.

p2panda-spaces/src/identity.rs
sandreae: Ok, so to be clear, there is no "additional" validation (beyond calling `key_bundle.verify()`) that we need to do in spaces when receiving key bundle messages?

p2panda-spaces/src/identity.rs
sandreae: Yes, that's also an option, if you're not happy with the current solution (only `from_rng(..)` method is public) we can do it that way.

p2panda-spaces/src/identity.rs
sandreae: Ah yeh, true, I'll make that change. Already looked at it a bit and see that actually in the case where, on initialisation, we detect that keys should be rotated we would want to publish that new key bundle. This would mean that Manager::new(..) can actually itself return an M message. Believe this is fine, but hadn't considered that case in the api yet.

sandreae: Same would occur on first start.

p2panda-spaces/src/manager.rs
sandreae: Done

p2panda-spaces/src/manager.rs
sandreae: I'll go for naming it key_bundle_message() for now as this keeps the naming consistent (we refer to key bundles elsewhere).

p2panda-spaces/src/space.rs
sandreae: This has been changed now in #825

p2panda-spaces/src/identity.rs
sandreae: We've discussed moving this onto the encryption state, in which case it would then be persisted.

Issue #810, opened 2025-09-25
Is it ever valid for Header.extensions to be None?


maackle:
https://github.com/p2panda/p2panda/blob/62f8c93c8b092a004dee46e4256cb46060ccca9b/p2panda-core/src/operation.rs#L206-L207

I see a lot of code handling the case that `extensions` is None, including returning empty lists of dependencies, no LogId, etc., but I don't see `extensions` ever being None except in tests. Is it valid for it to be None? If not can this be made non-optional?

maackle:
> A header without extensions

I imagine a header without extensions would just be `Header<()>`. But I don't understand why `extensions` needs to be `Option<E>` instead of just `E`.

> Real applications probably want to consider an operation invalid if the extension is `None`

This is my point, is there really ever a valid use case for extensions to be `None`? Even if used in tests, why not just use `Header<()>` if no extensions are needed? Feels like a code smell, so I'm trying to follow my nose to see if there's something I'm missing.

PR #811, opened 2025-09-27
Make extensions non-optional

maackle:
I did this experimental refactor to help me understand the implications of my question in https://github.com/p2panda/p2panda/issues/810#issuecomment-3341117623. All tests pass except for one with hardcoded bytes values.

Please don't merge this if it holds up whatever else you're working on, this was more for my understanding, but if it's helpful feel free.

## 📋 Checklist

### DIFF
--- a/p2panda-core/src/serde.rs
+++ b/p2panda-core/src/serde.rs
@@ -161,9 +161,7 @@ where
 
         seq.serialize_element(&self.previous)?;
 
-        if let Some(extensions) = &self.extensions {
-            seq.serialize_element(extensions)?;
-        }
+        seq.serialize_element(&self.extensions)?;
 
         seq.end()
     }
@@ -258,9 +256,10 @@ where
                     .map_err(|_| SerdeError::custom("invalid previous links, expected array"))?
                     .ok_or(SerdeError::custom("previous array missing"))?;
 
-                let extensions: Option<E> = seq
+                let extensions: E = seq
                     .next_element()
-                    .map_err(|err| SerdeError::custom(format!("invalid extensions: {err}")))?;
+                    .map_err(|err| SerdeError::custom(format!("invalid extensions: {err}")))?
+                    .ok_or(SerdeError::custom("extensions missing"))?;
 
                 Ok(Header {
                     version,
@@ -438,23 +437,7 @@ mod tests {
                 seq_num: 0,
                 backlink: None,
                 previous: vec![],
-                extensions: Some(extensions.clone()),
-                signature: None,
-            },
-            &private_key,
-        );
-
-        assert_serde_roundtrip(
-            Header::<CustomExtensions> {
-                version: 1,
-                public_key: private_key.public_key(),
-                payload_size: 0,
-                payload_hash: None,
-                timestamp: 0,
-                seq_num: 7,
-                backlink: Some(Hash::new(vec![1, 2, 3])),
-                previous: vec![],
-                extensions: None,
+                extensions: extensions.clone(),
                 signature: None,
             },
             &private_key,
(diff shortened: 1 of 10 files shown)


adzialocha:
This is a really nice API improvement! Let's get this merged soon so we can reflect it well in any of our ongoing refactorings.

From looking at the failing test one can also see that there's now a `null` value being serialized, I guess we now need a trick to avoid serializing the `()` (void) value.

p2panda-core/src/serde.rs
```diff
@@ -258,9 +258,15 @@ where
...
-                    .next_element()
-                    .map_err(|err| SerdeError::custom(format!("invalid extensions: {err}")))?;
+                let extensions: E = if std::mem::size_of::<E>() > 0 {
+                    seq.next_element()
+                        .map_err(|err| SerdeError::custom(format!("invalid extensions: {err}")))?
+                        .ok_or(SerdeError::custom("extensions missing"))?
+                } else {
+                    // For zero-sized types, we don't need to read anything from the sequence
+                    // as the serializer doesn't write anything for them
+                    unsafe { std::mem::zeroed() }
```
maackle: Maybe there is a "safer" way to get a zero-sized value here but I haven't found it yet.

maackle:
@adzialocha how about this?

p2panda-core/src/serde.rs
adzialocha: Maybe https://docs.rs/serde/latest/serde/de/struct.IgnoredAny.html ?

maackle: Skipping the field is no problem, I think the missing piece is a safe way to populate `E` without knowing what it is. The riddle is, what do `()`, `struct A`, `enum A { B(()) }`, etc. all have in common?

maackle: I can look into it a bit, maybe what is there now really is safe. The type is guaranteed to be zero bytes in this case after all.

maackle: looking into ZSTs some more (zero-sized types) it looks like this is the way. There is this https://github.com/rust-lang/rust/pull/146479 to look forward to but I guess it won't be stable for a while.

My takeaway at doing this unsafely is that it's fine as long as the ZST is not uninhabited, like `enum E {}`, but in that case you couldn't serialize it in the first place. Of course somebody could find a way to abuse this.

What do you think, is it acceptable to just have this extra byte hanging around for now? Will anyone be using a zero sized extension type in the real world anyway?

adzialocha: > What do you think, is it acceptable to just have this extra byte hanging around for now? Will anyone be using a zero sized extension type in the real world anyway?

Yeah, I agree, it's good to leave the code "safe" and accept the extra byte for this edge case. We can also argue that _if_ people really want to use a type without extensions they can provide one themselves (with their own serde methods).

adzialocha: Thank you for looking into this!

maackle:
should be good to merge :+1:

p2panda-stream/src/operation.rs
```diff
@@ -168,7 +168,7 @@ mod tests {
             seq_num: 0,
             backlink: None,
             previous: vec![],
-            extensions: None,
+            extensions: crate::test_utils::Extensions::default(),
```
mycognosist:
```suggestion
            extensions: Extensions::default(),
```

p2panda-stream/src/operation.rs
```diff
@@ -188,7 +188,7 @@ mod tests {
             seq_num: 12, // we'll be missing 11 operations between the first and this one
             backlink: Some(Hash::new(b"mock operation")),
             previous: vec![],
-            extensions: None,
+            extensions: crate::test_utils::Extensions::default(),
```
mycognosist:
```suggestion
            extensions: Extensions::default(),
```

PR #813, opened 2025-09-29
Refactored `p2panda-store`: SQLite store with transactions API

adzialocha:
## 📋 Checklist


p2panda-store-next/src/sqlite.rs
```diff
@@ -0,0 +1,490 @@
...
+/// Please note that SQLite strictly serializes transactions with _writes_ and will block any
+/// parallel attempt to begin another one. Processes starting a transaction will acquire a
+/// `TransactionPermit` and keep it until the transaction was committed or rolled back. If the
+/// query only involves _reads_ it is recommended to not use transactions and use the `execute`
+/// method directly as acquiring transactions will potentially block other processes to do work.
+///
+/// ## Design decisions
+///
+/// This storage API design was chosen to make the dynamics of the underlying SQLite database
+/// explicit. Internally any process can access the transaction object to do writes and
```
sandreae: One more sentence here succinctly explaining why exposing the underlying SQLite transaction patterns is required could be good.

p2panda-store-next/src/memory.rs
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::fmt::Debug;
+
+/// In-memory store.
```
sandreae, line 5: Why do we still need a separate `MemoryStore` impl rather than using SQLite in-memory database?

adzialocha: Good question! It's a little bit less boilerplate maybe but otherwise no real reason. So far I've been just copying old store implementations over (orderer), but for the future we can not do that and de-prioritize the in-memory implementation.

PR #814, opened 2025-09-29
Refactored `OperationStore` with SQLite and In-Memory implementation

adzialocha:
## 📋 Checklist


sandreae (approved):

p2panda-store-next/src/memory.rs
```diff
@@ -2,21 +2,41 @@
...
+where
+    T: Debug,
+    ID: Debug,
+{
     fn default() -> Self {
         Self::new()
     }
 }
+
+// Trait implementations are in the regaring modules, see for example `orderer` or `operation` etc.
```
mycognosist:
```suggestion
// Trait implementations are in the relevant modules, see for example `orderer` or `operation` etc.
```

mycognosist (approved):

PR #815, opened 2025-09-29
Refactored `OrdererStore` in-memory implementation & SQLite support

adzialocha:
## 📋 Checklist

### DIFF
--- a/p2panda-store-next/migrations/20250928132846_create_ordering.sql
+++ b/p2panda-store-next/migrations/20250928132846_create_ordering.sql
@@ -1,0 +1,71 @@
+-- SPDX-License-Identifier: MIT OR Apache-2.0
+
+CREATE TABLE IF NOT EXISTS orderer_ready_v1 (
+    id                      TEXT            NOT NULL    PRIMARY KEY,
+    queue_index             INTEGER         NOT NULL    UNIQUE,
+    in_queue                BOOLEAN         NOT NULL    DEFAULT TRUE
+);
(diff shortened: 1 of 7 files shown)


p2panda-store-next/migrations/20250928132846_create_ordering.sql
```diff
@@ -0,0 +1,71 @@
...
+--
+-- id: 2
+-- child_id: 4
+-- parent_id: 2
+-- set_digest: Digest(4, 2)
+CREATE TABLE IF NOT EXISTS orderer_pending_v1 (
+    id                     TEXT            NOT NULL,
+    child_id               TEXT            NOT NULL,
+    parent_id              TEXT            NOT NULL,
+    set_digest             TEXT            NOT NULL
```
sandreae, line 54: It's not clear to me how exactly the set_digest field is used, could you explain a little more?

adzialocha: It's a way to group related entries in the table under one unique identifier, another way to do it would be to have a separate table which tracks that but that seems overkill

adzialocha: It roughly follows the design of `HashMap<ID, HashSet<(ID, Vec<ID>)>` in the in-memory store. I couldn't come up with something else as it seems the orderer was designed around that data type, so the SQLite implementation follows that a bit as well.

PR #822, opened 2025-10-01
Introduce actor supervision tree

mycognosist:
## Added

This PR introduces the root-level network actor and defines the supervision tree for the entire system. "Child" actors are spawned and supervision events are captured. Failed actors are respawned to ensure continued system functionality.

## Not Added 

No internal actor logic has been introduced in this PR (besides some minor changes for argument passing).  

## Network Supervision Tree

network_supervision_tree

Actor spawn order is downward and rightward (ie. first `Network`, then `Events`, `Endpoint`, `Subscription`, `Gossip` etc.).

The hierarchy is partially determined by the current dependencies of our sub-systems. Router, gossip and sync all require an iroh `Endpoint`, so it makes sense to have these as "children" of the endpoint actor. This means that any failure of the endpoint actor will result in a recreation of all downstream actors. It's likely that our supervision tree will change slightly during the course of this rewrite. What's important is that we keep actors neatly encapsulated; that will make it trivial to rearrange them in the hierarchy as needed.

## Tests

I'm trying out the [tracing-test](https://crates.io/crates/tracing-test) crate as a way of ensuring "child" actors have been started (by asserting that certain log statements appear). Since `ractor` actors are registered in a global namespace, we run into errors if tests are run in parallel (because, for instance, the `"subscription"` actor has already been registered when the second test starts). I've introduced the [serial_test](https://crates.io/crates/serial_test) to get around this issue.

## 📋 Checklist

### DIFF
--- a/p2panda-net-next/src/actors/router.rs
+++ b/p2panda-net-next/src/actors/router.rs
@@ -1,12 +1,6 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 //! An iroh-specific router actor for registering ALPNs.
-
-use std::time::Duration;
-
-use iroh::protocol::Router as IrohRouter;
-use iroh::Endpoint as IrohEndpoint;
-use iroh_gossip::ALPN as GOSSIP_ALPN;
 use ractor::{Actor, ActorProcessingErr, ActorRef, Message};
 
 // TODO: Remove once used.


p2panda-net-next/src/actors/subscription.rs
```diff
@@ -1,27 +1,66 @@
...
+use crate::actors::gossip::{Gossip, ToGossip};
+use crate::actors::sync::{Sync, ToSync};
 
 pub enum ToSubscription {}
 
 impl Message for ToSubscription {}
 
-pub struct SubscriptionState {}
+pub struct SubscriptionState {
+    endpoint: IrohEndpoint,
```
sandreae, line 23: There's been the new suggestion that we have a connection actor which holds the QUIC connection for one/many child actors. I can see the benefit of having a single point of failure per connection. How would that change the actor args/states like this one where the endpoint is currently passed into all actors that need it?

p2panda-net-next/src/actors/events.rs
```diff
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 //! Events actor.
+//!
+//! Receives events from other actors, aggregating and enriching them before informing
```
sandreae, line 5: Just a note for the future on composing events from what I've observed doing it in p2panda-spaces. Rather than the actors themselves composing events and sending them "ready formed" to the events actor, it might be a nicer pattern to send `prev_state` and `next_state` objects, along with maybe any message which was just processed, and then have the events actor itself composing the events, and optionally adding any aggregate data as well. 

Would keep all the event logic nicely in one place, instead of repeating a lot in all actors.

p2panda-net-next/src/actors/network.rs
```diff
@@ -0,0 +1,245 @@
...
+                            (),
+                            myself.clone().into(),
+                        )
+                        .await?;
+
+                        state.events_actor_failures += 1;
+                        state.events_actor = events_actor;
+                    }
+                    Some("endpoint") => {
+                        warn!("network actor: endpoint actor failed: {}", panic_msg);
```
sandreae, line 147: What causes an endpoint actor to fail?

p2panda-net-next/src/actors/subscription.rs
mycognosist: Great question. The iroh endpoint only really needs to be passed to the router and gossip actors. That's because the iroh `Router` and `Gossip` builders need to be instantiated with an `Endpoint` before they can be spawned. Another pattern is for those actors to gain access to an `Endpoint` by sending a message to the endpoint actor (which would then return a clone of the `Endpoint`). My feeling is to lean towards passing in arguments when they are required for the initial configuration and spawning of the actor; data or types which are required during the actual functioning of the actor can be fetched via message passing.

The connection actor approach is compatible with what I've described above. Requests for a connection would always be made to the endpoint actor (not directly from within another actor which may have received an `Endpoint`).

p2panda-net-next/src/actors/events.rs
mycognosist: Ah cool, that's helpful. I'll ping you for assistance when designing the event flows 🙏🏻

p2panda-net-next/src/actors/network.rs
mycognosist: Good question. I don't foresee any failure cases in the endpoint actor. There are some errors which could emerge when initialising the actor (during the `pre_start()` method) but those would not invoke the supervision strategy. So for now it's fair to say that this failure check is "just in case" and could help with debugging unexpected behaviour.

PR #824, opened 2025-10-02
`p2panda-spaces`: api for detetcting and publishing missing auth pointers to spaces

sandreae:
Introduces `spaces_repair_required` and `repair_spaces` with which users can 1) detect if any of their local spaces are out-of-sync with the global auth state, and 2) apply any missing operations (bring spaces in sync) from the global auth state to out-of-sync spaces.

## 📋 Checklist

### DIFF
--- a/p2panda-spaces/src/space.rs
+++ b/p2panda-spaces/src/space.rs
@@ -313,20 +313,32 @@ where
             return Ok(vec![]);
         }
 
+        let duplicate_pointer = y.auth_y.inner.operations.contains_key(&auth_message.id());
+
         let mut current_members = secret_members(y.auth_y.members(y.group_id));
         current_members.sort();
 
         // Process auth message on space auth state.
-        y.auth_y = AuthGroup::process(y.auth_y, auth_message).map_err(SpaceError::AuthGroup)?;
-
-        // Get next space members.
-        let mut next_members = secret_members(y.auth_y.members(y.group_id));
-        next_members.sort();
+        //
+        // Skip processing if this auth message has already been processed. This can happen when
+        // multiple peers concurrently publish pointers to some auth message into the space.
+        let next_members = if !duplicate_pointer {
+            y.auth_y = AuthGroup::process(y.auth_y, auth_message).map_err(SpaceError::AuthGroup)?;
+            // Get next space members.
+            let mut next_members = secret_members(y.auth_y.members(y.group_id));
+            next_members.sort();
+            next_members
+        } else {
+            current_members.clone()
+        };
 
         // Make the dgm aware of the new space members.
         y.encryption_y.dcgka.dgm.members = HashSet::from_iter(next_members.clone());
 
         // Construct encryption message.
+        //
+        // We do this even when the auth message was not processed above, to make sure that we
+        // still consume all direct messages.
         let my_id = self.manager.id();
         let encryption_message = EncryptionMessage::from_membership(
             space_message,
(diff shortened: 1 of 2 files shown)


p2panda-spaces/src/manager.rs
```diff
@@ -241,6 +242,25 @@ where
...
+        let mut events = vec![];
+        for message in messages {
+            let events_inner = self.process(message).await?;
+            events.extend(events_inner);
+        }
+
+        // @TODO: We _could_ repair any out-of-sync spaces here after processing a batch of
+        // messages. However this would introduce inconsistent behaviour with process() as we
+        // don't want to repair there in order to avoid redundant messages. So for now making
+        // repairing a concern of the user is maybe best.
```
adzialocha: (Not sure I understand the "inconsistent behaviour" part so I'm not sure if what I say makes sense.)

Since no high-level logic will look at any other methods than "process" and "process_batch" would it be possible to just _always_ call "should repair" and then "repair" if true after processing everything in-batch? That seems to resemble a "throttling" logic and was the idea of introducing a batch process method.

adzialocha: Alternative is to remove this here because it doesn't really otherwise add any functionality

sandreae: What you describe is possible, and my comment referred to the fact that if we don't also do that for `process` then the API would exhibit inconsistent behaviour. Calling `process_batch` results in spaces auto-repairing, only calling `process` does not.

p2panda-spaces/src/manager.rs
```diff
@@ -305,8 +325,92 @@ where
...
+                in_need_of_repair.push(id);
+            }
+        }
+
+        Ok(in_need_of_repair)
+    }
+
+    /// Apply any missing messages from the shared auth state into the passed spaces.
+    ///
+    /// This is required when a space has become "out-of-sync" with the shared auth state.
```
adzialocha: Do you want to add here the performance / messyness implications of calling this method too much? As I understood you _don't_ want to directly call it when `spaces_repair_required` is set to true but wait a bit and this dynamic is not clear. A diagram would also help. It's very complicated what's going on and we might forget already after 2 weeks what this was about 😆

p2panda-spaces/src/manager.rs
adzialocha: I see! Yeah, I think then it's either remove `process_batch` or add this here, both is fine with me as long as the behaviour is well documented.

sandreae: Why would you want to remove `process_batch` if it doesn't call the repair methods internally? (if this is what you're suggesting). I still see it as a useful method.

sandreae: Aside from that point, I do think that the desired behaviour _is_ to repair within `process_batch` so as not to push this consideration onto the user, but I just didn't feel the api inconsistency this introduces was acceptable. So basically both options (don't repair anywhere vs. only repair in process_batch) are not satisfactory.... I'd like to find a 3rd happy path but not sure what that is.

adzialocha: > Why would you want to remove process_batch if it doesn't call the repair methods internally? (if this is what you're suggesting). I still see it as a useful method.

We can't change the fact that processing operations is strictly sequential, so there's not much we can do about improving "performance" around that. The benefit of that method I can currently only see in that it is _maybe_ allowing to run multiple database queries within one transaction and since starting transactions is costly that _can_ be a benefit. Hard to tell right now what the real benefit is, probably something for a benchmark one day, not sure how measurable it will be.

If we would have done some "second order" expensive work like repairs (which one can see as a form of "batch processing" aka, grouping single pieces of expensive work into one piece) then I would see a clear benefit of having that method.

I'm not against keeping it, we can keep it for now, similar to you, I have the feeling we have all of these methods right now mostly because we don't know what that "3rd happy path" is and maybe that's a good reason to keep the surface very simple?

If I would ever get to implement a `process_batch` on a higher level ("spaces processor") I would implement it exactly like that:

1. Process all items in that array
2. Call "should repair"
3. If true was returned, call "repair"

Since no other methods will be exposed this will not lead to confusion for anyone else and is nicely hidden. We can expose them for now on that level and do the "hiding" in the processor. But then we really don't need the `process_batch` method.

PR #826, opened 2025-10-03
`p2panda-spaces`: allow processing the same message multiple times (idempotency)

sandreae:
## 📋 Checklist


sandreae:
Ok, on third look there is an approach we can take without adding any additional state tracking which does achieve the expected behaviour. 

1) processed control messages are already tracked inside p2panda-auth and the encryption orderer, we can tap into this to ignore (do not process, do not emit state change events) for already processed control messages.
2) application messages do not change state, and therefore can be processed multiple times with no unexpected side effects. This is actually the expected behaviour, as it supports "replays" of application messages. Each time an application message is processed an application event will be emitted.

adzialocha (approved):

PR #830, opened 2025-10-08
Fix long-term bundle keys by allowing to manage multiple ones

adzialocha:
The previous implementation of `KeyManager` and `KeyRegistry` around `LongTermKeyBundle` was over-simplified (probably due to copy & paste from the "message encryption" scheme where long-term key bundles are not used). Some bugs needed fixing to make long-term key-bundle management possible:

- [x] 1. Key registry currently only gives us the "highest" key bundle from the stack instead of the actual "latest" one (so it's _not_ looking at the lifetime currently). Also it's popping the item off the stack right now which is not what we want, we want to keep it there instead
- [x] 2. Key manager doesn't keep "almost expired" key bundles around and just removes them on rotation, this breaks use of different published key bundles for other peers:  Peer A publishes key bundles K1 and later K2 (because K1 is about to expire), now Peer B wants to establish a key agreement with A and _still_ uses K1 (as they haven't been aware of K2) yet and A will need to reject it even though technically it's been a valid key bundle still. It might be that we need to change the methods to query the registry based on public pre keys instead, as this can be used as an unique id for each key bundle 
- [x] 3. Rejecting direct messages from other members during X3DH which used one of our expired key bundles for key agreement

I've also made the separation between the "static" identity secret and the "dynamic" pre-key secrets clearer. They can also be now independently stored, since the identity secret might come from a different source / store. This makes it's use easier and fits better into the needs of `p2panda-spaces`.

Related PR: #808 

## 📋 Checklist

### DIFF
--- a/p2panda-encryption/src/data_scheme/tests.rs
+++ b/p2panda-encryption/src/data_scheme/tests.rs
@@ -102,15 +102,18 @@ fn group_operations() {
     let alice_group_secret_0 = SecretBundle::generate(&alice_bundle, &rng).unwrap();
     let (alice_dcgka, output) =
         Dcgka::create(alice_dcgka, vec![alice, bob], &alice_group_secret_0, &rng).unwrap();
-    let (alice_dcgka, _) = Dcgka::process(alice_dcgka, ProcessInput {
-        seq: MessageId {
+    let (alice_dcgka, _) = Dcgka::process(
+        alice_dcgka,
+        ProcessInput {
+            seq: MessageId {
+                sender: alice,
+                seq: 0,
+            },
             sender: alice,
-            seq: 0,
+            control_message: output.control_message.clone(),
+            direct_message: None,
         },
-        sender: alice,
-        control_message: output.control_message.clone(),
-        direct_message: None,
-    })
+    )
     .unwrap();
     let alice_bundle = SecretBundle::insert(alice_bundle, alice_group_secret_0.clone());
     assert_eq!(alice_bundle.len(), 1);
@@ -127,15 +130,18 @@ fn group_operations() {
         .find(|dm| dm.recipient == bob)
         .expect("direct message for bob");
 
-    let (bob_dcgka, output) = Dcgka::process(bob_dcgka, ProcessInput {
-        seq: MessageId {
+    let (bob_dcgka, output) = Dcgka::process(
+        bob_dcgka,
+        ProcessInput {
+            seq: MessageId {
+                sender: alice,
+                seq: 0,
+            },
             sender: alice,
-            seq: 0,
+            control_message: output.control_message.clone(),
+            direct_message: Some(direct_message),
         },
-        sender: alice,
-        control_message: output.control_message.clone(),
-        direct_message: Some(direct_message),
-    })
+    )
     .unwrap();
 
     let GroupSecretOutput::Secret(bob_group_secret_0) = output else {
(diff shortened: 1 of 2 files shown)


p2panda-encryption/src/two_party/two_party.rs
```diff
@@ -321,7 +321,9 @@ where
                 let plaintext = x3dh_decrypt(
                     &ciphertext,
                     KMG::identity_secret(&y_manager_i),
-                    KMG::prekey_secret(&y_manager_i),
+                    KMG::prekey_secret(&y_manager_i, &ciphertext.prekey_id)
+                        // Fails when sender used an invalid / expired pre-key for initial X3DH.
+                        .map_err(|_| TwoPartyError::UnknownPreKeyUsed(ciphertext.prekey_id))?,
```
adzialocha, line 326: This satisfies point 3 and is checking if the remote peer / sender used an expired pre-key from us.

p2panda-encryption/src/two_party/x3dh.rs
```diff
@@ -48,6 +51,7 @@ pub fn x3dh_encrypt<KB: KeyBundle>(
     their_prekey_bundle: &KB,
     rng: &Rng,
 ) -> Result<X3dhCiphertext, X3dhError> {
+    // Ensure we're not using an expired bundle of receiver.
     their_prekey_bundle.verify()?;
```
adzialocha, line 55: This satisfies point 4: We've already checked here to not use an expired pre-key when sending a message to the other peer

p2panda-encryption/src/two_party/x3dh.rs
```diff
@@ -34,6 +34,9 @@ pub struct X3dhCiphertext {
     /// in long-term key bundles).
     pub onetime_prekey_id: Option<OneTimePreKeyId>,
 
+    /// Identifier of the used long-term pre-key.
+    pub prekey_id: PreKeyId,
```
adzialocha, line 38: This satisfies partly point 2: We know now _which_ pre-key was used during X3DH key agreement

p2panda-encryption/src/traits/key_manager.rs
```diff
@@ -20,15 +20,18 @@ pub trait PreKeyManager {
 
     type Error: Error;
 
-    fn prekey_secret(y: &Self::State) -> &SecretKey;
+    fn prekey_secret<'a>(
+        y: &'a Self::State,
+        id: &'a PreKeyId,
+    ) -> Result<&'a SecretKey, Self::Error>;
```
adzialocha, line 26: This satisfies Point 2: The API allows now to store multiple pre-key secrets of us, so we can keep around all valid ones (and not only one)

p2panda-encryption/src/traits/key_manager.rs
```diff
@@ -20,15 +20,18 @@ pub trait PreKeyManager {
...
 
     fn rotate_prekey(
         y: Self::State,
         lifetime: Lifetime,
         rng: &Rng,
     ) -> Result<Self::State, Self::Error>;
 
-    fn prekey_bundle(y: &Self::State) -> LongTermKeyBundle;
+    fn prekey_bundle(y: &Self::State) -> Result<LongTermKeyBundle, Self::Error>;
```
adzialocha, line 34: Related to Point 2 we allow now an error to be returned when no valid bundle exists. This is the point where the application should generate new keys

p2panda-encryption/src/key_registry.rs
```diff
@@ -111,13 +111,14 @@ where
...
     ) -> Result<(Self::State, Option<LongTermKeyBundle>), Self::Error> {
-        let bundle = y
-            .longterm_bundles
-            .get_mut(id)
-            .and_then(|bundles| bundles.pop());
+        let Some(bundles) = y.longterm_bundles.get(id) else {
+            return Ok((y, None));
+        };
+
+        let bundle = latest_key_bundle(bundles).cloned();
```
adzialocha: This satisfies Point 1 and returns now always the "latest" key bundle if it exists or none if invalid.

p2panda-encryption/src/key_bundle/key_bundle.rs
```diff
@@ -137,6 +137,56 @@ impl KeyBundle for LongTermKeyBundle {
     }
 }
 
+/// Helper method to identify the "latest" (valid and with furthest expiry date) key bundle from a
```
sandreae, line 140: Thanks for the diagrams :+1:

PR #831, opened 2025-10-09
Update key manager methods after upstream fixes, improve automatic rotation

adzialocha:
This PR fixes the breaking changes introduced by https://github.com/p2panda/p2panda/pull/830 and makes pre-key management more robust.

### DIFF
--- a/p2panda-spaces/src/tests.rs
+++ b/p2panda-spaces/src/tests.rs
@@ -2130,7 +2130,7 @@ async fn process_operation_from_expired_member() {
     bob.manager.register_member(&expired_bob).await.unwrap();
 
     // Alice creates a space with Bob.
-    let (_space, messages) = alice
+    let (_space, messages, _events) = alice
         .manager
         .create_space(0, &[(expired_bob.id(), Access::write())])
         .await
@@ -2139,9 +2139,9 @@ async fn process_operation_from_expired_member() {
     // Sleep to make bundle expire.
     thread::sleep(Duration::from_secs(2));
 
-    // Bob processes Alice's "create", but unfortunately Bob's key bundle expired and they can't
-    // decrypt the initial key agreement (X3DH) in the direct message anymore.
-
-    // @TODO: This is failing, but I'm not sure why ..
-    // assert!(bob.manager.process(&messages[0]).await.is_err());
+    // Bob processes Alice's "create group" message.
+    bob.manager.process(&messages[0]).await.unwrap();
+    // Bob processes Alice's "create space", but unfortunately Bob's key bundle expired and they
+    // can't decrypt the initial key agreement (X3DH) in the direct message anymore.
+    assert!(bob.manager.process(&messages[1]).await.is_err());
 }


p2panda-spaces/src/traits/key_store.rs
```diff
@@ -30,11 +30,13 @@ pub trait KeyRegistryStore {
...
-    fn key_manager(&self) -> impl Future<Output = Result<KeyManagerState, Self::Error>>;
+    fn prekey_secrets(&self) -> impl Future<Output = Result<PreKeyBundlesState, Self::Error>>;
 
-    fn set_key_manager(&self, y: &KeyManagerState)
-    -> impl Future<Output = Result<(), Self::Error>>;
+    fn set_prekey_secrets(
+        &self,
+        y: &PreKeyBundlesState,
+    ) -> impl Future<Output = Result<(), Self::Error>>;
 }
```
adzialocha, line 42: The store is now called "KeySecretsStore", both "KeyRegistryStore" and this one will _only_ be concerned with pre-keys and _not_ identity secrets (this is handled independently via the new `Credentials` struct). Like this the separation is hopefully clearer now.

p2panda-spaces/src/credentials.rs
```diff
@@ -8,11 +8,12 @@ use serde::{Deserialize, Serialize};
...
 /// then both keys should be rotated together.
-///  
+///
 /// A peers' identity secret is used for key agreement in p2panda-encryption. Their private key is
 /// used to sign messages and the associated public key is used to identify the peer (eg. for
-/// access control purposes).  
-#[derive(Clone, Debug, Serialize, Deserialize)]
+/// access control purposes).
+#[derive(Debug, Serialize, Deserialize)]
+#[cfg_attr(test, derive(Clone))]
```
adzialocha: I've also removed `Clone` from `Credentials` to make moving sensitive secrets explicit (I could only see clone being required in tests).

p2panda-spaces/src/member.rs
```diff
@@ -5,6 +5,13 @@ use p2panda_encryption::traits::KeyBundle;
 
 use crate::types::ActorId;
 
+// @NOTE(adz) **Security:** This struct does _not_ guarantee if the member's handle / id is
+// authentic. We or applications will need to provide an authentication scheme and validate
+// `Member` before using it anywhere to prevent impersonation attacks.
+//
+// Since we're currently not allowing to construct `Member` from "the outside" (all instances are
+// provided by our API which derived everything from signed messages) I don't see an issue yet, but
+// care will be required as soon as `Member` becomes serializable etc.
```
adzialocha: I think this is okay for now, just adding a note to future-ourselves.

p2panda-spaces/src/identity.rs
```diff
@@ -2,171 +2,143 @@
...
+/// **Warning:** Neither of a peers keys should be rotated individually, this would result in
+/// undefined behavior. Rotating both keys is possible but will result in the loss of access to
+/// existing spaces.
 #[derive(Debug)]
 pub struct IdentityManager<ID, K, M, C> {
     key_store: K,
     credentials: Credentials,
-    pre_key_lifetime: Duration,
-    pre_key_rotate_after: Duration,
-    my_keys_rotated_at: u64,
```
adzialocha, line 41: No explicit state needs to be kept around anymore for managing automatic pre-key rotations as we can rely on the lifetime state in the store instead.

p2panda-spaces/src/identity.rs
```diff
@@ -2,171 +2,143 @@
...
-        if identity_secret != &self.credentials.identity_secret() {
-            return Err(IdentityError::IdentitySecretRotated);
-        }
-        let public_key = self.key_store.public_key();
-        if public_key != self.credentials.public_key() {
-            return Err(IdentityError::PrivateKeyRotated);
-        }
-
-        Ok(())
-    }
```
adzialocha, line 88: I had to remove the validation logic as the credentials are now detached from the pre-key store and I didn't see a quick way to keep that functionality. Since it's rather a "nice to have" I deprioritized bringing it back for now, hope that's okay.

p2panda-spaces/src/identity.rs
```diff
@@ -434,16 +381,22 @@ mod tests {
...
-        {
-            alice_identity_manager.my_keys_rotated_at = 0;
-        }
+        // Override max. lifetime of 90 days (default) with pre-rotation window to force rotation.
+        alice_identity_manager.config.pre_key_rotate_after =
+            Duration::from_secs(60 * 60 * 24 * 1024);
+
+        // Make lifetime of next key longer to "win" over the previous one, in case it is still
+        // considered valid due to a race condition (both keys can be generated "at the same time").
+        alice_identity_manager.config.pre_key_lifetime = Duration::from_secs(60 * 60 * 24 * 2048);
```
adzialocha, line 395: This should fix the race condition where this test sometimes failed or not.

p2panda-spaces/src/tests.rs
```diff
@@ -1786,3 +1786,15 @@ async fn idempotent_api() {
...
+async fn key_store_expired() {
+    let peer = TestPeer::<i32>::new(0).await;
+
+    // Any just created instance will need a pre-key in the beginning.
+    assert!(peer.manager.key_bundle_expired().await.unwrap());
+
+    // Publish a new key bundle with newly generated pre-keys and we should be fine.
+    let _message = peer.manager.key_bundle_message().await.unwrap();
+    assert!(!peer.manager.key_bundle_expired().await.unwrap());
+}
```
adzialocha, line 2050: Added that test here to check if `key_bundle_expired` works. Also it shows how on "fresh init" the manager will _not_ have any pre-keys yet. The other tests "magically" work because we call `me` everywhere which causes key generation.

p2panda-spaces/src/identity.rs
```diff
@@ -2,171 +2,143 @@
...
+    /// Note: Key bundle will be rotated if the latest is reaching it's configured expiry date.
     pub(crate) async fn me(&mut self) -> Result<Member, IdentityError<ID, K, M, C>> {
-        let my_id = self.id();
+        Ok(Member::new(self.id(), self.key_bundle().await?))
+    }
 
+    /// Returns "latest", publishable key bundle of us or automatically generates a new one if
+    /// either nothing was generated yet, if previous bundles expired or are about to be expired
+    /// (given an additional "pessimistic" rotation window).
+    async fn key_bundle(&mut self) -> Result<LongTermKeyBundle, IdentityError<ID, K, M, C>> {
```
adzialocha, line 81: Factored that logic out into it's own method.

Right now there is only two methods which can trigger generating new pre-keys and both of them always return the new material to the user directly: `me` and `key_bundle_message`. We don't generate on init anymore after the refactoring in `p2panda-encryption`. So this is much more clear and explicit now.

p2panda-spaces/src/tests.rs
```diff
@@ -1786,3 +1780,112 @@ async fn idempotent_api() {
...
+        .unwrap();
+
+    // Sleep to make bundle expire.
+    thread::sleep(Duration::from_secs(2));
+
+    // Bob processes Alice's "create", but unfortunately Bob's key bundle expired and they can't
+    // decrypt the initial key agreement (X3DH) in the direct message anymore.
+
+    // @TODO: This is failing, but I'm not sure why ..
+    // assert!(bob.manager.process(&messages[0]).await.is_err());
```
adzialocha: @sandreae Can't figure out yet why this is failing (it should return an error, but it doesn't ..), Bob doesn't even have any valid key material in it's store. I must be doing something wrong ..

sandreae: The first of Alice's messages is a "create group" message which bob can process as it doesn't yet relate to an encrypted space. The second message is the "create space" message which does indeed fail as we expect.

sandreae: (I updated the test :+1: )

p2panda-spaces/src/identity.rs
sandreae: Yes, that's fine :+1:

p2panda-spaces/src/member.rs
sandreae: Yeh, good to keep a not about this around.

PR #833, opened 2025-10-09
Configure endpoint actor and remove router actor

mycognosist:
This PR is primarily concerned with configuring and spawning the endpoint. Configuration parameters are exposed via methods on the provided `NetworkBuilder`.

- Bump iroh to latest (`v0.93.x`)
- Introduce fields for `EndpointConfig`
- Listen for the the initial relay connection in the `endpoint` actor
- Remove the `router` actor
- Instantiate the `router` within the `endpoint` actor
- Hash the protocol ID with the network ID when registering protocols
- Introduce a minimal `NetworkBuilder` API
- Introduce `NetworkId` and `ProtocolId`

## 📋 Checklist

### DIFF
--- a/p2panda-net-next/src/network.rs
+++ b/p2panda-net-next/src/network.rs
@@ -74,7 +74,7 @@ impl NetworkBuilder {
     }
 
     /// Adds a custom protocol for communication between two peers.
-    pub fn protocol(mut self, id: &ProtocolId, handler: impl ProtocolHandler) -> Self {
+    fn _protocol(mut self, id: &ProtocolId, handler: impl ProtocolHandler) -> Self {
         // Hash the protocol ID with the network ID.
         //
         // The hashed ID is what will be registered with the iroh `Endpoint`.


p2panda-net-next/src/actors/endpoint.rs
```diff
@@ -0,0 +1,250 @@
...
+
+#[derive(Debug)]
+// TODO: Remove once used.
+#[allow(dead_code)]
+pub(crate) struct EndpointConfig {
+    pub(crate) bind_ip_v4: Ipv4Addr,
+    pub(crate) bind_port_v4: u16,
+    pub(crate) bind_ip_v6: Ipv6Addr,
+    pub(crate) bind_port_v6: u16,
+    pub(crate) private_key: PrivateKey,
```
adzialocha: Maybe we should make that a separate argument, I think it's good to always explicitly deal with sensitive keys and not hide it in "configs" or "defaults"

p2panda-net-next/src/network.rs
```diff
@@ -0,0 +1,97 @@
...
+        //
+        // The XOR'd ID is what will be registered with the iroh `Endpoint`.
+        let identifier_xor = protocols::protocol_id_xor(id, self.network_id);
+
+        self.network_config
+            .endpoint_config
+            .protocols
+            .insert(identifier_xor, Box::new(handler));
+        self
+    }
```
adzialocha, line 88: Is the endpoint anywhere exposed to the outside (to connect with that custom protocol). I'm wondering what users hope to see in such an interface

mycognosist: Previously we spoke about directly exposing a `connect(node_addr, protocol_id)` method on the `Network` API (this came up around system service and other discussions). But I don't think we considered this as part of the current scope so I'm fine with leaving it out of the external API for now.

p2panda-net-next/src/protocols.rs
```diff
@@ -0,0 +1,35 @@
...
+/// identifier.
+pub type ProtocolId = [u8];
+
+/// Hash the concatenation of the given protocol and network identifiers (using a `+` delimiter).
+pub(crate) fn hash_protocol_id_with_network_id(
+    protocol_id: &ProtocolId,
+    network_id: &NetworkId,
+) -> [u8; 32] {
+    let ids = [protocol_id, b"+", network_id].concat();
+    let hash = blake3::hash(&ids);
```
adzialocha: We can use p2panda_core::Hash here (which uses blake3 internally)

PR #836, opened 2025-10-13
`p2panda-spaces`: Data encryption for groups and multiple devices

sandreae:
This PR introduces the `p2panda-spaces` crate which provides an API for establishing and managing encryption contexts with dynamic groups of actors. The p2panda-encryption [Data Encryption](https://docs.rs/p2panda-encryption/latest/p2panda_encryption/data_scheme/index.html) scheme is used for key agreement and group management is achieved through an integration with [p2panda-auth groups](https://docs.rs/p2panda-auth/latest/p2panda_auth/).

## Features

* Decentralised group key agreement with forward secrecy and encrypted messaging with
  post-compromise security
* Decentralised group management with robust conflict resolution strategies
* Private space identifiers
* Re-use of groups across encryption boundaries
* Nested groups allowing for modelling multi-device profiles
* Generic over message type

## 📋 Checklist

### DIFF
--- a/p2panda-spaces/src/encryption/dgm.rs
+++ b/p2panda-spaces/src/encryption/dgm.rs
@@ -3,8 +3,8 @@
 use std::collections::HashSet;
 use std::convert::Infallible;
 
-use serde::{Deserialize, Serialize};
 use p2panda_encryption::traits::GroupMembership;
+use serde::{Deserialize, Serialize};
 
 use crate::types::{ActorId, OperationId};
 
@@ -22,9 +22,7 @@ pub struct EncryptionMembershipState {
     pub(crate) members: HashSet<ActorId>,
 }
 
-impl GroupMembership<ActorId, OperationId>
-    for EncryptionGroupMembership
-{
+impl GroupMembership<ActorId, OperationId> for EncryptionGroupMembership {
     type State = EncryptionMembershipState;
 
     type Error = Infallible;
--- a/p2panda-spaces/src/manager.rs
+++ b/p2panda-spaces/src/manager.rs
@@ -21,13 +21,13 @@ use crate::message::SpacesArgs;
 use crate::space::{Space, SpaceError};
 use crate::traits::{
     AuthStore, AuthoredMessage, Forge, KeyRegistryStore, KeySecretStore, MessageStore, SpaceId,
-    SpacesStore, SpacesMessage,
+    SpacesMessage, SpacesStore,
 };
 use crate::types::{ActorId, AuthResolver, OperationId};
 use crate::{Config, Credentials};
 
-/// API for creating and managing groups and spaces. 
-/// 
+/// API for creating and managing groups and spaces.
+///
 /// There should be only one manager instance per application. Any messages received from other
 /// instances must be processed on the manager. All methods which mutate state locally return M
 /// message(s) which must be replicated to and processed by other instances. Any action which
--- a/p2panda-spaces/src/group.rs
+++ b/p2panda-spaces/src/group.rs
@@ -19,8 +19,8 @@ use crate::manager::Manager;
 use crate::message::SpacesArgs;
 use crate::traits::SpaceId;
 use crate::traits::{
-    AuthStore, AuthoredMessage, Forge, KeyRegistryStore, KeySecretStore, MessageStore, SpacesStore,
-    SpacesMessage,
+    AuthStore, AuthoredMessage, Forge, KeyRegistryStore, KeySecretStore, MessageStore,
+    SpacesMessage, SpacesStore,
 };
 use crate::types::{
     ActorId, AuthControlMessage, AuthGroup, AuthGroupAction, AuthGroupError, AuthGroupState,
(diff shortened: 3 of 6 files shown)


p2panda-encryption/src/key_bundle/lifetime.rs
```diff
@@ -56,6 +56,39 @@ impl Lifetime {
             Err(LifetimeError::InvalidLifetime)
         }
     }
+
+    /// Returns true if this lifetime with an additional window is valid.
+    ///
+    /// This method can be used to calculate if an lifetime _is about_ to expire, given a defined
```
mycognosist:
```suggestion
    /// This method can be used to calculate if a lifetime _is about_ to expire, given a defined
```

p2panda-spaces/src/encryption/dgm.rs
```diff
@@ -0,0 +1,66 @@
...
+/// p2panda-auth.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct EncryptionGroupMembership {}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct EncryptionMembershipState {
+    pub(crate) members: HashSet<ActorId>,
+}
+
+impl p2panda_encryption::traits::GroupMembership<ActorId, OperationId>
```
mycognosist: Just a minor nitpick but I'd suggest importing with `use p2panda_encryption::traits::GroupMembership;` and then changing this to `impl GroupMembership<..>`.

p2panda-spaces/src/encryption/message.rs
```diff
@@ -0,0 +1,239 @@
...
+            args: encryption_args,
+        }
+    }
+
+    /// Construct an encryption message from a corresponding space message and required additional
+    /// arguments.
+    ///
+    /// This method is required when we receive a space message and associated auth message and we
+    /// want to adjust our local encryption state accordingly. The main requirement is that we
+    /// process our own direct messages (contained in the space message), in many cases the actual
```
mycognosist:
```suggestion
    /// process our own direct messages (contained in the space message); in many cases the actual
```

p2panda-spaces/src/encryption/message.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// encryption control message type and content is redundant as the DGM state is always
+    /// manually replaced with the latest membership state provided by p2panda-auth. The only case
+    /// where it does matter is if we ourselves were added or removed from the group, here we
```
mycognosist:
```suggestion
    /// where it does matter is if we ourselves were added or removed from the group; here we
```

p2panda-spaces/src/manager.rs
```diff
@@ -0,0 +1,612 @@
...
+use crate::message::SpacesArgs;
+use crate::space::{Space, SpaceError};
+use crate::traits::{
+    AuthStore, AuthoredMessage, Forge, KeyRegistryStore, KeySecretStore, MessageStore, SpaceId,
+    SpaceStore, SpacesMessage,
+};
+use crate::types::{ActorId, AuthResolver, OperationId};
+use crate::{Config, Credentials};
+
+/// API for creating and managing groups and spaces. There should be only one manager instance per
```
mycognosist: I'd suggest introducing a newline after "API for creating and managing groups and spaces."

Then there's a nice one-liner in the docs manager module docs next to the `Manager` `struct`. Clicking through to the `Manager` page then reveals the rest of the doc string.

p2panda-spaces/src/manager.rs
```diff
@@ -0,0 +1,612 @@
...
+        message: &M,
+    ) -> Result<(), ManagerError<ID, S, K, F, M, C, RS>> {
+        let manager = self.inner.write().await;
+        manager
+            .store
+            .set_message(&message.id(), message)
+            .await
+            .map_err(ManagerError::MessageStore)?;
+        Ok(())
+    }
```
maackle, line 561: I just came across this. Could you explain by what mechanism messages are supposed to be already persisted, when we are receiving and processing messages from another node? I see the TestForge persists messages so I assume that's what all Forges should do, but that only covers the authoring side, what about on the receiving side? Are receivers just supposed to call `set_message` before processing?

I upgraded p2panda to the latest version of this branch after not doing anything with it for a few weeks, and now I'm getting `MissingAuthMessage` errors, because that message is not persisted, even though it is being processed.

maackle: I wound up having to hold an extra copy of the MessageStore just to manually persist message before processing them, because I couldn't find any other public way to do it

sandreae: Messages are persisted in the ingest processor of `p2panda-stream`, and as we expect that `p2panda-spaces` will ultimately be used mainly as another processor in this pipeline the assumption is that operations are persisted to the store before they get processed here.

Just to add that the `MessageStore` currently defined as a trait interface here in p2panda-spaces will be replaced with `OperationStore` from `p2panda-store`  which makes more sense as this is used across the rest of the stack. For now it's assumed these two stores are equivalent.

sandreae: > I wound up having to hold an extra copy of the MessageStore just to manually persist message before processing them, because I couldn't find any other public way to do it

This sounds like a fine solution for now though, when using p2panda-spaces in a standalone manner.

PR #839, opened 2025-10-16
Introduce ephemeral messaging

mycognosist:
This PR introduces ephemeral messaging by adding two new types: `EphemeralTopicStream` and `EphemeralTopicStreamSubscription`. These types hook into the subscription actor which in turn communicates with the gossip actor to establish the required gossip session(s). Several top-level API methods have been added to expose this ephemeral messaging functionality.

The API looks like this this:

```rust
let node_builder = NetworkBuilder::new(&network_id);
let node = node_builder.build().await?;

let stream = node.ephemeral_stream(&topic_id).await?;

stream.publish(b"I am the nothingness at the centre of creation").await?;

let mut stream_subscription = stream.subscribe().await?;

let msg = stream_subscription.try_recv();

stream.unsubscribe()?;
```

Subscribing to the same `topic_id` multiple times via the top-level API does not result in additional gossip sessions being created. Instead, the underlying channels are simply cloned. This is made possible by using a `tokio` broadcast channel which ensures that all receivers are sent a clone of the inner value.

I've introduced a sketch integration test to show the API. This will require more work once the address book functionality has been merged (an empty `Vec` is currently provided when subscribing to the gossip, meaning that we will never join other peers).

Another thing you'll notice are the `with_suffix()` and `without_suffix()` helper functions around actor names. Since named actors are registered in a global registry with ractor, we run into naming collisions (and panics) if two or more nodes are running in a single process. To get around this, I've introduce a public-key-derived suffix for each actor name. It's not the prettiest of solutions but it works. Once we have all logic in place (through other PRs) we can take a look at alternative approaches (such as passing around `ActorRef`s instead of querying actors by name). 

## 📋 Checklist

### DIFF
--- a/p2panda-net-next/src/actors/subscription.rs
+++ b/p2panda-net-next/src/actors/subscription.rs
@@ -163,10 +163,8 @@ impl Actor for Subscription {
                     if !reply.is_closed() {
                         let _ = reply.send(Some(subscription));
                     }
-                } else {
-                    if !reply.is_closed() {
-                        let _ = reply.send(None);
-                    }
+                } else if !reply.is_closed() {
+                    let _ = reply.send(None);
                 }
             }
             ToSubscription::UnsubscribeEphemeral(topic_id) => {


p2panda-net-next/src/actors/gossip/mod.rs
```diff
@@ -84,9 +85,10 @@ impl Message for ToGossip {}
 pub struct GossipState {
```
sandreae: Is it possible to group some of these mappings together? It's a little hard to see which mappings are interdependent / related when looking at them like this. For example, maybe a GossipSessionState object holding the required channels and actor refs? Sorry, don't have a concrete suggestion yet, but hopefully my general point is clear.

p2panda-net-next/src/actors/subscription.rs
```diff
@@ -25,6 +44,8 @@ pub struct SubscriptionState {
     gossip_actor_failures: u16,
     sync_actor: ActorRef<ToSync>,
     sync_actor_failures: u16,
+    ephemeral_senders: HashMap<TopicId, Sender<ToNetwork>>,
+    gossip_senders: HashMap<TopicId, BroadcastSender<FromNetwork>>,
```
sandreae: Slightly hard to grok the relation between these two maps from the naming, some kind of `to_network_senders` and `from_network_senders` could be better (not sure about the "network" part...).

p2panda-net-next/src/topic_streams.rs
```diff
@@ -0,0 +1,158 @@
...
+    #[error(transparent)]
+    Recv(#[from] RecvError),
+
+    #[error("subscription actor is not available to create topic subscription")]
+    SubscriptionNotAvailable,
+}
+
+/// A handle to an ephemeral messaging stream.
+///
+/// The stream can be used to publish messages or to request a subscription.
```
sandreae, line 49: What's the reasoning behind only providing the subscription receiver when calling `.subscribe()` on the `EphemeralTopicStream`? As both the sender and receiver are already `Clone` I can't immediately see the reason. I actually find the API a little unintuitive as it is now, I would normally expect to call `recv` directly on a "stream" object.

p2panda-net-next/src/topic_streams.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// Subscribes to the stream.
+    ///
+    /// The returned `EphemeralTopicStreamSubscription` provides a means of receiving messages from
+    /// the stream.
+    pub async fn subscribe(
+        &self,
+    ) -> Result<EphemeralTopicStreamSubscription, TopicStreamError<()>> {
+        // Get a reference to the subscription actor.
+        if let Some(subscription_actor) =
```
sandreae: Aha, I didn't know you could "fetch" an actor reference from the registry anywhere in the codebase. That explains the need for a global registry.

p2panda-net-next/src/actors/subscription.rs
```diff
@@ -25,6 +44,8 @@ pub struct SubscriptionState {
...
 }
 
 pub struct Subscription;
```
adzialocha, line 51: follow up: This is not anymore a "subscription actor" but probably more a "stream actor" (tbd), since it manages handles into streams which can be subscribed and published to

p2panda-net-next/src/actors/subscription.rs
```diff
@@ -39,25 +60,37 @@ impl Actor for Subscription {
         myself: ActorRef<Self::Msg>,
         endpoint: Self::Arguments,
     ) -> Result<Self::State, ActorProcessingErr> {
+        let public_key_suffix = to_public_key(endpoint.node_id()).to_hex()[..6].to_string();
+
         // Spawn the gossip actor.
         let (gossip_actor, _) = Actor::spawn_linked(
-            Some("gossip".to_string()),
+            Some(with_suffix("gossip", &public_key_suffix)),
```
adzialocha: follow up: These actor identifiers feel like constants and should maybe be better expressed as such

p2panda-net-next/src/actors/subscription.rs
```diff
@@ -8,14 +8,33 @@
...
 
-pub enum ToSubscription {}
+    /// Return a subscription handle for the given topic ID.
+    ReturnEphemeralSubscription(TopicId, RpcReplyPort<EphemeralTopicStreamSubscription>),
+
+    /// Unsubscribe from an ephemeral stream for the given topic ID.
+    UnsubscribeEphemeral(TopicId),
+}
 
 impl Message for ToSubscription {}
```
adzialocha, line 39: follow up: This is a feature we don't need of `ractor` and can be removed (next to the `cluster` flag)

p2panda-net-next/src/actors/subscription.rs
```diff
@@ -82,9 +115,67 @@ impl Actor for Subscription {
...
+                    // Store the gossip sender. This can be used to create a broadcast receiver
+                    // when the user calls `.subscribe()` on `EphemeralTopicStream`.
+                    state.gossip_senders.insert(topic_id, from_gossip_tx);
+
+                    EphemeralTopicStream::new(topic_id, to_gossip_tx, public_key_suffix)
+                };
+
+                if !reply.is_closed() {
+                    let _ = reply.send(stream);
+                }
```
adzialocha, line 155: Is there a difference between checking before if the channel is closed before sending or just ignoring the error on send?

p2panda-net-next/src/actors/subscription.rs
```diff
@@ -82,9 +115,67 @@ impl Actor for Subscription {
...
+            }
+            ToSubscription::ReturnEphemeralSubscription(topic_id, reply) => {
```
adzialocha: This never replies if there is no entry in `gossip_senders` for this `topic_id`. I assume some code somewhere else might hang here forever now awaiting a reply. Probably you want to return an `Option` here instead and return `None` in that case.

p2panda-net-next/src/actors/supervisor.rs
```diff
@@ -1,43 +1,43 @@
...
-pub struct NetworkState {
+pub struct SupervisorState {
     events_actor: ActorRef<ToEvents>,
     events_actor_failures: u16,
     endpoint_actor: ActorRef<ToEndpoint>,
     address_book_actor: ActorRef<ToAddressBook>,
     address_book_actor_failures: u16,
     discovery_actor: ActorRef<ToDiscovery>,
     discovery_actor_failures: u16,
+    public_key_suffix: String,
```
adzialocha: Maybe we can give this an own type and rename this to something like "Actor Name Suffix" or maybe even "Actors Namespace" because this is what it is used for and that it's using a public key is less important.

The own type helps to give it a doc-string so developers, if they stumble upon this, can quickly check what all of this is about (as it's a bit surprising to see it popping up in an unrelated struct when reading the code).

p2panda-net-next/src/actors/subscription.rs
mycognosist: Yeah these names are awful haha. I'll make some improvements.

p2panda-net-next/src/network.rs
```diff
@@ -99,4 +106,130 @@ impl NetworkBuilder {
...
+        {
+            let actor: ActorRef<ToSubscription> = subscription_actor.into();
+
+            // Ask the subscription actor for an ephemeral stream.
+            let stream = call!(actor, ToSubscription::CreateEphemeralStream, *topic_id)
+                .map_err(|_| NetworkError::StreamCreation)?;
+
+            Ok(stream)
+        } else {
+            Err(NetworkError::StreamCreation)
```
adzialocha: This is a different type of error: I don't think we want to panic here but this is certainly a critical failure if the "subscription actor" is down (it doesn't contain any code which would panic and we consider it a core system of `p2panda-net`).

adzialocha: .. this is different from an error where the subscription actor itself not being able to do work we requested it to do

p2panda-net-next/src/network.rs
```diff
@@ -99,4 +106,130 @@ impl NetworkBuilder {
...
+/// Message received from the network.
+#[derive(Debug, Clone, PartialEq)]
+pub enum FromNetwork {
+    EphemeralMessage {
+        bytes: Vec<u8>,
+        delivered_from: PublicKey,
+    },
+    Message {
+        header: Vec<u8>,
+        payload: Option<Vec<u8>>,
```
adzialocha, line 207: follow up: I'm not sure we want this distinction between "header" and "payload"

p2panda-net-next/src/topic_streams.rs
```diff
@@ -0,0 +1,158 @@
...
+use tokio::sync::mpsc::Sender;
+use tokio::sync::mpsc::error::SendError;
+
+use crate::TopicId;
+use crate::actors::subscription::ToSubscription;
+use crate::network::{FromNetwork, ToNetwork};
+use crate::utils::with_suffix;
+
+#[derive(Debug, Error)]
+pub enum TopicStreamError<T> {
```
adzialocha: nit: Errors are usually the least interesting so good to keep them at the bottom of the file instead

p2panda-net-next/src/topic_streams.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub fn topic_id(&self) -> TopicId {
+        self.topic_id
+    }
+
+    /// Unsubscribes from the ephemeral messaging stream.
+    pub fn unsubscribe(self) -> Result<(), TopicStreamError<()>> {
+        if let Some(subscription_actor) =
+            registry::where_is(with_suffix("subscription", &self.public_key_suffix))
+        {
+            let actor: ActorRef<ToSubscription> = subscription_actor.into();
```
adzialocha, line 121: This is slightly annoying boilerplate which is repeated here and there, it would make sense to factor it into a dedicated method.

adzialocha: nit: It could make sense to use more "guard clauses" or "early returns" to avoid nesting (can be applied to other parts of this PR as well)

p2panda-net-next/src/topic_streams.rs
mycognosist: We haven't yet had our in-depth API design meeting so I went ahead and implemented adz' API suggestion from the network meeting we had a while ago. In that meeting he proposed this clear separation of publishing and subscribing handles (unless I misinterpreted something).

Personally, I like the API we had in the previous version - where a single `.subscribe(topic_id)` returned a tuple of `(tx, rx)` for the stream. If it's not a major blocker, I suggest we keep this as-is for now and then I can revise based on our collective discussions.

p2panda-net-next/src/topic_streams.rs
adzialocha: This silently fails if the actor is not available. Do we want to make that an explicit error?

Similar to my comment somewhere else there's a distinction between an criticial part of the system being down (which should never happen) and the command itself not failing. Don't have a feeling for that right now when and where it makes sense to silence this, but we can discuss that tomorrow.

p2panda-net-next/src/topic_streams.rs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+/// A handle to an ephemeral messaging stream subscription.
+///
+/// The stream can be used to receive messages from the stream.
+pub struct EphemeralTopicStreamSubscription {
```
adzialocha: Maybe a slighly shorter naming: `EphemeralStreamSubscription` (can maybe be applied to other types as well)

p2panda-net-next/src/topic_streams.rs
```diff
@@ -0,0 +1,158 @@
...
+    topic_id: TopicId,
+    from_topic_rx: BroadcastReceiver<FromNetwork>,
+}
+
+// TODO: Implement `BroadcastStream`.
```
adzialocha: What is this about?

p2panda-net-next/src/topic_streams.rs
```diff
@@ -0,0 +1,158 @@
...
+
+impl EphemeralTopicStreamSubscription {
```
adzialocha: follow up: Tokio receivers can be wrapped around a `Stream` (via https://docs.rs/tokio-stream/0.1.17/tokio_stream/wrappers/struct.BroadcastStream.html) which makes them compatible with many other crates which is awesome.

Our types should also implement `Stream`.

p2panda-net-next/src/topic_streams.rs
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    /// Returns the topic ID of the stream.
+    pub fn topic_id(&self) -> TopicId {
+        self.topic_id
+    }
+
+    /// Unsubscribes from the ephemeral messaging stream.
+    pub fn unsubscribe(self) -> Result<(), TopicStreamError<()>> {
```
adzialocha: With ephemeral messaging this is a bit special since "unsubscribing" also means to leave the gossip overlay and when leaving the gossip overlay we can't publish anymore, so I'd rather call this `close` instead.

p2panda-net-next/src/actors/endpoint.rs
```diff
@@ -82,6 +83,8 @@ impl Actor for Endpoint {
     ) -> Result<Self::State, ActorProcessingErr> {
         let (private_key, config) = args;
 
+        let public_key_suffix = private_key.public_key().to_hex()[..6].to_string();
```
adzialocha: Similar to my comment somewhere else: This is special code only required for `ractor` and maybe worth factoring this into a separate method and giving it an explicit place, even if it's used here then afterwards to improve code readability

p2panda-net-next/src/actors/subscription.rs
```diff
@@ -8,14 +8,33 @@
 //! An iroh `Endpoint` is held as part of the internal state of this actor. This allows an
 //! `Endpoint` to be passed into the gossip and sync actors in the event that they need to be
 //! respawned.
```
adzialocha: My feeling is that there's an `Endpoint` instance held here because `iroh-gossip` needs it. For sync, discovery, etc. we'll use actors instead to manage connections and keep the layers nicely separated.

p2panda-net-next/src/actors/subscription.rs
```diff
@@ -39,25 +60,37 @@ impl Actor for Subscription {
...
+
+        let ephemeral_senders = HashMap::new();
+        let gossip_senders = HashMap::new();
 
         let state = SubscriptionState {
             endpoint,
             gossip_actor,
             gossip_actor_failures: 0,
             sync_actor,
             sync_actor_failures: 0,
```
adzialocha, line 91: follow up: I think this is mixing up actor concerns slightly here and this will require the meeting we talked about to clarify and sync up our understanding of how to design actors. My two cents:

"subscription actor" is a worker which manages gossip overlays because it's needed to "do the work" around handling ephemeral streams.

It is _not_ concerned with the health of the gossip and sync actor, this is the role of a "supervisor actor". 

Both the gossip and sync actor exist _all the time_ (one can understand them as "managers") and do not get only temporarily spawned for subscriptions.

p2panda-net-next/src/actors/subscription.rs
```diff
@@ -8,14 +8,33 @@
...
+use crate::utils::{with_suffix, without_suffix};
+use crate::{TopicId, to_public_key};
+
+pub enum ToSubscription {
+    /// Subscribe to the topic ID and return a publishing handle.
+    CreateEphemeralStream(TopicId, RpcReplyPort<EphemeralTopicStream>),
 
-pub enum ToSubscription {}
+    /// Return a subscription handle for the given topic ID.
+    ReturnEphemeralSubscription(TopicId, RpcReplyPort<EphemeralTopicStreamSubscription>),
```
adzialocha: Is this a "get"? Then I think it can just be named "EphemeralSubscription"

p2panda-net-next/src/actors/supervisor.rs
mycognosist: That's a really nice suggestion. Good observation regarding this being a namespacing concern / solution.

p2panda-net-next/src/topic_streams.rs
mycognosist: I disagree on this one 🤪 Types which are defined in a module and appear in the code below should be introduced first - at the top; just like other new types, structs, enums etc. That way, you already know what they are when you read through the code. It often confuses me when I'm reading code which uses an error that hasn't been introduced; then I need to go hunting for it.

For me it's not a "hill I'm willing to die on", but I wanted to at least say that I dislike errors at the bottom.

p2panda-net-next/src/topic_streams.rs
mycognosist: Cool, I like that. `close` communicates more clearly what is happening.

p2panda-net-next/src/topic_streams.rs
mycognosist: This is about exactly what you wrote just below; that we can implement `Broadcast` stream for these channels.

p2panda-net-next/src/topic_streams.rs
adzialocha: For me the most important thing is to understand what a code _does_ and error types usually don't define behaviour, so one would just scroll until the interesting bit comes. But I agree, it's a nitpick and you do you :-)

p2panda-net-next/src/topic_streams.rs
adzialocha: Ah, maybe then you change the TODO to "Implement `Stream` for `BroadcastReceiver`"?

p2panda-net-next/src/topic_streams.rs
adzialocha: Yes, this is another overdue meeting to coordinate! Generally I'm open to both approaches.

Some random thoughts on that high-level API topic:

We will need _some_ distinction between publishing and subscribing and not everyone is happy with using channels (this came from the Reflection feedback), also we should consider if we want this to be consistent with other parts of the API (eventual consistent streams), but maybe here it's overkill.

If we go with `(tx, rx)` I would then not call this method "subscribe" returning a channel to publish.

p2panda-net-next/src/actors/gossip/mod.rs
mycognosist: Took my best shot at improving this (following your suggestion I've split some state into `Sessions`). Hopefully that makes it a bit clearer!

p2panda-net-next/src/actors/subscription.rs
mycognosist: There's not a major difference. I copied this code snippet directly from ractor's codebase when I was getting started. I don't mind either way (using the `is_closed()` check as a shortcut or simply bundling that case with the ignored error).

PR #840, opened 2025-10-22
Establish and accept connections

adzialocha:


p2panda-net-next/src/actors/discovery/discovery.rs
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use ractor::{Actor, ActorProcessingErr, ActorRef};
+
+pub const DISCOVERY: &str = "net.discovery";
```
adzialocha, line 13: I thought it was nice to not repeat the actor name in different places (and break tests when it changes) and have it as a constant defined next to each actor instead.

p2panda-net-next/Cargo.toml
```diff
@@ -4,16 +4,18 @@ version = "0.1.0"
...
 
 [dependencies]
-anyhow = "1.0.97"
 futures-lite = "2.6.1"
 iroh = { version = "0.93.2", default-features = false }
 iroh-base = "0.93.2"
 iroh-gossip = "0.93.1"
+n0-future = "0.3.0"
 p2panda-core = { path = "../p2panda-core", version = "0.4.0" }
-ractor = { version = "0.15.8", features = ["cluster"] }
```
adzialocha, line 12: The `cluster` feature flag is only required if we have "remote" actors (on other machines). This and the connected `Message` trait can be removed.

p2panda-net-next/src/actors/supervisor.rs
```diff
@@ -0,0 +1,247 @@
...
+    events_actor_failures: u16,
+    address_book_actor: ActorRef<ToAddressBook<T>>,
+    address_book_actor_failures: u16,
+    discovery_actor: ActorRef<ToDiscovery>,
+    discovery_actor_failures: u16,
+    endpoint_supervisor: ActorRef<()>,
+    endpoint_supervisor_failures: u16,
+}
+
+pub struct Supervisor<S, T> {
```
adzialocha, line 32: Renamed the `Network` actor to a `Supervisor` (the main one) to follow Erlang-style naming conventions

p2panda-net-next/src/args.rs
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use p2panda_core::PrivateKey;
+
+use crate::config::Config;
+
+#[derive(Clone, Debug)]
+pub struct ApplicationArguments {
```
adzialocha, line 8: Restarting actors always requires to give them the same arguments again, to make that less verbose I thought it's nice to just have an object around we can always re-use.

p2panda-net-next/src/actors/endpoint/iroh.rs
```diff
@@ -0,0 +1,203 @@
...
+        let accept_handle = {
+            let endpoint = endpoint.clone();
+
+            tokio::spawn(async move {
+                loop {
+                    let Some(incoming) = endpoint.accept().await else {
+                        break; // Endpoint is closed.
+                    };
+
+                    if let Some(router_actor) = registry::where_is(IROH_ROUTER.into()) {
```
adzialocha, line 157: Just calling any router globally out of the blue feels very weird to me and hard to reason about what the expected behaviour is when it is _not_ available.

p2panda-net-next/src/actors/endpoint/supervisor.rs
```diff
@@ -0,0 +1,130 @@
...
+    application_args: ApplicationArguments,
+    iroh_actor: ActorRef<ToIroh>,
+    iroh_actor_failures: usize,
+    iroh_router_actor: ActorRef<ToIrohRouter>,
+    iroh_router_actor_failures: usize,
+    connection_manager_actor: ActorRef<ToConnectionManager>,
+    connection_manager_actor_failures: usize,
+}
+
+pub struct EndpointSupervisor;
```
adzialocha, line 19: Looking at Erlang it seems good practice to separate supervisor actors from actors doing actual work, but I don't know how far we can bring that with the `ractor` API as we're loosing the reference to the re-spawned actor then and need to use the (global) registry to call that actor instead from work-actors which is ugly in another way ..

p2panda-net-next/src/actors/endpoint/connection_manager.rs
```diff
@@ -0,0 +1,104 @@
...
+
+                let Ok(connecting) = result else {
+                    // @TODO: connection attempt failed.
+                    return Ok(());
+                };
+
+                let (connection_actor, _) =
+                    Actor::spawn_linked(None, Connection, (connecting,), myself.into()).await?;
+
+                let _ = reply.send(Ok(connection_actor));
```
adzialocha: This reads very crazy to me 😝

p2panda-net-next/src/actors/endpoint/router.rs
```diff
@@ -0,0 +1,124 @@
...
+        match message {
+            ToIrohRouter::AcceptAlpn(protocol_id, protocol_handler) => {
+                let mut protocols = state.protocols.write().await;
+                protocols.insert(protocol_id, protocol_handler);
+            }
+            ToIrohRouter::Incoming(incoming) => {
+                let protocols = state.protocols.clone();
+                state.accepted.spawn(async move {
+                    handle_connection(incoming, protocols).await;
+                });
```
adzialocha, line 83: That can maybe be it's own actor, on the other hand this seems to be totally fine as well

p2panda-net-next/src/actors/gossip/listener.rs
```diff
@@ -48,21 +46,12 @@ impl Actor for GossipListener {
         Ok(state)
     }
 
-    async fn post_start(
-        &self,
-        _myself: ActorRef<Self::Msg>,
-        _state: &mut Self::State,
-    ) -> Result<(), ActorProcessingErr> {
-        Ok(())
-    }
```
adzialocha, line 57: All of these seem to be auto-implemented by `ractor`, feels good to remove them for less boilerplate and visual noise

p2panda-net-next/src/actors/endpoint/supervisor.rs
mycognosist: Hmm I think I understand what you mean. One possible solution is to have a message variant on the worker actor which can be used to receive a passed-in, updated actor reference. You could do that in the supervisor right after calling the (re)spawn logic. The worker actor would receive the message and update it's internal state. I'll think about some other options.

PR #841, opened 2025-10-23
Check "previous" during op ingestion

maackle:
I know you're heavily reworking this part and maybe this will go away anyway, but it looks like this part is missing:

`backlink` was being counted as a dependency, but not `previous`.

## 📋 Checklist

### DIFF
--- a/p2panda-stream/src/operation.rs
+++ b/p2panda-stream/src/operation.rs
@@ -192,6 +192,7 @@ mod tests {
         };
         header.sign(&private_key);
         let header_bytes = header.to_bytes();
+        let hash1 = header.hash();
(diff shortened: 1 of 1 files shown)


maackle:
Hey, thank you for the info! I had not noticed `PartialOrder` before. I assumed that partial ordering was meant to work automatically if `previous` is set, so I looked for the simplest way to make it happen, but that was a wrong assumption and I see it's not missing!

It took me a while to figure out how to hook this up, but it works! One PartialOrder per topic, authors have to process their new operations through the PartialOrder, and recipients of operations have to process and then drain the list of ready operations. Whew!

PR #842, opened 2025-10-27
Discovery traits and implementations for in-memory address book and "random walk" strategy

adzialocha:
Introduces new `p2panda-discovery` crate (will replace the previous one later) which implements the first part of p2panda's new discovery strategy: Random Walk network traversal (later combined with Confidential Discovery Protocol for private exchange of node infos and topics).

`p2panda-discovery` is _not_ anymore a crate of a collection of different discovery strategies but one single concrete algorithm which will be "hard-coded" in `p2panda-net`. More transport-specific strategies like mDNS (Internet Protocol) will be instead handled directly in `-net` using iroh's implementation.

A good intro into this PR is probably the crate-level doc-string.

### DIFF
--- a/p2panda-discovery-next/src/address_book.rs
+++ b/p2panda-discovery-next/src/address_book.rs
@@ -81,7 +81,7 @@ pub trait AddressBookStore<T, ID, N> {
         topics: impl IntoIterator<Item = T>,
     ) -> impl Future<Output = Result<(), Self::Error>>;
 
-    /// Sets the list of "topic identifiers" this node is "interested" in in
+    /// Sets the list of "topic identifiers" this node is "interested" in.
     ///
     /// Topic ids for gossip overlays (used for ephemeral messaging) are usually shared privately
     /// and directly with nodes, this is why implementers usually want to simply overwrite the
--- a/p2panda-discovery-next/src/lib.rs
+++ b/p2panda-discovery-next/src/lib.rs
@@ -70,7 +70,7 @@
 //! ## Inspiration
 //!
 //! To use Private Equality Testing in this "category" of peer-to-peer protocols was (to our
-//! knowledge) first suggested by the [Willow](https://willowprotocol.org/) and we believe that
+//! knowledge) first suggested by [Willow](https://willowprotocol.org/) and we believe that
 //! this is the only way towards ["more fine"](https://newdesigncongress.org/en/pub/this-is-fine/)
 //! peer-to-peer systems.
 //!


p2panda-discovery-next/src/naive.rs
```diff
@@ -0,0 +1,213 @@
...
+    type Error = NaiveDiscoveryError<S, P, T, ID, N>;
+
+    type Message = NaiveDiscoveryMessage<T, ID, N>;
+
+    async fn alice(
+        &self,
+        tx: Sender<Self::Message>,
+        mut rx: Receiver<Self::Message>,
+    ) -> Result<DiscoveryResult<T, ID, N>, Self::Error> {
+        // 1. Alice sends Bob all their topics and topic ids.
```
sandreae, line 61: The lib level description is really great and I get a clear understanding of the core types and features of this crate. However when I read this line and code here I get a little confused as the `T` topic is now introduced whereas in the doc strings a topic was only ever referred to as " a secret, randomly generated hash, similar to a shared symmetric key". 

From my perspective working in p2panda-sync (where the T topic is used) I'm not sure what the reason for sending it here in discovery is. That is maybe for a wider conversation about how the systems fit together though, which we will have soon.

In any case, it would be good to include some explanation in the docs here of what the T topic is and what peers gain from discovering it here.

sandreae: I'm assuming that even thought this is a "naive" discovery protocol to be replaced with PET implementation that the node / topic information which is exchanged will be the same.

p2panda-discovery-next/src/random_walk.rs
```diff
@@ -0,0 +1,383 @@
...
+        let reset = {
+            if previous.is_none() {
+                // Always reset at the beginning to initialise everything and start with
+                // "bootstrap" nodes.
+                true
+            } else if self.state.read().await.unvisited.is_empty() {
+                // We've visited all nodes.
+                true
+            } else {
+                // Flip a coin.
```
sandreae, line 217: :coin: :smile:

p2panda-discovery-next/src/random_walk.rs
```diff
@@ -0,0 +1,383 @@
...
+    }
+}
+
+/// Breadth-first "random walk" peer sampling strategy.
+///
+/// Strategy to traverse a network of nodes of possibly unknown size and shape combined with a
+/// discovery protocol to exchange known nodes.
+///
+/// A random next node is selected to initiate the discovery protocol, starting from a locally
+/// configured set of "bootstrap nodes" if available. The "walker" continues with another random
```
sandreae, line 43: Could be good to explain what differentiates a "bootstrap node" from a "normal node" and clarify if both are included in the random walk algorithm.

p2panda-discovery-next/src/traits.rs
```diff
@@ -0,0 +1,87 @@
...
+        &self,
+        tx: Sender<Self::Message>,
+        rx: Receiver<Self::Message>,
+    ) -> impl Future<Output = Result<DiscoveryResult<T, ID, N>, Self::Error>>;
+
+    fn bob(
+        &self,
+        tx: Sender<Self::Message>,
+        rx: Receiver<Self::Message>,
+    ) -> impl Future<Output = Result<DiscoveryResult<T, ID, N>, Self::Error>>;
```
sandreae, line 45: Is it important to distinguish between initiator (alice) and receiver (bob) in the discovery protocol? With the current implementation? I can't see a need for this (both could immediately send their topic and topic id sets). Maybe it's needed for the PET implementation though.

p2panda-discovery-next/src/traits.rs
```diff
@@ -0,0 +1,87 @@
...
+/// Result containing node information and topics of a session between Alice and Bob running a
+/// discovery protocol.
+#[derive(Clone, Debug)]
+pub struct DiscoveryResult<T, ID, N>
+where
+    N: NodeInfo<ID>,
+{
+    pub remote_node_id: ID,
+    pub node_transport_infos: BTreeMap<ID, N::Transports>,
+    pub node_topics: HashSet<T>,
```
sandreae, line 57: My usual question around the reason for `T` topics to be included here :rofl:

p2panda-discovery-next/src/traits.rs
adzialocha: Yes exactly, it is not always symmetrical. Also there's a semantic difference between starting and accepting a protocol (the accepting side is usually "forced" into the protocol and shouldn't do unnecessary work upfront, in a better implementation the accepting side wants to be lazy and only react when the initiating side can proof some work).

p2panda-discovery-next/src/naive.rs
adzialocha: Agreed that the terminology around topics is still unclear, let's have a meeting on that as you say

p2panda-discovery-next/src/traits.rs
adzialocha: There's also a technical difference on the iroh layer which is good to be aware of: https://docs.rs/iroh/latest/iroh/index.html

> "To keep streams cheap, they are lazily created on the network: only once a sender starts sending data on the stream will the receiver become aware of a stream. This means only calling [Connection::open_bi](https://docs.rs/iroh/latest/iroh/endpoint/struct.Connection.html#method.open_bi) is not sufficient for the corresponding call to [Connection::accept_bi](https://docs.rs/iroh/latest/iroh/endpoint/struct.Connection.html#method.accept_bi) to return. The sender must send data on the stream before the receiver’s [Connection::accept_bi](https://docs.rs/iroh/latest/iroh/endpoint/struct.Connection.html#method.accept_bi) call will return"

p2panda-discovery-next/src/random_walk.rs
adzialocha: Somehow have a hard time explaining "bootstrap nodes" different than in this sentence right now (they are prioritized on the initial first walk), maybe you have an idea how to frame this better?

p2panda-discovery-next/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+//!
+//! It's also worth mentioning that the random walk can be parallelized in applications (run
+//! multiple "walkers" at the same time) and that it is an "ambient" process, constantly running in
+//! the background, informing other parts of the stack about confidentially exchanged sets of
+//! topics and nodes.
+//!
+//! ## Inspiration
+//!
+//! To use Private Equality Testing in this "category" of peer-to-peer protocols was (to our
+//! knowledge) first suggested by the [Willow](https://willowprotocol.org/) and we believe that
```
mycognosist:
```suggestion
//! knowledge) first suggested by [Willow](https://willowprotocol.org/) and we believe that
```

p2panda-discovery-next/src/address_book.rs
```diff
@@ -0,0 +1,470 @@
...
+    ///
+    /// Topics are usually shared privately and directly with nodes, this is why implementers
+    /// usually want to simply overwrite the previous topic set (_not_ extend it).
+    fn set_topics(
+        &self,
+        id: ID,
+        topics: impl IntoIterator<Item = T>,
+    ) -> impl Future<Output = Result<(), Self::Error>>;
+
+    /// Sets the list of "topic identifiers" this node is "interested" in in
```
mycognosist:
```suggestion
    /// Sets the list of "topic identifiers" this node is "interested" in.
```

p2panda-discovery-next/src/naive.rs
```diff
@@ -0,0 +1,213 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
```
mycognosist, line 2: Would be good to add a short module-level doc-string here to explain what the "naive" implementation does (it's not clear just from the name).

adzialocha: This will probably be removed as soon as we provide the "real" PET implementation

p2panda-discovery-next/src/random_walk.rs
sandreae: If there is no difference except for them being chosen by the local node, then I think this description is maybe enough.

PR #843, opened 2025-10-28
Node information and transport-specific address types 

adzialocha:
This PR introduces the specific types (for example `NodeId` and `NodeInfo`) we need for identifying nodes and storing information about them (to connect to them via different transports). It is a first step towards implementing an address book and discovery protocol which takes place here: https://github.com/p2panda/p2panda/pull/842 (using generics).

Transport information to connect to that node is stored in `TransportInformation` and `TransportAddress`. We give it a timestamp to understand which information is the "latest" and a signature for authentication to prevent attacks.

During discovery we want to transmit `TransportInformation` to other nodes, next to the node's identifier. Information if the node is a bootstrap is a _local setting_ and is not transmitted to other nodes. This is because we can't really verify if a node is a bootstrap or not, this is a "trusted" assumption and every node can decide themselves if they want to consider someone a bootstrap or not (that means it gets prioritized during discovery). Later we might want to extend this with actual metrics and prioritize nodes based on this as well.

The address book actor can be removed I believe as it seems to be mostly a wrapper around a database. We can think about it a bit more and see if we identify non-database concerns we need from it, that would then be an argument to bring it into an actor. If we can't find one, I would propose to remove the actor and instead pass an store reference instead.

## 📋 Checklist

### DIFF
--- a/p2panda-net-next/src/addrs.rs
+++ b/p2panda-net-next/src/addrs.rs
@@ -187,7 +187,7 @@ pub struct TransportInfo {
     /// This can be used to find out which information is the "latest".
     pub timestamp: u64,
 
-    /// Signature to proof authenticity of this transport information.
+    /// Signature to prove authenticity of this transport information.
     ///
     /// Other nodes can validate the authenticity by checking this signature against the associated
     /// node id and info.


p2panda-net-next/src/addrs.rs
```diff
@@ -1,106 +1,405 @@
...
-    pub direct_addresses: Vec<SocketAddr>,
-    pub relay_url: Option<RelayUrl>,
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct TransportInfo {
+    /// UNIX timestamp from when this transport information was published.
+    ///
+    /// This can be used to find out which information is the "latest".
+    pub timestamp: u64,
+
+    /// Signature to proof authenticity of this transport information.
```
mycognosist:
```suggestion
    /// Signature to prove authenticity of this transport information.
```

PR #844, opened 2025-10-29
`p2panda-sync`: modular core protocols and live-mode support

sandreae:
Introduces a new crate `p2panda-sync-next` which contains initial components for a re-write of the current `p2panda-sync` crate.

A generic `Protocol` trait for implementing two-party protocols is introduced with implementations of `TopicHandshakeProtocol`, `LogSyncProtocol` and `TopicLogSyncProtocol`. The latter of these is a wrapper around the first two with additional  "live-mode" behaviour after topic handshake and sync have been performed. Metrics containing total bytes sent/to be sent and received are reported on the event stream.

Two store methods have been added to `LogStore` to improve sync protocol performance and lower memory use:
- `get_log_size(..)` : get the total byte size of operations in a specific log from an optional starting point
- `get_log_hashes(..)` : get only the hashes of operations in a log from an optional starting point

## 📋 Checklist

### DIFF
--- a/p2panda-store/src/operations.rs
+++ b/p2panda-store/src/operations.rs
@@ -129,7 +129,6 @@ pub trait LocalLogStore<LogId, Extensions> {
         from: Option<u64>,
     ) -> Result<Option<u64>, Self::Error>;
 
-
     /// Get hashes from an authors' log ordered by sequence number.
     ///
     /// The `from` value will be used as the starting index for log retrieval, if supplied,


p2panda-sync-next/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+pub mod cbor;
+pub mod log_sync;
```
adzialocha: Can be nice to have this behind a feature flag and have that feature activated by default. To help understand what are a) core types and traits b) "features on top of that"

p2panda-sync-next/src/protocols/topic_log_sync.rs
```diff
@@ -0,0 +1,857 @@
+use std::fmt::Debug;
```
adzialocha, line 3: License headers missing (maybe in other places as well)

p2panda-sync-next/src/protocols/topic_log_sync.rs
```diff
@@ -0,0 +1,857 @@
...
+/// If we implement `TopicQuery` to express that we're interested in syncing over a specific chat
+/// group, for example "Chat Group 2" we would implement `TopicLogMap` to give us all append-only
+/// logs of all members inside this group, that is the entries inside logs `A2`, `B2` and `C2`.
+pub trait TopicLogMap<T, L>
+where
+    T: TopicQuery,
+{
+    type Error: Debug;
+
+    fn get(&self, topic: &T) -> impl Future<Output = Result<Logs<L>, Self::Error>>;
```
adzialocha, line 388: A more descriptive method name would help here, `logs_for_topic` or similar?

p2panda-sync-next/src/protocols/log_sync.rs
```diff
@@ -0,0 +1,973 @@
...
+
+/// Sync session status variants carried on log sync session events.
+#[derive(Clone, Debug, PartialEq)]
+pub enum StatusEvent {
+    Started {
+        metrics: Metrics,
+    },
+    Progress {
+        metrics: Metrics,
+    },
```
adzialocha, line 450: Progress sounds like a constant update about the current "progress" but I had a look at the code and it seems more like an event occurring at specific steps in the state machine? Is there a better name for it maybe?

Or maybe it's not required (so far we only need to know about what's going on when the session begins and ends)?

p2panda-sync-next/src/log_sync.rs
```diff
@@ -0,0 +1,973 @@
...
+        metrics: Metrics,
+    },
+}
+
+/// Operations carried on log sync session events.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Operation<E> {
+    pub(crate) header: Header<E>,
+    pub(crate) body: Option<Body>,
+}
```
adzialocha: Is it possible to use `Operation` coming from core? It will be confusing to introduce another, independent type with the same name (and we should in general avoid introducing too many different representations of the same thing, that at least was coming from feedback)

p2panda-sync-next/src/protocols/topic_log_sync.rs
```diff
@@ -0,0 +1,857 @@
...
+/// "Chat group 2 with members A, B and C"
+/// - Log A2
+/// - Log B2
+/// - Log C2
+/// ```
+///
+/// If we implement `TopicQuery` to express that we're interested in syncing over a specific chat
+/// group, for example "Chat Group 2" we would implement `TopicLogMap` to give us all append-only
+/// logs of all members inside this group, that is the entries inside logs `A2`, `B2` and `C2`.
+pub trait TopicLogMap<T, L>
```
adzialocha, line 382: Nice, this looks much simpler now :+1:

p2panda-sync-next/src/traits.rs
```diff
@@ -0,0 +1,60 @@
...
+    type Output;
+    type Error;
+    type Message;
+
+    fn run(
+        self,
+        sink: &mut (impl Sink<Self::Message, Error = Self::Error> + Unpin),
+        stream: &mut (impl Stream<Item = Result<Self::Message, Self::Error>> + Unpin),
+    ) -> impl Future<Output = Result<Self::Output, Self::Error>>;
+}
```
adzialocha: Would that be then part of `p2panda-net` instead? I'm not sure what this is about / for

p2panda-sync-next/src/traits.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// Identify the particular dataset a peer is interested in syncing.
+///
+/// Exactly how this is expressed is left up to the user to decide. During sync the "initiator"
+/// sends their topic query to a remote peer where it is be mapped to their local dataset.
+/// Additional access-control checks can be performed. Once this "handshake" is complete both
```
adzialocha, line 27: Here it mentions "access control" checks? Are we doing this already? Is this part of the next PR?

Cargo.toml
```diff
@@ -11,7 +11,7 @@ members = [
     "p2panda-net-next",
     "p2panda-store",
     "p2panda-stream",
-    "p2panda-sync",
+    "p2panda-sync", "p2panda-sync-next",
```
adzialocha, line 14: Nit: newline

p2panda-sync-next/src/traits.rs
```diff
@@ -0,0 +1,60 @@
+use std::fmt::Debug;
+use std::hash::Hash as StdHash;
+
+use futures::Sink;
+use futures_util::{AsyncRead, AsyncWrite, Stream};
+use serde::{Deserialize, Serialize};
+
+
```
adzialocha: Nit: newline

p2panda-sync-next/src/protocols/log_sync.rs
```diff
@@ -0,0 +1,973 @@
...
+    Data(Operation<E>),
+}
+
+/// Sync session metrics emitted in session event messages.
+#[derive(Clone, Debug, PartialEq, Default)]
+pub struct Metrics {
+    pub total_operations_local: Option<u64>,
+    pub total_operations_remote: Option<u64>,
+    pub total_bytes_local: Option<u64>,
+    pub total_bytes_remote: Option<u64>,
```
adzialocha, line 437: What does the `Option` mean here?

p2panda-sync-next/src/test_utils.rs
```diff
@@ -0,0 +1,72 @@
...
+        timestamp,
+        seq_num,
+        backlink,
+        previous: vec![],
+        extensions: Some(()),
+    };
+    header.sign(private_key);
+    let header_bytes = header.to_bytes();
+    (header, header_bytes)
+}
```
adzialocha, line 272: Soon we can use this: https://github.com/p2panda/p2panda/blob/stream-next/p2panda-core/src/test_utils.rs :-)

p2panda-sync-next/src/protocols/topic_log_sync.rs
```diff
@@ -0,0 +1,857 @@
...
+use thiserror::Error;
+use tokio::sync::broadcast;
+
+pub struct TopicLogSync<T, S, M, L, E> {
+    pub store: S,
+    pub topic_map: M,
+    pub role: Role<T>,
+    pub event_tx: mpsc::Sender<TopicLogSyncEvent<T, E>>,
+    pub live_mode_rx: Option<broadcast::Receiver<LiveModeMessage<E>>>,
+    pub _phantom: PhantomData<(T, L, E)>,
```
adzialocha: nit: T and E is used

p2panda-sync-next/src/topic_log_sync.rs
```diff
@@ -0,0 +1,857 @@
...
+            protocol
+                .run(&mut log_sync_sink, &mut log_sync_stream)
+                .await?;
+        }
+
+        // Enter live mode.
+        if let Some(mut live_mode_rx) = self.live_mode_rx {
+            if let Ok(message) = live_mode_rx.recv().await {
+                match message {
+                    LiveModeMessage::FromSub { header, body } => {
```
adzialocha: What does `FromSub` stand for?

p2panda-sync-next/src/topic_log_sync.rs
```diff
@@ -0,0 +1,857 @@
...
+                        sink.send(TopicLogSyncMessage::Live(header, body)).await?;
+                    }
+                    LiveModeMessage::FromSync { header, body } => {
+                        // @TODO: deduplicate messages.
```
adzialocha: Ah, so this PR doesn't contain any ring buffers yet, right?

adzialocha (commented):
Realized that this PR is still in draft so I'm not sure what will be worked on more or if this is already the "feature set" for this instance.

In any case, here are some initial comments, happy to properly review when it's ready :-)

p2panda-sync-next/src/cbor.rs
```diff
@@ -0,0 +1,291 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Utility methods to encode or decode wire protocol messages in [CBOR] format.
```
adzialocha, line 3: Wondering if this is a new `p2panda-io` crate? We will need the same functionality in the discovery protocol (and also for a gossiping protocol in the future)

p2panda-sync-next/src/protocols/topic_log_sync.rs
```diff
@@ -0,0 +1,857 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// Map raw message sink and stream into topic handshake protocol specific channels.
+#[allow(clippy::complexity)]
+pub(crate) fn topic_channels<'a, T, L, E>(
```
adzialocha, line 183: Maybe move into topic handshake mod

p2panda-sync-next/src/log_sync.rs
```diff
@@ -0,0 +1,973 @@
...
+                        metrics,
+                    };
+                }
+                State::SendPreSyncOrDone {
+                    operations,
+                    mut metrics,
+                } => {
+                    let total_operations = operations.len() as u64;
+                    if total_operations > 0 {
+                        let total_bytes = count_cbor_bytes(&operations) as u64;
```
adzialocha: I think this should actually be a database query (where the header and body size will be stored)

p2panda-sync-next/src/lib.rs
sandreae: Yeh, nice, I'll do that.

p2panda-sync-next/src/protocols/log_sync.rs
sandreae: I'm planning to send progress messages on the event channel on a schedule while receiving/sending operations during sync. I think the name works then?

p2panda-sync-next/src/protocols/log_sync.rs
sandreae: The option means that we don't know the values yet (they are not `0`, but rather unknown).

p2panda-sync-next/src/topic_log_sync.rs
sandreae: No, I'm gunna bring that in with a follow-up PR (it's already implemented).

p2panda-sync-next/src/cbor.rs
sandreae: Interesting, we can follow-up this discussion on our chat (as we started there). In general I'm pro this direction (moving encoding/decoding into one place).

p2panda-sync-next/src/protocols/topic_log_sync.rs
sandreae: This is actually a concern of the "wrapper" protocol rather than the sub-protocols, we're making sure the types from the combined stream of messages fit what the sub-protocol expects. For a different "wrapper" these types would be different. For those reasons I believe it should stay here.

p2panda-sync-next/src/log_sync.rs
sandreae: Nice, I implemented that now, I can pull it out into an own PR to main if we prefer to record changes to the store api there.

p2panda-sync-next/src/traits.rs
sandreae: This is for enforcing consistency across different protocol implementations which can then be "wrapped" by a top-level protocol. There is no encoding/decoding of message streams, that's expected to be taken care of by the wrapper. So the `Protocol` trait wouldn't be used in `p2panda-net` (only internally here), _unless_ we decide to break all decoding/encoding out into `p2panda-io`, then this would become the top level user facing trait for this crate, and be used in `p2panda-net`.

p2panda-sync-next/src/traits.rs
sandreae: No, this isn't part of this PR, will take that once we have the core features back in `p2panda-net`.

p2panda-sync-next/src/topic_log_sync.rs
sandreae: Added doc string.

adzialocha (approved):
So cool you've done the encoding changes so quickly! I think this looks great now, the code is very useful beyond sync, will make future work much easier (also for the discovery part).

For me this is ready to be merged now.

mycognosist (approved):

PR #848, opened 2025-11-04
`p2panda-sync`: introduce `SyncManager` trait and `TopicSyncManager` implementation

sandreae:
Introduce `SyncManager` traits and `TopicSyncManager` implementation.

## 📋 Checklist

### DIFF
--- a/p2panda-sync-next/src/lib.rs
+++ b/p2panda-sync-next/src/lib.rs
@@ -3,7 +3,7 @@
 #[cfg(any(test, feature = "cbor"))]
 pub mod cbor;
 pub mod dedup;
-pub mod manager;
+pub mod managers;
 pub mod protocols;
 pub mod session_topic_map;
 #[cfg(any(test, feature = "test_utils"))]
@@ -11,6 +11,7 @@ pub mod test_utils;
 pub mod traits;
 
 pub use dedup::{DEFAULT_BUFFER_CAPACITY, Dedup};
+pub use managers::topic_sync_manager::TopicSyncManager;
 pub use protocols::{log_sync, topic_handshake, topic_log_sync};
 pub use session_topic_map::SessionTopicMap;
 
--- a/p2panda-sync-next/src/managers/mod.rs
+++ b/p2panda-sync-next/src/managers/mod.rs
@@ -1,0 +1,3 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+pub mod topic_sync_manager;
--- a/p2panda-sync-next/src/managers/topic_sync_manager.rs
+++ b/p2panda-sync-next/src/managers/topic_sync_manager.rs
@@ -256,8 +256,8 @@ mod tests {
     use p2panda_core::Header;
     use p2panda_core::{Body, cbor::encode_cbor};
 
+    use crate::TopicSyncManager;
     use crate::log_sync::{LogSyncEvent, StatusEvent};
-    use crate::manager::TopicSyncManager;
     use crate::test_utils::{LogIdExtension, Peer, TestTopic, TestTopicSyncEvent, run_protocol};
     use crate::topic_handshake::TopicHandshakeEvent;
     use crate::topic_log_sync::TopicLogSyncEvent;
--- a/p2panda-sync-next/src/test_utils.rs
+++ b/p2panda-sync-next/src/test_utils.rs
@@ -13,14 +13,13 @@ use rand::rngs::StdRng;
 use serde::{Deserialize, Serialize};
 use tokio::join;
 
+use crate::TopicSyncManager;
 use crate::log_sync::{LogSyncError, LogSyncEvent, LogSyncMessage, LogSyncProtocol, Logs};
-use crate::manager::TopicSyncManager;
 use crate::topic_log_sync::TopicLogMap;
 use crate::topic_log_sync::{
     LiveModeMessage, Role, TopicLogSync, TopicLogSyncError, TopicLogSyncEvent, TopicLogSyncMessage,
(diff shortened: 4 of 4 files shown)


p2panda-sync-next/src/traits.rs
```diff
@@ -20,6 +22,33 @@ pub trait Protocol {
...
+    fn next_event(&mut self) -> impl Future<Output = Result<Option<Self::Event>, Self::Error>>;
+}
+
+/// Generic interface for configurable types.
+pub trait Configurable {
+    type Error;
+    type Config;
+
+    fn configure(&mut self, config: &Self::Config) -> Result<(), Self::Error>;
+}
```
adzialocha: This feels a bit overly generic while you maybe had a concrete use-case in mind (I think the `Protocol + Configurable<Config = SyncSessionConfig<T>>;` hints on that), why not have the configuration on the concrete implementation instead?

sandreae: The config is created and applied to a sync session in p2panda-net, so it needs to be a known type not limited to only one implementation of `SyncManager` or `Protocol`. However I agree that this is an extremely generic trait for what we need.

sandreae: As we have it now any manager implementation used in p2panda-net needs to produce sessions which can be configured with the concrete `SyncSessionConfig` type exported from `p2panda-sync`.

sandreae: Think this is a reasonable requirement, and I'd be happy with any sensible API that achieves this.

p2panda-sync-next/src/manager.rs
```diff
@@ -0,0 +1,328 @@
...
+        )
+    }
+
+    fn session_handle(
+        &self,
+        session_id: u64,
+    ) -> Option<impl Sink<ToSync, Error = Self::Error> + 'static> {
+        self.session_tx_map.get(&session_id).cloned().map(
+            |sink: mpsc::Sender<LiveModeMessage<E>>| {
+                sink.with(|to_sync| {
```
mycognosist: Ah, cool. This is my first time seeing / noticing `with()`.

p2panda-sync-next/src/manager.rs
```diff
@@ -0,0 +1,328 @@
...
+                    ready({
+                        match to_sync {
+                            ToSync::Payload(bytes) => {
+                                // @TODO(sam): not sure what will be happening at this interface
+                                // yet, this code assumes that bytes are sent from p2panda-net and
+                                // we decode them here as the topic sync protocol expects messages
+                                // to contain decoded types. We could change that so that we send
+                                // bytes to the remote and they decode it. Maybe this is
+                                // preferable to avoid an extra decoding/encoding round.
+                                let (header, body): (Header<E>, Option<Body>) =
```
mycognosist: I'm in favour of avoiding the extra decoding / encoding pass.

p2panda-sync-next/src/traits.rs
adzialocha: Maybe as a "guideline", I think there is two reasons to introduce generics:

1. Nice separation of systems to integrate this in, for example actors
2. Different types being used in different places

I don't see right now how either of this is needed for configurations (as you are using only one concrete configuration type and it will not change in the context of actors either).

adzialocha: Let's make it generic when the need arises and keep it "simple" for now?

p2panda-sync-next/src/manager.rs
```diff
@@ -0,0 +1,598 @@
...
+                    });
+                return Ok(Some(manager_event));
+            }
+            _ => return Ok(Some(manager_event)),
+        };
+
+        if let Some((header, body)) = operation {
+            let Some(topic) = self.session_topic_map.topic(session_id) else {
+                panic!();
+                // TODO: Error("received unexpected operation")
```
sandreae: Need to handle errors here, forgot about that.

p2panda-sync-next/src/manager.rs
```diff
@@ -0,0 +1,598 @@
...
+                // Don't forward messages back to the session they came from.
+                if id == session_id {
+                    continue;
+                }
+
+                // Forward live operation to all concurrent sessions. If they have indeed seen
+                // this operation before they will deduplicate it themselves.
+                let Some(tx) = self.session_topic_map.sender_mut(id) else {
+                    panic!();
+                    // Error("missing session channel")
```
sandreae: Here too.

p2panda-sync-next/src/session_topic_map.rs
```diff
@@ -0,0 +1,207 @@
...
+
+#[derive(Debug)]
+pub struct SessionTopicMap<T, TX> {
+    pub(crate) accepting_sessions: HashSet<u64>,
+    pub(crate) session_tx_map: HashMap<u64, TX>,
+    pub(crate) session_topic_map: HashMap<u64, T>,
+    pub(crate) topic_session_map: HashMap<T, HashSet<u64>>,
+}
+
+impl<T, TX> Default for SessionTopicMap<T, TX> {
```
sandreae, line 15: Needs doc strings.

sandreae:
One general point which is not addressed in this PR is ugliness around decoding operation header and payloads inside the manager and sync protocols. I need to review all message types and see where we can handle only bytes instead of decoded `Header` and `Body` pair. The only concern I have is that in some places we need the header hash in order to perform message deduplication. It might be worth deduplicating using simply a hash over opaque message bytes in order to avoid additional (otherwise unnecessary) decoding of messages.

PR #849, opened 2025-11-05
Address book actor

adzialocha:
## 📋 Checklist

### DIFF
--- a/p2panda-net-next/src/actors/address_book.rs
+++ b/p2panda-net-next/src/actors/address_book.rs
@@ -172,21 +172,17 @@ where
                 //
                 // If a node info already exists, only update the "transports" aspect of it and
                 // keep any other "local" configuration, otherwise create a new "default" node info.
-                let current = state.store.node_info(&node_id).await?;
-                let node_info = match current {
-                    Some(mut current) => current.update_transports(transport_info).map(|_| current),
-                    None => Ok(NodeInfo::new(node_id)),
+                let mut node_info = match state.store.node_info(&node_id).await? {
+                    Some(current) => current,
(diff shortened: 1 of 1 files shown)


p2panda-net-next/src/args.rs
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use p2panda_core::PrivateKey;
+
+#[derive(Clone, Debug)]
+pub struct ApplicationArguments<S> {
```
sandreae, line 6: Nice, I assume these are arguments we expect to pass into many/most root-ish actors/supervisors.

p2panda-net-next/src/actors/address_book.rs
```diff
@@ -1,59 +1,403 @@
...
+            _marker: PhantomData,
+        }
+    }
+}
+
+impl<S, T> ThreadLocalActor for AddressBook<S, T>
+where
+    S: AddressBookStore<T, NodeId, NodeInfo> + Send + 'static,
+    S::Error: StdError + Send + Sync + 'static,
+    // @TODO: Remove "Send" as soon as we're removing the "cluster" feature of ractor.
```
sandreae: Ah, nice, hadn't figured out how we could remove the Send requirement here.

sandreae (approved):
This all looks good to me, the RPC pattern in the address book actor is very clear.

p2panda-net-next/src/actors/address_book.rs
adzialocha: Yeah, it actually came back. As soon as it's used as part of the state it needs to be `Send` again.

PR #851, opened 2025-11-06
Create and accept connections

adzialocha:


p2panda-net-next/src/actors/iroh/endpoint.rs
```diff
@@ -0,0 +1,183 @@
...
+            }
+            ToIrohEndpoint::Connect(node_addr, alpn, reply) => {
+                let mixed_protocol_id =
+                    hash_protocol_id_with_network_id(&alpn, &state.args.network_id);
+
+                // This actor will shut down immediately after the connection was established. The
+                // responsibility to handle the connection object is shifted to the caller from
+                // this point on, so using this actor to reason about the state of the connection
+                // is not possible here.
+                IrohConnection::spawn_linked(
```
mycognosist: Since there's no supervision relationship with the spawned actor, I believe you should be able to call `Actor::spawn(...)` instead and omit the `myself.into()` argument.

mycognosist: I'm a bit confused by the syntax here...we use `Actor::spawn_linked()` everywhere else in the codebase so it seems odd to call `IrohConnection::spawn_linked()` here. But maybe that has something to do with this being a `ThreadLocalActor`? I haven't worked with that type before so I may be missing something. If it's possible to have a unified approach (which matches the rest of the code), that would be best. If not, that's also fine - I just have to read some more to get used to this approach.

p2panda-net-next/src/actors/iroh/endpoint.rs
```diff
@@ -0,0 +1,183 @@
...
+                        reply,
+                    },
+                    myself.into(),
+                    state.worker_pool.clone(),
+                )
+                .await?;
+            }
+            ToIrohEndpoint::Incoming(incoming) => {
+                // This actor will run as long as the protocol session.
+                IrohConnection::spawn_linked(
```
mycognosist: Same comment as above: if there's no supervision then use `spawn()`.

p2panda-net-next/src/actors/iroh/endpoint.rs
adzialocha: > Since there's no supervision relationship with the spawned actor

Yes, we can remove the "linking" here for now, though, it's not necessarily only because of "supervision". We can use the events returned back to report to the "events" actor for example which I think it's better to do in the parent "worker factory" (endpoint being one over "connection workers") than inside the connection itself (it will get very noisy / much boilerplate to send something to "events" everytime a connection fails for some reason etc.

Since this PR doesn't really treat connection actors as actual connection "sessions", except when "accepting" I agree it's not necessary to link, but if we bring more of the session logic into it (which would be great in the future), I think we want to have that back again.

> I'm a bit confused by the syntax here

Yes, the `ThreadLocalActor` have a slightly different syntax here as far as I understand.

PR #852, opened 2025-11-06
Discovery manager, walker and session actors

adzialocha:


p2panda-net-next/src/actors/discovery/manager.rs
```diff
@@ -0,0 +1,236 @@
...
+                self.manager_ref.clone(),
+                DiscoverySessionArguments::Accept { connection },
+            ),
+            self.manager_ref.clone().into(),
+            self.pool.clone(),
+        )
+        .await
+        .map_err(|err| iroh::protocol::AcceptError::from_err(err))?;
+
+        // Wait until discovery session ended (failed or successful).
```
sandreae: What's the reason for waiting here until the discovery session has completed? From reading the iroh docs I can see it mentions that the connection is be dropped when `accept` completes, however that it is kept alive if a clone exists, which is the case for us here.

Just interested if this isn't actually the case or if there's another reason to await here.

adzialocha: I think this came from my thinking that it would be nice to manage the connection elsewhere, but for now we can just copy the object around

p2panda-net-next/src/actors/discovery/manager.rs
```diff
@@ -0,0 +1,507 @@
...
+
+#[derive(Clone, Debug, Default)]
+pub struct DiscoveryMetrics {
+    /// Failed discovery sessions.
+    pub failed_discovery_sessions: usize,
+
+    /// Successful discovery sessions.
+    pub successful_discovery_sessions: usize,
+
+    /// Number of discovered transport infos which we're actually new for us.
```
mycognosist:
```suggestion
    /// Number of discovered transport infos which were actually new for us.
```

p2panda-net-next/src/actors/iroh/endpoint.rs
```diff
@@ -141,6 +159,19 @@ impl ThreadLocalActor for IrohEndpoint {
                     .bind()
                     .await?;
 
+                // Watch for changes of our own endpoint address.
```
mycognosist, line 162: We may want to split this listener out at some point, since the address updates will need to be communicated to several sub-systems (for example, they will be used to notify the user via eventually consistent topic streams).

mycognosist (approved):

p2panda-net-next/src/actors/iroh/endpoint.rs
adzialocha: Why would the eventual consistent topic stream (sync) be interested in a change of our own iroh address?

PR #854, opened 2025-11-07
Use non-send "local" actors

adzialocha:
Making (almost) all actors Non-Send by using ractor's https://docs.rs/ractor/latest/ractor/thread_local/trait.ThreadLocalActor.html

This results in the following structure:

```
- THREAD
    - supervisor
        - address book
        - events
        - endpoint supervisor
            - iroh endpoint
                - THREAD
                    - iroh connection
            - discovery manager
                - THREAD
                    - discovery session
                    - discovery walker
            - stream supervisor
                    - sync manager
                        - THREAD
                            - ...
                    - stream
                    - gossip (non-local Actor*)
                        - THREAD
                            - gossip session
                            - gossip listener
                            - gossip receiver
                            - ...
```

### DIFF
--- a/p2panda-net-next/src/actors/endpoint_supervisor.rs
+++ b/p2panda-net-next/src/actors/endpoint_supervisor.rs
@@ -138,6 +138,9 @@ impl ThreadLocalActor for EndpointSupervisor {
                         // 2. Respawn the iroh endpoint actor
                         // 3. Respawn the stream supervisor and discovery actors
                         state
+                            .stream_supervisor
+                            .stop(Some("{IROH_ENDPOINT} actor failed".to_string()));
+                        state
                             .discovery_manager_actor
                             .stop(Some("{IROH_ENDPOINT} actor failed".to_string()));
 
@@ -153,17 +156,6 @@ impl ThreadLocalActor for EndpointSupervisor {
                         state.iroh_endpoint_actor_failures += 1;
                         state.iroh_endpoint_actor = iroh_endpoint_actor;
 
-                        // Respawn the stream supervisor.
-                        let (stream_supervisor, _) = StreamSupervisor::spawn_linked(
-                            Some(with_namespace(STREAM_SUPERVISOR, &state.actor_namespace)),
-                            state.args.clone(),
-                            myself.clone().into(),
-                            state.args.root_thread_pool.clone(),
-                        )
-                        .await?;
-
-                        state.stream_supervisor = stream_supervisor;
-
                         // Respawn the discovery manager actor.
                         let (discovery_manager_actor, _) = Discovery::spawn_linked(
                             Some(with_namespace(DISCOVERY, &state.actor_namespace)),
@@ -174,6 +166,16 @@ impl ThreadLocalActor for EndpointSupervisor {
                         .await?;
 
                         state.discovery_manager_actor = discovery_manager_actor;
+                        // Respawn the stream supervisor.
+                        let (stream_supervisor, _) = StreamSupervisor::spawn_linked(
+                            Some(with_namespace(STREAM_SUPERVISOR, &state.actor_namespace)),
+                            state.args.clone(),
+                            myself.clone().into(),
+                            state.args.root_thread_pool.clone(),
+                        )
+                        .await?;
+
+                        state.stream_supervisor = stream_supervisor;
                     } else if name == with_namespace(DISCOVERY, &state.actor_namespace) {
                         warn!(
                             "{ENDPOINT_SUPERVISOR} actor: {DISCOVERY} actor failed: {}",


p2panda-net-next/src/actors/gossip/receiver.rs
```diff
@@ -20,36 +21,31 @@ pub enum ToGossipReceiver {
...
-    session: ActorRef<ToGossipSession>,
-}
-
-impl GossipReceiver {
-    pub fn new(session: ActorRef<ToGossipSession>) -> Self {
-        Self { session }
-    }
-}
+#[derive(Default)]
+pub struct GossipReceiver;
```
adzialocha, line 28: To use `ThreadLocalActor` `Self` needs to implement `Default`, this is why I've needed to move the `ActorRef` into the state instead, which also feels like a good place.

adzialocha: Needed to do the same for the other actors as well

p2panda-net-next/src/actors/gossip/mod.rs
```diff
@@ -94,14 +96,15 @@ pub struct GossipState {
     sessions: Sessions,
     neighbours: HashMap<TopicId, HashSet<PublicKey>>,
     topic_delivery_scopes: HashMap<TopicId, Vec<IrohDeliveryScope>>,
+    gossip_thread_pool: ThreadLocalActorSpawner,
 }
 
+#[derive(Default)]
 pub struct Gossip;
 
 impl Actor for Gossip {
```
adzialocha: `Gossip` can _not_ implement `ThreadLocalActor` as otherwise the test will fail (see comment there)

p2panda-net-next/src/actors/gossip/mod.rs
```diff
@@ -375,10 +385,10 @@ impl Actor for Gossip {
...
 
                     // Drop all state associated with the failed gossip session.
-                    let _ = state.sessions.sessions_by_topic_id.remove(&topic_id);
-                    let _ = state.sessions.from_gossip_senders.remove(&topic_id);
-                    let _ = state.sessions.gossip_joined_senders.remove(&actor_id);
-                    let _ = state.neighbours.remove(&topic_id);
+                    state.sessions.sessions_by_topic_id.remove(&topic_id);
+                    state.sessions.from_gossip_senders.remove(&topic_id);
+                    state.sessions.gossip_joined_senders.remove(&actor_id);
+                    state.neighbours.remove(&topic_id);
```
adzialocha, line 399: The `let _` reads a bit as if we're silencing a `Result`, so I removed it here. If it did something I didn't see, let me know

p2panda-net-next/src/actors/gossip/mod.rs
```diff
@@ -534,7 +546,7 @@ mod tests {
         let Ok(ant_gossip_actor_result) = ant_supervisor_rx.await else {
             panic!("expected result from gossip actor")
         };
-        let ActorResult::Terminated(state, _reason) = ant_gossip_actor_result else {
+        let ActorResult::Terminated(state, reason) = ant_gossip_actor_result else {
             panic!("expected clean termination of gossip actor")
         };
         let Some(mut boxed_state) = state else {
```
adzialocha: The returned state when an actor got killed will _always_ be `None` when using local actors. The reason is that they can't send the state across thread boundaries. See here: https://github.com/slawlor/ractor/blob/59f3354f12091360b68f947a90763973524bad62/ractor/src/thread_local/inner.rs#L337-L338

For this reason I've decided to keep `Gossip` a non-local actor for now. We will need to find another way to test the actor's state on termination then ..

p2panda-net-next/src/args.rs
```diff
@@ -50,6 +52,7 @@ impl ArgsBuilder {
             public_key: private_key.public_key(),
             private_key,
             iroh_config: self.iroh_config.unwrap_or_default(),
+            root_thread_pool: ThreadLocalActorSpawner::new(),
```
adzialocha, line 55: This thread pool is currently used for _all_ actors, except `Gossip` (see comment in test) and all "workers", like `GossipSession` and iroh connections - later the discovery- and sync equivalents.

p2panda-net-next/src/actors/endpoint_supervisor.rs
```diff
@@ -122,49 +136,79 @@ impl Actor for EndpointSupervisor {
...
+                        let (iroh_endpoint_actor, _) = IrohEndpoint::spawn_linked(
+                            Some(with_namespace(IROH_ENDPOINT, &state.actor_namespace)),
+                            state.args.clone(),
+                            myself.clone().into(),
+                            state.args.root_thread_pool.clone(),
+                        )
+                        .await?;
+
+                        state.iroh_endpoint_actor_failures += 1;
+                        state.iroh_endpoint_actor = iroh_endpoint_actor;
```
adzialocha, line 157: I've brought in restarting the iroh endpoint actor

p2panda-net-next/src/actors/endpoint_supervisor.rs
```diff
@@ -122,49 +136,79 @@ impl Actor for EndpointSupervisor {
...
+                        let (discovery_manager_actor, _) = Discovery::spawn_linked(
                             Some(with_namespace(DISCOVERY, &state.actor_namespace)),
-                            Discovery {},
                             (),
                             myself.clone().into(),
+                            state.args.root_thread_pool.clone(),
                         )
                         .await?;
 
-                        state.discovery_actor_failures += 1;
```
adzialocha, line 149: Counting up the discovery actor failures seems to be an copy & paste bug

p2panda-net-next/src/actors/endpoint_supervisor.rs
```diff
@@ -90,8 +103,9 @@ impl Actor for EndpointSupervisor {
         let reason = Some("endpoint supervisor is shutting down".to_string());
 
         // Stop all the actors which are directly supervised by this actor.
-        state.endpoint_actor.stop(reason.clone());
-        state.discovery_actor.stop(reason);
+        state.iroh_endpoint_actor.stop(reason.clone());
+        state.discovery_manager_actor.stop(reason.clone());
+        state.stream_supervisor.stop(reason);
```
adzialocha, line 107: I've added shutting down the stream supervisor if the endpoint supervisor goes down

p2panda-net-next/src/actors/endpoint_supervisor.rs
```diff
@@ -122,49 +136,68 @@ impl Actor for EndpointSupervisor {
...
-                            Discovery {},
                             (),
                             myself.clone().into(),
+                            state.args.root_thread_pool.clone(),
+                        )
+                        .await?;
+
+                        state.discovery_manager_actor_failures += 1;
+                        state.discovery_manager_actor = discovery_manager_actor;
+                    } else if name == with_namespace(STREAM_SUPERVISOR, &state.actor_namespace) {
```
adzialocha, line 196: I've added a supervision logic to restart the stream supervisor, is this correct?

p2panda-net-next/src/actors/endpoint_supervisor.rs
```diff
@@ -2,16 +2,24 @@
...
-//! This supervisor monitors the health of the endpoint actor, as well as the stream and discovery
-//! actors. If the endpoint actor fails, all child actors of the endpoint supervisor are respawned
-//! (including the stream and discovery actors); this is necessary because stream and discovery
-//! are indirectly reliant on a functioning endpoint actor. If either the stream or discovery actors
-//! fail in isolation, they are simply respawned in a one-for-one manner.
+//! ```plain
+//! - "Endpoint" Supervisor
+//!     - "Stream" Supervisor
+//!     - "Iroh Endpoint" Actor
+//!     - "Discovery Manager" Actor
```
mycognosist: Reordering to reflect the spawn order.
```suggestion
//!     - "Iroh Endpoint" Actor
//!     - "Discovery Manager" Actor
//!     - "Stream" Supervisor
```

p2panda-net-next/src/actors/supervisor.rs
```diff
@@ -2,8 +2,15 @@
...
-//! The root of the entire system supervision tree; it's only role is to spawn and
-//! supervise other actors.
+//! The root of the entire system supervision tree; it's only role is to spawn and supervise other
+//! actors.
+//!
+//! ```plain
+//! - "Root" Supervisor
+//!     - "Endpoint" Supervisor
+//!     - "Address Book" Actor
+//!     - "Events" Actor
```
mycognosist: Spawn order.
```suggestion
//!     - "Events" Actor
//!     - "Address Book" Actor
//!     - "Endpoint" Supervisor
```

p2panda-net-next/src/actors/endpoint_supervisor.rs
```diff
@@ -122,49 +135,68 @@ impl Actor for EndpointSupervisor {
                         // If the endpoint actor fails then we need to:
                         //
                         // 1. Stop the stream supervisor and discovery actors
-                        // 2. Respawn the endpoint actor
+                        // 2. Respawn the iroh endpoint actor
                         // 3. Respawn the stream supervisor and discovery actors
```
adzialocha, line 139: We're not respawning the stream supervisor here, is that missing @mycognosist ?

adzialocha: I've added that now, lmk if that was incorrect

p2panda-net-next/src/actors/gossip/mod.rs
```diff
@@ -767,6 +782,9 @@ mod tests {
         let (_bat_to_gossip, bat_from_gossip) =
             call!(bat_gossip_actor, ToGossip::Subscribe, topic_id, bat_peers).unwrap();
 
+        // Briefly sleep to allow overlay to form.
+        sleep(Duration::from_millis(100)).await;
+
```
adzialocha, line 787: Added some sleeps here and there after "subscribe" (as in the other tests) as some of the tests failed _sometimes_. This seems to fix it

PR #858, opened 2025-11-12
Ephemeral and eventually consistent streams

mycognosist:
This PR introduces the types associated with ephemeral and eventually consistent streams. It also introduces actors which coordinate with several sub-systems to provide stream functionality.

The streams actor is now split in two: one for ephemeral and one for eventually consistent.

Checks are failing because we're using some types which are not yet available on the current branch; these are being introduced with the `SyncManager` work by @sandreae .

## 📋 Checklist

### DIFF
--- a/p2panda-net-next/src/network.rs
+++ b/p2panda-net-next/src/network.rs
@@ -10,16 +10,21 @@ use p2panda_core::{PrivateKey, PublicKey};
 use p2panda_discovery::address_book::AddressBookStore;
 use ractor::errors::SpawnErr;
 use ractor::thread_local::{ThreadLocalActor, ThreadLocalActorSpawner};
-use ractor::{call, registry, ActorRef};
+use ractor::{ActorRef, call, registry};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio::task::JoinHandle;
 
-use crate::actors::streams::ephemeral::{ToEphemeralStreams, EPHEMERAL_STREAMS};
-use crate::actors::supervisor::{Supervisor, SUPERVISOR};
-use crate::actors::{generate_actor_namespace, with_namespace, ActorNamespace};
+use crate::actors::streams::ephemeral::{EPHEMERAL_STREAMS, ToEphemeralStreams};
+use crate::actors::streams::eventually_consistent::{
+    EVENTUALLY_CONSISTENT_STREAMS, ToEventuallyConsistentStreams,
+};
+use crate::actors::supervisor::{SUPERVISOR, Supervisor};
+use crate::actors::{ActorNamespace, generate_actor_namespace, with_namespace};
 use crate::args::{ApplicationArguments, ArgsBuilder};
+use crate::streams::StreamError;
 use crate::streams::ephemeral::EphemeralStream;
+use crate::streams::eventually_consistent::EventuallyConsistentStream;
 use crate::{NetworkId, NodeId, NodeInfo, TopicId};
 
 /// Builds an overlay network for eventually-consistent pub/sub.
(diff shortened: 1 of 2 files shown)


p2panda-net-next/src/actors/stream_supervisor.rs
```diff
@@ -2,14 +2,14 @@
 
 //! Stream supervisor actor.
 //!
-//! This actor is responsible for spawning the sync, gossip and stream actors. It also performs
+//! This actor is responsible for spawning the gossip and streams actors. It also performs
 //! supervision of the spawned actors, restarting them in the event of failure.
 //!
 //! ```plain
 //! - "Stream" Supervisor
-//!     - "Sync Manager" Actor
```
mycognosist, line 10: Sync manager actors are now spawned as-needed by the eventually consistent streams actor (ie. each time that actor receives a `Create` message).

p2panda-net-next/src/actors/streams/ephemeral.rs
```diff
@@ -0,0 +1,142 @@
...
+    Create(TopicId, RpcReplyPort<EphemeralStream>),
+
+    /// Return an ephemeral subscription handle for the given topic ID.
+    Subscribe(TopicId, RpcReplyPort<Option<EphemeralSubscription>>),
+
+    /// Close all ephemeral streams for the given topic ID.
+    Close(TopicId),
+
+    /// Return `true` if there are any active ephemeral streams for the given topic ID.
+    IsActive(TopicId, RpcReplyPort<bool>),
```
mycognosist, line 37: I realised that we only want to trigger a gossip unsubscribe from the eventually consistent streams actor if there isn't any active _ephemeral_ stream. This allows us to check for active ephemeral streams when an eventually consistent stream is closed and when a sync manager actor fails.

p2panda-net-next/src/actors/streams/eventually_consistent.rs
```diff
@@ -0,0 +1,342 @@
...
+                {
+                    sync_manager_actor.send_message(ToSyncManager::Close(node_id, topic_id))?;
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn handle_supervisor_evt(
```
mycognosist, line 290: We handle supervision events for the sync manager actors here. This replicates the approach we take in the root gossip actor (which supervises the underlying gossip session actors).

p2panda-net-next/src/streams/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+    #[error(transparent)]
+    Send(#[from] SendError<T>),
+
+    #[error(transparent)]
+    Recv(#[from] RecvError),
+
+    #[error(transparent)]
+    TryRecv(#[from] TryRecvError),
+
+    #[error("failed to subscribe to topic {0:?} due to system error")]
```
mycognosist, line 26: I'd love some feedback on these error types around streams...I've struggled to arrive at clear semantics for the user.

p2panda-net-next/src/network.rs
```diff
@@ -111,15 +116,12 @@ impl NetworkBuilder {
...
 
 #[derive(Debug, Error)]
-pub enum NetworkError {
-    #[error("failed to create topic stream")]
-    StreamCreation,
-
-    #[error("failed to call {0} actor; it may be in the process of restarting")]
-    ActorNotFound(String),
-
+pub enum NetworkError<T> {
```
mycognosist, line 131: This needs work! I've just put some placeholder variants here for now; we can improve things during a collective API session ^_^

adzialocha (approved):

PR #860, opened 2025-11-17
Join gossip with nodes from the address book

mycognosist:
Here we introduce a gossip healer actor for rejoining gossip overlays whenever new node info is discovered for a topic.

Any time we create a new gossip session (could be initiated by ephemeral or eventually consistent stream creation) we spawn a healer actor: it subscribes to the address book for node info events associated with the relevant topic and then listens on the receiver - when an event comes in it forwards it up to the session as a `JoinPeers` event.

When the stream is closed, the session is killed and the healer goes away.

### DIFF
--- a/p2panda-net-next/src/actors/gossip/healer.rs
+++ b/p2panda-net-next/src/actors/gossip/healer.rs
@@ -2,15 +2,17 @@
 
 //! Subscribe to address book updates and rejoin the gossip with the given nodes if we're actively
 //! interested in the associated topic IDs.
+use p2panda_discovery::address_book::NodeInfo as _;
 use ractor::thread_local::ThreadLocalActor;
 use ractor::{ActorProcessingErr, ActorRef, call, registry};
-use tokio::sync::mpsc::Receiver;
-use tracing::warn;
+use tokio::sync::broadcast::Receiver;
+use tracing::debug;
 
 use crate::TopicId;
-use crate::actors::address_book::{ADDRESS_BOOK, ToAddressBook};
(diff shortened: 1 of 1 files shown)


p2panda-net-next/src/actors/gossip/healer.rs
```diff
@@ -0,0 +1,127 @@
...
+        match message {
+            ToGossipHealer::SubscribeToAddressBook(topic_id) => {
+                if let Some(address_book_actor) =
+                    registry::where_is(with_namespace(ADDRESS_BOOK, &state.actor_namespace))
+                {
+                    let actor: ActorRef<ToAddressBook<()>> = address_book_actor.into();
+
+                    // TODO: Uncomment this once merged with the latest address book PR.
+                    /*
+                    let receiver = call!(actor, ToAddressBook::SubscribeTopicChanges, topic_id)
```
mycognosist, line 78: Subscribe to address book changes so we can get a receiver.

p2panda-net-next/src/actors/gossip/healer.rs
```diff
@@ -0,0 +1,127 @@
...
+
+                    // Invoke the handler to wait for the first event on the receiver.
+                    let _ = myself.cast(ToGossipHealer::WaitForEvent);
+                } else {
+                    panic!("address book actor unavailable")
+                };
+            }
+            ToGossipHealer::WaitForEvent => {
+                if let Some(receiver) = &mut state.receiver {
+                    match receiver.recv().await {
```
mycognosist, line 90: Match on received events and notify the gossip session (parent actor) that we need to (re)join the overlay with these latest nodes.

PR #877, opened 2025-11-21
Improvements to sync poller actor with tests

sandreae:
- loop over `WaitForMessage` calls in handle with await point
- spawn poller as unsupervised actor
- e2e sync test using `TopicSyncManager`


p2panda-net-next/src/actors/sync/poller.rs
```diff
@@ -71,7 +72,8 @@ where
             state.sender.send(event)?;
         }
 
-        // The stream closed, which is caused by the sync session terminated.
+        tokio::time::sleep(Duration::from_millis(20)).await;
+        let _ = myself.cast(ToSyncPoller::WaitForMessage);
```
sandreae: @adzialocha this is the point where I've added a sleep.

PR #879, opened 2025-11-23
Types, trait and API follow-ups of sync manager integration

sandreae:
- `publish()` on eventually consistent stream takes sync protocol concrete message type (`Operation`)
- flatten sync event types
- add trait requirements from `p2panda-net` onto `p2panda-sync` traits
- remove generics from error types to avoid explosion of trait requirements
- wait for the sync poller queue to be drained after a sync manager has been terminated

### DIFF
--- a/p2panda-net-next/src/actors/streams/eventually_consistent.rs
+++ b/p2panda-net-next/src/actors/streams/eventually_consistent.rs
@@ -360,11 +360,6 @@ where
                 // Inform address book about removed topic.
                 state.update_address_book();
 
-                // @TODO(sam): If we drop the sync manager here we lose some final messages coming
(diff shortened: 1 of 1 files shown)


p2panda-net-next/src/actors/sync/manager.rs
```diff
@@ -124,10 +129,26 @@ where
...
             pool,
         })
     }
 
+    async fn post_stop(
+        &self,
+        _myself: ActorRef<Self::Msg>,
+        state: &mut Self::State,
+    ) -> Result<(), ActorProcessingErr> {
+        // Drain the sync poller to ensure that all sync session messages are forwarded before it
```
mycognosist, line 142: @sandreae This was the change required to prevent sync events from being dropped.

PR #883, opened 2025-11-25
Non-blocking manager event streams

sandreae:
This PR introduces a refactor of the topic sync manager to remove the need for a `Mutex` (and blocking behaviour) around the internal state.

It also introduces a few other minor changes, such as supervision logging for the sync poller actor and logging in the topic log sync protocol.

### DIFF
--- a/p2panda-sync-next/src/protocols/topic_log_sync.rs
+++ b/p2panda-sync-next/src/protocols/topic_log_sync.rs
@@ -12,6 +12,7 @@ use p2panda_store::{LogId, LogStore, OperationStore};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio::sync::broadcast;
+use tracing::{debug, warn};
 
 use crate::log_sync::{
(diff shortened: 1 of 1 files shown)


p2panda-sync-next/src/managers/topic_sync_manager.rs
```diff
@@ -144,38 +166,86 @@ where
...
-        let inner = self.inner.lock().await;
-        inner.session_topic_map.sender(session_id).map(|tx| {
+        self.session_topic_map.sender(session_id).map(|tx| {
             Box::pin(tx.clone().with(map_fn))
                 as Pin<Box<dyn Sink<ToSync<Operation<E>>, Error = Self::Error>>>
         })
     }
 
-    fn subscribe(&self) -> impl Stream<Item = FromSync<Self::Event>> + Send + Unpin + 'static {
+    fn subscribe(&mut self) -> impl Stream<Item = FromSync<Self::Event>> + Send + Unpin + 'static {
```
adzialocha: I can imagine the code would benefit from being easier to read and having less lines if we would skip the `Stream` implementation part and just go for channels instead.

PR #888, opened 2025-11-26
Reset backoff when subscribing to topic

adzialocha:


p2panda-net-next/src/actors/address_book/address_book.rs
```diff
@@ -310,6 +326,15 @@ where
...
+                let topics = state.topics_for_node(&node_id).await?;
+                let rx = state.node_topics_watchers.subscribe(
+                    node_id,
+                    updates_only,
+                    WatchedNodeTopics::from_topics(node_id, topics),
+                );
+                let _ = reply.send(rx);
+            }
 
             // Mostly a wrapper around the store ..
```
mycognosist:
```suggestion
            // Mostly a wrapper around the store.
```

p2panda-net-next/src/actors/discovery/manager.rs
```diff
@@ -50,6 +54,11 @@ pub enum ToDiscoveryManager {
         Box<dyn StdError + Send + Sync + 'static>,
     ),
 
+    /// Reset backoff logic of all walkers and make them start from the bootstrap set again. This
+    /// will allow them to faster do their work and can be used to improve the user experience in
+    /// moments where the application needs discovery.
```
mycognosist:
```suggestion
    /// will allow them to do their work faster and can be used to improve the user experience in
    /// moments where the application needs discovery.
```

p2panda-net-next/src/actors/discovery/manager.rs
```diff
@@ -236,18 +253,32 @@ where
...
+
     async fn handle(
         &self,
         myself: ActorRef<Self::Msg>,
         message: Self::Msg,
         state: &mut Self::State,
     ) -> Result<(), ActorProcessingErr> {
         match message {
-            ToDiscoveryManager::RegisterProtocol => {
+            ToDiscoveryManager::Initiate => {
```
mycognosist: If this event is only intended to occur once during the life-cycle of the node then the associated code can be moved into the `post_start()` handler instead. That removes the need for the `Initiate` message variant and makes it clear that this is once-off initialisation.

adzialocha: Oh interesting, I wasn't aware of `post_start`

adzialocha: Nice, could be many other places we actually want to do the same then.

mycognosist: Definitely. I can think of at least one gossip actor which could use this.

PR #889, opened 2025-11-27
Remove transport info when connection attempt failed

adzialocha:


p2panda-net-next/src/actors/address_book/address_book.rs
```diff
@@ -84,6 +84,15 @@ pub enum ToAddressBook {
     #[allow(unused)]
     RemoveNodeInfo(NodeId, RpcReplyPort<bool>),
 
+    /// Removes transport information for a node which will make it "not connectable" for us. This
+    /// method can be used to mark nodes as "stale", for example after a failed connection attempt.
+    ///
+    /// Transport info can be set again either manually or if that node connected us and we've
+    /// exchanged their info again during discovery.
```
mycognosist:
```suggestion
    /// Transport info can be set again either manually or if that node connected to us and we've
    /// received their info again during discovery.
```

mycognosist (approved):

PR #895, opened 2025-12-05
Hash-based private set intersection for topic discovery

jackaperkins:
Introduce a protocol for private set intersection between two peers using hashes to protect the topics.

Current tests prove the successful run of the protocol for both alice and bob, and some basic failure cases where either side sends out of order protocol messages.

### DIFF
--- a/p2panda-discovery-next/src/psi_hash.rs
+++ b/p2panda-discovery-next/src/psi_hash.rs
@@ -50,19 +50,11 @@ where
     },
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct Config {
     /// If enabled the protocol will only share transport infos of nodes which have at least one
     /// common topic with the remote node.
     pub share_nodes_with_common_topics: bool,
-}
-
-impl Default for Config {
-    fn default() -> Self {
-        Self {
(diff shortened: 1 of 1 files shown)


adzialocha:
I did a rebase & force-push against the latest `net-rewrite` commit @jackaperkins. I hope it didn't mess up anyhing, to me it looks okay.

p2panda-discovery-next/src/lib.rs
```diff
@@ -84,6 +84,8 @@ pub mod address_book;
 // protocol.
 #[cfg(any(test, feature = "test_utils"))]
 pub mod naive;
+#[cfg(any(test, feature = "test_utils"))]
```
adzialocha: This doesn't need to be behind any flags

p2panda-discovery-next/src/tests.rs
```diff
@@ -249,3 +251,129 @@ async fn naive_protocol() {
 
     handle.await;
 }
+
+#[tokio::test]
+async fn topic_discovery() {
```
adzialocha: I would maybe move these tests into `psi_hash` as on lib-level `tests` we're usually handling E2E tests spanning multiple systems (random walker, discovery protocol, etc.).

PR #899, opened 2025-12-11
Pause and resume discovery walkers based on local connectivity status

mycognosist:
This is an interim fix to prevent nodes from being incorrectly marked as stale because of failed discovery sessions during periods when the local node is offline.

We now pause discovery walkers when the local node goes offline and resume them when online again. No new actors or supervision logic is introduced.

## 📋 Checklist

### DIFF
--- a/p2panda-net-next/src/actors/discovery/manager.rs
+++ b/p2panda-net-next/src/actors/discovery/manager.rs
@@ -66,6 +66,14 @@ pub enum ToDiscoveryManager {
     /// moments where the application needs discovery.
     ResetWalkers,
 
+    /// Pause all active walkers, preventing new sessions from being initiated. This is called when
+    /// the local node has gone offline.
+    PauseWalkers,
+
+    /// Resume all walkers, allowing iinitiation of new sessions. This is called when new transport
+    /// info has been received for the local node, indicating that we are online.
+    ResumeWalkers,
+
     /// Subscribe to system events.
     Events(RpcReplyPort<broadcast::Receiver<DiscoveryEvent>>),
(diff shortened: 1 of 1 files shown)


p2panda-net-next/src/actors/discovery/walker.rs
```diff
@@ -153,61 +163,87 @@ where
     ) -> Result<(), ActorProcessingErr> {
         match message {
             ToDiscoveryWalker::NextNode(mut walk_from_here) => {
```
mycognosist, line 165: The diff for this message handler looks messy but all that has changed is the `if state.local_node_is_online` check (which then impacts the indentation for the rest of the message handling logic).

p2panda-net-next/src/actors/discovery/manager.rs
```diff
@@ -350,13 +358,21 @@ where
                                 }
                             }
                             Some(event) = node_info_rx.recv() => {
-                                // Reset walkers if new transport info was set, ignore if we're
-                                // offline (and transport info is `None`).
                                 match event.value {
-                                    Some(node_info) => if node_info.transports.is_none() {
+                                    Some(node_info) => if node_info.transports.is_some() {
```
mycognosist, line 362: @adzialocha I'd be stoked if you could double-check my logic here.

If `node_info.transport.is_none()` or we receive `None` on `event.value` then we pause walkers, otherwise we continue (which will result in the walkers being reset).

mycognosist:
Closed in favour of https://github.com/p2panda/p2panda/pull/900

PR #900, opened 2025-12-12
Terminate and (re)spawn walkers based on network connectivity

mycognosist:
This PR is intended to replace https://github.com/p2panda/p2panda/pull/899

- [x] Create a ractor pub-sub port for broadcasting changes to the set of local endpoint addresses
- [x] Subscribe to address changes in the discovery manager
- [x] Parse address changes to determine the level of connectivity
- [x] If connectivity becomes degraded, terminate all walkers
- [x] If connectivity becomes improved, respawn all walkers

## 📋 Checklist

### DIFF
--- a/p2panda-net-next/src/actors/discovery/tests.rs
+++ b/p2panda-net-next/src/actors/discovery/tests.rs
@@ -1,9 +1,11 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use std::sync::Arc;
+
 use p2panda_core::PrivateKey;
 use p2panda_discovery::address_book::AddressBookStore as _;
 use ractor::thread_local::{ThreadLocalActor, ThreadLocalActorSpawner};
-use ractor::{ActorRef, call};
+use ractor::{ActorRef, OutputPort, call};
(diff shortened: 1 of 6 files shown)


p2panda-net-next/src/actors/iroh/endpoint.rs
```diff
@@ -186,6 +195,17 @@ impl ThreadLocalActor for IrohEndpoint {
...
+                // Listen for changes to our endpoint's addresses.
+                let watch_addr_handle = {
+                    let endpoint = endpoint.clone();
+                    let myself = myself.clone();
+                    tokio::spawn(async move {
+                        while let Ok(addr) = endpoint.watch_addr().updated().await {
+                            let _ = myself.send_message(ToIrohEndpoint::AddressUpdate(addr.addrs));
+                        }
+                    })
+                };
```
adzialocha, line 210: This logic is already handled through the node info watchers. We receive this information the address book discovery service ("resolve" and "publish") and can watch our own address to receive this data.

Designing it like this allows other services to just look at the address book without knowing about the endpoint specifically.

PR #908, opened 2025-12-21
Allow endpoint to gracefully shut down on drop

adzialocha:
Allows all instances where we use iroh's `Endpoint` to gracefully close connections and sockets to not run into the issue where we can't re-connect to a relay after restarting (see https://github.com/n0-computer/iroh/issues/3798).

To make this work we need to spawn tasks on the `post_stop` handlers which allow the endpoint instance to wind down, independent of when the actors are actually dropped.

Lastly I've needed to change the methods from `stop` to `stop_and_wait` in ractor, as otherwise the `post_stop` handler was actually never called (at least for nested actors).

### DIFF
--- a/p2panda-net-next/src/actors/supervisor.rs
+++ b/p2panda-net-next/src/actors/supervisor.rs
@@ -126,12 +126,23 @@ where
         _myself: ActorRef<Self::Msg>,
         state: &mut Self::State,
     ) -> Result<(), ActorProcessingErr> {
+        trace!("supervisor shutting down");
+
         let reason = Some("network system is shutting down".to_string());
 
         // Stop all the actors which are directly supervised by this actor.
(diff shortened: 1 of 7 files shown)


p2panda-net-next/src/actors/supervisor.rs
```diff
@@ -126,12 +126,23 @@ where
...
+            .stop_and_wait(reason.clone(), None)
+            .await?;
+        state
+            .address_book_actor
+            .stop_and_wait(reason.clone(), None)
+            .await?;
+        state
+            .endpoint_supervisor
+            .stop_and_wait(reason.clone(), None)
+            .await?;
```
adzialocha, line 145: When just calling `stop` in ractor apparently the `post_stop` handler is not really called, at least this seems to be the case for nested actors.

PR #909, opened 2025-12-24
Modular API for new `p2panda-net`

adzialocha:
This PR turns the new `p2panda-net` API "inside out" and exposes all inner modules as independent objects with a builder pattern.

This should allow for a clearer and transparent API surface, less generics (since every object has a clear function and represents a variant of something), run different, multiple sync protocols etc. in parallel, clear system-level events subscription for each object and the user only needs to pull in what they need.

With this approach it'll be easier in the future for applications to switch between different endpoints (multi-transport), discovery strategies (because some might be less useful for certain transports), etc.

## Example

```rust
use p2panda_core::PrivateKey;
use p2panda_net::iroh_mdns::MdnsDiscoveryMode;
use p2panda_net::test_utils::{TestMemoryStore, TestTopicMap};
use p2panda_net::{AddressBook, Discovery, Endpoint, Gossip, LogSync};

let private_key = PrivateKey::new();

let address_book = AddressBook::builder().spawn().await?;

let endpoint = Endpoint::builder(address_book.clone())
    .private_key(private_key)
    .network_id([42; 32])
    .spawn()
    .await?;

let _mdns = MdnsDiscovery::builder(address_book.clone(), endpoint.clone())
    .mode(MdnsDiscoveryMode::Active)
    .spawn()
    .await?;

let _discovery = Discovery::builder(address_book.clone(), endpoint.clone())
    .spawn()
    .await?;

let gossip = Gossip::builder(address_book.clone(), endpoint.clone())
    .spawn()
    .await?;

let handle = gossip.stream([1; 32]).await?;
handle.publish(b"Hello, Panda!").await?;
let mut rx = handle.subscribe();

tokio::spawn(async move {
    while let Some(_bytes) = rx.next().await {
        // ...
    }
});

let store = TestMemoryStore::new();
let topic_map = TestTopicMap::new();

let sync = LogSync::builder(store, topic_map, address_book, endpoint, gossip)
    .spawn()
    .await?;

let handle = sync.stream([1; 32], true).await?;
let mut rx = handle.subscribe().await?;

tokio::spawn(async move {
    while let Some(_event) = rx.next().await {
        // ...
    }
});
```

While the diff looks scary not much has changed other than moving files around and removing boilerplate which was required by the previous approach. These are the actual changes:

- Replace outdated crates with new `p2panda-net`, `p2panda-sync` and `p2panda-discovery`
- Move each "module" into own folder with clear builder and high-level API, actors are now nested inside
- Removed actor namespacing, actors are now addresses directly by their `ActorRef`
- Change tests to use objects directly (less boilerplate when writing tests)
- "Streams" become part of the regarding gossip or sync module API
- Automatically close sync / gossip sessions when all references to subscriptions and / or handles are dropped
- Improved error handling for streams, reduced type size
- Fixed re-use of topic across gossip- and log sync streams
- Removed unused trait bound and generics in -sync a bit
- Simplified address book & discovery protocol to not distinct between ephemeral messaging- and eventually consistent sync topics
- Implement `Stream` for all subscribers

## Todo

- [x] Address book
- [x] iroh Endpoint
- [x] iroh mDNS
- [x] Confidential Random-Walk Discovery
- [x] Gossip peer-sampling and ephemeral messaging (unreliable multicast)
- [x] Log Sync "Namakemono" (using p2panda's "operations")
- [x] Bring back supervision as an explicit module and feature flag which can be imported


mycognosist:
> I would rather move this task into an issue and tackle it later instead of "promising" this feature to users when it's not doing what would be expected

Agreed.

mycognosist (approved):

PR #929, opened 2026-01-15
Chat example for `p2panda-net` using modular API

mycognosist:
- [x] Gossip: used to send heartbeat messages
- [x] Sync: used to send chat messages

Includes the ability to set your nickname using `/nick <NAME>`.

Prints online / offline status updates:

`-> <NICKNAME / SHORT FMT PUBLIC KEY> came online / went offline`

**Usage example:**

Node 1: `cargo run --example chat`

Node 2: `cargo run --example chat -- -b <PUBLIC_KEY_OF_NODE_1> -c <CHAT_TOPIC_ID_OF_NODE_1>`

## 📋 Checklist

### DIFF
--- a/p2panda-net/examples/chat.rs
+++ b/p2panda-net/examples/chat.rs
@@ -1,6 +1,30 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 //! Example chat application using p2panda-net.
+//!
+//! Here we showcase the sync and gossip features of the p2panda-net API and demonstrate the
+//! configuration steps required to successfully run the required sub-systems.
+//!
+//! Chat messages are sent using the p2panda log sync protocol; this allows catching up on past
+//! messages which were published before you came online. The gossip protocol is used to broadcast
+//! heartbeat messages which are used to display online / offline messages for other peers in the
+//! chat.
+//!
+//! ## Usage
+//!
+//! Run the example on the first node:
+//!
+//! `cargo run --example chat`
+//!
+//! Run the example on a second node, using the chat topic ID and public key of the first node:
+//!
+//! `cargo run --example chat -- -c <CHAT_TOPIC_ID> -b <FIRST_NODE_PUBLIC_KEY>`
+//!
+//! Once several nodes are in the same chat, any node's public key can be supplied as the `-b`
+//! parameter (bootstrap node ID).
+//!
+//! Type into the terminal and press <ENTER> to send messages. Type `/nick <NICKNAME>` and press
+//! <ENTER> to set your desired nickname.
 use std::collections::HashMap;
 use std::convert::Infallible;
 use std::sync::Arc;


p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,351 @@
...
+
+#[derive(Parser)]
+struct Args {
+    /// Identifier of a bootstrap node.
+    #[arg(short = 'b', long, value_name = "BOOTSTRAP_ID")]
+    bootstrap_id: Option<NodeId>,
+
+    /// Seed for deterministic private key generation.
+    #[arg(short = 's', long, value_name = "SEED")]
+    seed: Option<u8>,
```
adzialocha: I'd be up for removing this, it was useful for troubleshooting in the past but for an example it's maybe unnecessary noise? I believe it's good to keep the code complexity as low as possible.

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,351 @@
...
+use tokio::sync::{RwLock, mpsc};
+use tracing::info;
+use tracing_subscriber::EnvFilter;
+use tracing_subscriber::prelude::*;
+
+type LogId = u64;
+
+const CHAT_TOPIC: [u8; 32] = [1; 32];
+
+const HEARTBEAT_TOPIC: [u8; 32] = [2; 32];
```
adzialocha: With the new `p2panda-net` we can re-use topics across `Gossip` and `LogSync`.

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,351 @@
...
+use p2panda_sync::traits::TopicLogMap;
+use p2panda_sync::{Logs, TopicLogSyncEvent};
+use tokio::sync::{RwLock, mpsc};
+use tracing::info;
+use tracing_subscriber::EnvFilter;
+use tracing_subscriber::prelude::*;
+
+type LogId = u64;
+
+const CHAT_TOPIC: [u8; 32] = [1; 32];
```
adzialocha: I'm undecided here but what do you think about encouraging best-practices by using a truly random topic already in this example? It probably means that the person creating a chat generates a topic with a CSPRNG and the other nodes need to mention it as a `--chat-id` argument?

adzialocha: We obviously would make the example slightly "harder" to use. On the other hand you get multiple chat-rooms for free and true confidential discovery done right 😎

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,351 @@
...
+            .insert_node_info(NodeInfo::from(endpoint_addr).bootstrap())
+            .await?;
+    }
+
+    let endpoint_config = IrohConfig {
+        bind_ip_v4: Ipv4Addr::LOCALHOST,
+        bind_port_v4: 0,
+        bind_ip_v6: Ipv6Addr::LOCALHOST,
+        bind_port_v6: 0,
+        relay_urls: vec![RELAY_URL.parse().unwrap()],
```
adzialocha: Hmn .. this is interesting. I think I forgot to add this in the modular API PR. One needs to dig pretty deep now to add relay urls. Maybe we want to add something on the builder for it?

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,351 @@
...
+    if let Some(id) = args.bootstrap_id {
+        let endpoint_addr = EndpointAddr::new(from_public_key(id));
+        let endpoint_addr = endpoint_addr.with_relay_url(RELAY_URL.parse().unwrap());
+        address_book
+            .insert_node_info(NodeInfo::from(endpoint_addr).bootstrap())
+            .await?;
+    }
+
+    let endpoint_config = IrohConfig {
+        bind_ip_v4: Ipv4Addr::LOCALHOST,
```
adzialocha: Just in case: Setting the interfaces to `LOCALHOST` will not enable this chat example to be used across different devices

p2panda-net/examples/chat.rs
adzialocha: Would be nice if one doesn't need to configure the endpoint in this example, it makes it unnecessarily noisy

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,351 @@
...
+        .unwrap();
+
+    let _discovery = Discovery::builder(address_book.clone(), endpoint.clone())
+        .config(DiscoveryConfig::default())
+        .spawn()
+        .await
+        .unwrap();
+
+    if args.mdns {
+        let _mdns = MdnsDiscovery::builder(address_book.clone(), endpoint.clone())
```
adzialocha: The `MdnsDiscovery` instance gets dropped instantly and never runs since it it's scope is limited by the if-clause

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,351 @@
...
+                            String::from_utf8(operation.body.as_ref().unwrap().to_bytes()).unwrap();
+
+                        // Check if the text of this operation is setting a nickname.
+                        if let Some(nick) = text.strip_prefix("/nick ") {
+                            if let Some(previous_nick) = nicknames.get(&remote_public_key) {
+                                print!("-> {} is now known as: {}", previous_nick, nick);
+                            } else {
+                                print!("-> {} is now known as: {}", remote_id, nick);
+                            }
+                            nicknames.insert(remote_public_key, nick.trim().to_owned());
```
adzialocha: This is pretty cool, instantly gives it "real" chat vibes!

adzialocha (commented):
Had a quick look at this already for early feedback and thoughts. It's cool!

`+1` from me for using `Gossip` for an "Awareness / Presence" feature. That's a great usecase!

p2panda-net/examples/chat.rs
mycognosist: ^_^

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,426 @@
...
+use serde::{Deserialize, Serialize};
+use tokio::sync::{RwLock, mpsc};
+use tokio::time::Instant;
+use tracing::info;
+use tracing_subscriber::EnvFilter;
+use tracing_subscriber::prelude::*;
+
+type LogId = u64;
+
+const NETWORK_ID: [u8; 32] = [7; 32];
```
adzialocha: We can also use the default network id in `p2panda-net` and skip setting our own

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,426 @@
...
+    setup_logging();
+
+    let args = Args::parse();
+
+    let private_key = PrivateKey::new();
+    let public_key = private_key.public_key();
+
+    // Retrieve the chat topic ID from the provided arguments, otherwise generate a new, random,
+    // cryptographically-secure identifier.
+    let topic_id: [u8; 32] = if let Some(topic) = args.chat_topic_id {
```
adzialocha: Maybe it's nice to get the `TopicId` type from the crate, just to showcase the types

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,426 @@
...
+    println!("relay url: {}", RELAY_URL);
+
+    // Set up sync for p2panda operations.
+    let mut store = ChatStore::new();
+
+    let topic_map = ChatTopicMap::default();
+    topic_map.insert(topic_id, public_key, LOG_ID).await;
+
+    // Prepare address book.
+    let address_book = AddressBook::builder().spawn().await.unwrap();
```
adzialocha: There's many places where we `unwrap`, we can propagate the error up instead

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,426 @@
...
+    let endpoint = Endpoint::builder(address_book.clone())
+        .private_key(private_key.clone())
+        .network_id(NETWORK_ID)
+        .relay_url(RELAY_URL.parse().unwrap())
+        .spawn()
+        .await
+        .unwrap();
+
+    let _discovery = Discovery::builder(address_book.clone(), endpoint.clone())
+        .config(DiscoveryConfig::default())
```
adzialocha: Using `config` here is not required

p2panda-net/examples/chat.rs
```diff
@@ -0,0 +1,426 @@
...
+        .spawn()
+        .await
+        .unwrap();
+
+    let mdns_discovery_mode = if args.mdns {
+        MdnsDiscoveryMode::Active
+    } else {
+        MdnsDiscoveryMode::Disabled
```
adzialocha: I've just removed `Disabled` https://github.com/p2panda/p2panda/commit/40bc184c0c9e9fadad3f766f1c4d3ebb35aa6bbd, so maybe you can switch it to `Passive` instead?

adzialocha (approved):

PR #943, opened 2026-01-16
Test for incorrectly "raw" topics being added to address book

sandreae:
This occurs due to the sync manager incorrectly adding the "raw" topic to the address book.

## 📋 Checklist


adzialocha:
https://github.com/p2panda/p2panda/pull/947 should fix the underlying issue this PR is getting at here. The test is testing a different layer (including `LogSync`) so I'd be up for getting this in our test suite as well. What do you think?

PR #961, opened 2026-01-21
Avoid leaking low-level p2panda-sync types in p2panda-net API 

sandreae:
This fixes the fact that the `SyncManager` trait was being leaked due to it being a requirement
on a generic parameter in error types and the sync stream handle.

With this change error types and the SyncHandle now have generic parameters for the concrete
message and event types defined at application level, rather than requiring the concrete manager
type to be defined. Some minor improvements to generic parameters in p2panda-sync have also
been included.

Additionally a BroadcastStreamRecvError is returned directly from SyncSubscription and
GossipSubscription streams which avoids unnecessary error unwrapping.

## 📋 Checklist

### DIFF
--- a/p2panda-net/src/sync/log_sync/api.rs
+++ b/p2panda-net/src/sync/log_sync/api.rs
@@ -1,13 +1,13 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use std::fmt::Debug;
+use std::marker::PhantomData;
 use std::sync::Arc;
 
-use p2panda_core::Extensions;
+use p2panda_core::{Extensions, Operation};
 use p2panda_store::{LogId, LogStore, OperationStore};
-use p2panda_sync::manager::TopicSyncManager;
-use p2panda_sync::protocols::Logs;
-use p2panda_sync::traits::{Manager as SyncManagerTrait, TopicMap};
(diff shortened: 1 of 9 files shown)


p2panda-net/src/sync/log_sync/api.rs
```diff
@@ -1,13 +1,13 @@
...
+use std::marker::PhantomData;
 use std::sync::Arc;
 
-use p2panda_core::Extensions;
+use p2panda_core::{Extensions, Operation};
 use p2panda_store::{LogId, LogStore, OperationStore};
-use p2panda_sync::manager::TopicSyncManager;
-use p2panda_sync::protocols::Logs;
-use p2panda_sync::traits::{Manager as SyncManagerTrait, TopicMap};
+use p2panda_sync::protocols::{Logs, TopicLogSyncEvent};
```
adzialocha, line 9: This is unrelated to this PR, just a naming thought:

All sync protocols are over a "topic", so maybe we can skip mentioning that in the name? This will also make the types shorter / more readable and nicely consistent with `LogSync`:

`TopicLogSyncEvent` -> `LogSyncEvent`

adzialocha (approved):
Great cleanup!

PR #973, opened 2026-02-03
Cleanup state in gossip unsubscribe handler

mycognosist:
The `ActorTerminated` event cannot be relied upon to fire when a child actor is stopped by the parent. Since we were cleaning up some gossip session state and sending the `Left` event from the termination handler, this resulted in the event not being sent on some occasions.

The cleanup logic has been moved into the `Unsubscribe` handler and the `ActorTerminated` event handler has been removed completely to avoid further confusion in the future.

## 📋 Checklist


sandreae (approved):
Nice find and fix!

PR #974, opened 2026-02-03
Gracefully handle concurrently pruned operations in log sync protocol

sandreae:
It is possible that operations are pruned from a log after a sync session has started and hashes of all operations a remote needs calculated, but before the operations themselves were retrieved from the store. Prior to this change this caused a panic in the log sync protocol as it expected all operations to be still present in the store. With this change we gracefully handle the case where an operation is no longer present, logging the fact that it is missing.

## 📋 Checklist

### DIFF
--- a/p2panda-sync/src/protocols/log_sync.rs
+++ b/p2panda-sync/src/protocols/log_sync.rs
@@ -13,6 +13,7 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio::select;
 use tokio::sync::broadcast;
+use tracing::{trace, warn};
 
 use crate::dedup::{DEFAULT_BUFFER_CAPACITY, Dedup};
 use crate::traits::Protocol;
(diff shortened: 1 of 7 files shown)


p2panda-sync/src/protocols/log_sync.rs
```diff
@@ -296,26 +297,48 @@ where
...
+
+                                // For testing, where the store is an in memory implementation
+                                // which may not yield work to other tasks, we force the task to
+                                // yield in order to test concurrent database actions occurring
+                                // during sync.
+                                #[cfg(test)]
+                                tokio::task::yield_now().await;
+
+                                if let Some((header, body)) = result {
+                                    debug!(
```
mycognosist: What do you think about using `trace!` here instead? I imagine the `debug` log output will get very noisy in production with lots of sync sessions on the go; could be good to keep that behind the `trace`.

The `None` case could maybe also move to `trace` but I'm not as confident in that suggestion.

PR #987, opened 2026-02-10
Introduce p2panda high-level API skeleton

adzialocha:
Here we are introducing the skeleton of the upcoming `p2panda` high-level API.

This is still a draft, minor things might change, but we're setting this off to already start working on GObject and FFI.

### DIFF
--- a/p2panda/src/network.rs
+++ b/p2panda/src/network.rs
@@ -1,0 +1,140 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::collections::HashSet;
+
+use p2panda_core::PrivateKey;
+use p2panda_net::address_book::AddressBookError;
+use p2panda_net::addrs::NodeInfo;
+use p2panda_net::discovery::{DiscoveryConfig, DiscoveryError};
+use p2panda_net::gossip::{GossipConfig, GossipError};
+use p2panda_net::iroh_endpoint::{EndpointError, IrohConfig, RelayUrl};
+use p2panda_net::iroh_mdns::{MdnsDiscoveryError, MdnsDiscoveryMode};
+use p2panda_net::{
+    AddressBook, DEFAULT_NETWORK_ID, Discovery, Endpoint, Gossip, MdnsDiscovery, NetworkId, NodeId,
+};
+use thiserror::Error;
+
+#[derive(Clone)]
+pub struct Network {
+    pub address_book: AddressBook,
+    pub mdns: MdnsDiscovery,
+    pub endpoint: Endpoint,
+    pub discovery: Discovery,
+    pub gossip: Gossip,
+}
+
+impl Network {
+    pub async fn spawn(
+        config: NetworkConfig,
+        private_key: PrivateKey,
+    ) -> Result<Self, NetworkError> {
+        // TODO: Pass in store.
+
+        // TODO: Supervision of actors.
+
+        let address_book = AddressBook::builder()
+            // TODO: Move address book store into p2panda-store-next
+            // .store(address_book_store)
+            .spawn()
+            .await?;
+
+        for bootstrap in &config.bootstraps {
+            address_book.insert_node_info(NodeInfo::new(*bootstrap).bootstrap());
+        }
+
+        let mut endpoint = Endpoint::builder(address_book.clone())
+            .config(config.iroh)
+            .private_key(private_key)
+            .network_id(config.network_id);
+
+        for url in &config.relay_urls {
+            endpoint = endpoint.relay_url(url.clone());
+        }
+
+        let endpoint = endpoint.spawn().await?;
+
+        let mdns = MdnsDiscovery::builder(address_book.clone(), endpoint.clone())
+            .mode(config.mdns_mode)
+            .spawn()
+            .await?;
+
+        let discovery = Discovery::builder(address_book.clone(), endpoint.clone())
+            .config(config.discovery)
+            .spawn()
+            .await?;
+
+        let gossip = Gossip::builder(address_book.clone(), endpoint.clone())
+            .config(config.gossip)
+            .spawn()
+            .await?;
+
+        // TODO: Add log sync with topic map and stores.
+
+        Ok(Self {
+            address_book,
+            endpoint,
+            mdns,
+            discovery,
+            gossip,
+        })
+    }
+
+    pub fn id(&self) -> NodeId {
+        self.endpoint.node_id()
+    }
+
+    pub fn network_id(&self) -> NetworkId {
+        self.endpoint.network_id()
+    }
+
+    pub async fn insert_bootstrap(&self, node_id: NodeId) -> Result<(), NetworkError> {
+        let node_info = NodeInfo::new(node_id).bootstrap();
+        self.address_book.insert_node_info(node_info).await?;
+        Ok(())
+    }
+
+    // TODO: Do we need methods to get the transport info (with ip addresses etc.)?
+}
+
+#[derive(Clone, Debug)]
+pub struct NetworkConfig {
+    pub network_id: NetworkId,
+    pub relay_urls: HashSet<RelayUrl>,
+    pub bootstraps: HashSet<NodeId>,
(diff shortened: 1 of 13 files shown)


p2panda/src/streams/stream.rs
```diff
@@ -0,0 +1,121 @@
...
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Message<M> {
+    header: Header,
+    topic: Topic,
+    body: M,
+}
+
+impl<M> Message<M> {
+    // TODO: Or rather have a method to get the topic? Subject is not really useful here anymore.
```
sandreae: outdated comment

p2panda/src/streams/stream.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// Subscribe to an event stream where each message must be committed manually by calling
+    /// StreamHandle::commit().
+    ///
+    /// This is for users who want to commit each message manually themselves. Useful if
+    /// application validation needs to pass before a message is accepted.
+    pub async fn subscribe_no_commit(&self) -> Result<StreamSubscription<M>, StreamError> {
+        unimplemented!()
+    }
+
+    /// Manually commit a message.
```
sandreae, line 39: Do we still want this method if there is a `commit()` method on the message itself? I can imagine it would still be useful, so I'm fine to keep both, just wondering.

p2panda/src/streams/stream.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    pub fn timestamp(&self) -> u64 {
+        self.header.timestamp
+    }
+
+    pub fn header(&self) -> &Header {
+        &self.header
+    }
+
+    pub fn body(&self) -> &M {
```
sandreae, line 80: We need to account for operations where the body was deleted. Not a nice API to have to deal with Options all the time though....

sandreae: It could be a different event type in the stream and have it's own concrete type. I don't even know if we would want to expose them at all really.

p2panda/src/streams/stream.rs
adzialocha: Yeah, I think its good to allow multiple ways to do the same thing

p2panda/src/streams/stream.rs
adzialocha: Good question! On this level (application layer) you will probably not receive them, underneath we'll be dealing with them for system layer stuff

p2panda-net/src/iroh_endpoint/mod.rs
```diff
@@ -12,6 +12,9 @@ mod supervisor;
 mod tests;
 pub(crate) mod user_data;
 
+// Re-export useful iroh types.
+pub use iroh::RelayUrl;
```
jsparber, line 16: this probably should be a seperate MR.

p2panda/src/streams/stream.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /// Subscribe to an event stream where each message must be committed manually by calling
+    /// StreamHandle::commit().
+    ///
+    /// This is for users who want to commit each message manually themselves. Useful if
+    /// application validation needs to pass before a message is accepted.
+    pub async fn subscribe_no_commit(&self) -> Result<StreamSubscription<M>, StreamError> {
+        unimplemented!()
+    }
```
adzialocha: Wondering if we shouldn't have an manual_commit option on the NodeBuilder instead and have it globally set for all streams, or at least on the handle? It gets a little bit weird if it's possible to create an auto-commit and manual-commit subscription for the same topic.

sandreae: Yeah, I like that.

p2panda/src/network.rs
```diff
@@ -0,0 +1,140 @@
...
+        self.address_book.insert_node_info(node_info).await?;
+        Ok(())
+    }
+
+    // TODO: Do we need methods to get the transport info (with ip addresses etc.)?
+}
+
+#[derive(Clone, Debug)]
+pub struct NetworkConfig {
+    pub network_id: NetworkId,
```
mycognosist, line 101: Maybe we keep the fields private and expose methods for setting them individually? Could take the same approach for `Network` to retrieve a direct handle to an internal service if required.

p2panda/tests/api.rs
```diff
@@ -0,0 +1,33 @@
...
+// TODO: Remove this later.
+#![allow(unused)]
+
+use futures_core::Stream;
+use p2panda::Topic;
+use p2panda_core::PrivateKey;
+
+#[tokio::test]
+async fn it_works() -> Result<(), Box<dyn std::error::Error>> {
+    let node = p2panda::spawn().await?;
```
mycognosist, line 12: So fresh and so clean! ✨

p2panda/src/node.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    pub async fn spawn() -> Result<Self, NodeError> {
+        // Generates new private key using CSPRNG from system.
+        let private_key = PrivateKey::new();
+
+        // Initialises an in-memory SQLite database.
+        let store = SqliteStoreBuilder::default().build().await?;
+
+        // Use default config, this will _not_ include a bootstrap and relay and reduces the
```
mycognosist, line 34: I imagine we want to expose these methods on the `Node` eventually to allow (re-)setting bootstrap and relay at runtime. Same goes for other network-related actions such as changing the mode that mDNS discovery is running in.

p2panda/src/streams/stream.rs
```diff
@@ -0,0 +1,115 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum StreamEvent<M> {
+    // TODO: Add more topic-related system events here which can come from node.
+    Message(Message<M>),
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Message<M> {
+    header: Header,
+    topic: Topic,
```
mycognosist, line 55: Here we discussed the potential of dropping the `topic` field since it can be inferred from contextual information (e.g. a message arrives on the `bamboo_appreciation` topic and is therefore associated with that topic).

p2panda/src/streams/ephemeral_stream.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum EphemeralStreamEvent<M> {
+    Message(EphemeralMessage<M>),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct EphemeralMessage<M> {
+    topic: Topic,
```
mycognosist, line 45: Same comment as `Message<M>`: perhaps `topic` is not required here.

p2panda/src/network.rs
adzialocha: `NetworkConfig` is so far an internal struct, not public for anyone outside of the crate, would that solve it for you?

p2panda/src/streams/stream.rs
adzialocha: Yes, that's true for internal workings but this is what the user receives, so they might want to know what topic this message was associated with.

`Message` does not represent the serialized version but the "enriched" one for the high-level API :+1:

p2panda/src/streams/ephemeral_stream.rs
adzialocha: Same as above

p2panda/src/node.rs
adzialocha: Yes, good idea. So far they can do all of this via `NodeBuilder` or `p2panda::builder` which I believe is sufficient for all the use-cases we have right now.

p2panda/src/network.rs
mycognosist: Ah yes, that's fine.

sandreae (approved):
Yap, think we got to a good place. Exciting!

PR #994, opened 2026-02-12
Sketches for new store APIs

sandreae:
## 📋 Checklist

### DIFF
--- a/p2panda-store-next/src/operations/traits.rs
+++ b/p2panda-store-next/src/operations/traits.rs
@@ -129,6 +129,7 @@ pub trait LogStore<T, L, ID> {
         public_key: &PublicKey,
         log_id: &L,
         after: Option<SeqNum>,
+        // to: Option<SeqNum>,
     ) -> impl Future<Output = Result<Option<(u64, u64)>, Self::Error>>;
 
     /// Get all entries in a log after an optional starting point.
@@ -140,9 +141,11 @@ pub trait LogStore<T, L, ID> {
         // "accidentally" making queries based on the seq number of a log branch that we don't
         // actually have locally (it would save us one db query).
         after: Option<SeqNum>,
+        // to: Option<SeqNum>,
     ) -> impl Future<Output = Result<Option<Vec<T>>, Self::Error>>;
 
     /// Efficiently prune a log.
+    /// // @TODO: rename this method to something better.... prune/prefex deletion related.
     fn delete_entries(
         &self,
         author: &PublicKey,


p2panda-store-next/src/operations/traits.rs
```diff
@@ -22,6 +26,20 @@ pub trait OperationStore<T, ID> {
...
     fn get_operation(&self, id: &ID) -> impl Future<Output = Result<Option<T>, Self::Error>>;
 
+    /// Get a batch of operations.
+    fn get_operations(&self, id: &[ID]) -> impl Future<Output = Result<Vec<T>, Self::Error>>;
+
+    /// Get a batch of raw operation header and body bytes.
+    fn get_raw_operations(
+        &self,
+        ids: &[ID],
+    ) -> impl Future<Output = Result<Vec<RawOperation>, Self::Error>>;
```
adzialocha: That's an interesting one as the operation type is generic in that API, so maybe it's not necessary to distinct between these two types, just have one single operation type which wraps the encoded and decoded version?

We would need to look closer how this plays out in the ingest and forge flows, avoiding a type mess + repetitive decoding etc.

p2panda-store-next/src/operations/traits.rs
```diff
@@ -32,4 +50,43 @@ pub trait OperationStore<T, ID> {
...
+/// The concrete message type contained on each log "entry" is not known in this API. It is
+/// assumed there is another store for retrieving these during sync (eg. OperationStore).
+pub trait LogStore<L, ID> {
+    type Error: Error;
+
+    /// Efficiently get all log heights for a set of author logs.
+    fn get_frontiers(
+        &self,
+        author: &PublicKey,
+        logs: Vec<L>,
```
adzialocha: Maybe this doesn't need to be an owned vector

p2panda-store-next/src/operations/traits.rs
```diff
@@ -32,4 +50,43 @@ pub trait OperationStore<T, ID> {
...
+    fn log_entries(
+        &self,
+        author: &PublicKey,
+        log_id: &L,
+        from: Option<SeqNum>,
+    ) -> impl Future<Output = Result<Option<Vec<(ID, SeqNum)>>, Self::Error>>;
+
+    /// Efficiently prune a log.
+    fn delete_entries(
+        &mut self,
```
adzialocha: The new approach for store APIs is to _not_ require `mut` for writes. Technically we don't need exclusive access to the store here, it's mostly an annoying thing to deal with in Rust for no reason

p2panda-store-next/src/operations/traits.rs
sandreae: Yeah, having a wrapped type here solving all use cases sounds like a good option actually. I also didn't like that this brought opinionated ideas (separate header and body) to the operation store.

The only reason we had for making it possible to get the raw bytes directly was to avoid unnecessary serialisation.

p2panda-store-next/src/operations/traits.rs
```diff
@@ -32,4 +39,130 @@ pub trait OperationStore<T, ID> {
...
+        &self,
+        log_id: &L,
+        id: &ID,
+        seq_num: SeqNum,
+    ) -> impl Future<Output = Result<bool, Self::Error>>;
+
+    /// Efficiently get current frontiers for a set of author logs.
+    ///
+    /// The frontiers of a set of logs are needed when communicating our current state to a remote
+    /// peer during set. Including the "frontiers" (set of hash+seq_num tuples) allows for
```
sandreae:
```suggestion
    /// peer during sync. Including the "frontiers" (set of hash+seq_num tuples) allows for
```

sandreae:
Elements of the work in this branch are being pulled out into self-contained PRs, for starters: #1004

PR #996, opened 2026-02-12
Ephemeral Messaging for Node API

adzialocha:
Implements ephemeral messaging functionality for new p2panda Node API.

```rust
let node = p2panda::spawn().await?;

let chat_id = Topic::new();
let chat = node.ephemeral_stream(chat_id).await?;

chat.publish("Hi!".into()).await?;

let mut rx = chat.subscribe().await;
let message = rx.next().await?;

println!("{}", message.body());
println!("{}", message.timestamp());
println!("{}", message.author());
```

This introduces an internally used "wrapped message" format which is encoded as CBOR and sent among nodes during gossip. It wraps the application's message, providing it with integrity and provenance guarantees with a signing scheme and makes sure that messages are unique with the help of an hybrid timestamp (microsecond-precision UNIX timestamp and logical lamport clock).

The uniqueness is required as iroh-gossip aka p2panda-net ignores duplicate messages which shown to be confusing for users.

Closes: #1000 

## 📋 Checklist

- [x] ~~Add this PR to the _Unreleased_ section in `CHANGELOG.md`~~

### DIFF
--- a/p2panda/Cargo.toml
+++ b/p2panda/Cargo.toml
@@ -17,6 +17,12 @@ all-features = true
 [lints]
 workspace = true
 
+[features]
+default = []
+test_utils = [
+  "dep:tracing-subscriber",
+]
+
 [dependencies]
 futures-util = { version = "0.3.31", default-features = false }
 hex = "0.4.3"
@@ -28,6 +34,9 @@ rand = "0.10.0"
 rand_chacha = "0.10.0"
 serde = { version = "1.0.228", features = ["derive"] }
 thiserror = "2.0.18"
+tracing = "0.1.41"
+tracing-subscriber = { version = "0.3.20", features = ["env-filter"], optional = true }
 
 [dev-dependencies]
+p2panda = { path = ".", features = ["test_utils"] }
 tokio = { version = "1.49.0", features = ["macros", "rt"] }
--- a/p2panda/src/lib.rs
+++ b/p2panda/src/lib.rs
@@ -11,6 +11,8 @@ mod network;
 pub mod node;
 pub mod operation;
 pub mod streams;
+#[cfg(any(test, feature = "test_utils"))]
+pub mod test_utils;
 pub mod topic;
 
 pub use node::Node;
--- a/p2panda/src/test_utils.rs
+++ b/p2panda/src/test_utils.rs
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+pub fn setup_logging() {
+    if std::env::var("RUST_LOG").is_ok() {
+        let _ = tracing_subscriber::fmt()
+            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+            .try_init();
+    }
+}


p2panda/src/streams/ephemeral_stream.rs
```diff
@@ -76,20 +283,74 @@ where
...
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        match ready!(self.inner.poll_next_unpin(cx)) {
+            // Check encoding & supported version and signature during deserialisation.
+            Some(Ok(bytes)) => match RawMessage::from_bytes(&bytes) {
+                Ok(raw) => Poll::Ready(Some(EphemeralMessage {
+                    topic: self.topic,
+                    inner: raw,
+                })),
+                Err(_err) => Poll::Pending, // TODO(adz): Do we want to inform the user about an
+                                            // invalid message?
```
adzialocha: I'm not entirely sure what to do with exposing our `p2panda` users of the ephemeral stream to errors which are concerned about "invalid messages".

Do users want to be informed if they received a message which had an invalid signature or format (according to our "wrapper" specification)? Or do we simply want to ignore this?

This type is not visible to users, so not really a public concern and an error potentially confusing?

I think that most of the time users will not be bothered about this. The only cases I can think where this helps are:

1. I want to disconnect from that "misbehaving" / spammy node
2. I want to find out about a bug I've introduced where I used a different version / implementation of p2panda etc. where the message is formatted wrongly

adzialocha: I went now with not surfacing this error but logging a warning. Let me know what you think!

p2panda/src/streams/ephemeral_stream.rs
```diff
@@ -2,72 +2,280 @@
...
+            timestamp,
+            body,
+        })
     }
 
-    pub async fn publish(&self, _message: M) -> Result<(), EphemeralStreamError> {
-        unimplemented!()
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, WrappedMessageError> {
+        // Attempt deserializing message tuple. This fails if the encoding is invalid.
+        let (version, public_key, signature, timestamp, logical, body): (
```
sandreae, line 73: Like the tuple approach here 👍

p2panda/src/streams/ephemeral_stream.rs
```diff
@@ -2,72 +2,280 @@
...
 
-/// Handle onto an ephemeral stream, exposes API for publishing messages and subscribing to the
-/// event stream.
-pub struct EphemeralStreamHandle<M> {
-    topic: Topic,
-    _marker: PhantomData<M>,
+/// Message specification version to create and encode messages for ephemeral streams.
+const MESSAGE_VERSION: u64 = 1;
+
+/// Message being dissiminated to other nodes via gossip.
```
mycognosist:
```suggestion
/// Message being disseminated to other nodes via gossip.
```

p2panda/src/streams/ephemeral_stream.rs
```diff
@@ -2,72 +2,280 @@
...
+        let (timestamp, logical) = self.timestamp.to_parts();
+        let message = (
+            self.version,
+            self.public_key,
+            timestamp,
+            logical,
+            &self.body,
+        );
+
+        // Treat encoding error for verifying the signature as an "invalid signature" since we
```
mycognosist:
```suggestion
        // Treat encoding error for verifying the signature as an "invalid signature" since
```

p2panda/src/streams/ephemeral_stream.rs
```diff
@@ -2,72 +2,280 @@
...
+        // the data came from an remote (potentially malicious) node.
```
mycognosist:
```suggestion
        // the data came from a remote (potentially malicious) node.
```

p2panda/src/node.rs
```diff
@@ -84,6 +91,14 @@ impl Node {
     }
 }
 
+/// Broken / closed communication channel with the internal gossip actor in `p2panda-net`. This can
+/// be due to the actor crashing.
+///
+/// Users may re-attempt creating a new ephemeral stream handle in case the actor restarted later.
+#[derive(Error, Debug)]
+#[error("error in internal gossip actor occurred: {0}")]
```
mycognosist:
```suggestion
#[error("error occurred in internal gossip actor: {0}")]
```

PR #1001, opened 2026-02-16
Export more network config via the builder

mycognosist:
Configuration options have been exported for the iroh (endpoint), discovery and gossip sub-systems. These are all exported as methods on the node builder; no node-runtime configuration is yet available.

- [x] Look at all configs and expose them in `NodeBuilder`

I've purposefully kept runtime configuration out of this PR for now. This is because 1) we want to keep the API surface as small as possible for this initial version, 2) it's not yet clear exactly which runtime config we wish to expose, and 3) some things are not possible due to missing functionality at the lower levels (for example, our mDNS discovery service does not currently expose a means of changing the mode during runtime...one would need to drop the service and recreate it with the desired mode configured). 

## 📋 Checklist

### DIFF
--- a/p2panda/src/builder.rs
+++ b/p2panda/src/builder.rs
@@ -1,7 +1,11 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use std::net::{Ipv4Addr, Ipv6Addr};
+
 use p2panda_core::PrivateKey;
-use p2panda_net::iroh_endpoint::RelayUrl;
+use p2panda_net::discovery::DiscoveryConfig;
+use p2panda_net::gossip::GossipConfig;
+use p2panda_net::iroh_endpoint::{IrohConfig, RelayUrl};
 use p2panda_net::iroh_mdns::MdnsDiscoveryMode;
 use p2panda_net::{NetworkId, NodeId};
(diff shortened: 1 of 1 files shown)


p2panda/src/builder.rs
```diff
@@ -68,7 +71,50 @@ impl NodeBuilder {
...
+
+    pub fn gossip_broadcast(mut self, broadcast: PlumTreeConfig) -> Self {
+        self.config.network.gossip.broadcast = broadcast;
+        self
+    }
+
+    pub fn gossip_max_message_size(mut self, size: usize) -> Self {
+        self.config.network.gossip.max_message_size = size;
+        self
+    }
```
adzialocha: It could make sense to only have commonly-used configs as "quick" builder methods while the lesser-used ones are more "hidden" in something like `gossip_config(mut self, config: GossipConfig)`, simply to reduce noise of the API for highly advanced features which can be confusing.

mycognosist: Good call. I've exposed the `IrohConfig` as separate methods for now (my thinking is that it's more common to want to set custom a custom IP and port).

adzialocha (approved):

PR #1004, opened 2026-02-16
`LogStore` SQLite implementation

sandreae:
Introduces `LogStore` trait in `p2panda-store` and implements an opinionated SQLite version for use in p2panda `Node`, along with tests.
    
We also add the `delete_operation_payload()` trait to the `OperationStore` and supply implementations for the memory and SQLite stores.

https://github.com/p2panda/p2panda/issues/1003

## 📋 Checklist

### DIFF
--- a/p2panda-core/src/test_utils.rs
+++ b/p2panda-core/src/test_utils.rs
@@ -4,7 +4,7 @@ use std::cell::RefCell;
 use std::rc::Rc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use crate::{Body, Extensions, Hash, Header, Operation, PrivateKey, Topic};
+use crate::{Body, Extensions, Hash, Header, Operation, PrivateKey, PublicKey, Topic};
 
 #[derive(Clone, Default)]
 pub struct TestLog {
@@ -24,8 +24,18 @@ impl TestLog {
         }
     }
 
+    pub fn from_private_key(private_key: PrivateKey) -> Self {
+        let mut log = TestLog::new();
+        log.private_key = private_key;
+        log
+    }
+
     pub fn id(&self) -> Topic {
         self.log_id
+    }
+
+    pub fn author(&self) -> PublicKey {
+        self.private_key.public_key()
     }
 
     pub fn operation<E: Extensions>(&self, body: &[u8], extensions: E) -> Operation<E> {
--- a/p2panda-store-next/src/logs/mod.rs
+++ b/p2panda-store-next/src/logs/mod.rs
@@ -1,0 +1,7 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+#[cfg(feature = "sqlite")]
+mod sqlite;
+mod traits;
+
+pub use traits::LogStore;
--- a/p2panda-store-next/src/logs/sqlite/mod.rs
+++ b/p2panda-store-next/src/logs/sqlite/mod.rs
@@ -1,0 +1,252 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+mod models;
+#[cfg(test)]
+mod tests;
+
+use std::collections::HashMap;
+
+use p2panda_core::cbor::encode_cbor;
+use p2panda_core::{Extensions, Hash, LogId, Operation, PublicKey};
+use sqlx::{query, query_as};
+
+use crate::logs::LogStore;
+use crate::logs::sqlite::models::{LatestEntryRow, LogHeightRow, LogMetaRow};
+use crate::operations::OperationRow;
+use crate::sqlite::{SqliteError, SqliteStore};
+
+pub type SeqNum = u64;
+
+impl<'a, L, E> LogStore<Operation<E>, PublicKey, L, SeqNum, Hash> for SqliteStore<'a>
+where
+    E: Extensions,
+    L: LogId,
+{
+    type Error = SqliteError;
+
+    /// Retrieve the hash and sequence number of the latest entry in an author's log.
+    async fn get_latest_entry(
+        &self,
+        author: &PublicKey,
+        log_id: &L,
+    ) -> Result<Option<(Hash, SeqNum)>, Self::Error> {
+        if let Some(latest_entry) = query_as::<_, LatestEntryRow>(
+            "
+            SELECT
+                hash,
+                seq_num
+            FROM
+                operations_v1
+            WHERE
+                public_key = ?
+                AND log_id = ?
+            ORDER BY
+                CAST(seq_num AS NUMERIC) DESC LIMIT 1
+            ",
+        )
+        .bind(author.to_string())
+        .bind(encode_cbor(&log_id).map_err(|err| SqliteError::Encode("log id".to_string(), err))?)
+        .fetch_optional(&self.pool)
+        .await?
+        {
+            let hash_seq_num = latest_entry.try_into()?;
+
+            Ok(Some(hash_seq_num))
+        } else {
+            Ok(None)
+        }
+    }
+
+    /// Retrieve the latest sequence number for a set of author's logs.
+    async fn get_log_heights(
+        &self,
+        author: &PublicKey,
+        logs: &[L],
+    ) -> Result<Option<HashMap<L, SeqNum>>, Self::Error> {
+        let mut encoded_log_ids = Vec::new();
+        for log in logs {
+            let encoded_log_id =
+                encode_cbor(&log).map_err(|err| SqliteError::Encode("log id".to_string(), err))?;
+            encoded_log_ids.push(encoded_log_id);
+        }
+
+        // This query formation approach is required since there is currently no
+        // way to directly bind arrays as comma-separated lists in sqlx.
+        let params = format!("?{}", ", ?".repeat(encoded_log_ids.len() - 1));
+        let query_str = format!(
+            "
+            SELECT
+                log_id,
+                CAST(MAX(CAST(seq_num AS NUMERIC)) AS TEXT) as seq_num
+            FROM
+                operations_v1
+            WHERE
+                public_key = ?
+                AND log_id IN ( {} )
+            GROUP BY
+                log_id
+            ",
+            params
+        );
+
+        let mut query = query_as::<_, LogHeightRow>(&query_str).bind(author.to_string());
+
+        for log_id in encoded_log_ids {
+            query = query.bind(log_id)
+        }
+
+        let log_heights_query = query.fetch_all(&self.pool).await?;
+
+        let log_heights = if log_heights_query.is_empty() {
+            None
+        } else {
+            let mut log_heights = HashMap::new();
+
+            for row in log_heights_query {
+                let (log_id, seq_num) = row.try_into()?;
+                log_heights.insert(log_id, seq_num);
+            }
+
+            Some(log_heights)
+        };
+
+        Ok(log_heights)
+    }
+
+    /// Retrieve the count and total byte size of all operations in an author's log.
+    async fn get_log_size(
+        &self,
+        author: &PublicKey,
+        log_id: &L,
+        after: Option<SeqNum>,
+        until: Option<SeqNum>,
+    ) -> Result<Option<(u64, u64)>, Self::Error> {
+        // We need to use an inclusive greater-than to ensure our
+        // query includes the operation with sequence number 0.
+        let after_operator = if after.is_none() { ">=" } else { ">" };
+        let query_str = format!(
+            "
+            SELECT
+                CAST(SUM(CAST(header_size AS NUMERIC)) AS TEXT) AS total_header_bytes,
+                CAST(SUM(CAST(payload_size AS NUMERIC)) AS TEXT) AS total_payload_bytes,
+                CAST(COUNT(*) AS TEXT) AS total_operation_count
+            FROM
+                operations_v1
+            WHERE
+                public_key = ?
+                AND log_id = ?
+                AND CAST(seq_num AS NUMERIC) {} CAST(? as NUMERIC)
+                AND CAST(seq_num AS NUMERIC) <= CAST(? as NUMERIC)
+            ",
+            after_operator
+        );
+
+        let log_meta: Option<LogMetaRow> = query_as::<_, LogMetaRow>(&query_str)
+            .bind(author.to_string())
+            .bind(
+                encode_cbor(&log_id)
+                    .map_err(|err| SqliteError::Encode("log id".to_string(), err))?,
+            )
+            .bind(after.unwrap_or(0).to_string())
+            .bind(until.unwrap_or(u64::MAX).to_string())
+            .fetch_optional(&self.pool)
+            .await?;
+
+        if let Some(row) = log_meta {
+            let (total_header_bytes, total_payload_bytes, total_operation_count) =
+                row.try_into()?;
+
+            return Ok(Some((
+                total_operation_count,
+                total_header_bytes + total_payload_bytes,
+            )));
+        }
+
+        Ok(None)
+    }
+
+    /// Retrieve log entries representing operations from an author's log.
+    async fn get_log_entries(
+        &self,
+        author: &PublicKey,
+        log_id: &L,
+        after: Option<SeqNum>,
+        until: Option<SeqNum>,
+    ) -> Result<Option<Vec<(Operation<E>, Vec<u8>)>>, Self::Error> {
+        // We need to use an inclusive greater-than to ensure our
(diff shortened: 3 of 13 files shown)


p2panda-store-next/src/logs/sqlite.rs
```diff
@@ -0,0 +1,80 @@
...
+        &self,
+        public_key: &p2panda_core::PublicKey,
+        log_id: &L,
+    ) -> Result<Option<(Hash, SeqNum)>, Self::Error> {
+        todo!()
+    }
+
+    async fn get_log_heights(
+        &self,
+        author: &p2panda_core::PublicKey,
```
adzialocha: Here he argument says author, in other places public_key. I don't have a preference, but we should agree on a canonical name for "operation signers"

p2panda-store-next/src/logs/sqlite.rs
```diff
@@ -0,0 +1,80 @@
...
+        logs: &[L],
+    ) -> Result<Option<std::collections::HashMap<L, SeqNum>>, Self::Error> {
```
adzialocha: Nitpick: Either use inlined imports or not but maybe not mix it. There's a HashMap already imported and other p2panda_core types

p2panda-store-next/src/logs/traits.rs
```diff
@@ -0,0 +1,121 @@
...
+/// application may choose to create a separate log for each author-channel pairing. In such a
+/// scenario, `LogId` might be implemented for a `struct` containing a `String` representation of
+/// the channel name.
+///
+/// Finally, please note that implementers of `LogId` must take steps to ensure their log design is
+/// fit for purpose and that all operations have been thoroughly validated before being persisted.
+/// No such validation checks are provided by `p2panda-store`.
+pub trait LogId: Clone + Debug + Eq + std::hash::Hash {}
+
+impl<T> LogId for T where T: Clone + Debug + Eq + std::hash::Hash {}
```
adzialocha: Could be interesting to move this into `p2panda-core`?

p2panda-store-next/src/logs/traits.rs
```diff
@@ -0,0 +1,121 @@
...
+
+pub type SeqNum = u64;
```
adzialocha: I'm not sure if we want to have already a concrete `SeqNum` type, is that a generic on the `LogStore` instead?

adzialocha: Our concrete implementation would of course use `Operation` and `u64` etc.

p2panda-store-next/src/logs/traits.rs
```diff
@@ -0,0 +1,121 @@
...
+    // @NOTE: If these batch queries are simple to implement then we don't actually need the
+    // equivalent "single log" queries above.  
+
+    /// Get all entries in a set of logs.
+    /// 
+    /// An optional range can be provided for every requested log.
+    fn get_log_entries_batch(
+        &self,
+        public_key: &PublicKey,
+        ranges: &HashMap<L, (Option<SeqNum>, Option<SeqNum>)>,
```
adzialocha: Can we use https://doc.rust-lang.org/std/ops/struct.Range.html to express ranges?

p2panda-store-next/src/logs/sqlite.rs
```diff
@@ -0,0 +1,80 @@
...
+impl<'a, L, E> LogStore<Operation<E>, L, Hash> for SqliteStore<'a>
+where
+    E: Extensions,
+    L: LogId,
+{
+    type Error = SqliteError;
+
+    async fn get_log_height(
+        &self,
+        public_key: &p2panda_core::PublicKey,
```
adzialocha: What do you think about also making `public_key` a generic? We already do it for the operation id / hash digest (`ID`)

p2panda-store-next/src/logs/sqlite.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    async fn prune_entries(
+        &self,
+        author: &p2panda_core::PublicKey,
+        log_id: &L,
+        until: &SeqNum,
+    ) -> Result<bool, Self::Error> {
+        todo!()
+    }
```
adzialocha: Maybe return a `u64` instead to indicate how many entries have been pruned?

p2panda-store-next/src/logs/traits.rs
```diff
@@ -0,0 +1,121 @@
...
+///
+/// Here we briefly outline several implementation scenarios:
+///
+/// An application relying on a one-log-per-author design might choose to implement `LogId` for a thin
+/// wrapper around an Ed25519 public key; this effectively ties the log to the public key of the
+/// author. Secure Scuttlebutt (SSB) is an example of a protocol which relies on this model.
+///
+/// In an application where one author may produce operations grouped into multiple logs,
+/// `LogId` might be implemented for a `struct` which includes both the public key of the author
+/// and a unique number for each log instance.
```
adzialocha: Is this still the case? We're having `public_key` as a separate argument from `log_id` in the trait methods, so either this is confusing or not correct anymore

p2panda-store-next/src/logs/traits.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// Get the current height of a log.
+    ///
+    /// Returns the ID and sequence number for the latest entry in a log.
+    ///
+    /// Returns None when the author or a log with the requested id was not found.
+    fn get_log_height(
+        &self,
+        public_key: &PublicKey,
+        log_id: &L,
+    ) -> impl Future<Output = Result<Option<(ID, SeqNum)>, Self::Error>>;
```
adzialocha: Maybe we can make the state vector an associated type or generic?

For classic logs it's an (Hash, SeqNum), for forkable logs it's two HashSets with Hash, one for the frontier, one for the "pruning frontier" (optional).

If we make this generic we can re-use these methods for different log types.

p2panda-store-next/src/logs/sqlite.rs
```diff
@@ -0,0 +1,80 @@
...
+    E: Extensions,
+    L: LogId,
+{
+    type Error = SqliteError;
+
+    async fn get_log_height(
+        &self,
+        public_key: &p2panda_core::PublicKey,
+        log_id: &L,
+    ) -> Result<Option<(Hash, SeqNum)>, Self::Error> {
```
adzialocha: Maybe we can make the state vector an associated type or generic?

For classic logs it's an (Hash, SeqNum), for forkable logs it's two HashSets with Hash, one for the frontier, one for the "pruning frontier" (optional).

If we make this generic we can re-use these methods for different log types.

p2panda-store-next/src/logs/traits.rs
adzialocha: For forkable logs it wouldn't be a `SeqNum` anymore, but a vector of operation id's ("cut" in the graph). So maybe this shouldn't even be called sequence number

p2panda-store-next/src/logs/sqlite.rs
sandreae: Yep, cool, let's go with `author` across the board.

p2panda-store-next/src/logs/traits.rs
sandreae: Yep cool, let's make sequence number generic in the trait.

p2panda-store-next/src/logs/traits.rs
sandreae: Sounds good!

p2panda-store-next/src/logs/sqlite.rs
sandreae: Yep, can go generic in the trait.

p2panda-store-next/src/logs/traits.rs
sandreae: The return type doesn't actually represent a state vector here. The ID is included so that we can get the required backlink (as well as the seq number) for an operation in the forge. I wouldn't make this generic right now, although I agree that when we do move to using state vectors, these should be generic.

sandreae: I've renamed the method to `get_latest_entry` to make that clearer.

adzialocha: For me what you're describing is exactly a state vector (a vector representing the current state of this log). How do you define a state vector?

sandreae: I agree with your definition of a state vector in general. 

In all methods in the current API (except `get_latest_entry`) we refer to a point in a log by only it's sequence number. If we introduce a generic state vector then for consistency I would say it should to be used in all these methods, and if we say the state vector is (for now) `(Hash, SeqNum)` then this would introduce the redundant requirement for a hash to be passed and returned in all places. I don't dislike this entirely, as it certainly points in the direction we expect to go, but it feels a little premature, before we actually require use of the hash (or any other state vector, eg. `[(Hash, SeqNum)]`.

I'm open to this discussion though. I had considered using Hash in place of SeqNum to refer to entries in a log. This was assuming the current sql implementations would then fetch the seq number internally for efficient queries. 

Right now for me it feels ok that `get_latest_entry` has a slightly different purpose (specifically for getting operation arguments) and therefore returns the hash and sequence number, in all other places we simply use sequence number in arguments and return types as this is all we need right now. This is done understanding in a later step we will likely update the API and queries.

p2panda-store-next/src/logs/sqlite.rs
```diff
@@ -0,0 +1,81 @@
...
+        &self,
+        author: &PublicKey,
+        log_id: &L,
+        after: Option<SeqNum>,
+        until: Option<SeqNum>,
+    ) -> Result<Option<(u64, u64)>, Self::Error> {
+        todo!()
+    }
+
+    async fn get_log_entries(
```
sandreae: @mycognosist we didn't talk about this on the call before, but rather than returning an `Operation<E>` here we discussed returning a wrapper type which includes the operation as well as the header bytes themselves. This saves the need for having a separate `get_raw_log_entries(..)` method which both reduces the API surface and allows for fewer db calls during sync while avoiding encoding to bytes unnecessarily during sync.

sandreae: The same goes for the batch version of this method if that ends up being implemented.

p2panda-store-next/src/logs/traits.rs
adzialocha: That makes sense!

How about renaming the method then to `get_latest_entry_id` as this is what you're returning (not the operation) + the state vector?

p2panda-store-next/src/logs/traits.rs
adzialocha: And maybe we can revisit the trait bounds, like `Debug` when we've implemented the concrete SQLite store

p2panda-store-next/src/logs/traits.rs
sandreae: Yeah, sounds good!

p2panda-core/src/test_utils.rs
```diff
@@ -24,10 +24,20 @@ impl TestLog {
         }
     }
 
+    pub fn new_from_private_key(private_key: PrivateKey) -> Self {
```
adzialocha: I think the Rusty way of naming constructors as such would be `from_private_key`

p2panda-store-next/src/logs/sqlite/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+        for log in logs {
+            let encoded_log_id =
+                encode_cbor(&log).map_err(|err| SqliteError::Encode("log id".to_string(), err))?;
+            encoded_log_ids.push(encoded_log_id);
+        }
+
+        // This query formation approach is required since there is currently no
+        // way to directly bind arrays as comma-separated lists in sqlx.
+        let params = format!("?{}", ", ?".repeat(encoded_log_ids.len() - 1));
+        println!("{params}");
```
adzialocha: Remove println

p2panda-store-next/src/logs/sqlite/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+        until: Option<SeqNum>,
+    ) -> Result<Option<(u64, u64)>, Self::Error> {
+        // We need to use an inclusive greater-than to ensure our
+        // query includes the operation with sequence number 0.
+        let after_operator = if after.is_none() { ">=" } else { ">" };
+        let query_str = format!(
+            "
+            SELECT
+                CAST(SUM(CAST(header_size AS NUMERIC)) AS TEXT) AS total_header_bytes,
+                CAST(SUM(CAST(payload_size AS NUMERIC)) AS TEXT) AS total_payload_bytes,
```
adzialocha, line 131: Converting from u64 to TEXT is really annoying when using SQLite (because it's not supported) and it should work most of the time, except when we do math on them _inside_ of SQLite. I assume this will break if the sum of all header_size and payload_size exceeds the max. interger in SQLite.

This is very much an edge-case but I think we should acknowledge that this is broken and either document it or disallow any large integers for the affected fields already on Rust level?

p2panda-store-next/src/logs/sqlite/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+        );
+
+        let log_meta: Option<LogMetaRow> = query_as::<_, LogMetaRow>(&query_str)
+            .bind(author.to_string())
+            .bind(
+                encode_cbor(&log_id)
+                    .map_err(|err| SqliteError::Encode("log id".to_string(), err))?,
+            )
+            .bind(after.unwrap_or(0).to_string())
+            .bind(until.unwrap_or(u64::MAX).to_string())
```
adzialocha, line 151: What is `u64::MAX` when cast as `NUMERIC` in SQLite (as it can't be represented)?

I assume SQLite will convert it into a REAL (floating point number) and do the comparison, which would work as assumed, as long as the other number is not very high as well.

Similar to the other comment I believe this is a behaviour we should document.

p2panda-store-next/src/logs/mod.rs
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+mod sqlite;
```
adzialocha, line 4: This should be behind the `sqlite` feature flag

p2panda-store-next/src/operations/sqlite.rs
```diff
@@ -55,10 +57,13 @@ where
                 .bind(operation.header.payload_size.to_string())
                 .bind(operation.header.payload_hash.map(|hash| hash.to_hex()))
                 .bind(operation.header.timestamp.to_string())
+                .bind(operation.header.seq_num.to_string())
                 .bind(
                     encode_cbor(&operation.header)
                         .map_err(|err| SqliteError::Encode("header".to_string(), err))?,
                 )
+                // TODO: Is this the correct way to determine header size?
+                .bind(operation.header.to_bytes().len().to_string())
```
adzialocha: Yes, it is!

p2panda-store-next/src/logs/sqlite/mod.rs
adzialocha: I've tried it out:

u64::MAX is 18446744073709551615

```
sqlite> SELECT CAST('18446744073709551615' AS NUMERIC);
1.84467440737096e+19
```

p2panda-store-next/src/logs/sqlite/mod.rs
sandreae: Oops, that was me :-)

p2panda-store-next/src/logs/traits.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// Some applications might require semantic grouping of operations. For example, a chat
+/// application may choose to create a separate log for each author-channel pairing. In such a
+/// scenario, `LogId` might be implemented for a `struct` containing a `String` representation of
+/// the channel name.
+///
+/// Finally, please note that implementers of `LogId` must take steps to ensure their log design is
+/// fit for purpose and that all operations have been thoroughly validated before being persisted.
+/// No such validation checks are provided by `p2panda-store`.
+pub trait LogId: Clone + Debug + Eq + std::hash::Hash {}
```
sandreae: This is in `p2panda-core` now so we can import it from there.

p2panda-store-next/src/logs/sqlite/mod.rs
mycognosist: Good call! For now I've opened an issue and we can tackle this as part of a wider documentation effort.

https://github.com/p2panda/p2panda/issues/1021

PR #1007, opened 2026-02-17
Address book SQLite implementation and refactorings

adzialocha:
- [x] Move `AddressBookStore` trait, tests and in-memory implementation from `p2panda-discovery` to `p2panda-store-next`
- [x] Implement SQLite store variant

### DIFF
--- a/p2panda-store-next/src/address_book/memory.rs
+++ b/p2panda-store-next/src/address_book/memory.rs
@@ -3,6 +3,7 @@
 use std::cell::RefCell;
 use std::collections::{BTreeMap, HashSet};
 use std::convert::Infallible;
+use std::fmt::Debug;
 use std::rc::Rc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
@@ -10,6 +11,7 @@ use rand::Rng;
 use rand::seq::IteratorRandom;
 
 use crate::address_book::{AddressBookStore, NodeInfo};
+use crate::memory::MemoryStore;
 
 #[derive(Clone, Debug)]
 pub struct AddressBookMemoryStore<R, ID, N> {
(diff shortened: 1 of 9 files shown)


p2panda-store-next/src/address_book/memory.rs
```diff
@@ -0,0 +1,198 @@
...
+use crate::address_book::{AddressBookStore, NodeInfo};
+use crate::memory::MemoryStore;
+
+#[derive(Clone, Debug)]
+pub struct AddressBookMemoryStore<R, ID, N> {
+    rng: Rc<RefCell<R>>,
+    node_infos: Rc<RefCell<BTreeMap<ID, N>>>,
+    node_infos_last_changed: Rc<RefCell<BTreeMap<ID, u64>>>,
+    topics: Rc<RefCell<BTreeMap<ID, HashSet<[u8; 32]>>>>,
+}
```
adzialocha, line 22: I've moved the in-memory implemention over as well as it already existed in `p2panda-discovery`.

This might likely be removed in the future next to all other in-memory stores.

p2panda-store-next/src/address_book/tests.rs
```diff
@@ -0,0 +1,242 @@
...
+        .await
+        .unwrap();
+
+    <SqliteStore<'_> as AddressBookStore<TestNodeId, TestNodeInfo>>::set_topics(
+        &store,
+        billie,
+        HashSet::from_iter([cats, dogs, rain]),
+    )
+    .await
+    .unwrap();
```
adzialocha, line 70: Setting the unknown types for Rust makes this very cumbersome, I hope that when having proper trait bounds and type descriptions in a method signature where a store gets called, this goes away.

p2panda-store-next/src/address_book/traits.rs
```diff
@@ -0,0 +1,121 @@
...
+    ///
+    /// Stale nodes should not be considered for connection attempts anymore and should not be
+    /// shared during discovery with other nodes.
+    fn is_stale(&self) -> bool;
+
+    /// Returns attached transport information for this node, if available.
+    fn transports(&self) -> Option<Self::Transports>;
+}
+
+/// Interface for storing, managing and querying information about nodes.
```
sandreae, line 37: Just a comment about naming really, but "address book" maybe doesn't do full justice to the functionality contained in this interface. Holding node info is very address book like, but mapping nodes to their topics of interest is closer to the `TopicStore` functionality. Don't think any immediate change is required as once you look at the APIs it's clear what's going on here.

adzialocha: Could be separated into two stores maybe, one is concerned with bootstrapping, metrics, staleness etc. and the other with mapping nodes to topics for "topic / service discovery"

sandreae: Yeah, that did cross my mind. Don't really feels it's necessary right now though.

adzialocha: Made a ticket here: https://github.com/p2panda/p2panda/issues/1015

Issue #1008, opened 2026-02-18
Move `LogId` and `Topic` into `p2panda-core`


sandreae:
Both are generally useful and used in multiple other crates.

PR #1010, opened 2026-02-18
Move LogId and Topic into p2panda-core

sandreae:
Move `LogId` trait from `p2panda-store` and `Topic` struct from p2panda into `p2panda-core`.

closes: #1008 

## 📋 Checklist

- [ ] ~Add tests that cover your changes~
- [ ] ~Add this PR to the _Unreleased_ section in `CHANGELOG.md`~


p2panda-core/src/logs.rs
```diff
@@ -0,0 +1,33 @@
...
+///
+/// Some applications might require semantic grouping of operations. For example, a chat
+/// application may choose to create a separate log for each author-channel pairing. In such a
+/// scenario, `LogId` might be implemented for a `struct` containing a `String` representation of
+/// the channel name.
+///
+/// Finally, please note that implementers of `LogId` must take steps to ensure their log design
+/// is fit for purpose and that all operations have been thoroughly validated before being
+/// persisted. No such validation checks are provided by `p2panda-store`.
+pub trait LogId: Clone + Eq + StdHash + Serialize + for<'de> Deserialize<'de> {}
```
adzialocha, line 31: Maybe this should go into the `traits` module instead

p2panda-net/src/sync/log_sync/api.rs
```diff
@@ -43,7 +42,7 @@ use crate::sync::log_sync::Builder;
 pub struct LogSync<S, L, E, TM>
 where
     S: OperationStore<L, E> + LogStore<L, E> + Send + 'static,
-    L: LogId + Serialize + for<'de> Deserialize<'de> + Send + 'static,
+    L: LogId + Debug + Send + 'static,
```
mycognosist, line 45: Any downside to moving the `Debug` bound onto `LogId`?

adzialocha: There's not many places right now where this is used, I wouldn't make it global yet

p2panda-core/src/logs.rs
sandreae: I did that at first, but then I thought it's an important enough concept to have it's own module. This follows the same convention we use for the `extensions` module which also only contains traits.

sandreae: We could move all traits into the `traits` module.

adzialocha: I see! .. Or move everything from traits into their regarding modules. Let's look into this when things have settled a bit and then we clean up!

PR #1011, opened 2026-02-18
`TopicStore` trait and SQLite implementation

### LINKED ISSUE #1009: SQLite implementation for `TopicStore`
Previously known as `TopicMap`, the concrete implementation will be designed specifically for use in p2panda `Node`.

sandreae:
Introduce `TopicStore` trait in `p2panda-store` (previously `TopicMap` in `p2panda-sync`) and implement an opinionated SQLite version for use in p2panda `Node`. 

closes #1009 

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ Highlights are marked with a pancake 🥞
 
 - Fix missing gossip events in sync manager [#988](https://github.com/p2panda/p2panda/pull/988)
 
+### Added
+
+- Add `TopicStore` trait and SQLite implementation [#1011](https://github.com/p2panda/p2panda/pull/1011)
+
 ## [0.5.1] - 09/02/2026
 
 ### Changed
--- a/p2panda-store-next/migrations/20260217222329_create_topics.sql
+++ b/p2panda-store-next/migrations/20260217222329_create_topics.sql
@@ -1,0 +1,9 @@
+-- SPDX-License-Identifier: MIT OR Apache-2.0
+
+CREATE TABLE IF NOT EXISTS topics_v1 (
+    topic                   TEXT            NOT NULL,
+    author                  VARCHAR(64)     NOT NULL,
+    data_id                 BLOB            NOT NULL,
+
+    UNIQUE (topic, author, data_id)
+);
--- a/p2panda-store-next/src/topics/mod.rs
+++ b/p2panda-store-next/src/topics/mod.rs
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! `TopicStore` trait for managing mappings of application data to topics as well as a concrete
+//! `SqliteStore` implementation.
+mod sqlite;
+#[cfg(test)]
+mod tests;
+mod traits;
+
+pub use traits::TopicStore;
--- a/p2panda-store-next/src/topics/sqlite.rs
+++ b/p2panda-store-next/src/topics/sqlite.rs
@@ -1,0 +1,124 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::collections::HashMap;
+
+use p2panda_core::cbor::{decode_cbor, encode_cbor};
+use p2panda_core::{LogId, PublicKey, Topic};
+use sqlx::{query, query_as};
+
+use crate::sqlite::{SqliteError, SqliteStore};
+use crate::topics::TopicStore;
+
+/// SQLite `TopicStore` implementation that can be used to map a topic to a set of (generic)
+/// per-author data identifiers.
+impl<'a, S> TopicStore<Topic, PublicKey, S> for SqliteStore<'a>
+where
+    S: LogId,
+{
+    type Error = SqliteError;
+
+    async fn associate(
+        &self,
+        topic: &Topic,
+        author: &PublicKey,
+        data_id: &S,
+    ) -> Result<bool, SqliteError> {
+        let result = self
+            .tx(async |tx| {
+                query(
+                    "
+                    INSERT OR IGNORE
+                    INTO
+                        topics_v1 (
+                            topic,
+                            author,
+                            data_id
+                        )
+                    VALUES
+                        (?, ?, ?)
+                    ",
+                )
+                .bind(topic.to_string())
+                .bind(author.to_string())
+                .bind(
+                    encode_cbor(&data_id)
+                        .map_err(|err| SqliteError::Encode("data_id".to_string(), err))?,
+                )
+                .execute(&mut **tx)
+                .await
+                .map_err(SqliteError::Sqlite)
+            })
+            .await?;
+        Ok(result.rows_affected() > 0)
+    }
+
+    async fn remove(
+        &self,
+        topic: &Topic,
+        author: &PublicKey,
+        data_id: &S,
+    ) -> Result<bool, SqliteError> {
+        let result = self
+            .tx(async |tx| {
+                query(
+                    "
+                    DELETE FROM
+                        topics_v1
+                    WHERE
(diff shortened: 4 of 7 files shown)


p2panda-store-next/migrations/20260217222329_create_topics.sql
```diff
@@ -0,0 +1,9 @@
+-- SPDX-License-Identifier: MIT OR Apache-2.0
+
+CREATE TABLE IF NOT EXISTS topics_v1 (
+    topic                   TEXT            NOT NULL,
```
adzialocha, line 4: How about calling this field `topic_digest`?

adzialocha: Or just use the string representation of `Topic` itself, afaik it should implement `Display` and if not we should maybe just implement it?

p2panda-store-next/src/topics/sqlite.rs
```diff
@@ -0,0 +1,92 @@
...
+                        topics_v1 (
+                            topic_digest,
+                            public_key,
+                            scope
+                        )
+                    VALUES
+                        (?, ?, ?)
+                    ",
+                )
+                .bind(calculate_hash(topic).to_string())
```
adzialocha: I believe you can use the string representation of Topic itself, afaik it should implement Display?

adzialocha: (then you can rename the field back to `topic` as well)

p2panda-store-next/src/topics/traits.rs
```diff
@@ -0,0 +1,51 @@
...
+/// for example "Chat Group 2" we would implement `TopicStore` to give us all append-only logs of
+/// all members inside this group, that is the entries inside logs `A2`, `B2` and `C2`.
+pub trait TopicStore<T, A, S, R> {
+    type Error: Error;
+
+    fn associate(
+        &self,
+        topic: &T,
+        author: &A,
+        scope: &S,
```
adzialocha: Scope is the set of data-types (for example logs) which is associated with this topic, right?

I understand topic as something like a scope, maybe we can find something which is less similar? Don't have any good ideas right now, but "data" or "related_items" etc.?

And why again is scope `S` and not `&[R]`? I remember we've discussed something, but don't have access to the notes.

sandreae: > And why again is scope S and not &[R]? I remember we've discussed something, but don't have access to the notes.

In the case of our initial p2panda `Node` implementation `S` will be `Topic` (standing in for `Path` later) and `R` is `HashMap<PublicKey, Vec<Topic>>` (where `Topic` is used as the log id, again standing in for `Path`). For them to be the same (generic) type I believe we'd need to change the `associate` signature to be something more like `associate(topic: &T, data: &R)`. Does that answer your question?

sandreae: I agree that the `scope` argument name is bad though.... maybe `data` is better :thinking:

sandreae: Or just `id` ?

sandreae: Have a look at the trait definition now @adzialocha I made one change. I think what you were getting at with the return type re-using the `S` (now `ID`) generic type was a good one. Although this makes the trait definition less generic, I think it's to an acceptable level considering the possible use cases we have in mind and the fact that we want to offer a generally useful implementation, and for that must decide on sensible database layouts.

p2panda-store-next/src/topics/mod.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! `TopicStore` trait for managing mappings of application data to topics and concrete
+//! `SqliteStore` implementation.
```
adzialocha: Mini: ".. and concrete `SqliteStore` implementation." reads a bit as if we're mapping application data to topics AND a concrete store implementation. Maybe this can be two separate sentences.

p2panda-store-next/migrations/20260217222329_create_topics.sql
```diff
@@ -0,0 +1,9 @@
...
+    public_key              TEXT            NOT NULL,
```
adzialocha: Could be renamed to `author` to use the same naming as in the traits etc.

adzialocha: You can also write `VARCHAR(64)` for the type. It will be internally changed to a `TEXT` by SQLite but could be nice to indicate more concretly what it is in the schema.

p2panda-store-next/src/topics/sqlite.rs
```diff
@@ -0,0 +1,118 @@
...
+        let mut result: HashMap<PublicKey, Vec<S>> = HashMap::new();
+
+        for (author, data_id) in data_ids {
+            let author: PublicKey = author
+                .parse()
+                .expect("values from database encoded correctly");
+
+            let data_id =
+                decode_cbor(&data_id[..]).expect("values from database encoded correctly");
+
```
adzialocha, line 117: We should also handle decode errors (see other sqlite impls), they are highly unlikely but it'll be good to have that in case applications roll out updates, encodings changed, and we don't want them to panic but handle migration paths instead etc.

p2panda-store-next/src/topics/sqlite.rs
```diff
@@ -0,0 +1,118 @@
...
+            result.entry(author).or_default().push(data_id);
```
adzialocha, line 119: Maybe nice to note here that the returned result will only contain a set of data id's where each item is unique since the SQL UNIQUE constraint takes care of that.

p2panda-store-next/src/topics/traits.rs
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::{collections::HashMap, error::Error};
```
adzialocha: Mini: Formatting

p2panda-store-next/src/topics/traits.rs
```diff
@@ -0,0 +1,61 @@
...
+/// "Chat group 2 with members A, B and C"
+/// - Log A2
+/// - Log B2
+/// - Log C2
+/// ```
+///
+/// If we implement `T` to express that we're interested in syncing over a specific chat group,
+/// for example "Chat Group 2" we would implement `TopicStore` to give us all append-only logs of
+/// all members inside this group, that is the entries inside logs `A2`, `B2` and `C2`.
+pub trait TopicStore<T, A, ID> {
```
adzialocha, line 41: I like this version of `TopicStore`! It's certainly easier to read and I can see the upgrade path as well for paths, etc.

p2panda-store-next/src/topics/sqlite.rs
```diff
@@ -0,0 +1,118 @@
...
+                data_id
+            FROM
+                topics_v1
+            WHERE
+                topic = ?
+            ",
+        )
+        .bind(topic.to_string())
+        .fetch_all(&self.pool)
+        .await?;
```
adzialocha: Ah yes, making the pool field public is one way of doing this, another is to use `execute`. I'm fine with either.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Highlights are marked with a pancake 🥞
 ### Changed
 
 - Update iroh to v0.96.1 [#978](https://github.com/p2panda/p2panda/pull/978)
+- Add `TopicStore` trait and SQLite implementation [#1011](https://github.com/p2panda/p2panda/pull/1011)
```
adzialocha: I believe this should go under "Added"

PR #1014, opened 2026-02-19
Introduce `LogId` to `OperationStore`

mycognosist:
Here we add an additional generic type parameter `C` as a requirement on the `OperationStore` trait. The generic specifies a collection id which has been added as a parameter to the `insert_operation()` method. This is useful for the SQLite operation store so that we can easily query by log id when making log-related queries.

https://github.com/p2panda/p2panda/issues/1013

## 📋 Checklist

- [x] ~~Add this PR to the _Unreleased_ section in `CHANGELOG.md`~~

### DIFF
--- a/p2panda-core/src/test_utils.rs
+++ b/p2panda-core/src/test_utils.rs
@@ -4,13 +4,14 @@ use std::cell::RefCell;
 use std::rc::Rc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use crate::{Body, Extensions, Hash, Header, Operation, PrivateKey};
+use crate::{Body, Extensions, Hash, Header, Operation, PrivateKey, Topic};
 
 #[derive(Clone, Default)]
 pub struct TestLog {
     private_key: PrivateKey,
     backlink: Rc<RefCell<Option<Hash>>>,
     seq_num: Rc<RefCell<u64>>,
+    log_id: Topic,
 }
 
 impl TestLog {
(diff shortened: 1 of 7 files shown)


p2panda-store-next/src/operations/traits.rs
```diff
@@ -6,7 +6,7 @@ use std::error::Error;
 ///
 /// The concrete type of an "operation" is generic and implementors can use the same interface for
 /// different approaches: sets, append-only logs, hash-graphs (DAG) etc.
-pub trait OperationStore<T, ID> {
+pub trait OperationStore<T, ID, L = u64> {
```
sandreae: Haha, I know I suggested doing this, but actually maybe it's better to not have a default on this level.

p2panda-store-next/src/operations/traits.rs
```diff
@@ -17,6 +17,7 @@ pub trait OperationStore<T, ID> {
         &self,
         id: &ID,
         operation: T,
+        log_id: L,
```
sandreae: I wonder if we could rename this argument `collection` or something less log specific? This would then not tie the implementation to logs necessarily, as is currently the case.

sandreae: Or `collection_id` ?

p2panda-core/src/test_utils.rs
```diff
@@ -1,16 +1,44 @@
...
+    pub fn from_rng<R: Rng>(rng: &mut R) -> Self {
+        Self(rng.r#gen())
+    }
+}
+
+impl Default for TestLogId {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
adzialocha: Minor suggestion, we could also use the new Topic struct here

PR #1016, opened 2026-02-19
Remove in-memory store in `p2panda-store-next` and use SQLite in `p2panda-stream-next`

adzialocha:
Towards https://github.com/p2panda/p2panda/issues/1017

As part of this PR I also worked a bit on database tooling since it became clear quite fastly that we need a bit more:

- A way to fastly create a temporary database for tests -> `SqliteStore::temporary` builder
- A way to fastly express transactions, avoiding all of the boilerplate -> `tx` and `tx_unwrap` macro
- A way to bring in a transaction API into places where the store is generic -> `Transaction` trait

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Highlights are marked with a pancake 🥞
 ### Changed
 
 - Address book SQLite implementation and refactorings [#1007](https://github.com/p2panda/p2panda/pull/1007)
+- Remove in-memory store in `p2panda-store-next` and use SQLite in `p2panda-stream-next` [#1016](https://github.com/p2panda/p2panda/pull/1016)
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


p2panda-store-next/src/sqlite.rs
```diff
@@ -265,35 +308,10 @@ impl<'a> SqliteStore<'a> {
...
-    where
-        F: AsyncFnOnce(&sqlx::SqlitePool) -> Result<R, SqliteError>,
-    {
-        f(&self.pool).await
-    }
 }
 
 #[allow(unused)]
-pub struct TransactionPermit<'a>(SemaphorePermit<'a>);
+pub struct TransactionPermit(OwnedSemaphorePermit);
```
adzialocha, line 314: Using an owned semaphore permit here instead allows us to omit the lifetime. This makes building the `Transaction` trait a bit easier (no GAT, etc.)

PR #1022, opened 2026-02-23
Use new `SqliteStore` for log, operation and topic stores in `p2panda-sync` and `p2panda-net`

sandreae:
Use `LogStore`, `OperationStore` and `TopicStore` from `p2panda-store-next` in `p2panda-sync` and `p2panda-net`. The topic store replaces `TopicMap` which has now been removed. 

- [x] use new SQLite stores in `p2panda-net` and `p2panda-sync` 
- [x] remove `TopicMap`
- [x] use `BTreeMap` in store and sync APIs to represent sets of author logs in order to have consistent ordering
- [x] bonus: move log diffing method from `p2panda-sync` to `p2panda-core` 

## 📋 Checklist


p2panda-core/src/logs.rs
```diff
@@ -28,6 +29,205 @@ use serde::{Deserialize, Serialize};
...
+///
+/// Local and remote states are represented by a map of authors to logs, where the logs are
+/// represented by their unique identifier and current log height. If the remote is not aware of a
+/// log, then the range containing all entries in the local log will be contained in the diff, if
+/// the remote knows of some entries in a log, then the range representing only the entries they
+/// need will be included.
+///
+/// Log ranges are represented by `(Option<u64>, Option<u64>)` tuples where the first value is an
+/// exclusive "from" sequence number and the later is an inclusive "until" sequence number. If
+/// either values are `None` then signifies that all entries from the start, or to the end, are
```
mycognosist:
```suggestion
/// either values are `None` that signifies that all entries from the start, or to the end, are
```

p2panda-net/examples/chat.rs
```diff
@@ -151,10 +122,10 @@ async fn main() -> Result<()> {
     };
 
     // Set up sync for p2panda operations.
```
mycognosist: This comment could be updated to something like:

```rust
// Create a temporary SQLite database to store topic associations and operations.
let store = SqliteStore::temporary().await;

// Associate our local log with the chat topic and commit it to the database.
let permit = store.begin().await?;
store.associate(&topic_id, &public_key, &LOG_ID).await?;
store.commit(permit).await?;
````

File: p2panda-net/examples/chat.rs [hunk 3]
```diff
@@ -260,7 +231,7 @@ async fn main() -> Result<()> {
         });
     }
 
-    let sync = LogSync::builder(store.clone(), topic_map.clone(), endpoint, gossip)
+    let sync = LogSync::<_, u64, _, _>::builder(store.clone(), store.clone(), endpoint, gossip)
```
mycognosist: Could we use `LogId` in place of `u64` to help communicate what the turbofish is defining?

p2panda-net/examples/chat.rs
```diff
@@ -283,7 +254,13 @@ async fn main() -> Result<()> {
                         );
                     }
                     SyncEvent::Operation(operation) => {
-                        if store.has_operation(operation.hash).await.unwrap() {
+                        if <SqliteStore as OperationStore<Operation, Hash, u64>>::has_operation(
```
mycognosist: Here's another place where we could replace `u64` with `LogId`.

p2panda-stream/src/orderer/orderer.rs
```diff
@@ -3,7 +3,7 @@
 use std::marker::PhantomData;
 
 // @TODO: Change this to p2panda_store when ready.
```
mycognosist: Remove `TODO` comment.

p2panda-net/src/sync/log_sync/api.rs
```diff
@@ -39,15 +39,15 @@ use crate::sync::log_sync::Builder;
...
 #[derive(Clone)]
-pub struct LogSync<S, L, E, TM>
+pub struct LogSync<S, L, E, TS>
 where
-    S: OperationStore<L, E> + LogStore<L, E> + Send + 'static,
+    S: LogStore<Operation<E>, PublicKey, L, u64, Hash> + Clone + Send + 'static,
     L: LogId + Debug + Send + 'static,
     E: Extensions + Send + 'static,
-    TM: TopicMap<TopicId, Logs<L>> + Send + 'static,
+    TS: TopicStore<TopicId, PublicKey, L> + Clone + Send + 'static,
```
adzialocha: Is it possible to merge `S` with `TS`?

p2panda-store-next/src/topics/tests.rs
```diff
@@ -89,6 +89,28 @@ async fn path_based_log_ids() {
     assert_eq!(logs, expected_logs);
 }
 
+#[tokio::test]
+async fn integer_log_ids() {
```
adzialocha: What does this test check against?

p2panda-stream/src/test_utils.rs
```diff
@@ -4,9 +4,14 @@ use std::collections::VecDeque;
 use std::task::Poll;
 
 use futures_test::task::noop_context;
+use p2panda_core::PublicKey;
+use serde::{Deserialize, Serialize};
 use tokio::pin;
 use tokio::sync::Notify;
 
+#[derive(Clone, Debug, Default, Hash, Eq, PartialOrd, Ord, PartialEq, Serialize, Deserialize)]
+pub struct StreamName(PublicKey, Option<String>);
```
adzialocha: Where is this used?

p2panda-sync/src/protocols/log_sync.rs
```diff
@@ -306,64 +304,81 @@ where
...
+                                        );
+                                        continue;
+                                    };
+
+                                    // For testing, where the store is an in memory implementation
+                                    // which may not yield work to other tasks, we force the task to
+                                    // yield in order to test concurrent database actions occurring
+                                    // during sync.
+                                    #[cfg(test)]
+                                    tokio::task::yield_now().await;
```
adzialocha: How does this relate to the in-memory implementation? I assume in the past it was because there was no .await point in the store code where the scheduler could switch futures, is that still the case though?

p2panda-net/Cargo.toml
```diff
@@ -71,9 +71,8 @@ iroh = { version = "0.96.1", default-features = false, optional = true }
 iroh-base = { version = "0.96.1", optional = true }
 iroh-gossip = { version = "0.96.0", optional = true }
 p2panda-core = { path = "../p2panda-core", version = "0.5.1" }
-p2panda-discovery = { path = "../p2panda-discovery", version = "0.5.1", optional = true }
-p2panda-store = { path = "../p2panda-store", version = "0.5.1", optional = true }
-p2panda-store-next = { path = "../p2panda-store-next", version = "0.5.1", optional = true }
+p2panda-discovery = { path = "../p2panda-discovery", version = "0.5.1", features = ["test_utils"], optional = true }
+p2panda-store = { package = "p2panda-store-next", path= "../p2panda-store-next", version = "0.5.1", features = ["test_utils"], optional= true }
```
adzialocha: Is the `test_utils` feature necessary for release builds?

p2panda-net/src/sync/log_sync/api.rs
sandreae: I believe we need to keep them as two separate separate types for the case where users want to provide their own `TopicStore` implementation, but still use our `SqliteStore` for everything else. 

If we don't want to support that feature we could combine them.

adzialocha: The TopicStore itself is already pretty flexible, for now I would simplify things and if the need arises consider the options?

sandreae: Nice, that's also my preferred approach.

p2panda-stream/src/test_utils.rs
sandreae: It's not! Think it's a rebase artefact, I'll remove it.

p2panda-store-next/src/topics/tests.rs
sandreae: I added this test when I was trying to figure out a decode error coming from the db, it wasn't related in the end though, so could remove this test as it doesn't add any particular value.

p2panda-sync/src/protocols/log_sync.rs
sandreae: Yeah, this is probably not required any more, I'll check and then remove it if we can.

p2panda-net/Cargo.toml
sandreae: No, you're right, not for release builds. It's needed for "test_utils" feature, but we can activate the features on these two crates conditionally.

Issue #1027, opened 2026-02-24
Define and implement `Forge`


mycognosist:
- generic trait with a concrete impl (using new store)
- creates and holds the private key
- create and sign operations
- provides public key
- atomic

PR #1032, opened 2026-02-25
Define and implement `Forge`

mycognosist:
Introduces a `Forge` trait interface for creating signed operations and holding
the signing keypair.

We also introduce a concrete `OperationForge` which implements the `Forge` trait
and provides a convenient interface which internally composes operations, signs
them, inserts them into the store and associates the operation log id with the
topic.

The `OperationForge` has been incorporated into the `Node` and is used internally
by the `StreamHandle` when messages are published.

Closes: https://github.com/p2panda/p2panda/issues/1027

## 📋 Checklist

### DIFF
--- a/p2panda/src/builder.rs
+++ b/p2panda/src/builder.rs
@@ -5,13 +5,13 @@ use std::net::{Ipv4Addr, Ipv6Addr};
 use p2panda_core::PrivateKey;
 use p2panda_net::discovery::DiscoveryConfig;
 use p2panda_net::gossip::GossipConfig;
-use p2panda_net::iroh_endpoint::{IrohConfig, RelayUrl};
+use p2panda_net::iroh_endpoint::RelayUrl;
 use p2panda_net::iroh_mdns::MdnsDiscoveryMode;
 use p2panda_net::{NetworkId, NodeId};
 use p2panda_store::sqlite::SqliteStoreBuilder;
 
 use crate::Node;
-use crate::network::NetworkConfig;
+use crate::forge::OperationForge;
 use crate::node::{AckPolicy, Config, NodeError};
 
 #[derive(Default)]
@@ -105,7 +105,8 @@ impl NodeBuilder {
     pub async fn spawn(self) -> Result<Node, NodeError> {
         let private_key = self.private_key.unwrap_or_default();
         let store = self.store.build().await?;
+        let forge = OperationForge::from_private_key(private_key, store.clone());
 
-        Node::spawn_inner(self.config, private_key, store).await
+        Node::spawn_inner(self.config, store, forge).await
     }
 }
--- a/p2panda/src/forge.rs
+++ b/p2panda/src/forge.rs
@@ -1,0 +1,210 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::error::Error as StdError;
+use std::time::{SystemTime, SystemTimeError, UNIX_EPOCH};
+
+use p2panda_core::{Body, Hash, PrivateKey, PublicKey, SeqNum, Topic};
+use p2panda_store::logs::LogStore;
+use p2panda_store::operations::OperationStore;
+use p2panda_store::topics::TopicStore;
+use p2panda_store::{SqliteError, SqliteStore, tx};
+use thiserror::Error;
+
+use crate::operation::{Extensions, Header, Operation};
+
+/// Interface for obtaining a keypair and creating signed operations.
+pub trait Forge<T, C, E> {
+    type Error: StdError;
+
+    fn private_key(&self) -> &PrivateKey;
+
+    fn public_key(&self) -> PublicKey;
+
+    fn create_operation(
+        &mut self,
+        topic: T,
+        collection_id: C,
+        body: Option<Vec<u8>>,
+        extensions: E,
+    ) -> impl Future<Output = Result<Option<p2panda_core::Operation<E>>, Self::Error>>;
+}
+
+#[derive(Clone, Debug)]
+pub struct OperationForge {
+    private_key: PrivateKey,
+    store: SqliteStore<'static>,
+}
+
+impl OperationForge {
+    /// Create a forge for inserting signed operations into the database and associating topics
+    /// with logs.
+    ///
+    /// The forge holds the private key used to sign operations. This method generates a new key
+    /// using CSPRNG from the system.
+    pub fn new(store: SqliteStore<'static>) -> Self {
+        Self {
+            private_key: PrivateKey::new(),
+            store,
+        }
+    }
+
+    /// Create a forge using an existing private key.
+    pub fn from_private_key(private_key: PrivateKey, store: SqliteStore<'static>) -> Self {
+        Self { private_key, store }
+    }
+}
+
+type LogId = [u8; 32];
+
+impl Forge<Topic, LogId, Extensions> for OperationForge {
+    type Error = ForgeError;
+
+    fn private_key(&self) -> &PrivateKey {
+        &self.private_key
+    }
+
+    fn public_key(&self) -> PublicKey {
+        self.private_key.public_key()
+    }
+
+    /// Create a signed operation and insert it into the store.
+    ///
+    /// This method performs several actions: it first queries the store to determine the latest
+    /// entry for the given author and log id. It then composes an operation and signs it. Finally,
+    /// the relevant log is associated with the topic and the signed operation is inserted into the
+    /// store. Both the log-topic association and operation insertion are executed as part of a
+    /// single transaction, thereby ensuring atomicity.
+    async fn create_operation(
+        &mut self,
+        topic: Topic,
+        log_id: LogId,
+        body: Option<Vec<u8>>,
+        extensions: Extensions,
+    ) -> Result<Option<Operation>, Self::Error> {
+        let (seq_num, backlink) = <SqliteStore<'static> as LogStore<
+            Operation,
+            PublicKey,
+            [u8; 32],
+            SeqNum,
+            Hash,
+        >>::get_latest_entry(
+            &self.store, &self.private_key.public_key(), &log_id
+        )
+        .await?
+        .map(|(hash, seq_num)| (seq_num + 1, Some(hash)))
+        .unwrap_or((0, None));
+
+        let body: Option<Body> = body.map(|bytes| bytes.into());
+
+        let timestamp = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
+
+        let mut header = Header {
+            version: 1,
+            public_key: self.private_key.public_key(),
+            signature: None,
+            payload_size: body.as_ref().map(|body| body.size()).unwrap_or(0),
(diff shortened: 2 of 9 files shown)


p2panda/src/forge.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// single transaction, thereby ensuring atomcitiy.
+    async fn create_operation(
+        &mut self,
+        topic_id: TopicId,
+        log_id: LogId,
+        body: Option<Body>,
+        backlink: Option<Hash>,
+        extensions: Extensions,
+    ) -> Result<Option<(Hash, Operation)>, Self::Error> {
+        let (seq_num, backlink) = <SqliteStore<'static> as LogStore<
```
adzialocha, line 84: `backlink` replaces the passed in argument here, I believe we can remove the argument from the method.

p2panda/src/forge.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// store. Both the log-topic association and operation insertion are executed as part of a
+    /// single transaction, thereby ensuring atomcitiy.
+    async fn create_operation(
+        &mut self,
+        topic_id: TopicId,
+        log_id: LogId,
+        body: Option<Body>,
+        backlink: Option<Hash>,
+        extensions: Extensions,
+    ) -> Result<Option<(Hash, Operation)>, Self::Error> {
```
adzialocha: We receive the Hash from `Operation` itself, so maybe no need to return it extra

p2panda/src/streams/stream.rs
```diff
@@ -41,8 +47,19 @@ where
...
+
+        let encoded_message = encode_cbor(&message)?;
+        let body = Body::new(&encoded_message[..]);
+
+        let operation = self
+            .forge
+            .create_operation(self.topic(), self.topic().into(), Some(body), extensions)
+            .await?;
+
+        // TODO: How to handle the case where no insertion occurred?
```
sandreae: I believe at this level it shouldn't be possible to create and insert a duplicate operation (which is the case where `None` is returned). I would say we can either unwrap here, or maybe better to be safe and have a dedicated error variant which would indicate there is a bug in the store / forge implementation itself.

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
...
+    ///
+    /// This method performs several actions: it first queries the store to determine the latest
+    /// entry for the given author and log id. It then composes an operation and signs it. Finally,
+    /// the relevant log is associated with the topic and the signed operation is inserted into the
+    /// store. Both the log-topic association and operation insertion are executed as part of a
+    /// single transaction, thereby ensuring atomcitiy.
+    async fn create_operation(
+        &mut self,
+        topic: Topic,
+        // TODO: Should this also be `Topic`?
```
sandreae: This will be `Topic` in `Node` but I guess in the forge API itself we need it as a separate argument as you have it here.

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
...
+    type Error: StdError;
+
+    fn private_key(&self) -> PrivateKey;
+
+    fn public_key(&self) -> PublicKey;
+
+    fn create_operation(
+        &mut self,
+        topic: T,
+        log_id: L,
```
sandreae: Could we rename this argument `collection_id: C` here to match naming generic naming in the store?

sandreae (changes requested):

p2panda/src/streams/stream.rs
```diff
@@ -41,8 +47,19 @@ where
     }
 
     /// Publish a message.
-    pub async fn publish(&self, _message: M) -> Result<Hash, StreamError> {
-        unimplemented!()
+    pub async fn publish(&mut self, message: M) -> Result<Hash, StreamError> {
+        let extensions = Extensions { version: VERSION };
```
adzialocha: Not important for now but it could be nice to not expose the details (like versioning) of `Extensions` on this level but just have an constructor called instead

p2panda/src/streams/stream.rs
```diff
@@ -41,8 +47,19 @@ where
...
+
+        let encoded_message = encode_cbor(&message)?;
+        let body = Body::new(&encoded_message[..]);
+
+        let operation = self
+            .forge
+            .create_operation(self.topic(), self.topic().into(), Some(body), extensions)
+            .await?;
+
```
adzialocha: We need to publish the operation also into the log sync stream, in case there's a running live-mode session somewhere.

p2panda/src/streams/stream.rs
```diff
@@ -139,4 +156,10 @@ where
 }
 
 #[derive(Debug, Error)]
-pub enum StreamError {}
+pub enum StreamError {
```
adzialocha: Maybe we make this a dedicated `PublishError` type? To prevent our error types to become too large again.

p2panda/src/builder.rs
```diff
@@ -105,7 +106,8 @@ impl NodeBuilder {
     pub async fn spawn(self) -> Result<Node, NodeError> {
         let private_key = self.private_key.unwrap_or_default();
         let store = self.store.build().await?;
+        let forge = OperationForge::from_private_key(private_key.clone(), store.clone());
```
adzialocha: Cloning the private key is not necessary here

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::cell::RefCell;
+use std::error::Error as StdError;
+use std::rc::Rc;
+use std::time::{SystemTime, SystemTimeError, UNIX_EPOCH};
```
adzialocha: I can see some unused types here which is because we've disabled the clippy warning on lib-level. I can recommend commenting it out for a short moment to clean this up.

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
...
+
+type LogId = [u8; 32];
+
+/// Interface for obtaining a keypair and creating signed operations.
+pub trait Forge<T, L, E> {
+    type Error: StdError;
+
+    fn private_key(&self) -> PrivateKey;
+
+    fn public_key(&self) -> PublicKey;
```
adzialocha: Both methods should return a borrowed version of the key, at least the `PrivateKey`, users can always clone if they really need it

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
...
+        self.private_key.public_key()
+    }
+
+    /// Create a signed operation and insert it into the store.
+    ///
+    /// This method performs several actions: it first queries the store to determine the latest
+    /// entry for the given author and log id. It then composes an operation and signs it. Finally,
+    /// the relevant log is associated with the topic and the signed operation is inserted into the
+    /// store. Both the log-topic association and operation insertion are executed as part of a
+    /// single transaction, thereby ensuring atomcitiy.
```
adzialocha: "atomcitiy" typo

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
...
+use p2panda_store::logs::LogStore;
+use p2panda_store::operations::OperationStore;
+use p2panda_store::topics::TopicStore;
+use p2panda_store::{SqliteError, SqliteStore, Transaction, tx};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+use crate::operation::{Extensions, Header, Operation};
+
+type LogId = [u8; 32];
```
adzialocha: Shouldn't this be `Topic`?

adzialocha: maybe good to move it _below_ the trait definition, just to indicate this is already a "specialised" type for a concrete forge implementation

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
...
+            SeqNum,
+            Hash,
+        >>::get_latest_entry(
+            &self.store, &self.private_key.public_key(), &log_id
+        )
+        .await?
+        .map(|(hash, seq_num)| (seq_num + 1, Some(hash)))
+        .unwrap_or((0, None));
+
+        let timestamp = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
```
adzialocha, line 99: I was wondering if we should actually have micro-second precision here instead of seconds?

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
...
+            payload_hash: body.as_ref().map(|body| body.hash()),
+            timestamp,
+            seq_num,
+            backlink,
+            previous: vec![],
+            extensions,
+        };
+
+        header.sign(&self.private_key);
+        let header_bytes = header.to_bytes();
```
adzialocha: Header bytes are not used anymore

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
...
+                &self.private_key.public_key(),
+                &log_id,
+            )
+            .await?;
+
+            self.store
+                .insert_operation(&hash.clone(), operation.clone(), log_id)
+                .await?
+                // Return the operation if the insertion occurred.
+                .then_some(operation)
```
adzialocha: Just a minor code style suggestion: Returning the operation inside the transaction here (even though it came from outside of it) is somewhat hard to read. You could remove the `let result` and `then_some` to keep things separated.

p2panda/src/operation.rs
```diff
@@ -6,8 +6,11 @@ pub type Header = p2panda_core::Header<Extensions>;
 
 pub type Operation = p2panda_core::Operation<Extensions>;
 
+/// Internal value for versioning operations via extensions.
+pub(crate) const VERSION: u64 = 0;
```
adzialocha: This version number is for our own extensions format, operations are already versioned separately

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    fn private_key(&self) -> PrivateKey;
+
+    fn public_key(&self) -> PublicKey;
+
+    fn create_operation(
+        &mut self,
+        topic: T,
+        log_id: L,
+        body: Option<Body>,
```
adzialocha: What do you think about not exposing `Body` here and asking for bytes instead?

p2panda/src/forge.rs
```diff
@@ -0,0 +1,215 @@
...
+        extensions: E,
+    ) -> impl Future<Output = Result<Option<p2panda_core::Operation<E>>, Self::Error>>;
+}
+
+#[derive(Clone, Debug)]
+pub struct OperationForge {
+    private_key: PrivateKey,
+    store: SqliteStore<'static>,
+}
```
adzialocha, line 36: I believe we want the forge to be a singleton, so there's only ever one place where operations are created even when it was cloned (using reference counters).

It doesn't matter too much for now, but latest when we look at the consistency guarantees of the log we might need to revisit this

adzialocha (changes requested):
Well done! ✨ 

Most of my comments are minor questions, publishing to the log sync stream is the only real blocker I think!

PR #1035, opened 2026-02-25
Introduce processor module for ordering and consuming groups control messages

sandreae:
## Context

`p2panda-auth` contains a generic group crdt with messaging protocol, messages must be partially ordered before they are processed in the provided APIs. With all the generics it's flexible enough to be integrated into systems with different core types and approaches to message ordering. However this makes it a little hard to "just use", even with core p2panda types and provided ordering approaches, some scaffolding is required.

## Decision 

We want to introduce a more opinionated API and types which allow a user to satisfy basic ordering requirements with built in support for core p2panda data types. It should still not be opinionated regarding application layer logic around how groups are created and managed (who are the initial members etc..) and it should be designed in a way where messages are forged outside of the API itself and simply passed in for processing. Message persistence and sync should also handled elsewhere.

## Changes

Introduce `processor` module containing types and methods  which can be used for ordering and processing auth groups control messages. 

- `AuthState` state object containing an operation buffer, orderer state and the groups crdt state itself
- `Store` providing a simple in-memory get/set API for the auth state object
- `Processor` struct with one `process(store, operation)` method which processes all group control messages
- `args` and `operation` modules which contain additional arguments expected on operations containing groups control messages and conversion trait for `Operation<E>` into concrete `GroupsOperation` 
- `groups` example demonstrating use with `p2panda-net` and `p2panda-sync` 

## Next steps
- extend store API to allow handing multiple auth states
- changes required to use new SQLite stores being worked on in `development`

## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/processor/mod.rs
+++ b/p2panda-auth/src/processor/mod.rs
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Types and methods for ordering and processing groups control messages.
+mod args;
+mod operation;
+#[allow(clippy::module_inception)]
+mod processor;
+mod store;
+
+pub use args::GroupsArgs;
+pub use operation::GroupsOperation;
+pub use processor::{GroupsProcessor, GroupsProcessorError};
+pub use store::{AuthState, Store};
--- a/p2panda-auth/Cargo.toml
+++ b/p2panda-auth/Cargo.toml
@@ -20,18 +20,34 @@ all-features = true
 workspace = true
 
 [features]
-default = []
-test_utils = ["dep:rand", "serde"]
+default = ["processor"]
+processor = ["dep:tokio", "serde", "dep:p2panda-core", "dep:p2panda-stream"]
+test_utils = ["dep:rand", "serde", "dep:p2panda-core"]
 serde = ["dep:serde"]
+test = ["test_utils"]
 
 [dependencies]
 p2panda-core = { path = "../p2panda-core", version = "0.5.1", optional = true }
+p2panda-stream = { path = "../p2panda-stream", version = "0.5.1", optional = true }
 petgraph = { version = "0.8.1", features = ["serde-1"] }
 rand = { version = "0.9.0", features = ["alloc"], optional = true }
 serde = { version = "1.0.219", features = ["derive"], optional = true }
+tokio = { version = "1.49.0", features = ["sync"], optional = true }
 thiserror = "2.0.12"
 
 [dev-dependencies]
+p2panda-core = { path = "../p2panda-core", version = "0.5.1" }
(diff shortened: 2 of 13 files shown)


p2panda-auth/src/processor/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl IdentityHandle for PublicKey {}
+impl OperationId for Hash {}
+
+/// Process a new groups operation.
+///
+/// Processed operations are first partially ordered, and only processed on the auth groups state
+/// if all their dependencies have been met. If other operations become "ready" by this one, then
+/// they will be all processed in order.
+pub async fn process<M, C, RS>(
```
sandreae: Think I'd rather make this a struct which holds a reference to the store with one `process` method. Later this can be wrapped and implement `Processor` for use in the `p2panda-stream` pipeline.

p2panda-auth/src/processor/extensions.rs
```diff
@@ -0,0 +1,47 @@
...
+            self.header.extension().expect("auth extension is present");
+        extension.group_id
+    }
+
+    fn action(&self) -> GroupAction<PublicKey, C> {
+        let extension: AuthExtension<C> =
+            self.header.extension().expect("auth extension is present");
+        extension.action
+    }
+}
```
maackle: minor nit, it would be a bit more convenient (and less crashy) if the processor just filtered out the non-auth operations itself so that the implementor doesn't have to make sure to do this

p2panda-auth/src/processor/mod.rs
maackle: Are race conditions a concern here, in the time between get and set?

p2panda-auth/src/processor/extensions.rs
sandreae: This is a little misleading, as resolving this extension will never return `None`. We know the concrete `AuthExtension` type, and this will always be present for an operation which has an `E` implementing `Extension<AuthExtension<C>>`. I believe this is a failing of the extensions API. I'll have a think about that. For now I can add an explicit extension impl (instead of relying on the default) to make it explicit that `None` is not possible in this case.

sandreae: I'll consider the filtering out of "non-auth" operations though, this is still a valid point. The case is when you have a single extension for your app and `AuthExtension` are not always present. We can make changes to handle this more elegantly in this API.

sandreae: Ok, in the end I've changed the flow a bit to handle this case. Now there is a concrete `GroupsOperation` type which can be used in the processor arguments. This type can be converted into from the actual p2panda operation, which would fail if the required args are not present in the extensions. There is a blanket `TryFrom` implementation for any operation which have an `E: Extension<GroupsArgs>`. 

Downside to this approach is that it introduces some unnecessary cloning, but I think there might be ways around that.

PR #1045, opened 2026-02-27
Event processing pipeline in Node stream

adzialocha:
For processing incoming operations via eventually consistent "log sync", streams we introduce a `Processor` here. This "event processor" is an async pipeline which looks at every incoming operation and executes work.

Currently this is simply just the `Ingest` processor which validates the operation and moves it into the database. Later we want to add pruning, encryption, access control, .. and other processors here.

Closes: #1025 

## 📋 Checklist

### DIFF
--- a/p2panda-store/migrations/20260217222329_create_topics.sql
+++ b/p2panda-store/migrations/20260217222329_create_topics.sql
@@ -1,7 +1,7 @@
 -- SPDX-License-Identifier: MIT OR Apache-2.0
 
 CREATE TABLE IF NOT EXISTS topics_v1 (
-    topic                   TEXT            NOT NULL,
+    topic                   BLOB            NOT NULL,
     author                  VARCHAR(64)     NOT NULL,
     data_id                 BLOB            NOT NULL,
 
--- a/p2panda-store/src/topics/traits.rs
+++ b/p2panda-store/src/topics/traits.rs
@@ -38,7 +38,7 @@ use std::error::Error;
 /// If we implement `T` to express that we're interested in syncing over a specific chat group,
 /// for example "Chat Group 2" we would implement `TopicStore` to give us all append-only logs of
 /// all members inside this group, that is the entries inside logs `A2`, `B2` and `C2`.
-pub trait TopicStore<T, A, ID> {
+pub trait TopicStore<T, A, S> {
     type Error: Error;
 
     /// Associate an author and data id pair with a topic.
@@ -46,7 +46,7 @@ pub trait TopicStore<T, A, ID> {
         &self,
         topic: &T,
         author: &A,
-        data_id: &ID,
+        data_id: &S,
     ) -> impl Future<Output = Result<bool, Self::Error>>;
 
     /// Remove an association with a topic.
@@ -54,10 +54,9 @@ pub trait TopicStore<T, A, ID> {
         &self,
         topic: &T,
         author: &A,
-        data_id: &ID,
+        data_id: &S,
     ) -> impl Future<Output = Result<bool, Self::Error>>;
 
     /// Get identifiers for all associated
-    fn resolve(&self, topic: &T)
-    -> impl Future<Output = Result<BTreeMap<A, Vec<ID>>, Self::Error>>;
+    fn resolve(&self, topic: &T) -> impl Future<Output = Result<BTreeMap<A, Vec<S>>, Self::Error>>;
 }


p2panda/src/streams/stream.rs
```diff
@@ -21,24 +27,114 @@ pub struct StreamHandle<M> {
...
+                            continue;
+                        };
+
+                        StreamEvent::Message(Message {
+                            header: processed.header,
+                            topic,
+                            body: message,
+                        })
+                    }
+                    TopicLogSyncEvent::SyncStatus(metrics) => StreamEvent::SyncStarted {
```
adzialocha, line 124: This event is received multiple times, is this the expected behaviour @sandreae? And if yes, what is the reason?

p2panda/tests/api.rs
```diff
@@ -67,3 +67,34 @@ async fn ephemeral_messaging() {
...
+#[tokio::test]
+async fn eventually_consistent_stream() {
+    let chat_id = Topic::new();
+
+    let panda = p2panda::spawn().await.unwrap();
+    let icebear = p2panda::spawn().await.unwrap();
+
+    // Panda joins the chat and sends a message to icebear.
+    let mut panda_chat = panda.stream::<String>(chat_id).await.unwrap();
+    panda_chat.publish("Hello, Icebear!".into()).await.unwrap();
```
adzialocha, line 85: Nothing is synced here, need to investigate why this not working.

@mycognosist @sandreae

p2panda/src/streams/stream.rs
sandreae: Yes, this is sent several times, for now it is when 1) we compute the local operation and byte count for a session 2) we receive the remote operation and byte count. The intention was that it might periodically get sent later to inform of sync progress as well, but it may be that this is not actually what we want.

adzialocha: I see, thanks!

Maybe we can make a flow diagram for these processes and ideally document it / place it inside the code-base.

With this we can design a state machine moving forward in the sync process, making sure we transition to the next state whenever we have the data we need to forward to the application layer as "system events" and at the same time be able to ignore redundant events.

The state machine can live in `p2panda`, I don't mind with the events being more detailled in `p2panda-sync`, the most important thing is that they're documented to understand how to use them.

adzialocha: > The intention was that it might periodically get sent later

For this I'd expect a separate struct which is a bit more "chatty", I think iroh blobs has a good pattern for that: https://docs.rs/iroh-blobs/latest/iroh_blobs/api/downloader/struct.DownloadProgress.html

p2panda/src/processor/tasks.rs
```diff
@@ -0,0 +1,206 @@
...
+///
+/// Tasks can be added to the manager. Event processors concurrently begin work on these tasks and
+/// whenever they've finished they mark the task as "ready". All processes which observe this task
+/// will be notified on this "ready" signal.
+///
+/// Tasks are automatically de-duplicated internally, so whenever multiple processes add the same
+/// task, they will look at the same instance and get notified about it's readyness at the same
+/// time.
+#[derive(Clone, Debug)]
+pub struct TaskManager<T, ID>(Arc<RwLock<HashMap<ID, Task<T, ID>>>>);
```
sandreae: Took me a while to grok that the task manager is mainly just a locking mechanism, with the additional feature that the result of the task can be awaited. Might just be the naming, but I was somehow expecting task "work" to actually be occurring here somewhere. I realise that's not the case, and the design makes total sense, it just wasn't immediately clear. Some terrible alternative names: `TaskLockRegistry`, `OperationLockRegistry`, `TaskOrchestrator`, `PipelineOrchestrator`, `TaskCoordinator`, `Coordinator` ...

sandreae: Not a blocker at all 👍

p2panda/src/processor/tasks.rs
```diff
@@ -0,0 +1,206 @@
...
+
+use tokio::sync::{Mutex, Notify, RwLock};
+
+/// Manages tasks for event processing pipelines.
+///
+/// Tasks can be added to the manager. Event processors concurrently begin work on these tasks and
+/// whenever they've finished they mark the task as "ready". All processes which observe this task
+/// will be notified on this "ready" signal.
+///
+/// Tasks are automatically de-duplicated internally, so whenever multiple processes add the same
```
sandreae, line 15: If I'm understanding correctly, the de-duplication would occur across multiple pipelines. I can't quite understand how this would work. If there was a "simple" pipeline and a "spaces" pipeline, then they would be performing different work, so I can't see how deduplication could be employed here.

adzialocha: Yes, you're right: Tasks should be deduplicated per operation and per pipeline "type". Since we don't have different pipeline types yet, this is not implemented, but we'll need it as soon as we have that.

Later we can have 4 "spaces pipelines" and 2 "regular pipelines" etc. and the tasks will be de-duplicated _within_ the regarding types.

sandreae: Aha, ok, that makes sense!

p2panda/src/processor/tasks.rs
adzialocha: Maybe a wrapper around all of these things (Pipeline + TaskManager) etc. can help to express the boundaries of these systems better?

mycognosist: Another option: `TaskTracker`.

p2panda/src/streams/stream.rs
```diff
@@ -21,24 +29,130 @@ pub struct StreamHandle<M> {
...
+
+                        // Do not forward operations to the application-layer if there's no body.
+                        let Some(body) = processed_event.body() else {
+                            continue;
+                        };
+
+                        // Attempt decoding application-layer message. This takes place _after_
+                        // system-level processing completed and the operation was ingested.
+                        //
+                        // In case decoding fails due to an application bug, users have the option
```
mycognosist, line 99: These are great comments; in the future it would be good to surface them to the level of doc comments so users can be fully aware of the behaviour and options they have available when things fail. Not a blocker for now.

p2panda/src/streams/stream.rs
```diff
@@ -115,52 +258,61 @@ impl<M> Message<M> {
...
-        self.header.timestamp
+        self.processed_event.header().timestamp
     }
 
     pub fn header(&self) -> &Header {
-        &self.header
+        self.processed_event.header()
     }
 
+    // TODO: Consider better naming here. It is confusing that I have to call body on Message to
```
mycognosist, line 276: Maybe `content()`? Calling `message.content()` doesn't seem too bad. Or `data()`.

adzialocha: I've handled this in this PR, let me know what you think!

https://github.com/p2panda/p2panda/pull/1061

Issue #1038, opened 2026-02-26
Consider making `validate_operation` and `validate_backlink` generic as well


adzialocha:
So it can be used by different data types and methods relying on it (like the "ingest" processor) can also be flexible.

PR #1050, opened 2026-03-03
More generic API for ingest & operation validation

sandreae:
Closes: #1038 

## 📋 Checklist

### DIFF
--- a/p2panda-stream/src/ingest/processor.rs
+++ b/p2panda-stream/src/ingest/processor.rs
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use std::borrow::Borrow;
 use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::marker::PhantomData;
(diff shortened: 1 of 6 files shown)


p2panda-stream/src/ingest/processor.rs
```diff
@@ -51,30 +52,33 @@ where
...
     E: Extensions,
 {
-    type Output = Operation<E>;
+    type Output = M;
 
     type Error = IngestError;
 
-    async fn process(&self, args: IngestArguments<L, E, TP>) -> Result<(), Self::Error> {
+    async fn process(&self, args: IngestArguments<M, L, E, TP>) -> Result<(), Self::Error> {
+        let inner_operation = args.message.clone().into();
```
sandreae: This approach requires a clone here as `Into` consumes the underlying type, it's not possible to implement conversion traits from an owned into a borrowed type. I can't immediately see a way around this.

sandreae: If we can pass the args in borrowed, then it would work, but I assume now that isn't possible.

sandreae: We can use `Borrow` in place of `Into `.

p2panda-stream/src/ingest/processor.rs
```diff
@@ -51,30 +42,32 @@ where
...
 
-impl<S, L, E, TP> Processor<IngestArguments<L, E, TP>> for Ingest<S, L, E, TP>
+impl<S, M, L, E, TP> Processor<M> for Ingest<S, M, L, E, TP>
 where
     S: Transaction
         + OperationStore<Operation<E>, Hash, L>
         + LogStore<Operation<E>, PublicKey, L, SeqNum, Hash>
         + TopicStore<TP, PublicKey, L>,
+    // @TODO: remove Clone after https://github.com/p2panda/p2panda/issues/1040
+    M: IngestArgs<L, TP> + Borrow<Operation<E>> + Clone,
```
sandreae: An alternative approach here instead of requiring `Borrow<Operation<E>>` here we could add an `fn operation(&self) -> &Operation<E>;` method to `IngestArgs` trait. Not sure which I prefer.

adzialocha: The borrow is somewhat less explicit, but I'm also undecided.

Nice idea with the args trait, it could be that ingest is the only place requiring these arguments, so that's a good place!

Also can't think of any other processor requiring to know the whole operation, another plus for an explicit trait.

PR #1058, opened 2026-03-04
Use `Topic` everywhere

adzialocha:
We have a useful `Topic` struct and we should make use of it in all places, instead of `TopicId` aka `[u8; 32]`.

`Topic` uses `serde_bytes` internally which serializes the bytestring optimally.

Closes: #1059 

## 📋 Checklist

### DIFF
--- a/p2panda-core/src/topic.rs
+++ b/p2panda-core/src/topic.rs
@@ -10,7 +10,7 @@ use thiserror::Error;
 
 use crate::{Hash, PublicKey};
 
-const TOPIC_LENGTH: usize = 32;
+pub const TOPIC_LENGTH: usize = 32;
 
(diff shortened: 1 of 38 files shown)


p2panda-core/src/topic.rs
```diff
@@ -10,8 +10,19 @@ use thiserror::Error;
 
 use crate::{Hash, PublicKey};
 
-const TOPIC_LENGTH: usize = 32;
-
+pub const TOPIC_LENGTH: usize = 32;
+
+/// Identifier for an gossip- or sync topic.
```
mycognosist:
```suggestion
/// Identifier for a gossip- or sync topic.
```

Issue #798, opened 2025-09-17
Ensure atomicity of operation store / integrity of log when creating new operations


maackle:
I'm wanting to ensure atomicity when working with an OperationStore. For example, there is at least one race condition in meshpit, where the [store is read to find the latest operation](https://github.com/adzialocha/meshpit/blob/2bfa572564132c51bf210b7334d25631599cea40/src/operation.rs#L50) and then [that's used to write a new operation](https://github.com/adzialocha/meshpit/blob/2bfa572564132c51bf210b7334d25631599cea40/src/node.rs#L213-L225), but the state is not locked in between the read and the write so the read can become stale. It's minor and probably unlikely to happen but would be nice to avoid.

I'm wondering if a quick fix would be to just implement OperationStore and LogStore on the `InnerMemoryStore` state, and on some wrapper around a sqlx Transaction, instead -- which is kind of what I thought was already happening until I realized it's not.

I also see [p2panda_store::transactions](https://github.com/p2panda/p2panda/blob/6749647ba8f9b20c00152e5402df4507f6c51f67/p2panda-store/src/transactions.rs), but it looks like the provided stores (mem and sql) don't use it yet. Is the intention to eventually do that?

Wondering if I'm seeing everything correctly, and what your thoughts have been on atomicity. If it's really straightforward maybe I could put up a PR.

adzialocha:
Thank you for this question!

Indeed we haven't touched upon all the atomic transaction work yet and especially the atomicity guarantees around creating operations are broken in it's current form.

You find our ideas on transactions here in this blog: https://p2panda.org/2025/07/09/streams-transactions-crash-resilience.html - I'm currently working on the refactored `p2panda-stream` API which will be a more "chainable" processing layer interface where the stores will be reworked and combined with a "stream controller" (see blog for more context), operations become "committable" and transactions strictly atomic.

Around creating operations specifically I was also pondering about a design which a) respects atomicity and b) allows storing the next operation arguments (backlink, seq_num) in memory, so we can fastly generate new ones without a database look up. The latter is an optimization and needs to be benchmarked, also it introduces complexity because of that "caching" layer, so this is optional / something for later.

[...]

PR #1060, opened 2026-03-05
Ensure log transactions are atomic

mycognosist:
This PR introduces two new trait methods and implementations: one to the
`OperationStore` and one to the `LogStore`. Both are versions of existing
methods which are designed to be used within the context of a
transaction. These include `get_latest_entry_tx` and `get_operation_tx`.

The updated methods are utilised within the `create_operation` (forge) and
`ingest_operation` (stream) methods.

Closes: https://github.com/p2panda/p2panda/issues/798

## 📋 Checklist

### DIFF
--- a/p2panda-store/src/logs/sqlite/mod.rs
+++ b/p2panda-store/src/logs/sqlite/mod.rs
@@ -15,6 +15,19 @@ use crate::logs::sqlite::models::{LatestEntryRow, LogHeightRow, LogMetaRow};
 use crate::operations::OperationRow;
 use crate::sqlite::{SqliteError, SqliteStore};
 
+const GET_LATEST_ENTRY: &str = "
+    SELECT
+        hash,
+        seq_num
+    FROM
+        operations_v1
+    WHERE
+        public_key = ?
+        AND log_id = ?
+    ORDER BY
+        CAST(seq_num AS NUMERIC) DESC LIMIT 1
+";
+
 impl<'a, L, E> LogStore<Operation<E>, PublicKey, L, SeqNum, Hash> for SqliteStore<'a>
 where
     E: Extensions,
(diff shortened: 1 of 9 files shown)


p2panda-store/src/logs/sqlite/mod.rs
```diff
@@ -55,6 +58,39 @@ where
         }
     }
 
+    // TODO: Improve doc comments where applicable.
```
adzialocha: Is this TODO something you wanted to handle as part of this PR?

p2panda-store/src/logs/sqlite/mod.rs
```diff
@@ -55,6 +58,39 @@ where
...
+    /// Retrieve the hash and sequence number of the latest entry in an author's log.
+    ///
+    /// This variant of the method is intended to be used in situations where atomicity of database
+    /// operations is needed. It requires a transaction context with an acquired permit.
+    async fn get_latest_entry_tx(
```
adzialocha, line 69: Not for this PR but I believe we can reduce the number of methods in a refactoring like that: https://github.com/p2panda/p2panda/issues/1065

p2panda-stream/src/ingest/operation.rs
```diff
@@ -48,18 +48,23 @@ where
         return Ok(false);
     }
```
adzialocha, line 54: It's not a critical issue / race condition, but if we make has_operation part of the transaction we could avoid attempting to insert duplicates.

The following might happen if it's two separate transactions:

1. Process A checks has_operations which returns false (TX1, auto-commit)
2. Process B checks has_operations which returns false (TX2, auto-commit)
3. Process A inserts operation (Tx3)
4. Process B attempts inserting operation as well, it causes a SQL CONFLICT but is resolved silently (Tx4)

If it's part of the same transaction:

1. Process A checks has_operations which returns false and inserts it (TX1)
2. Process B checks has_operations which returns true, early abort (TX2)

PR #1064, opened 2026-03-09
Replay stream from start based on offset rule

sandreae:
When a user calls stream_from with argument Offset::Start then before
processing operations arriving via sync all local operations for the
topic will be retrieved from the store and re-processed. This is done
inside a dedicated task where operations are retrieved in batches from
the store.

Closes: #1067 

## 📋 Checklist

### DIFF
--- a/p2panda/src/streams/replay.rs
+++ b/p2panda/src/streams/replay.rs
@@ -1,0 +1,90 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::collections::BTreeMap;
+
+use p2panda_core::{PublicKey, Topic};
+use p2panda_store::logs::LogStore;
+use p2panda_store::topics::TopicStore;
+use p2panda_store::{SqliteError, SqliteStore};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+use tokio::sync::mpsc;
+use tokio::sync::mpsc::Sender;
+use tokio::task::JoinHandle;
+
+use crate::node::AckPolicy;
+use crate::operation::{Extensions, Operation};
+use crate::processor::Pipeline;
+use crate::streams::StreamEvent;
+use crate::streams::stream::process_operation;
+
+/// Retrieve from the store and re-process all operations for a given topic.
+pub(crate) async fn replay_from_start<M>(
+    topic: Topic,
+    store: SqliteStore,
+    app_tx: Sender<StreamEvent<M>>,
+    pipeline: Pipeline<Topic, Extensions, Topic>,
+    ack_policy: AckPolicy,
+) -> Result<(), ReplayError>
+where
+    M: Serialize + for<'a> Deserialize<'a> + Send + 'static,
+{
+    let (replay_tx, mut replay_rx) = mpsc::unbounded_channel::<Operation>();
+
+    // Spawn task for retrieving operations from the store and sending them on a channel to be
+    // processed.
+    let replay_task: JoinHandle<Result<(), ReplayError>> = tokio::spawn(async move {
+        let author_logs: BTreeMap<PublicKey, Vec<Topic>> = store.resolve(&topic).await?;
+        for (author, logs) in author_logs {
+            for log_id in logs {
+                let Some(operations): Option<Vec<(Operation, _)>> =
+                    store.get_log_entries(&author, &log_id, None, None).await?
+                else {
+                    // If the log was concurrently deleted since calling TopicStore::resolve then
+                    // None is returned here. This is not considered an error, as no log integrity
+                    // is broken and deletes should be immediately respected.
+                    continue;
+                };
+
+                for (operation, _) in operations {
+                    replay_tx
+                        .send(operation)
+                        .map_err(|_| ReplayError::CriticalError)?;
+                }
+            }
+        }
+        Ok(())
+    });
+
+    // Pull operations from the replay channel and send them to the processing pipeline.
+    loop {
+        if let Some(operation) = replay_rx.recv().await {
+            match process_operation::<M>(operation, topic, &pipeline, ack_policy).await {
+                Some(event) => {
+                    app_tx
+                        .send(event)
+                        .await
+                        .map_err(|_| ReplayError::CriticalError)?;
+                }
+                None => continue,
+            }
+        };
+
+        if replay_task.is_finished() {
+            return replay_task
+                .await
+                .expect("panic in task")
+                .map_err(|_| ReplayError::CriticalError);
+        }
+    }
+}
+
+/// Error types which can occur during replay.
+#[derive(Debug, Error)]
+pub enum ReplayError {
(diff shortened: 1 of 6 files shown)


p2panda/src/streams/replay.rs
```diff
@@ -0,0 +1,88 @@
...
+use tokio::task::{JoinError, JoinHandle};
+
+use crate::node::AckPolicy;
+use crate::operation::{Extensions, Operation};
+use crate::processor::Pipeline;
+use crate::streams::StreamEvent;
+use crate::streams::stream::process_operation;
+
+/// Retrieve from the store and process all operations for a given topic.
+pub(crate) async fn replay_task<M>(
```
sandreae: I believe there will be cases where replaying a stream will result in the same operation being forwarded to the processor pipeline more than once, the case is as such:
1) topic T is already subscribed to with active sync sessions
2) now the user calls `stream_from` for topic T
3) a new from sync rx is taken (effectively this starts a buffer of any newly received operations)
==== operations X arrives and is ingested ====
4) all past operations are retrieved (which may include X) and sent to the app layer
5) all operations in the sync rx are drained (may also include X)

p2panda/src/streams/stream.rs
```diff
@@ -37,24 +40,44 @@ pub(crate) async fn processed_stream<M>(
...
     pipeline: Pipeline<Topic, Extensions, Topic>,
-    _offset: Offset,
+    offset: Offset,
 ) -> Result<
     (StreamPublisher<M>, StreamSubscription<M>),
     SyncHandleError<Operation, TopicLogSyncEvent<Extensions>>,
 >
 where
-    M: Serialize + for<'a> Deserialize<'a> + Send + 'static,
+    M: Debug + Serialize + for<'a> Deserialize<'a> + Send + 'static,
```
sandreae: This `Debug` requirement snuck in so that we can print errors returned from the relay task .

p2panda/src/streams/replay.rs
```diff
@@ -0,0 +1,92 @@
...
+#[derive(Debug, Error)]
+pub enum ReplayError<M> {
+    #[error("an error occurred while querying the store: {0}")]
+    Forge(#[from] SqliteError),
+
+    #[error("an error occurred sending on channel: {0}")]
+    ReplaySend(#[from] SendError<Operation>),
+
+    #[error("an error occurred sending on channel: {0}")]
+    AppSend(#[from] SendError<StreamEvent<M>>),
```
sandreae: If we remove this error variant, and instead ignore (or just log) errors sending on the app channel we could get rid of the M generic and in turn get rid of the Display trait bound.

p2panda/src/streams/replay.rs
```diff
@@ -0,0 +1,92 @@
...
+use tokio::task::{JoinError, JoinHandle};
+
+use crate::node::AckPolicy;
+use crate::operation::{Extensions, Operation};
+use crate::processor::Pipeline;
+use crate::streams::StreamEvent;
+use crate::streams::stream::process_operation;
+
+/// Retrieve from the store and re-process all operations for a given topic.
+pub(crate) async fn replay_task<M>(
```
adzialocha: Maybe rename this to `replay_all` or `replay_from_start` etc.? That it is a task running inside is maybe a secondary detail.

p2panda/src/streams/replay.rs
adzialocha: I'd sum up `ReplaySend` and `AppSend` into one error type (BrokenChannel, SendFailed, etc.) and remove the concrete type. I assume it's not interesting for the receiver of this error to know which operation / event got lost when the channel broke down as they want to only know if the replay worked out or not. With this we also save the `Debug` trait bound.

p2panda/src/streams/replay.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/// Error types which can occur during replay.
+#[derive(Debug, Error)]
+pub enum ReplayError<M> {
+    #[error("an error occurred while querying the store: {0}")]
+    Forge(#[from] SqliteError),
+
+    #[error("an error occurred sending on channel: {0}")]
+    ReplaySend(#[from] SendError<Operation>),
```
adzialocha: That the replay channel can fail seems highly unlikely (both halfs run right next to each other), I'd be fine with unwrapping here.

p2panda/src/streams/replay.rs
adzialocha: `AppSend` seems also very unlikely. I'm undecided if we should just unwrap here.

p2panda/src/streams/replay.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    #[error("an error occurred sending on channel: {0}")]
+    AppSend(#[from] SendError<StreamEvent<M>>),
+
+    #[error("an error occurred awaiting the replay tasks completion")]
+    JoinHandle(#[from] JoinError),
```
adzialocha: This can fail if the task panicked or got cancelled. Both seems very unlikely (unless we unwrap somewhere inside of p2panda-store and introduce cancellation somewhere else).

p2panda/src/streams/stream.rs
```diff
@@ -37,24 +40,44 @@ pub(crate) async fn processed_stream<M>(
...
+                ack_policy,
+            )
+            .await;
+
+            if let Err(err) = result {
+                // If an error occurs in the relay task then we abort and do not process any
+                // further messages received from sync.
+                warn!("error occurred in relay task: {err:?}");
+                return;
+            }
```
adzialocha, line 81: I believe we want to propagate the error up to the application layer here and not only log a warning. Users want to know if replaying failed, they likely want to stop doing whatever they were planning at this point (imagine Reflection starting up an existing document and replay fails, there will be nothing to materialize).

sandreae: Yeah, nice, making this change now. The error message might need to be passed up as a `String` as `StreamEvent` requires `Clone` and at least the sql store error does not implement this. But that seems fine to me.

p2panda/src/streams/replay.rs
sandreae: I'd prefer if we handle all errors explicitly, it's easy enough to do and I don't see any harm.

p2panda/src/streams/replay.rs
sandreae: Same for this, let's handle it explicitly.

p2panda/src/streams/stream.rs
```diff
@@ -51,10 +55,32 @@ where
...
+                topic,
+                store.clone(),
+                app_tx.clone(),
+                pipeline.clone(),
+                ack_policy,
+            )
+            .await;
+
+            if let Err(err) = result {
+                warn!("error occurred in relay task: {err:?}");
```
adzialocha: Typo: relay -> replay

PR #1073, opened 2026-03-10
Log-prefix pruning in `p2panda` API

adzialocha:
Closes: #1071 and #1066 

## 📋 Checklist

### DIFF
--- a/p2panda/src/processor/event.rs
+++ b/p2panda/src/processor/event.rs
@@ -3,8 +3,9 @@
 use std::borrow::Borrow;
 
 use p2panda_core::traits::Digest;
-use p2panda_core::{Body, Hash, Header, LogId, Operation, PruneFlag};
+use p2panda_core::{Body, Hash, Header, LogId, Operation, PruneFlag, PublicKey, SeqNum};
 use p2panda_stream::ingest::{IngestArgs, IngestError};
+use p2panda_stream::log_prune::{LogPruneArgs, LogPruneError, LogPruneResult};
 use thiserror::Error;
 
 /// Status of an event being processed by a _single_ processor in the pipeline.
@@ -24,11 +25,19 @@ pub enum ProcessorStatus<R, F> {
 /// Single event running through the "event processor" pipeline.
 #[derive(Clone, Debug)]
 pub struct Event<L, E, TP> {
-    /// Input arguments for the processing pipeline.
-    input: (Operation<E>, L, TP, PruneFlag),
+    operation: Operation<E>,
+
+    /// Input arguments for the "ingest" processor.
+    ingest_args: (L, TP, PruneFlag),
 
     /// Status of the "ingest" processor.
     pub(crate) ingest: ProcessorStatus<(), IngestError>,
+
+    /// Input arguments for the "log prune" processor.
+    log_prune_args: LogPruneArgs<PublicKey, L, SeqNum>,
+
+    /// Status of the "log prune" processor.
+    pub(crate) log_prune: ProcessorStatus<LogPruneResult, LogPruneError>,
 }
 
 impl<L, E, TP> Event<L, E, TP>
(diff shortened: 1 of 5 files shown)


p2panda/src/processor/event.rs
```diff
@@ -83,31 +117,41 @@ where
...
+        self.ingest_args.2.is_set()
     }
 
     fn operation(&self) -> impl Borrow<Operation<E>> {
-        &self.input.0
+        &self.operation
+    }
+}
+
+impl<L, E, TP> Borrow<LogPruneArgs<PublicKey, L, SeqNum>> for Event<L, E, TP>
```
adzialocha, line 136: I like this pattern supplying a processor with "arguments". Maybe we can do the same for `IngestArgs`?

p2panda/src/streams/stream.rs
```diff
@@ -183,16 +184,41 @@ where
...
+
+    /// Deletes all our previously published messages in this stream.
+    ///
+    /// This signals to all other nodes that they should remove them as well.
+    ///
+    /// An message can be optionally added when pruning, allowing to publish a "snapshot"
+    /// / state-based CRDT of the current state, so nodes can still consistently re-create all
+    /// state, even if previous messages are gone.
+    ///
+    /// Internally we're applying append-only log prefix deletion, meaning that the log's prefix
```
mycognosist, line 237: I'm not sure "prefix" makes sense in the context of append-only logs. Is this maybe a term which has leaked over from pruning in path-based contexts? I would suggest rather talking explicitly about log sequence numbers and deletion of all entries with a number lower than the message we're publishing.

p2panda-stream/src/log_prune/args.rs
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
mycognosist, line 1: Should this file maybe be named `args.rs` instead?

p2panda/src/streams/stream.rs
adzialocha: So far this term was used internally in our designs and how we talk about this feature. A log can be seen as a sequence (of operations), with this you can have a prefix and a suffix from a given point in that sequence. For me that term is not unusual, but happy to change it to something else if this is confusing, do you have something which sums it up in one word?

mycognosist: I thought there would be a mathematical term describing all elements below (or above) an index in a sequence but I can't seem to find one 😆 So unfortunately I cannot offer a single word. Perhaps we stick with prefix and reevaluate if others find it confusing.

adzialocha: Maybe I should at least explain the word in that doc-string or the one of the processor.

p2panda/src/streams/stream.rs
```diff
@@ -183,16 +184,41 @@ where
...
     pub async fn publish(&mut self, message: M) -> Result<Hash, PublishError> {
-        let encoded_message = encode_cbor(&message)?;
+        self.publish_inner(Some(message), false).await
+    }
+
+    /// Deletes all our previously published messages in this stream.
+    ///
+    /// This signals to all other nodes that they should remove them as well.
+    ///
+    /// An message can be optionally added when pruning, allowing to publish a "snapshot"
```
sandreae:
```suggestion
    /// A message can be optionally added when pruning, allowing to publish a "snapshot"
```

p2panda/src/processor/pipeline.rs
```diff
@@ -97,22 +99,33 @@ where
...
+                            Ok(mut event) => {
+                                event.ingest = ProcessorStatus::Completed(());
+                                event
+                            }
+                            Err((mut event, err)) => {
+                                event.ingest = ProcessorStatus::Failed(err);
+                                event
+                            }
+                        })
+                        .layer(log_prune)
```
sandreae, line 118: Yay, two processors!!!

PR #1080, opened 2026-03-12
Process local operations

adzialocha:
Locally created operations should run through the same internal event processing pipeline as remote, incoming ones.

This patch also introduces a way for users to optionally await the result of the processing pipeline when publishing locally:

```rust
let processing = tx.publish("Hello, World!".into()).await?;

// Get hash of resulting operation.
println!("{}", processing.hash());

// We can optionally await this and check the result.
let processing_result = processing.await?;
assert!(!processing_result.is_failed());
```

Closes: https://github.com/p2panda/p2panda/issues/1070

## 📋 Checklist

### DIFF
--- a/p2panda/src/streams/stream.rs
+++ b/p2panda/src/streams/stream.rs
@@ -224,20 +224,20 @@ where
 
                 // Send processing result to application layer.
                 //
-                // If channel stopped working, close the task as well.
-                if app_tx.send(event).await.is_err() {
-                    break;
-                }
+                // If channel stopped working because the subscriber got dropped, ignore it as we
+                // still might want to process locally published operations.
+                let _ = app_tx.send(event).await;
             }
         });
     }
 
-    // TODO: Find out if publishing breaks if StreamSubscription gets dropped as task and therefore
-    // channel might be gone.
+    // Keep around the sync handle on both the publisher and subscriber ends to keep it running
+    // even if one half got dropped.
+    let sync_handle = Arc::new(sync_handle);
 
     let tx = StreamPublisher {
         topic,
-        sync_handle: Arc::new(sync_handle),
+        sync_handle: sync_handle.clone(),
         forge,
         publish_tx,
         _marker: PhantomData,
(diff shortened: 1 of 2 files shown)


p2panda/src/streams/stream.rs
```diff
@@ -205,11 +318,53 @@ where
...
+
+    if event.is_failed() {
+        warn!(
+            id = %event.hash(),
+            "processing local operation failed: {}",
+            event.failure_reason().expect("error")
+        );
+    }
+
+    // TODO: Always ack your own operations?
```
sandreae, line 352: Intuitively I would say yes, always ack your own operations.

p2panda/src/streams/stream.rs
```diff
@@ -237,53 +392,92 @@ where
...
 
-        Ok(hash)
+        Ok(PublishFuture { hash, processed_rx })
+    }
+}
+
+/// Future which can be awaited to find out when locally published operation has finished
+/// processing.
+#[derive(Debug)]
+pub struct PublishFuture {
```
sandreae, line 449: This is really nice ⭐

p2panda/src/streams/stream.rs
```diff
@@ -237,53 +392,92 @@ where
...
+    hash: Hash,
```
sandreae, line 450: Slightly strange to call hash() on a future which has not even completed yet, but still this is probably the best approach available.

p2panda/src/streams/stream.rs
```diff
@@ -34,8 +33,53 @@ use crate::streams::replay::replay_from_start;
...
+///
+/// ## Stream design
+///
+/// Locally published operations are processed by the same event processor pipeline as incoming,
+/// remote operations.
+///
+/// If a replay was requested, processing local and remotely incoming operations is temporarily
+/// blocked until replay has finished.
+///
+/// ```text
```
sandreae, line 59: Lovely diagram ⭐

p2panda/src/streams/stream.rs
adzialocha: The operation exists already and is in the database, it is _not_ processed yet, so this is what you're awaiting

Issue #1069, opened 2026-03-10
Flaky test: `discovery::tests::smoke_test`


mycognosist:
Failure has been seen locally and in CI.

```text
thread 'discovery::tests::smoke_test' (90378) panicked at p2panda-net/src/discovery/tests.rs:49:5:
assertion `left == right` failed
  left: 1
 right: 0
```

```rust
// Alice didn't learn about new transport info of Bob as their manually added node info was
// already the "latest".
let alice_metrics = alice.discovery.metrics().await.unwrap();                               
assert_eq!(alice_metrics.newly_learned_transport_infos, 0);
```

PR #1081, opened 2026-03-12
Use MockInstant for Timestamp when running tests

mycognosist:
p2panda-core was not previously configured to use MockInstant when running in test contexts. This meant that the discovery smoke test (in p2panda-net) became flaky when the Timestamp type was moved to p2panda-core. This PR ensures that MockInstant is now used by p2panda-core when tests are run.

The change brought about a failure in the p2panda API test when comparing the timestamp of messages received by panda and icebear. This has been fixed by manually advancing the time after the first message has been received.

Fixes: https://github.com/p2panda/p2panda/issues/1069

## 📋 Checklist

### DIFF
--- a/p2panda/tests/api.rs
+++ b/p2panda/tests/api.rs
@@ -3,6 +3,7 @@
 use std::time::Duration;
 
 use futures_util::StreamExt;
+use mock_instant::thread_local::MockClock;
 use p2panda::operation::Operation;
 use p2panda::streams::{EphemeralMessage, Offset, ProcessedOperation, StreamEvent};
 use p2panda::test_utils::setup_logging;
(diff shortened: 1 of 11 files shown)


p2panda/tests/api.rs
```diff
@@ -49,6 +50,8 @@ async fn ephemeral_stream() {
 
         tokio::spawn(async move {
             let message = rx.next().await.unwrap();
+            // Advance the clock to ensure icebear's message is published later than panda's.
+            MockClock::advance_system_time(Duration::from_secs(1));
```
adzialocha, line 54: Did you experience race conditions here? I believe the microsecond-precision timestamp should already make sure that the previous message is in the past, but curious to hear if that wasn't enough.

mycognosist: In this case the test is strictly greater-than which failed because both had a timestamp of 0.

adzialocha: Ah, I see, it's because we're using mocked time now and didn't before

PR #1085, opened 2026-03-17
Process sync events and aggregate per-topic metrics in stream

sandreae:
Many sync sessions for a topic run concurrently and the events they emit are more fine-grained than what we want to forward to users of p2panda Node. We also want to offer insight into sync metrics for an entire topic (rather than per-session) and so should calculate aggregates and forward these in stream events.

This change makes some improvements to the lower level sync events and then introduces an aggregator which processes these events in the stream and emits intermediate events containing additional metrics and aggregate fields.

## 📋 Checklist

### DIFF
--- a/p2panda-sync/src/protocols/log_sync.rs
+++ b/p2panda-sync/src/protocols/log_sync.rs
@@ -30,24 +30,21 @@ enum State<L> {
     Start,
 
     /// Calculate local log heights and send Have message to remote.
-    SendHave { metrics: LogSyncMetrics },
+    SendHave,
 
     /// Receive have message from remote and calculate operation diff.
-    ReceiveHave {
-        local: LogHeights<PublicKey, L>,
-        metrics: LogSyncMetrics,
-    },
+    ReceiveHave { local: LogHeights<PublicKey, L> },
 
     /// Send PreSync message to remote or Done if we have nothing to send.
     SendPreSync {
         remote_needs: LogRanges<PublicKey, L>,
-        metrics: LogSyncMetrics,
     },
 
     /// Receive PreSync message from remote or Done if they have nothing to send.
     ReceivePreSyncOrDone {
         remote_needs: LogRanges<PublicKey, L>,
-        metrics: LogSyncMetrics,
+        outbound_operations: u64,
+        outbound_bytes: u64,
     },
 
     /// Enter sync loop where we exchange operations with the remote, moves onto next state when
@@ -116,26 +113,16 @@ where
 
         let metrics = loop {
             match self.state {
-                State::Start => {
-                    let metrics = LogSyncMetrics::default();
-                    self.event_tx
-                        .send(
-                            LogSyncEvent::Status(LogSyncStatus::Started {
-                                metrics: metrics.clone(),
-                            })
-                            .into(),
-                        )
-                        .map_err(|_| LogSyncError::BroadcastSend)?;
-                    self.state = State::SendHave { metrics };
-                }
-                State::SendHave { metrics } => {
+                State::Start => self.state = State::SendHave,
+                State::SendHave => {
                     let local = get_log_heights(&self.store, &self.logs).await?;
                     sink.send(LogSyncMessage::<L>::Have(local.clone()))
                         .await
                         .map_err(|err| LogSyncError::MessageSink(format!("{err:?}")))?;
-                    self.state = State::ReceiveHave { local, metrics };
+
+                    self.state = State::ReceiveHave { local };
                 }
-                State::ReceiveHave { local, mut metrics } => {
+                State::ReceiveHave { local } => {
                     let Some(message) = stream.next().await else {
                         return Err(LogSyncError::UnexpectedStreamClosure);
                     };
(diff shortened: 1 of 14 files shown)


p2panda/src/streams/mod.rs
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+mod aggregator;
```
adzialocha: Since this module contains more than just the `Aggregator`, maybe a more descriptive name would be `sync_events` or `sync_metrics`?

p2panda/src/streams/stream.rs
```diff
@@ -156,8 +158,19 @@ where
...
+                        };
+
+                        match event {
+                            aggregator::SyncEvent::SyncStarted { .. } => event.into(),
+                            aggregator::SyncEvent::SyncEnded { .. } => event.into(),
+                            // @TODO: we don't forward any metrics or aggregate data on
+                            // StreamEvent::ProcessedOperation yet. This requires a little more
+                            // thought if it's something we want to do as locally created and
+                            // processed operations are also forwarded up the stream and these
+                            // don't have any associated sync metrics.
```
adzialocha: How about an `Source` enum which can be either `Source::SyncSession` or `Source::LocalStore` (for locally published operations and replays)?

p2panda/src/streams/sync_metrics.rs
```diff
@@ -0,0 +1,351 @@
...
+/// Intermediate sync event type enriched with aggregate data.
+#[derive(Debug)]
+pub enum SyncEvent<E> {
+    SyncStarted {
+        remote: PublicKey,
+        session_id: u64,
+        incoming_operations: u64,
+        outgoing_operations: u64,
+        incoming_bytes: u64,
+        outgoing_bytes: u64,
```
adzialocha, line 175: incoming_operations -> received_operations
outgoing_operations -> sent_operations

etc.

p2panda/src/streams/aggregator.rs
```diff
@@ -0,0 +1,351 @@
...
+    pub fn total_bytes_received(&self) -> u64 {
+        self.total_bytes_received
+    }
+}
+
+/// Intermediate sync event type enriched with aggregate data.
+#[derive(Debug)]
+pub enum SyncEvent<E> {
+    SyncStarted {
+        remote: PublicKey,
```
adzialocha: Mini: I'd use `NodeId` here from `p2panda-net`, just to communicate a little bit the distinction between a key-pair used for transport encryption vs. used for signing operations (`PublicKey`)

p2panda/src/streams/sync_metrics.rs
```diff
@@ -0,0 +1,351 @@
...
+    total_bytes_sent: u64,
+
+    /// Total number of received bytes across all topic sessions.
+    total_bytes_received: u64,
+
+    /// Latest metrics for all sessions.
+    session_metrics: HashMap<SessionId, Metrics>,
+
+    /// Set of running sessions which have entered live mode.
+    live_mode: HashSet<SessionId>,
```
adzialocha, line 32: How is this used?

p2panda-sync/src/protocols/log_sync.rs
```diff
@@ -146,41 +133,29 @@ where
 
                     let remote_needs = compare(&local, &remote);
 
-                    let mut operation_count = 0;
-                    let mut byte_count = 0;
+                    self.state = State::SendPreSync { remote_needs };
+                }
+                State::SendPreSync { remote_needs } => {
```
adzialocha, line 139: Nice to re-use the state machine here to aggregate the metrics, this is what I had in mind to do in the "frontend" (`p2panda`) but it's much better to tackle it also for `p2panda-sync`

p2panda/src/streams/sync_metrics.rs
sandreae: I think this is correct as it's a SyncStarted event so no operations have been sent/received yet.

p2panda/src/streams/sync_metrics.rs
sandreae: Aha, I hadn't implemented it's use yet. I intended to add a `live_mode` field to the session metrics for each received operation, so the user can know if this operation came during initial sync or in live mode. I'll add that now.

p2panda-sync/src/protocols/log_sync.rs
sandreae: The `Aggregator` on the frontend has a slightly different feature set from what is needed during one sync session. It is aggregating metrics across many sync sessions running over the same topic, whereas in sync we're looking at just one sync session and we're aggregating bytes/count sent and received for every operation. I think what you're suggesting is maybe just a refactor of the logic in `log_sync`, rather than re-using what we have in `p2panda` now.

p2panda/src/streams/sync_metrics.rs
```diff
@@ -0,0 +1,364 @@
+use std::collections::{HashMap, HashSet};
```
adzialocha, line 3: License header missing

p2panda/src/streams/sync_metrics.rs
```diff
@@ -0,0 +1,364 @@
...
+
+use p2panda_core::{Extensions, Operation};
+
```
adzialocha: Remove newline

p2panda/src/streams/sync_metrics.rs
```diff
@@ -0,0 +1,364 @@
...
+                received_operations,
+                sent_bytes,
+                received_bytes,
+                sent_bytes_topic_total,
+                received_bytes_topic_total,
+                error,
+            },
+            // We can't convert operation events simply like this as they need to be processed and
+            // decoded first so this branch is never called.
+            SyncEvent::OperationReceived { .. } => unimplemented!(),
```
adzialocha: unimplemented -> unreachable

p2panda/src/streams/stream.rs
```diff
@@ -551,6 +590,44 @@ impl<M> ProcessedOperation<M> {
     }
 }
 
+/// The source of a processed operation.
+#[derive(Clone, Debug)]
+#[allow(clippy::large_enum_variant)]
+pub enum Source {
```
adzialocha, line 596: I believe this is not publicly exported on lib-level, might be true for other new types as well

p2panda-sync/src/protocols/log_sync.rs
adzialocha: Ah didn't mean to suggest anything :-)

PR #1087, opened 2026-03-18
System event stream API

mycognosist:
Here we introduce an `event_stream` method on the `Node` API which
serves as an aggregator for several network-related system event
streams. The event stream currently emits discovery events, with further
events still to be added.

The implementation uses `SelectAll` to allow merging several streams
into one.

Related to: https://github.com/p2panda/p2panda/issues/1083

## 📋 Checklist

### DIFF
--- a/p2panda/src/node.rs
+++ b/p2panda/src/node.rs
@@ -2,7 +2,8 @@
 
 use std::fmt::Debug;
 
-pub use p2panda_core::{Hash, PrivateKey, PublicKey, Topic};
+use futures_util::Stream;
+use p2panda_core::{Hash, Topic};
 use p2panda_store::sqlite::{SqliteError, SqliteStore, SqliteStoreBuilder};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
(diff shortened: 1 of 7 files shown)


p2panda/src/node.rs
```diff
@@ -150,8 +151,25 @@ impl Node {
...
-    pub async fn events(&self) -> Result<SystemEventStream, CreateStreamError> {
-        unimplemented!()
+    /// System event stream.
+    ///
+    /// The system event stream emits network-related events which are not directly associated with
+    /// a specific topic. This currently includes node discovery events which communicate the
+    /// start, end and / or failure of each discovery session.
+    ///
+    /// For events related to topic-scoped eventually consistent and ephemeral streams, call
+    /// `events()` on the relevant handle.
```
adzialocha: I think the idea was to have the events be part of the message stream itself (see EC stream), to keep them in-order with application messages

p2panda/tests/api.rs
```diff
@@ -146,6 +147,39 @@ async fn replay_stream() {
...
+    assert_eq!(
+        event,
+        Some(SystemEvent::Discovery(
+            p2panda_net::discovery::DiscoveryEvent::SessionStarted {
+                role: p2panda_net::discovery::SessionRole::Initiated,
+                remote_node_id: icebear.id()
+            }
+        ))
+    );
+    */
```
adzialocha: Dead code

p2panda/src/node.rs
mycognosist: Oh wow, I totally forgot that -_- Thanks for the reminder! I'll make the changes.

PR #1102, opened 2026-03-24
Add method for getting graph heads filtered by group ids

sandreae:
## Context

Operations in a single groups "context" (which contains operations for multiple groups) form a graph, and the current groups API only allows the user to get the ids of the current graph tips regardless of which groups the user may actually want to include (needed for computing operation dependencies). This results in the groups API forcing all groups in one context to be dependent on each other, which is not always the desired behaviour. If a context includes operations for groups A & B, but neither of these groups a children/parent of the other, then the user should be able to get dependencies (graph tips) selectively based on the group(s) they want to include.

## Decision

Allow the user of the groups API to specify which groups should be included when calculating dependencies via the graph tips.   

## Result

Introduce a `heads_filtered(&self, groups: &[ID]) -> HashSet<OP>` method where the user can specify the groups it wants to target when calculating graph tips. Sub-sections of the groups graph which do not contain operations for the passed groups will not be included in the calculation.

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Highlights are marked with a pancake 🥞
 - Get current network id from `Node` [#1143](https://github.com/p2panda/p2panda/pull/1143)
 - Deduplicate received operations in sync manager event stream [#1147](https://github.com/p2panda/p2panda/pull/1147)
 - Todo list example with LWW-CRDT for Node API [#1148](https://github.com/p2panda/p2panda/pull/1148)
+- Method for getting filtered groups CRDT heads [#1102](https://github.com/p2panda/p2panda/pull/1102)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


adzialocha (commented):
Looks all good, just needs .git history cleanup.

> "This results in the groups API forcing all groups in one context to be dependent on each other, which is not always the desired behaviour."

Just out of curiousity, what are example reasons? Is it performance?

adzialocha (approved):

sandreae:
> Looks all good, just needs .git history cleanup.
> 
> > "This results in the groups API forcing all groups in one context to be dependent on each other, which is not always the desired behaviour."
> 
> Just out of curiousity, what are example reasons? Is it performance?

I would say the main reason the assumption that we in general prefer to have fewer branches in the operation graph, their appearance normally suggests concurrent operations from remote peers. If operations are produced locally, the default should be that the known graph tips, so as not to introduce "concurrency" where there is actually none. There is a performance benefit to this, as resolving branches takes more work (less in the future with the upcoming changes to the resolver though). One other point is that parts of the graph which are _not_ pointed to (due to use of the new API) will reduce the effectiveness of the "implicit acking" property.

PR #1112, opened 2026-03-30
Processor for groups operations

### LINKED ISSUE #1097: SQLite `GroupsStore` store implementation
Work on the `p2panda-auth` processor for group control messages has introduced a "groups store" with the basic API requirements being to get and set state for a groups CRDT. Currently there is a temporary in-memory store but for persistence we need an SQLite implementation using the `SQLite` store exported from `p2panda-store`.

sandreae:
`p2panda-auth` contains a generic groups crdt with messaging protocol, messages must be
partially ordered before they are processed in the provided APIs. It is generic and flexible
enough to be integrated into systems with different core types and approaches to message
ordering. However this makes it a little hard to "just use", even with core p2panda types and
provided ordering approaches, some scaffolding is required.

We want to introduce a more opinionated API and types which allow a user to satisfy basic
ordering requirements with built in support for core p2panda data types. It should still not be
opinionated regarding application layer logic around how groups are created and managed (who
are the initial members etc..) and it should be designed in a way where messages are forged
outside of the API itself and simply passed in for processing. Message persistence and sync
should also be handled elsewhere.

This change adds a `processor` module containing types and methods which can be used for ordering and
processing auth groups control messages encoded on p2panda Operations.

- remove orderer generic type from groups API as this is not needed to be internal and generic
- introduce `GroupsStore` trait and `SqliteStore` implementations to `p2panda-store`
- `Processor` struct with one method for processes all groups control operations, internally
  ordering is performed before any operation is then applied to locally persisted groups state.
- `args` and `operation` modules which contain arguments expected on operations containing
  groups control messages and a concrete `GroupsOperation` with conversion trait from
  `Operation<E>`.
- `groups` example demonstrating use with `p2panda-net` and `p2panda-sync`

closes: #1097 

## 📋 Checklist


p2panda-store/src/groups/traits.rs
```diff
@@ -0,0 +1,12 @@
...
+
+use std::error::Error;
+
+/// Trait describing API for setting and getting groups CRDT state by id.
+pub trait GroupsStore<ID, S> {
+    type Error: Error;
+
+    fn set_state(&self, id: &ID, state: &S) -> impl Future<Output = Result<(), Self::Error>>;
+
+    fn get_state(&self, id: &ID) -> impl Future<Output = Result<Option<S>, Self::Error>>;
```
adzialocha: I'd pick less generic method names here, for example: `set_group_state`. While the current names make sense in the context of the `GroupsStore` it will lead to collisions if other stores start using the same name (and now you need to explicitly tell the compiler _which_ method you mean). On top it'll be less readable when being used, like: `store.set_state(id, state); // ???`

p2panda-store/src/groups/sqlite.rs
```diff
@@ -0,0 +1,72 @@
...
+    ID: for<'a> Deserialize<'a> + Serialize,
+    S: for<'a> Deserialize<'a> + Serialize,
+{
+    type Error = SqliteError;
+
+    async fn set_state(&self, id: &ID, state: &S) -> Result<(), SqliteError> {
+        self.tx(async |tx| {
+            query(
+                "
+                INSERT OR REPLACE
```
adzialocha, line 22: Oh, I didn't know about `INSERT OR REPLACE`, that's pretty cool.

p2panda-auth/src/processor/operation.rs
```diff
@@ -0,0 +1,41 @@
+use p2panda_core::{Hash, PublicKey};
```
adzialocha, line 3: License header missing

p2panda-auth/src/traits/mod.rs
```diff
@@ -16,14 +16,14 @@ pub use resolver::Resolver;
 ///
 /// Note that this needs to be unique within a group, can be a username, number or preferably a
 /// long byte string.
-pub trait IdentityHandle: Copy + Debug + PartialEq + Eq + StdHash {}
+pub trait IdentityHandle: Copy + Debug + PartialEq + Eq + Ord + StdHash {}
```
adzialocha, line 19: Could use `Author` trait in core here (it's new)

p2panda-auth/src/traits/mod.rs
```diff
@@ -16,14 +16,14 @@ pub use resolver::Resolver;
...
 
 /// Identifier for each group membership operation.
 ///
 /// Operations trigger changes of the group state and are usually sent in form of messages over the
 /// network. Each operation needs to be uniquely identifiable, preferably by a collision-resistant
 /// hash.
-pub trait OperationId: Copy + Debug + PartialEq + Eq + StdHash {}
+pub trait OperationId: Copy + Debug + PartialEq + Eq + Ord + StdHash {}
```
adzialocha, line 26: Could use `Digest` trait in core here

p2panda-auth/src/access.rs
```diff
@@ -150,6 +152,28 @@ impl<C: PartialOrd + Eq> Ord for Access<C> {
 
 impl Conditions for () {}
 
+#[derive(Debug, Error)]
+pub enum AccessError {
+    #[error("unknown access string: {0}")]
+    Unknown(String),
+}
```
adzialocha: If there's only one error type it's possible to represent the error as a `struct`

p2panda-auth/Cargo.toml
```diff
@@ -20,18 +20,49 @@ all-features = true
...
 serde = { version = "1.0.219", features = ["derive"] }
 ciborium = "0.2.2"
+tokio = "1.49.0"
+anyhow = "1.0.102"
+clap = { version = "4.5.54", features = ["derive"] }
+futures-util = "0.3.32"
+rand_chacha = "0.9.0"
+tracing = "0.1.44"
+hex = "0.4.3"
+tracing-subscriber = { version = "0.3.22", features = ["env-filter"] }
```
adzialocha: Sort dependencies

p2panda-auth/Cargo.toml
```diff
@@ -20,18 +20,49 @@ all-features = true
...
 [features]
-default = []
-test_utils = ["dep:rand", "serde"]
+default = ["processor"]
+processor = [
+  "dep:tokio",
+  "serde",
+  "dep:p2panda-core",
+  "dep:p2panda-stream",
+  "dep:p2panda-store",
```
adzialocha: Sort these as well, at least by type

p2panda-auth/Cargo.toml
adzialocha: You can require your own package here with the `test_utils` feature flag enabled and with this you save _not_ mentioning these crates again. We already do this in other crates, for example: https://github.com/p2panda/p2panda/blob/main/p2panda/Cargo.toml#L46

p2panda-auth/src/traits/operation.rs
```diff
@@ -1,11 +1,13 @@
...
 
+use crate::group::GroupAction;
+
 /// Interface to express required information from operations processed by any auth graph
 /// implementation.
 ///
 /// Applications implementing these traits should authenticate the original sender of each
 /// operation.
-pub trait Operation<ID, OP, P> {
+pub trait Operation<ID, OP, C = ()> {
```
adzialocha, line 10: This is not really part of this PR but I'd suggest a different name than "Operation" for this trait as this is already a well known term and re-using it here for something different can be confusing. How about `GroupMessage` or `GroupControlMessage` or something like this?

adzialocha: Or maybe even better, split up into multiple traits as we already do in other processors:

`id` -> `Digest` (`p2panda-core`)
`dependencies` -> `Ordering` (`p2panda-stream`)
`group_id` & `action` -> `GroupMessage` (`p2panda-auth`)

p2panda-auth/src/processor/processor.rs
```diff
@@ -0,0 +1,313 @@
...
+    where
+        SID: for<'a> Deserialize<'a> + Serialize,
+    {
+        // Convert this Operation<E> into a GroupsOperation. If this fails then the groups
+        // extension was not present and so we consider this a non-groups operation which does not
+        // require processing.
+        let Ok(groups_operation): Result<GroupsOperation<C>, _> = operation.clone().try_into()
+        else {
+            // @TODO: should this be an error rather than silently ignoring non-groups operations?
+            debug!(id = operation.hash.to_hex(), "ignore non-groups operation");
```
adzialocha: Good question!

I assume it depends in which context this is used and what the assumptions are.

Reading the API I somewhat see mixed signals here as a) You suggest that this processor can be used in pipelines which support other use-cases outside of groups and ignore it in these cases b) The conversion is fallible, leading to an error, even though I'd assume that this should only happen if something _went wrong_, using an `Option` is maybe suggesting better that we're doing something _optional_ here.

1. If this processor is _only_ for processing groups then I'd certainly fail here or even remove the conversion and rather make it an "extraction" / assume that the extension is always given (use `Borrow<GroupsArgs>` as a trait bound etc.)
2. If the processor can be used in a pipeline where we _sometimes_ want to process group messages and sometimes do other things, then I'd keep the early return, make the logging less verbose (`trace`) since it might happen _a lot_ in production and consider not using an error to indicate that the info is missing as it's not a failure case.

adzialocha: In `p2panda` we'll probably only have very concrete extension variants (`enum`, for example for application messages, group control messages, encryption control messages, etc.) which will _fail_ when they have the wrong format (during deserialization) or if they contain false data. So we want to verify the _kind_ of operation before and return `None` towards processors where processing is optional.

p2panda-auth/src/processor/processor.rs
```diff
@@ -0,0 +1,313 @@
...
+    E: Extensions + Extension<GroupsArgs<C>> + Extension<L>,
+    L: LogId,
+    C: Conditions + Serialize + for<'a> Deserialize<'a>,
+{
+    // Check if the operation is already in the store.
+    let has_operation = <SqliteStore as OperationStore<Operation<E>, Hash, L>>::has_operation(
+        store,
+        &operation.hash,
+    )
+    .await?;
```
adzialocha: Checking if the operation exists has to be part of the same transaction

p2panda-auth/src/processor/processor.rs
```diff
@@ -0,0 +1,313 @@
...
+        debug!(
+            group_id = groups_operation.group_id().to_hex(),
+            "current group membership: {:?}",
+            y.members(groups_operation.group_id())
+        );
+
+        // Process the operation in the orderer.
+        let operation_id = groups_operation.id();
+        let dependencies = groups_operation.dependencies();
+        let mut orderer = CausalOrderer::new(store.clone());
```
adzialocha: The `CausalOrderer` should probably live in the struct and not be constructed each time we process

p2panda-auth/src/processor/processor.rs
```diff
@@ -0,0 +1,313 @@
...
+        };
+
+        debug!(id = operation.hash.to_hex(), "process groups operation");
+
+        // Check if the operation is already in the store, if it isn't then we'll insert it
+        // ourselves.
+        insert_operation_checked::<E, L, C>(&store, &operation).await?;
+
+        // Start a transaction for all following database actions.
+        let permit = store.begin().await?;
```
adzialocha: Mostly just thinking loudly:

Wrapping the ordering _and_ groups processor into one transaction is important here. If we would have two transactions here (one per processor) and the groups processor would be interrupted mid-way, we'd loose that operation as it was already "ready-ed" by the orderer.

This is a strong argument to keep both of them together. Another, more flexible and modular, approach is to keep the processing state for each operation in a database. And finally, our automatic replays on nackes events solve this issue as well in `p2panda`.

p2panda-auth/src/processor/processor.rs
```diff
@@ -0,0 +1,313 @@
...
+        Some(operation) => { operation as Operation<E> }.try_into()?,
+        None => {
+            return Err(GroupsProcessorError::<C>::MissingOperation(*id));
+        }
+    };
+    Ok(operation)
+}
+
+/// Insert an operation to the store if it isn't already present.
+async fn insert_operation_checked<E, L, C>(
```
adzialocha: This feels like repetition of what ingest does but without testing and assuring atomicity. Can you use the method directly here instead from `p2panda-stream`?

p2panda-auth/examples/groups.rs
```diff
@@ -0,0 +1,455 @@
...
+    let mut buffer = String::new();
+    let stdin = std::io::stdin();
+    loop {
+        stdin.read_line(&mut buffer)?;
+        line_tx.blocking_send(buffer.clone())?;
+        buffer.clear();
+    }
+}
+
+fn create_operation(
```
adzialocha: We can use the `TestLog` from `p2panda-core` instead here

p2panda-auth/examples/groups.rs
```diff
@@ -0,0 +1,455 @@
...
+const RELAY_URL: &str = "https://euc1-1.relay.n0.iroh-canary.iroh.link.";
+
+pub fn setup_logging() {
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer().with_writer(std::io::stderr))
+        .with(EnvFilter::from_default_env())
+        .try_init()
+        .ok();
+}
+
```
adzialocha: Can we pull this from the duplicate in `test_utils`?

p2panda-auth/examples/groups.rs
```diff
@@ -0,0 +1,455 @@
...
+    fn extract(header: &Header<Self>) -> Option<LogId> {
+        Some(header.extensions.log_id)
+    }
+}
+
+#[derive(Parser)]
+struct Args {
+    /// Bootstrap node identifier.
+    #[arg(short = 'b', long, value_name = "BOOTSTRAP_ID")]
+    bootstrap_id: Option<NodeId>,
```
adzialocha: For an example I find this already pretty verbose / doing a lot non-groups related. For showcasing the functionality to a new user I'd suggest making examples which do not introduce too unrelated topics (networking, bootstrap nodes, topics, mdns, ...).

p2panda-auth/src/processor/processor.rs
```diff
@@ -0,0 +1,313 @@
...
+use crate::processor::{GroupsArgs, GroupsOperation};
+use crate::traits::{Conditions, IdentityHandle, Operation as GroupsOperationTrait, OperationId};
+
+type GroupsCrdtError<C> =
+    crate::group::GroupCrdtError<PublicKey, Hash, GroupsOperation<C>, C, StrongRemove<C>>;
+type StrongRemove<C> = crate::group::resolver::StrongRemove<PublicKey, Hash, GroupsOperation<C>, C>;
+type GroupsState<C> = GroupCrdtState<PublicKey, Hash, GroupsOperation<C>, C>;
+
+impl IdentityHandle for PublicKey {}
+impl OperationId for Hash {}
```
adzialocha, line 27: If you reuse the traits from `p2panda-core` then you get these impls for free

p2panda-auth/src/processor/processor.rs
```diff
@@ -0,0 +1,313 @@
...
+#[derive(Clone)]
+pub struct GroupsProcessor<E, L, C = ()> {
+    _marker: PhantomData<(E, L, C)>,
+}
+
+impl<E, L, C> GroupsProcessor<E, L, C>
+where
+    E: Extensions + Extension<GroupsArgs<C>> + Extension<L>,
+    L: LogId,
+    C: Conditions + Serialize + for<'a> Deserialize<'a>,
```
adzialocha, line 42: Maybe worth adding the serde trait bounds to the `Conditions` super-trait as well

p2panda-stream/src/orderer/mod.rs
```diff
@@ -7,6 +7,6 @@ mod processor;
 mod tests;
 mod traits;
 
-pub(crate) use orderer::CausalOrderer;
+pub use orderer::CausalOrderer;
```
adzialocha, line 10: Making this public exports two very similar sounding types now ("CausalOrderer" and "Orderer") which might confuse users

adzialocha (changes requested):
Nice to see how a "low-level" library can have a "high-level" interface as well! Maybe other "low-level" crates will follow some day!

I have a bunch of change requests which are mainly around repeating types or methods we already have in other places and some minor things + clippy seems to not be happy in CI. The semantics around non-existing groups args is probably something we should figure out before merging as well.

Generally speaking I can see a bit of repetition in functionality across `p2panda-stream` and `p2panda-auth` here (around ingest especially) and I wonder if it wouldn't be better if we'd build a pipeline (with ingest, orderer and a groups processor defined in `p2panda-stream`), that would allow us to keep the code rather thin here, make things more readable as they're less low-level and re-use already tested parts.

The tx issue across the orderer- and groups-processor is still an open question (see my comment) and before this isn't solved I'm fine with leaving it like this now or building a pipeline but without crash resilience guarantee for now, but latest when moving things into `p2panda` we should also maybe consider refactoring here as well? In the next weeks we might have different types of pipelines ready in the Node API of which one can be uses here!

p2panda-auth/src/processor/processor.rs
adzialocha: I guess re-using the existing ingest code here (see other comments) would prevent such footguns

p2panda-auth/examples/groups.rs
adzialocha: Not a blocker really for this PR, I want to flag this here though for discussing later what we expect examples to be as they're part of public documentation

p2panda-stream/src/orderer/mod.rs
adzialocha: I don't have a good suggestion here yet what to do about it but would see need in fixing this. Options I can see are:

1. Come up with better / more distinct names to separate these two
2. Use a pipeline with high-level processors instead of going low-level (more work, probably not what you want to do here)

p2panda-auth/src/processor/processor.rs
```diff
@@ -0,0 +1,313 @@
...
+use thiserror::Error;
+use tracing::debug;
+
+use crate::group::{GroupCrdt, GroupCrdtState};
+use crate::processor::args::GroupsArgsError;
+use crate::processor::{GroupsArgs, GroupsOperation};
+use crate::traits::{Conditions, IdentityHandle, Operation as GroupsOperationTrait, OperationId};
+
+type GroupsCrdtError<C> =
+    crate::group::GroupCrdtError<PublicKey, Hash, GroupsOperation<C>, C, StrongRemove<C>>;
```
mycognosist: Small stylistic suggestion: I'd first import the `GroupCrdtError` type with `use crate::group::GroupCrdtError` as part of the block spanning from L15 to L18. That should make the `GroupsCrdtError<C>` type definition slightly more readable.

The same thing can be done for `StrongRemove<C>` on the line below.

p2panda-spaces/src/auth/orderer.rs
```diff
@@ -1,105 +1,105 @@
...
-    }
-
-    fn next_ready_message(
-        _y: Self::State,
-    ) -> Result<(Self::State, Option<Self::Operation>), Self::Error> {
-        // We shift "dependency checked" message ordering to outside of `p2panda-spaces`.
-        unreachable!()
-    }
-}
+// // SPDX-License-Identifier: MIT OR Apache-2.0
```
mycognosist: Just wanted to flag that the contents of this entire file are commented-out. Was it maybe intended to be deleted?

p2panda-stream/src/orderer/mod.rs
mycognosist: `CausalOrderer` already has good documentation but `Orderer` has none. Perhaps a good first step would be to document `Orderer` (outside of this PR) and make clear the difference between the two.

p2panda-auth/examples/groups.rs
```diff
@@ -0,0 +1,455 @@
...
+    let topic: Topic = if let Some(topic) = args.topic {
+        let topic = hex::decode(topic)?;
+        topic.try_into().expect("topic id should be 32 bytes")
+    } else {
+        Topic::new()
+    };
+
+    // Set up sync for p2panda operations.
+    let store = SqliteStore::temporary().await;
+    store.associate(&topic, &public_key, &LOG_ID).await.unwrap();
```
mycognosist: When trying to run this example locally I hit a panic:

```
cargo run --example groups

thread 'main' (55913) panicked at p2panda-auth/examples/groups.rs:129:57:
called `Result::unwrap()` on an `Err` value: TransactionMissing
```

mycognosist (changes requested):
Followed along as best I could. I've left a couple of comments. I'd love to be able to try out the example! I agree with adz on some of the suggestions around refining naming; we have a lot of overlap of concepts and types between crates now so it makes sense that some points of confusion are emerging.

p2panda-auth/examples/groups.rs
sandreae: This example was intended to show how the auth processor can be integrated into the p2panda stack, so I don't see a way to make it much simpler (Node didn't exist at the time ;-p). But I agree that as a user facing documentation we want examples to be simpler, and when we bring the auth processor into p2panda-stream we'd need to update this example, or even move it elsewhere (as an e2e example of using auth groups) at which point it will probably look completely different. 

I'm going to comment further below about how to progress with the changes in this PR (I'm open to moving the processor straight into p2panda-stream) and this might have implications for the example too.

p2panda-auth/examples/groups.rs
sandreae: Ah, sorry, this must have broken when I brought in the SQLiteStore, I'll fix, then you can play with the example :-)

sandreae:
> Generally speaking I can see a bit of repetition in functionality across p2panda-stream and p2panda-auth here (around ingest especially) and I wonder if it wouldn't be better if we'd build a pipeline (with ingest, orderer and a groups processor defined in p2panda-stream), that would allow us to keep the code rather thin here, make things more readable as they're less low-level and re-use already tested parts.

Yeah, it's not surprising there's a lot of crossover... I'm very open to moving this work into p2panda-stream and implementing the processor there with the existing tools. It would be great to address general processor piping questions there and come to some hopefully re-useable patterns.

p2panda-auth/src/traits/mod.rs
sandreae: I appreciate that there's some duplication of trait definitions here and in p2panda-core now, and agree that we should aim for better reuse. However this is unrelated to changes introduced in this PR and I'd prefer to deal with them in an isolated PR once this work has been completed. Same goes for the other suggestions.

sandreae: I raised an issue here: https://github.com/p2panda/p2panda/issues/1133

p2panda-auth/src/traits/operation.rs
sandreae: https://github.com/p2panda/p2panda/issues/1134

p2panda-auth/src/processor/processor.rs
sandreae: Nice summary of the problem area, thanks. In this case (the stand-alone auth processor) I believe the primary use case will be to use it in a pipeline where all operations are processed, and it should pick up and process only those with auth args in the extensions. For that reason I'll adjust the misleading `TryFrom` pattern accordingly.

p2panda-auth/examples/groups.rs
sandreae: I've managed to hide away quite a lot of this "unnecessary" lower level detail from the example user now, and also in the code. It's still not a perfectly isolated example, but much better I think.

p2panda-auth/examples/groups.rs
sandreae: I've fixed this now, have a go with the example :-)

p2panda-auth/src/processor/processor.rs
sandreae: There's some naming conflict with the new types and imported types, so I still needed to import them in-place, I tried to improve it slightly by importing the `group` module first though. Hope you agree that's better.

p2panda-stream/src/orderer/mod.rs
sandreae: I still haven't dealt with this point yet, will look into it later today.

mycognosist (approved):
Thanks for the changes!

I've had a lot of fun playing with the example. While there's the option of using `Node` instead of manually assembling the networking stack, I don't mind the extra verbosity. Perhaps this could be reconsidered in the future PR which will focus on the processor.

The `DuplicateOperation` errors can be a bit noisy but on the other hand it can be useful to see that sync is indeed happening and the processor is doing the work of rejecting unnecessary operations. One option could be to move these into `error!()` logs to keep the `stdout` limited to group state transformations. Definitely not a blocker for me.

sandreae:
> The DuplicateOperation errors can be a bit noisy but on the other hand it can be useful to see that sync is indeed happening and the processor is doing the work of rejecting unnecessary operations. One option could be to move these into error!() logs to keep the stdout limited to group state transformations. Definitely not a blocker for me.

This is a good point, I think in this example it would be nice to silence that error. I'll make that change.

p2panda-stream/src/orderer/mod.rs
sandreae: I don't have a good suggestion for a fix here in this PR, I'm going to raise an issue to deal with the naming confusion and documentation.

p2panda-auth/src/processor/processor.rs
sandreae: https://github.com/p2panda/p2panda/issues/1133

p2panda-stream/src/orderer/mod.rs
sandreae: https://github.com/p2panda/p2panda/issues/1138

p2panda-auth/src/processor/processor.rs
sandreae: All serde traits are behind a feature flag and required only in the groups CRDT itself. We can raise a separate issue if this should be addressed in a different way, but it wasn't introduced in this PR in any case.

p2panda-stream/src/orderer/mod.rs
adzialocha: I can imagine we revert that change as soon as we use auth in the pipeline

adzialocha (approved):

adzialocha:
Needs .git history clean up, otherwise good to be merged now from my side!

Issue #1114, opened 2026-04-07
Improve APIs / types where `NodeId` and `RelayUrl` is combined


jsparber:
I think `NodeId` should store an optional `RelayUrl` since they are generally used together. See `NodeBuilder::bootstrap()`
There is already `node::EndpontAddr` which `iroh::EndpointAddr` which does almost what p2anda needs, but `node::EndpontAddr` uses iroh types so it is an implementation detail and shouldn't be public.

adzialocha:
I agree that some type combining these two things will be useful, I want to suggest a different approach though:

* `NodeId` is the identifier of a node (public key) and we certainly need to keep the type for it. The equivalent is `EndpointId` in iroh.
* iroh has an `EndpointAddr` combines the two things (id and addresses / relays) as you say, our equivalent is `TransportInfo` (it supports more things than just endpoint addr of iroh) but it lacks the id so far
* We have conversions to translate from `EndpointAddr` to `TransportInfo`: https://docs.rs/p2panda-net/latest/p2panda_net/addrs/struct.TrustedTransportInfo.html#impl-From%3CEndpointAddr%3E-for-TrustedTransportInfo

So my suggestion is:

* Add `NodeId` to both `TransportInfo` variants (trusted & authenticated)
* Change insert_bootstrap methods and variants to something along the lines of `impl Into<TrustedTransportInfo>`

PR #1125, opened 2026-04-11
Add insert_bootstrap to the node API as a thin wrapper around the net…

jadehopepunk:
I suggest the addition of the insert_bootstrap function from network to the node API.

I imagine you're going to have to carefully consider additions to the node API. When I migrated to the node, this one felt pretty immediate. I am allowing users to add bootstrap nodes at runtime, so I don't know how else I'd achieve this. Unlike some other parts of the network config, this is designed to be added on the fly fairly easily. 

There are no tests for this PR. It's such a thin wrapper that doesn't seem worth it to me, but I'm happy to add them. Mostly this is a prompt for discussion.

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ Highlights are marked with a pancake 🥞
 - Use Borrow to express required args in ingest [#1078](https://github.com/p2panda/p2panda/pull/1078)
 - Use MockInstant for Timestamp when running tests [#1081](https://github.com/p2panda/p2panda/pull/1081)
 - Derive log id from topic [#1082](https://github.com/p2panda/p2panda/pull/1082)
+- Expose insert_bootstrap on the node API [#1125](https://github.com/p2panda/p2panda/pull/1125)
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


adzialocha (changes requested):
Thank you for this addition, this is definitely a helpful method!

If you see the builder method we want to ask for the relay url as well to register bootstraps (with only the public key we don't know which relay to call to find out the direct address): https://github.com/p2panda/p2panda/blob/main/p2panda/src/builder.rs#L66

Later we want to maybe have a single type which merges these together (see https://github.com/p2panda/p2panda/issues/1114)

jadehopepunk:
Good point @adzialocha . Like this? Should the RelayURL be optional? We had a feature somewhere before that adding an bootstrap with no relay specified would use this node's relay by default. Should that be in the node or application level?

adzialocha:
> Good point @adzialocha . Like this? Should the RelayURL be optional? We had a feature somewhere before that adding an bootstrap with no relay specified would use this node's relay by default. Should that be in the node or application level? 

I'm undecided here, it encourages relying on the same relay / centralization maybe but it's a handy feature? What do you think @sandreae and @mycognosist ?

adzialocha (approved):
✨ 

Ready to merge from my end in any case. We can do a squash merge here to compact the git commits + add a CHANGELOG.md on our end.

jadehopepunk:
I've had a go at adding the changelog line, is that right? Go ahead and squash and merge as you like.

PR #1135, opened 2026-04-27
Functionality for importing external operation streams in node API

sandreae:
## Context

Being able to import already forged operations from an arbitrary source is an important feature for several use cases, these include importing from a USB stick or the filesystem, or syncing with a remote service, like a "support node". The current node API does not currently support this use case, it is only possible to forge operations locally or receive them over the built-in sync protocol which runs in the background. 

## Decision

Introduce an API for importing operations from an arbitrary source. Imported operations should be processed in the existing node pipeline and users should be notified when processing is complete via the event stream. The API should be idiomatic for the target use cases (reading from file, streaming from a remote service).

## Action

Add an `import` method to the `StreamPublisher` which a user receives when calling `node.stream(topic)`. The import method takes an `impl Stream<Item = Operation>` which is polled internally until completion, produced operations are processed in the node pipeline and events emitted via the event stream.

- [x] add import method to `StreamPublisher` which takes a `impl Stream<Item = Operation>`
- [x] associate each stream with a "session id" 
- [x] received operations are processed in the pipeline
- [x] issue lifetime events for each "batch" of operations (start, operation, end)
- [ ] attach basic metrics to events
    - optional but nice 
- [x] return `SessionId` on calls to `import()` 
- [x] support back-pressure for import operation streams
- [x] test

## 📋 Checklist

### DIFF
--- a/p2panda/src/streams/external_stream.rs
+++ b/p2panda/src/streams/external_stream.rs
@@ -1,0 +1,96 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use futures_util::stream::SelectAll;
+use futures_util::stream::{BoxStream, once};
+use futures_util::{FutureExt, Stream, StreamExt};
+use tokio::sync::oneshot;
+
+use crate::operation::Operation;
+
+pub type SessionId = u64;
+
+/// Set of currently active external streams.
+#[derive(Default)]
+pub(crate) struct ExternalStream {
+    streams: SelectAll<BoxStream<'static, ExternalStreamEvent>>,
+    next_session_id: SessionId,
+}
+
+impl ExternalStream {
+    /// Insert a new stream into the set, it is assigned the next available session id, and a
+    /// future is returned which will complete when the stream has ended and all operations are
+    /// processed.
+    pub fn insert(&mut self, stream: BoxStream<'static, Operation>) -> ExternalStreamFuture {
+        let session_id = self.next_session_id();
+        let (complete_tx, complete_rx) = oneshot::channel::<()>();
+        let future = ExternalStreamFuture {
+            session_id,
+            complete_rx,
+        };
+
+        let event_stream = once(async move { ExternalStreamEvent::Start { session_id } })
+            .chain(stream.map(move |operation| ExternalStreamEvent::Operation {
+                session_id,
+                operation: Box::new(operation),
+            }))
+            .chain(once(async move {
+                let _ = complete_tx.send(());
+                ExternalStreamEvent::End { session_id }
+            }));
+
+        self.streams.push(Box::pin(event_stream));
+
+        future
+    }
+
+    fn next_session_id(&mut self) -> SessionId {
+        let next = self.next_session_id;
+        self.next_session_id += 1;
+        next
+    }
+}
+
+impl Stream for ExternalStream {
+    type Item = ExternalStreamEvent;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.streams.poll_next_unpin(cx)
+    }
+}
+
+pub(crate) enum ExternalStreamEvent {
+    Start {
+        session_id: SessionId,
+    },
+    Operation {
+        session_id: SessionId,
+        operation: Box<Operation>,
+    },
+    End {
+        session_id: SessionId,
+    },
+}
+
+/// Future which can be awaited to find out when external stream has completed processing.
+#[derive(Debug)]
+pub struct ExternalStreamFuture {
+    session_id: SessionId,
+    complete_rx: oneshot::Receiver<()>,
+}
+
+impl ExternalStreamFuture {
+    pub fn session_id(&self) -> SessionId {
+        self.session_id
+    }
+}
+
+impl Future for ExternalStreamFuture {
+    type Output = Result<(), oneshot::error::RecvError>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        self.complete_rx.poll_unpin(cx)
+    }
+}
--- a/p2panda/src/streams/stream.rs
+++ b/p2panda/src/streams/stream.rs
@@ -7,6 +7,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
+use futures_util::stream::BoxStream;
 use futures_util::{FutureExt, Stream, StreamExt};
 use p2panda_core::cbor::{DecodeError, EncodeError, decode_cbor, encode_cbor};
 use p2panda_core::traits::Digest;
@@ -28,6 +29,9 @@ use crate::node::{AckPolicy, CreateStreamError};
 use crate::operation::{Extensions, Header, LogId, Operation};
 use crate::processor::{Event, Pipeline, ProcessorError};
 use crate::streams::acked::{Acked, AckedError};
+use crate::streams::external_stream::{
+    ExternalStream, ExternalStreamEvent, ExternalStreamFuture, SessionId,
+};
 use crate::streams::replay::{ReplayError, StreamFrom, replay_log_ranges};
 use crate::streams::sync_metrics::{self, Aggregator, SessionPhase, SyncError};
 
@@ -39,6 +43,10 @@ const BUFFER_SIZE: usize = 16;
 /// Number of items which can stay in the buffer before processing kicks in for locally published
 /// items. If buffer runs full, creating new operations will apply backpressure.
 const PUBLISH_BUFFER_SIZE: usize = 128;
+
+/// Number of streams which can stay in the import buffer. If the buffer runs full importing of
+/// new streams will have backpressure applied.
+const IMPORT_BUFFER_SIZE: usize = 16;
 
 /// Returns publish and subscribe halfs of an eventually consistent messaging stream for a given
 /// topic.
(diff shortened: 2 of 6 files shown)


p2panda/src/streams/stream.rs
```diff
@@ -490,6 +531,17 @@ where
         self.publish_inner(message, true).await
     }
 
+    /// Import operations from a stream.
+    pub async fn import(
```
mycognosist, line 548: I wonder if it might be nice to add an optional name / source when calling import? That would allow a user to tag each import stream with an identifier of their choice. This could be useful for understanding stream events when potentially multiple import streams have been provided; each variant of `ImportEvent` would include the name.

`handle.import(stream, Some("usb"))`

Very much a nice-to-have but figured I'd mention it here.

p2panda/src/streams/import_streams.rs
```diff
@@ -0,0 +1,61 @@
...
+use futures_util::stream::SelectAll;
+use futures_util::stream::{BoxStream, once};
+use futures_util::{Stream, StreamExt};
+
+use crate::operation::Operation;
+
+pub type BatchId = u64;
+
+#[derive(Default)]
+pub(crate) struct ImportStreams {
```
adzialocha: I like the "Import" naming, but giving it some more thought maybe it's too specific for how it's used? How about `ExternalStream`? It just indicates that the data is coming from outside our stack.

adzialocha: I'm fine with also doing it singular, since in the end it's technically still just a single stream (fan-in)

p2panda/src/streams/import_streams.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl Stream for ImportStreams {
+    type Item = ImportEvent;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.streams.poll_next_unpin(cx)
+    }
+}
+
+#[allow(clippy::large_enum_variant)]
```
adzialocha: Could make sense to Box the operation

p2panda/src/streams/stream.rs
```diff
@@ -7,6 +7,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
+use futures_util::stream::BoxStream;
 use futures_util::{FutureExt, Stream, StreamExt};
```
adzialocha, line 11: I think in other parts of our APIs we're using tokio's `StreamExt`. It's better we're using `futures` here (it's more feature rich) but maybe we should streamline it in a later refactor (let's make a ticket)

p2panda/src/streams/stream.rs
```diff
@@ -114,6 +116,13 @@ where
         oneshot::Sender<Event<LogId, Extensions, Topic>>,
     )>(PUBLISH_BUFFER_SIZE);
 
+    // Channel for sending operation streams to the processor for importing.
+    // @TODO: do we want to bound this channel and if so what would the value be?
+    let (import_tx, mut import_rx) = mpsc::unbounded_channel::<BoxStream<'static, Operation>>();
```
adzialocha: Yes! That would be great. Any number will be sufficient for now.

p2panda/src/streams/stream.rs
```diff
@@ -240,6 +249,36 @@ where
...
+
+                    Some(stream) = import_rx.recv() => {
+                        import_streams.insert(stream);
+                        continue;
+                    }
+
+                    // @TODO: we currently don't have any back-pressure here (like we do with
+                    // publishing operations). We could introduce a bounded channel and add an if
+                    // conditional to this select arm based on available capacity. Or some other
+                    // approach....
```
adzialocha: We could make `pipeline_tx` a bounded channel, then we would have backpressure here automatically, don't think it'll require adding any extra condition handling, the async runtime will automatically "await" on pressure.

If we add the bound on the pipeline it'll be global for all other streams as well which is nice.

adzialocha (commented):
Great! Looks really good already and has very handy functionality.

I can see how we want to maybe deliver some "out-of-the-box" importers already users can just plug in, like something reading from a file (for sneakernets) etc.

p2panda/src/streams/stream.rs
mycognosist: If we bound it then a sane limit would probably be quite small, considering this is a channel for sending streams (not their constituent operations). How many bulk import streams might an application realistically use? Surely not more than 32 or 64.

p2panda/src/streams/stream.rs
```diff
@@ -240,6 +249,36 @@ where
 
                         event
                     }
+
+                    Some(stream) = import_rx.recv() => {
+                        import_streams.insert(stream);
+                        continue;
+                    }
+
+                    // @TODO: we currently don't have any back-pressure here (like we do with
```
mycognosist: The call to `process_operation()` is essentially providing the backpressure here, right? That's the limiting factor on the rate at which we pull the next event from the stream.

mycognosist (commented):
Looking really good! Just added a couple of early review thoughts.

p2panda/src/streams/stream.rs
sandreae: I like the idea, definitely want the user to be able to associate stream events with a certain import stream. I'm not sure about adding another argument to the import method though :thinking: maybe it's fine, I like that the user could give a meaningful name then.

I'd intended that `BatchId`, which is generated per stream and attached to every stream event,  would be how a user could associate events with a certain stream. This isn't actually returned to the user though yet... What do you think about this being a solution? If we return the `BatchId` on a call to import so the user can make this mapping (to a meaningful source) themselves?

sandreae: I'd also suggest renaming `BatchId` to `SessionId`.

p2panda/src/streams/stream.rs
sandreae: Right, I find it a little hard to understand exactly where backpressure is being applied right now actually :sweat_smile:

sandreae: I have added the bounded channel to the pipeline tx/rx now, so I guess that's an additional point where backpressure can be fine tuned.

p2panda/src/streams/stream.rs
adzialocha: > I'd also suggest renaming `BatchId` to `SessionId`.

Yes! "Batch" usually implies processing multiple items at the same time, which is not the case here (we're importing one item after another)

mycognosist: > What do you think about this being a solution? If we return the `BatchId` on a call to import so the user can make this mapping (to a meaningful source) themselves?

That sounds good to me.

adzialocha: Similar to the publish call we can return an object which has the session id plus an optional Future implemented which can be awaited if the import has finished

p2panda/src/streams/external_stream.rs
```diff
@@ -0,0 +1,102 @@
...
+            .chain(stream.map(move |operation| ExternalStreamEvent::Operation {
+                session_id,
+                operation: Box::new(operation),
+            }))
+            .chain(once(async move {
+                if complete_tx.send(()).is_err() {
+                    warn!(
+                        session_id,
+                        "failed sending on external stream complete channel"
+                    );
```
adzialocha: This would always print a warning whenever a user does _not_ use the `ExternalStreamFuture` (thus dropping the rx half) which is likely often the case. I would just ignore any error here and not print anything.

p2panda/src/streams/stream.rs
```diff
@@ -240,6 +257,43 @@ where
...
+                        let session_id = external_stream_future.session_id();
+                        if ready_tx.send(external_stream_future).is_err() {
+                            warn!(session_id = session_id, "failed sending on import ready channel")
+                        };
+                        continue;
+                    }
+
+                    // Receive an operation from any imported external source.
+                    //
+                    // Receive
```
adzialocha: Missing doc-string?

p2panda/src/streams/stream.rs
```diff
@@ -240,6 +257,43 @@ where
 
                         event
                     }
+
+                    // Receive imported external source of operations.
+                    //
+                    // All external streams are polled concurrently and the next available event
+                    // processed.
```
adzialocha: Does this comment about processing events belong here to the "add an external stream" logic?

p2panda/src/streams/stream.rs
```diff
@@ -774,6 +868,9 @@ pub enum PublishError {
     #[error("an error occurred while publishing an operation to the log sync stream: {0}")]
     SyncHandle(String),
 
-    #[error("could not send operation to processor pipeline: {0}")]
+    #[error("could not send to processor pipeline: {0}")]
     SendToProcessor(String),
+
+    #[error("an error occurred awaiting message from processor: {0}")]
+    ReceiveFromProcessor(String),
```
adzialocha, line 876: Where is this error variant used?

p2panda/tests/api.rs
```diff
@@ -384,3 +384,111 @@ async fn replay_stream_from_cursor() {
...
+                hash: header.hash(),
+                header,
+                body,
+            };
+            exported.push(operation);
+            if exported.len() == 3 {
+                break;
+            }
+        }
+    }
```
adzialocha: Just an idea: You could also use `TestLog` here to create these operations to make test simpler

adzialocha (approved):
Looking all good!

I have some minor change requests, all code clean ups + .git history clean up related.

p2panda/src/streams/stream.rs
sandreae: It's used in the import method when awaiting receiving on the ready_rx channel.

adzialocha: Ah, thanks! Can you make an own ImportError for this? It gets harder to understand if they're all merged into a larger type with different detailed errors.

mycognosist (approved):
Great work. The test looks really good.

PR #1139, opened 2026-04-29
Ensure iroh Gossip is shut down gracefully

mycognosist:
We previously observed flakiness in the `leave_overlay_on_drop` gossip test, whose purpose is to ensure cleanup of state when gossip stream handles - or the entire gossip system - are dropped. The test would simply hang without completion.

I tracked the behaviour down to a failure to join the gossip overlay after the p2panda `Gossip` object had been dropped and recreated from scratch for one of the two nodes in the test. This suggested that the node which remained running may have held some incorrect connection state relating to the downed node; the assumption being that this prevented the latest overlay connection attempt to fail.

On further inspection, I realised that the call to gracefully shut down iroh gossip (from within the `post_stop` method of the gossip manager actor) was not being executed. Calling `gossip.shutdown()` directly in `post_stop`, without spawning it in a tokio task, resulted in the shut down but caused a panic of unknown origin. My suspicion is that one of the other gossip actors was causing the panic.

The implemented solution is three-fold (all in the `Drop` impl for p2panda):

1) Drain all children of the gossip manager actor
2) Send the newly-added Shutdown message to the gossip manager actor
3) Drain the gossip manager actor

The test now passes reliably without any panics.

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ Highlights are marked with a pancake 🥞
 - Race where replay_from misses processing ops [#1104](https://github.com/p2panda/p2panda/pull/1104)
 - Race where replay state was determined late [#1104](https://github.com/p2panda/p2panda/pull/1104)
 - React to node address changes in mDNS test [#1141](https://github.com/p2panda/p2panda/pull/1141)
+- Ensure iroh Gossip is shut down gracefully [#1139](https://github.com/p2panda/p2panda/pull/1139)
 
 ## [0.5.2] - 09/03/2026
 
--- a/p2panda-net/src/gossip/actors/manager.rs
+++ b/p2panda-net/src/gossip/actors/manager.rs
@@ -75,6 +75,9 @@ pub enum ToGossipManager {
 
     /// Subscribe to system events.
     Events(RpcReplyPort<broadcast::Receiver<GossipEvent>>),
+
+    /// Gracefully shut down the gossip actor, cleaning up connection state.
+    Shutdown,
 }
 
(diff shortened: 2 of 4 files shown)


adzialocha:
Ran `cargo test` in `p2panda-net` of the branch a couple of times and observed this error:

```plain
---- gossip::tests::join_without_bootstrap stdout ----

thread 'gossip::tests::join_without_bootstrap' (15444) panicked at p2panda-net/src/gossip/tests.rs:167:10:
called `Result::unwrap()` on an `Err` value: ActorSpawn(StartupFailed(StartupFailed(ActorRpc(Messaging(SendErr)))))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I'm not sure if this is related yet, but seems to be new?

adzialocha (approved):

adzialocha:
> Ran `cargo test` in `p2panda-net` of the branch a couple of times and observed this error:
> 
> ```
> ---- gossip::tests::join_without_bootstrap stdout ----
> 
> thread 'gossip::tests::join_without_bootstrap' (15444) panicked at p2panda-net/src/gossip/tests.rs:167:10:
> called `Result::unwrap()` on an `Err` value: ActorSpawn(StartupFailed(StartupFailed(ActorRpc(Messaging(SendErr)))))
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> I'm not sure if this is related yet, but seems to be new?

It only appeared once so far and this PR improves a lot already, so in any case I would merge it.

PR #1141, opened 2026-04-30
React to node address changes in mDNS test

### LINKED ISSUE #1140: mDNS Discovery Test Sometimes Fails
Since yesterday I've been observing flaky behaviour in our `iroh_mdns::tests::mdns_discovery` test.

```
thread 'iroh_mdns::tests::mdns_discovery' (18556) panicked at p2panda-net/src/iroh_mdns/tests.rs:63:5:
assertion failed: result.is_some()
```

The assertion is that Alice's address book contains the newly-discovered node info for Bob, and vice versa. I see both failure scenarios reflected in the test outcome: sometimes the first assertion fails and sometimes only the second.

This may have corresponded to the point at which I bumped the iroh dependencies. https://github.com/p2panda/p2panda/pull/1131/changes

Looking into it now.

mycognosist:
The mDNS discovery test, which relied on assertions that Alice and Bob had learned one another's node info, was sometimes failing. We were relying on a 1 second sleep after spawning the mDNS processes to allow enough time for discovery. I believe this duration is not always sufficient for both peers to discover one another and update their address books, thus resulting in inconsistent test outcomes.

The test now uses address book watchers; we wait for Bob's node info to be updated in Alice's address book (and vice versa) before running the assertions. This ensures that the mutual discovery process has enough time to complete.

Closes https://github.com/p2panda/p2panda/issues/1140

## 📋 Checklist

### DIFF
--- a/p2panda-net/src/iroh_mdns/tests.rs
+++ b/p2panda-net/src/iroh_mdns/tests.rs
@@ -1,9 +1,5 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use std::time::Duration;
-
-use tokio::time::sleep;
-
 use crate::address_book::AddressBook;
 use crate::iroh_endpoint::Endpoint;
 use crate::iroh_mdns::{MdnsDiscovery, MdnsDiscoveryMode};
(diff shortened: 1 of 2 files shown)


p2panda-net/src/iroh_mdns/tests.rs
```diff
@@ -47,7 +68,8 @@ async fn mdns_discovery() {
         .unwrap();
 
     // Wait until they find each other and exchange transport infos.
-    sleep(Duration::from_millis(1000)).await;
+    alice_address_book_bob.recv().await;
```
mycognosist, line 71: There's some redundancy here but I like how explicit the existing address book look-ups are.

sandreae (approved):

PR #1142, opened 2026-04-30
`README.md` and more doc-strings for high-level Node API

mycognosist:
- [x] `README.md` (adz)
- [x] crate-level doc-string (adz)
- [x] `builder()` (glyph)
- [x] `spawn()` (glyph)
- [x] `Node`
    - [x] `insert_bootstrap` / `id` / `network_id` (glyph)
    - [x] `event_stream` (glyph)
    - [x] `ephemeral_stream` (sam)
        - [x] `EphemeralMessage`
        - [x] `EphemeralStreamPublisher`
        - [x] `EphemeralStreamSubscription`
    - [x] `stream` / `stream_from` (adz)
        - [x] `ProcessedOperation`
        - [x] `StreamPublisher`
        - [x] `StreamSubscription`
        - [x] `StreamEvent`
- [x] `NodeBuilder` (glyph)

### DIFF
--- a/p2panda/src/builder.rs
+++ b/p2panda/src/builder.rs
@@ -17,6 +17,9 @@ use crate::network::MdnsDiscoveryMode;
 use crate::node::{AckPolicy, Config, SpawnError};
 use crate::processor::{Pipeline, TaskTracker};
 
+/// Builder for `Node`.
+///
+/// To create the `Node` call `NodeBuilder::spawn()`.
 #[derive(Default)]
 pub struct NodeBuilder {
     signing_key: Option<SigningKey>,
@@ -25,6 +28,7 @@ pub struct NodeBuilder {
 }
 
 impl NodeBuilder {
+    /// Creates a new `NodeBuilder` using default configuration values.
     pub fn new() -> Self {
         NodeBuilder {
             signing_key: None,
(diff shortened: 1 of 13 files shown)


p2panda/src/builder.rs
```diff
@@ -33,36 +37,100 @@ impl NodeBuilder {
...
+    ///
+    /// If left unset, a new private key will be randomly generated.
     pub fn private_key(mut self, private_key: PrivateKey) -> Self {
         self.private_key = Some(private_key);
         self
     }
 
+    /// Defines the database URL to be used by the store.
+    ///
+    /// If left unset, the node will default to using an ephemeral in-memory database.
```
adzialocha, line 60: Maybe good to mention that this needs to be an SQLite database URI: https://sqlite.org/uri.html and that we are running migrations automatically.

Users should use `database_pool` together with `sqlx` if manual migration management is required.

p2panda/src/builder.rs
```diff
@@ -72,41 +140,68 @@ impl NodeBuilder {
         self
     }
 
+    /// Sets the mDNS discovery mode.
+    ///
+    /// mDNS may be set to passive or active mode. If left unset, the mode defaults to passive and
+    /// this node will not actively advertise it's endpoint address on the local area network.
```
adzialocha: In this API the default is "active". We also have an option to disable it altogether.

p2panda/src/builder.rs
```diff
@@ -33,36 +37,100 @@ impl NodeBuilder {
...
+    /// If left unset, the policy defaults to `Automatic` and all messages emitted from topic
+    /// streams will be automatically acknowledged.
+    ///
+    /// See the `Node::stream(topic)` documentation for further information.
     pub fn ack_policy(mut self, value: AckPolicy) -> Self {
         self.config.ack_policy = value;
         self
     }
 
+    /// Sets the network identifier.
```
adzialocha, line 87: Should we add a note on the default network id?

p2panda/src/builder.rs
```diff
@@ -33,36 +37,100 @@ impl NodeBuilder {
         }
     }
 
+    /// Sets the private key.
+    ///
+    /// The public key derived from the given private key is used to identify the node in the
+    /// network. For example, this key can be used to directly connect to the node. The private key
+    /// is used to sign operations to ensure data integrity and authenticity.
```
adzialocha: The private key is also used to authenticate the node during the connection handshake (using TLS 1.3 over Ed25519). You say this implicitly here but maybe this is technically more correct than pure derivation.

p2panda/src/builder.rs
mycognosist: Ah, right, your API changes from PR#1143 are not yet included on this branch - hence the divergence. I'll leave this unresolved so I remember to make the change after we've rebased this branch onto `main`.

p2panda/src/streams/stream.rs
```diff
@@ -470,12 +527,16 @@ where
         self.topic
     }
 
-    /// Publish a message.
+    /// Publish a message in a topic stream.
```
mycognosist:
```suggestion
    /// Publish a message into a topic stream.
```

p2panda/src/streams/stream.rs
```diff
@@ -509,8 +570,8 @@ where
             .await?;
         let hash = operation.hash;
 
-        // Start processing operation in pipeline. Keep an oneshot receiver around to allow users
-        // to optionally await & get informed when processing has finished.
+        // Start processing operation in pipeline. Keep an oneshot receiver around to allow users to
```
mycognosist:
```suggestion
        // Start processing operation in pipeline. Keep a oneshot receiver around to allow users to
```

p2panda/src/streams/stream.rs
```diff
@@ -595,16 +690,24 @@ where
     }
 }
 
+/// Operations with application messages, system events and errors coming from an topic stream
```
mycognosist:
```suggestion
/// Operations with application messages, system events and errors coming from a topic stream
```

p2panda/src/streams/stream.rs
```diff
@@ -595,16 +690,24 @@ where
...
+/// subscription.
 #[derive(Clone)]
 #[allow(clippy::large_enum_variant)]
 pub enum StreamEvent<M> {
+    /// Operation with application message coming from a topic stream.
+    ///
+    /// Operations can arrive from various sources, for example from a sync session with another,
+    /// remote node, a locally created message or import.
```
mycognosist:
```suggestion
    /// Operations can arrive from various sources. For example, from a sync session with a remote
    /// node, a locally created message or import.
```

p2panda/src/streams/stream.rs
```diff
@@ -595,16 +690,24 @@ where
...
     Processed {
         /// Processed operation.
         operation: ProcessedOperation<M>,
 
         /// The source of the operation.
         source: Source,
     },
+
+    /// Sync session with a remote node started.
```
mycognosist:
```suggestion
    /// Sync session started with a remote node.
```

p2panda/src/streams/stream.rs
```diff
@@ -627,6 +730,8 @@ pub enum StreamEvent<M> {
         /// Total sessions currently running over the same topic.
         topic_sessions: u64,
     },
+
+    /// Sync session with a remote node ended.
```
mycognosist:
```suggestion
    /// Sync session ended with a remote node.
```

mycognosist (commented):
Looking great! 🥳 Left a couple of grammar-related suggestions.

p2panda/src/node.rs
mycognosist, line 25: Would be good to include high-level docs for the node. This can probably come from the crate-level docs or README.

adzialocha: Yes, definitely!

p2panda/src/lib.rs
```diff
@@ -1,5 +1,254 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+//! p2panda's high-level Node API is an opinionated, out-of-the-box peer-to-peer stack which
+//! orchestrates all individual [p2panda] modules.
+//!
+//! ```rust
+//! # #[tokio::main]
+//! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! # let topic = p2panda::Topic::new();
```
sandreae, line 9: I would maybe like to see the topic being generated in this example.

PR #1147, opened 2026-05-12
Deduplicate received operations in sync manager event stream

### LINKED ISSUE #1145: Expected behaviour around concurrent sync sessions and duplicate messages
During each sync session we're transmitting data bi-directionally. This means that A receives data from B and B receives data from A. When connecting to another node we're starting usually two sync sessions at the same time (incoming / outgoing) due to normal "race conditions" in a distributed system.

The downside is now that we're transmitting the same data potentially twice. Another option is to make the sync session uni-directionally, so A only sends data to B or vice-versa but this can lead to another problem:

If Node A goes offline, Node B will mark A as stale and will not attempt contacting them ever again. We need A to come up again and initialise a sync session to a) be unmarked by B and b) receive new data from B.

We might want to look into this again and see if we can solve a) and b) differently, to allow uni-directional / more efficient data exchange.

If we stick to the bi-directional approach we might want to de-duplicate messages arriving at the application layer as currently users will observe duplicates because of this behaviour. Since the pipeline strictly serializes every operation we should be able to ignore any following duplicate if it was already acked.

We might want to apply de-duplication in any case as an operation P might arrive from multiple sources at the same time anyhow, independent of the approach of the sync protocol.

sandreae:
This change fixes a bug where deduplication was only occurring per-session and when sending operations. An additional deduplication buffer is now created per sync manager event stream subscription and operations receive from any session are deduplicated.

## 📋 Checklist

### DIFF
--- a/p2panda-sync/src/manager/mod.rs
+++ b/p2panda-sync/src/manager/mod.rs
@@ -30,6 +30,7 @@ use tokio_stream::wrappers::BroadcastStream;
 use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
 use tracing::debug;
 
+use crate::dedup::Dedup;
 use crate::manager::event_stream::{ManagerEventStreamState, StreamDebug};
 use crate::protocols::{TopicLogSync, TopicLogSyncError, TopicLogSyncEvent};
 use crate::traits::Manager;
@@ -48,6 +49,13 @@ pub type ToTopicSync<E> = ToSync<Operation<E>>;
 /// A handle can be acquired to a sync session via the session_handle method for sending any live
 /// mode operations to a specific session. It's expected that users map sessions (by their id) to
 /// any topic subscriptions in order to understand the correct mappings.  
+///
+/// There are two points where message deduplication occurs; 1) per-subscription deduplication of
+/// received operations across all sessions for this manager occurs in the event stream 2)
+/// per-session deduplication occurs on sending operations to a remote, this is especially
+/// required during live-mode when we may receive the same new operation from several sources. The
+/// former stops any local consumers from receiving the same operation more than once, the latter
+/// reduces unnecessary noise on the network.
 #[allow(clippy::type_complexity)]
 #[derive(Debug)]
 pub struct TopicSyncManager<T, S, L, E>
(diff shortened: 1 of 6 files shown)


p2panda-sync/src/manager/mod.rs
```diff
@@ -551,51 +563,41 @@ mod tests {
...
                         if let TopicLogSyncEvent::OperationReceived { operation, .. } = event.event() {
-                            operations_a.push(*operation.clone());
+                            println!("A received operation: {}", operation.hash);
+                            operations_a.push(operation.header().clone());
                         }
                     }
                     Some(event) = event_stream_b.next() => {
                         if let TopicLogSyncEvent::OperationReceived { operation, .. } = event.event() {
-                            operations_b.push(*operation.clone());
+                            println!("B received operation: {}", operation.hash);
```
sandreae, line 572: Need to remove this and the above println.

p2panda-sync/src/manager/mod.rs
```diff
@@ -48,6 +49,13 @@ pub type ToTopicSync<E> = ToSync<Operation<E>>;
 /// A handle can be acquired to a sync session via the session_handle method for sending any live
 /// mode operations to a specific session. It's expected that users map sessions (by their id) to
 /// any topic subscriptions in order to understand the correct mappings.  
+///
+/// There are two points where message deduplication occurs; 1) per-subscription deduplication of
+/// received operations across all session for this manager occurs in the event stream 2)
```
mycognosist:
```suggestion
/// received operations across all sessions for this manager occurs in the event stream 2)
```

mycognosist (approved):
Looks good to me.

One note on naming which is out of scope for this PR but I'll mention it here anyway; I'd recommend renaming `Dedup` to `DeduplicationBuffer` and giving the `dedup` module the same treatment. It's a small change but it should make the code easier to understand and follow. In general I'm in favour of longer names when they support clarity.

sandreae:
> understand

Thanks for the feedback, I created an issue https://github.com/p2panda/p2panda/issues/1149

adzialocha (approved):

PR #1150, opened 2026-05-13
Move shared dependencies into workspace Cargo.toml

sandreae:
## 📋 Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,3 +16,26 @@ members = [
 ]
 
 [workspace.lints.rust]
+
+[workspace.dependencies]
+ciborium = { version = "0.2.2", default-features = false}
+futures-util = { version = "0.3.32", default-features = false }
+hex = { version = "0.4.3", default-features = false }
+p2panda = { path = "p2panda", version = "0.5.2", default-features = false }
+p2panda-auth = { path = "p2panda-auth", version = "0.5.2", default-features = false }
+p2panda-core = { path = "p2panda-core", version = "0.5.2", default-features = false }
+p2panda-discovery = { path = "p2panda-discovery", version = "0.5.2", default-features = false }
+p2panda-encryption = { path = "p2panda-encryption", version = "0.5.2", default-features = false }
+p2panda-net = { path = "p2panda-net", version = "0.5.2", default-features = false }
+p2panda-spaces = { path = "p2panda-spaces", version = "0.5.2", default-features = false }
+p2panda-store = { path = "p2panda-store", version = "0.5.2", default-features = false }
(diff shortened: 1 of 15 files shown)


p2panda-net/Cargo.toml
```diff
@@ -20,83 +20,55 @@ all-features = true
...
-  "dep:p2panda-store",
-  "dep:p2panda-sync",
-]
+default = ["address_book", "iroh_endpoint", "iroh_mdns", "discovery", "gossip", "sync"]
+address_book = ["dep:p2panda-store"]
+iroh_endpoint = ["address_book", "dep:iroh", "dep:iroh-base"]
+iroh_mdns = ["iroh_endpoint", "iroh/address-lookup-mdns"]
+discovery = ["address_book", "iroh_endpoint", "dep:p2panda-discovery"]
+gossip = ["address_book", "iroh_endpoint", "dep:iroh-gossip"]
+sync = ["address_book", "iroh_endpoint", "gossip", "dep:p2panda-store", "dep:p2panda-sync"]
```
adzialocha: If possible I would prefer to not "flatten" arrays like that, as they become harder to read

sandreae: Is it just for feature blocks that you'd like to avoid flattening like this? We didn't indented arrays in other places before except features.

adzialocha: Maps we didn't format multiple so far, I'd say all arrays which are less structured make more sense in a sorted list.

map:

```toml
tokio = { version = "1.47.1", default-features = false, features = ["rt", "sync", "macros"] }
```

vs. array

```toml
default = [
  "ingest",
  "log_prune",
  "orderer",
]
```

adzialocha: In general I'd just keep the format as it was so far :-)

Cargo.toml
```diff
@@ -16,3 +16,26 @@ members = [
...
+p2panda-stream = { path = "p2panda-stream", version = "0.5.2", default-features = false }
+p2panda-sync = { path = "p2panda-sync", version = "0.5.2", default-features = false }
+rand = { version = "0.10.0", default-features = false }
+rand_chacha = { version = "0.10.0", default-features = false }
+serde = { version = "1.0.228", default-features = false }
+thiserror = { version = "2.0.17", default-features = false }
+tokio = { version = "1.48.0", default-features = false }
+tokio-stream = { version = "0.1.18", default-features = false }
+tracing = { version = "0.1.41", default-features = false }
+tracing-subscriber = { version = "0.3.20", default-features = false }
```
adzialocha: A pass over using the latest versions would be cool before merging

PR #1153, opened 2026-05-14
Add docs and README for p2panda-store

mycognosist:


p2panda-store/src/macros.rs
```diff
@@ -1,5 +1,19 @@
...
+///
+/// Internally it acquires a permit, executes the given code block and then commits the transaction
+/// before returning with the result of the code block.
+///
+/// ## Example
+///
+/// ```rust,ignore
+/// # tx!(store, {
+/// #     store.set_cursor(&new_cursor).await?;
+/// # });
```
adzialocha: Are all of these lines hidden?

sandreae: I doubt it!

adzialocha (approved):
I think this is ready for mergin! :-) 

I've added the Cargo.toml fields required for release and moved the example into `lib.rs`, there we can check the code for correctness.

PR #1155, opened 2026-05-14
Future-proof extensions format in Node API

### LINKED ISSUE #1056: Make sure extensions encoding is canonical

### LINKED ISSUE #1057: Shave off some bytes in extensions encoding

### LINKED ISSUE #1146: Future-proof extensions encoding

adzialocha:
Open Questions:

* [x] Should log id and timestamp move into the extensions header?
* [x] Do we want to extend the log id field itself in the future or introduce additional fields?

Closes: #1146 and #1057 and #1056

## 📋 Checklist

### DIFF
--- a/p2panda/src/operation.rs
+++ b/p2panda/src/operation.rs
@@ -1,10 +1,17 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use std::collections::HashSet;
 use std::hash::Hash as StdHash;
 
 use p2panda_core::hash::{HASH_LEN, Hash};
 use p2panda_core::{PruneFlag, Timestamp, Topic};
+use serde::de::{Error as SerdeError, SeqAccess, Visitor};
+use serde::ser::SerializeSeq;
 use serde::{Deserialize, Serialize};
+
+pub type Version = u16;
+
+pub type VariantCode = u16;
 
 /// Header type with our system-level extensions.
 pub type Header = p2panda_core::Header<Extensions>;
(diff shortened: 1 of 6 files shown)


p2panda/src/operation.rs
```diff
@@ -1,52 +1,605 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+// ## Node Extensions Format
```
sandreae, line 3: I opened the docs for this module and was surprised none of the extension doc strings are there. Had you intended to keep them as code comments only?

p2panda/src/operation.rs
```diff
@@ -1,52 +1,605 @@
...
+// See: <https://en.wikipedia.org/wiki/Forward_compatibility>.
+//
+// - Old systems can accept and gracefully process input intended for newer versions.
+// - Old systems can ignore new/unknown parts without breaking.
+//
+// ### Specification
+//
+// #### Encoding
+//
+// NEF MUST be encoded as CBOR while an deterministic encoding SHOULD be preferred as defined in the
```
sandreae, line 44:
```suggestion
// NEF MUST be encoded as CBOR while a deterministic encoding SHOULD be preferred as defined in the
```

p2panda/src/operation.rs
```diff
@@ -1,52 +1,605 @@
...
+//
+// #### Allow excess fields when decoding
+//
+// Decoders MUST _never_ fail on excessive sequence fields in reasonable range as this would break
+// any backwards compatibility, this roughly follows the [robustness
+// principle](https://en.wikipedia.org/wiki/Robustness_principle). Decoders MAY fail when an
+// extraordinarily large number of fields should be allocated (like more than 100 etc.).
+//
+// ```plain
+// # Imagine an extensions variant introducing a new field `age`:
```
sandreae, line 117: Slightly confusing to use the word "variant" here as in terms of the spec this is not actually a new variant (the `variant_code` remains `0`).

p2panda/src/operation.rs
```diff
@@ -1,52 +1,605 @@
...
+// #### Design fallbacks when introducing new fields
+//
+// To assure that removing extension fields doesn't break anything for older versions (backwards
+// compatibility) it is recommended to implement code to gracefully fall-back when a field is zero
+// or null.
+//
+// If a graceful fallback can't be guaranteed it is recommended to document this in the variant's
+// fields specification.
+//
+// ```rust
```
sandreae, line 137: I like the examples however they are quite "application layer" specific. We don't actually expect to put username or age into the extensions (I don't believe). Are there any examples we can think of which are expected but still easily understandable? Not a blocker, but just wanted to mention it.

p2panda/src/operation.rs
```diff
@@ -66,16 +619,201 @@ impl LogId {
...
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                formatter.write_str("32 bytes representing log id")
+            }
+
+            fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                // This is a forward-compatible decoder implementation for future formats where we
```
sandreae: I understand that this is a forward compatible implementation of a deserializer (older systems can decode log id into a known format by ignoring the path) but I believe that in terms of the resulting input (operation) being processed in an older pipeline version, this would not be forward-compatible. The log integrity validation would fail as operations which on "newer" systems are intended for different logs (based on the path being different) would actually be appended to the same log (based on only the root hash). This would result in log forking.

I'm not sure you were suggesting this is possible though, I just record this here to help make it clearer again about this different concerns when we talk about forward-compatibility.

sandreae: It does make me wonder if this forward-compatible decoder is a bit misleading though, maybe it would be better to _not_ allow access bytes on the log id and instead expect that a new extensions variant will be introduced when paths become a feature.

sandreae: Unless the log id root is based on a hash of the topic + path, then this would actually work and everything I just said can be ignored... 😄

sandreae: In any case, the path could still be added as a new field rather than having special decoding logic, it feels slightly like trying to solve the same issue (adding new things to existing extension variants) in two different places.

p2panda/src/operation.rs
adzialocha: Not sure about this one, but roughly I'd say that doc-strings are for _users_ of an API while comments are for _developers_ of an API. Since introducing new extensions and changing them is a developers aspect I thought we shouldn't confuse _users_ with these details (why would anyone using our Node API like to read the specification)?

I guess the underlying issue is that we don't have a place to put our specification 😆

p2panda/src/operation.rs
adzialocha: There's no new variant being introduced here? The example (and title) talks about adding as new _field_ to an _existing_ variant, this is why the code doesn't change.

p2panda/src/operation.rs
adzialocha: > The log integrity validation would fail as operations which on "newer" systems are intended for different logs (based on the path being different) would actually be appended to the same log (based on only the root hash). This would result in log forking.

Let me rephrase this, just to be sure I understood your point:

To compute the hash / operation id of a header A we're serializing the header into CBOR bytes using serde to then finally digest them with BLAKE3. If an older system receives a new input it would encode the header _without_ the suffix since this was unknown to serde.

The resulting operation id would differ which in-itself is not a problem yet. Now: When a new operation B was appended to the same log, which points at A via it's backlink, the hash in the backlink will not match the computed id and the operation B will be rejected?

Another issue is the verification of the signature, the bytes the signature will be checked on differs and that will make operation A be rejected.

That's indeed an issue! I wonder though if the problem lies actually more in the encoding logic of the header itself which aids log validation and not here.

> In any case, the path could still be added as a new field rather than having special decoding logic.

I think it would actually break here as well. The added, unknown field will be ignored during serialization as well and the same will happen.

We can't offer any forwards-compatibility for _added fields_ or _added data inside a field_ within the same log. I believe removals and potentially some type changes are fine?

Actually this also is true for multiple logs: If a new field was added or it's decoding changed (because of added bytes, which is the same "strategy"), we could set the rule that any operation of that new system needs be created from seq_num 0. An older system will organise this in a separate log and reject the operations inside this log for the same reason as it can't compute the backlink or signature correctly, so same stopper here.

Another option is to look into how we are serializing operations for checking their signature and computing their digest. If we only extract the signature (nullify it) to check the signature and use the raw bytes for digest then it could work again. The trick would be to not use serde / a strict type system to re-serialize bytes but work with the CBOR bytes directly.

This needs more thinking!

p2panda/src/operation.rs
adzialocha: > I guess the underlying issue is that we don't have a place to put our specification 😆

I'm fine with keeping it in the code until we have though.

p2panda/src/operation.rs
adzialocha: I believe the solution is to disconnect operation and log integrity checks from concrete extension types. Or in other words: separate integrity from interpretation. This is the correct separation probably as log integrity shouldn't be a concern of extensions.

We already do the same for the body, where we only look at the raw bytes, independent of the type used inside.

Maybe we can pull some `serde::Any` tricks for the extensions to interpret them as raw bytes as well. It would require a new method which is only used when checking integrity of already signed operations.

adzialocha: > Let me rephrase this, just to be sure I understood your point.

I think I got your real point as well now (help me if not haha). Old systems can't treat it just like a "root" log id since it wasn't meant for the "root log" if a suffix was set by the new system.

Adding a field would not help us here either. Old systems would still assign a "sub log" falsly to the "root".

I guess we would need a new extensions variant here to support log id suffixes and not allow old systems to handle them 😿

p2panda/src/operation.rs
sandreae: Yeah, I came to this conclusion too. We could put it behind a `developer` feature flag which isn't compiled to docs by default, lolz :rofl:

p2panda/src/operation.rs
sandreae: Yes, I understand the example. But the text in line 117 says "Imagine an extensions variant introducing ..." and I read this to mean a _new_ variant. I think it would be clearer on first read to say "Imagine a new field `age` being introduced to an existing variant"

p2panda/src/operation.rs
sandreae: > > Let me rephrase this, just to be sure I understood your point.
> 
> I think I got your real point as well now (help me if not haha). Old systems can't treat it just like a "root" log id since it wasn't meant for the "root log" if a suffix was set by the new system.
> 
> Adding a field would not help us here either. Old systems would still assign a "sub log" falsly to the "root".
> 
> I guess we would need a new extensions variant here to support log id suffixes and not allow old systems to handle them 😿

Yes, this was my original point. I believe my suggestion that the root log id hash is made over `topic/path` would solve that particular issue though. And _all_ system versions only use this root hash for the log id. This would retain the future-compatible properties for both decoding received bytes and using them as input on both system versions.

sandreae: > Another option is to look into how we are serializing operations for checking their signature and computing their digest. If we only extract the signature (nullify it) to check the signature and use the raw bytes for digest then it could work again. The trick would be to not use serde / a strict type system to re-serialize bytes but work with the CBOR bytes directly.

I had actually assumed that computing the digest would need to be done over the received CBOR bytes directly, as otherwise different interpretations (by different system versions) of received bytes for the same operation could result in different digests. I hadn't actually considered that this wasn't accounted for yet though, I think making sure the raw bytes are available for hashing and used during validation should be a sound approach.

p2panda/src/operation.rs
adzialocha: > developer feature flag which isn't compiled to docs by default

Actually we compile all flags when pushing to docs.rs 😄

p2panda/src/operation.rs
adzialocha: > Yes, this was my original point. I believe my suggestion that the root log id hash is made over topic/path would solve that particular issue though. And all system versions only use this root hash for the log id. This would retain the future-compatible properties for both decoding received bytes and using them as input on both system versions.

Okay cool! I'll think about it a bit to see if we can express log ids in the future like that as well.

In any case it'll not be possible to append to the same log again (log ids will differ), even with this definition. We would need to write this into the specification that that change needs to involve starting a new log.

> I had actually assumed that computing the digest would need to be done over the received CBOR bytes directly

Unfortunately not, this is how it is implemented: https://github.com/p2panda/p2panda/blob/main/p2panda-core/src/operation.rs#L287-L292

I'll raise a separate task for this in our stabilistaion document. Unfortunately just using the raw bytes is also not sufficient as we'll need to remove the signature somehow to verify it:

https://github.com/p2panda/p2panda/blob/main/p2panda-core/src/operation.rs#L279

sandreae: > In any case it'll not be possible to append to the same log again (log ids will differ), even with this definition. We would need to write this into the specification that that change needs to involve starting a new log.

I don't understand this, could you expand on what you mean?

adzialocha: > I don't understand this, could you expand on what you mean?

If the rule for assuring forwards-compatibility in an old system is:

1. If field is 32 bytes use this as log id (or hash this as well but don't see why)
2. If field is larger than 32 bytes, take all bytes, compute hash and use this as log id

Then you end up with the following behaviour:

1. Old system receives operation 0 in log A with the old input (all good): `32 bytes, for example 0x2df1 => log id = 0x2df1`
2. Old system receives operation 1 in log A with the new input (here it breaks log integrity): `44 bytes, still 0x21df1 for the first 32 bytes and then 0xabcd for the rest => digest resulting in log id = 0x5821`

Byte values are obviously all completly made up and don't represent real length 😆 

The old system considers operation 1 (at seq_num = 1 and with backlink) now as part of log `0x5821` aka a "new log" from it's perspective. Since the previous entry is from a different log the database will not return anything to verify log integrity, the operation is considered invalid.

sandreae: Ok, right, that's not what I was suggesting actually. I suggest the following:

```rust
struct LogId(Hash);

struct TopicWithPath {
    root: Topic, 
    path: Path
);

impl TopicWithPath {
    fn into_log_id(&self) -> LogId {
        LogId(self.hash_both_topic_and_path())
    }
}

```

With this there is no additional "path" field on any log id version, and so both old and new systems can decode and process it no problem. Going from TopicWithPath to LogId is a lossy conversion (both topic and path are lost) but newer systems can still get the path from an additional field appended to the extensions (which old systems will ignore).

adzialocha: This looks what I had in mind as well! I'm not sure that solves being able to upgrade to the different format within the same log. Can you maybe walk through my example with your implementation and show how it is possible to upgrade to a topic with path without changing the log id?

adzialocha: Or is your suggestion to already implement a path into the code? That wouldn't be a spec change that would just be providing the feature already 😅

sandreae: Oh, "within the same log", I didn't mean to suggest this would be supported. Originally I was discussing (and proposing a solution for) "paths can be introduced to an existing topic without breaking log integrity". Sorry, I misinterpreted your example.

sandreae: I agree that there doesn't seem to be a way to "migrate" to a path based log_id within an existing log where earlier operations used the old log id format.

sandreae: > In any case it'll not be possible to append to the same log again (log ids will differ), even with this definition. We would need to write this into the specification that that change needs to involve starting a new log.

Ok, this comment makes total sense now, I just hadn't considered supporting this behaviour :smile:

sandreae: I think still my code example differs a bit from your suggestion of how to handle the format change. In my example the root log id is made future compatible when the publisher creates it (hashes the topic+path) whereas in your example the receiver calculates the digest over all received bytes (even if they then ignore them).

mycognosist: > > In any case it'll not be possible to append to the same log again (log ids will differ), even with this definition. We would need to write this into the specification that that change needs to involve starting a new log.
> 
> Ok, this comment makes total sense now, I just hadn't considered supporting this behaviour 😄

Presumably it would be possible for a custom migration extension to specify a flag to be included in the last operation appended to the old log? In a similar way to the to prune flag, you could have a migration flag which would point to the new log and maintain some continuity (despite the changing log id).

p2panda/src/operation.rs
```diff
@@ -1,52 +1,605 @@
...
+//
+// NEF MUST be encoded as CBOR while a deterministic encoding SHOULD be preferred as defined in the
+// CBOR specification:
+// <https://www.rfc-editor.org/rfc/rfc8949.html#name-deterministically-encoded-c>. All following
+// encoding definitions can be considered as CBOR. Refer to CBOR specification for byte-level
+// definitions.
+//
+// #### Header
+//
+// Every NEF is represented as a tuple and begins with an `u16` **version** field:
```
mycognosist:
```suggestion
// Every NEF is represented as a tuple and begins with a `u16` **version** field:
```

p2panda/src/operation.rs
```diff
@@ -1,52 +1,605 @@
...
+//
+// ```plain
+// (
+//    extensions_version[u16],
+// )
+// ```
+//
+// The latest (and only) supported NEF version is `1`. Version 1 requires another field indicating
+// the type of **extensions variant** in form of an `u16` code:
```
mycognosist:
```suggestion
// the type of **extensions variant** in form of a `u16` code:
```

p2panda/src/operation.rs
```diff
@@ -1,52 +1,605 @@
...
+// // Fall back to public key if group_id is not set.
+// let group_id = group_id.unwrap_or(verifying_key);
+// ```
+//
+// #### Design less strict decoders when introducing new fields
+//
+// Consider the possibility that your introduced field will change in the future and requires
+// additional data to support more options.
+//
+// If your decoder is satisfied with the given data you MAY want to stop here and don't require
```
mycognosist:
```suggestion
// If your decoder is satisfied with the given data you MAY want to stop here and not require
```

p2panda/src/operation.rs
```diff
@@ -1,52 +1,605 @@
...
+//     timestamp[u64],
+//     previous[Vec[32]],
+// )
+// ```
+//
+// ### Introducing Changes
+//
+// #### Header
+//
+// The a) sequence format, b) CBOR encoding and c) first version field MUST never be changed for
```
mycognosist: What precisely is meant by "sequence format" here? Is this the ordering of the items in the header? Or the array-like format without nesting?

mycognosist (approved):
Really nice work on the docs. It's great to have clear definitions plus examples of what to do and what not to do.

p2panda/src/operation.rs
sandreae: What's the benefit of maintaining continuity here? I'm not sure what the implications / downside of a new log "just appearing" are.

p2panda/src/operation.rs
adzialocha: I've change this to be hopefully a bit more precise. It means that the header is an CBOR type array.

p2panda/src/operation.rs
```diff
@@ -1,47 +1,600 @@
...
+// additional data to support more options.
+//
+// If your decoder is satisfied with the given data you MAY want to stop here and not require
+// additional strict validation checks to match the _exact_ format.
+//
+// ### Known extensions variants table (draft)
+//
+// #### `0x00_00 (0)`: "Basic Extensions"
+//
+// TODO: Mention deterministic topic -> log id digest
```
mycognosist, line 157: @adzialocha Did you want to tackle these two `TODO`s as part of this PR?

p2panda/src/operation.rs
```diff
@@ -1,47 +1,600 @@
...
+        match self.variant {
+            ExtensionsVariantV1::Basic(mut extensions) => {
+                extensions.prune_flag = prune_flag.into();
+                self.variant = ExtensionsVariantV1::Basic(extensions);
+                self
+            }
+            ExtensionsVariantV1::Causal(_) => {
+                // NOTE: We're using the causal variant only as an example placeholder for now, it
+                // is not integrated or even complete yet.
+                unimplemented!()
```
mycognosist, line 438: Do we leave this unimplemented for now?

mycognosist (approved):
A couple of questions, otherwise looks good to me!

p2panda/src/operation.rs
adzialocha: This is just an example, will probably be removed soon. I could also remove it from the PR

p2panda/src/operation.rs
adzialocha: I don't know if we've settled on the first variant format yet, might be too early

sandreae (approved):
I'm really happy with this, great foundation to start building from, thanks for guiding the whole process @adzialocha 🙏

PR #1157, opened 2026-05-14
Fix header CBOR encoding with correct field count

adzialocha:
If we serialize a header (without backlink) we say the array is 9 fields long, but it should be 8! I think this regression comes from removing `previous` from the Header but not changing the field_count logic.

Here is an example of a correct header:

```
880158202f55984b871445179a6a3a94b6d7b0a6c5c8b7b014e9b55bfe069f7a661e34295840d116f0936033960c3846e485185b4564edb9380918f5dae97126d3e3d610017df612960f485163ba79dde72f709001aa3aeba32c465bf09d39421661498976000458204878ca0425c739fa427f7eda20fe845f6b2e46ba5fe2a14df5b1e32f506032150000f6
```

The first byte 8 indicates this is a CBOR array
The second byte 8 indicates there's 8 items in that array

And this is what we were (wrongly) serializing:

```
890158202f55984b87... (same as above)
```

Apparently ciborium is not so picky if the expected number of items in the array doesn't match with what's actually inside of it, this is why this error didn't surface before.

This fix corrects the field count and adds an additional assertion to check if there's any remainder fields in the header tuple. This should work around ciboriums more lax interpretation of arrays and is a good validation check in any case (prevent excessive, invalid data).

## 📋 Checklist


mycognosist (approved):
Great find! Nice to have the check for unexpected remaining fields.

PR #1164, opened 2026-05-15
Fix: revert dependency feature flags to previously working state

sandreae:
Revert feature flags for dependencies to state before 7f2c9da4e8ddff7319725b3f66b14eb2d239c269 while still using workplace level dependencies.

Run this command to view the diff between Cargo.toml files: `git diff 1e8883dc 18472def */Cargo.toml`

## 📋 Checklist

### DIFF
--- a/p2panda-net/Cargo.toml
+++ b/p2panda-net/Cargo.toml
@@ -68,24 +68,22 @@ test_utils = [
 [dependencies]
 ciborium = { workspace = true }
 futures-channel = "0.3.31"
-futures-util = { workspace = true }
-hex = { workspace = true }
-iroh = { version = "0.98.2", default-features = false, features = [
-  "tls-ring",
-], optional = true }
+futures-util = { workspace = true, default-features = true }
+hex =  { workspace = true }
+iroh = { version = "0.98.2", default-features = false, features = ["tls-ring"], optional = true }
 iroh-base = { version = "0.98.0", optional = true }
 iroh-gossip = { version = "0.98.0", optional = true }
-p2panda-core = { workspace = true }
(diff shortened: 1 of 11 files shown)


p2panda-net/Cargo.toml
```diff
@@ -29,6 +29,7 @@ default = [
   "sync",
 ]
 address_book = [
+  "iroh_endpoint",
```
adzialocha: Lets check maybe if its really required to compile the whole endpoint codebase or if a single iroh dependency is enough (I assume iroh-base is)

sandreae: I did try adding both `iroh` and `iroh-base` but still came across a dependency error.

adzialocha: Okay, that's weird, let's look into it

adzialocha: I assume it's because the conversion methods for verifying key are behind a test flag but should also include iroh_endpoint feature flag

p2panda-stream/Cargo.toml
```diff
@@ -27,20 +27,20 @@ default = [
 ]
 ingest = [
   "dep:p2panda-core",
+  "dep:p2panda-store",
+  "dep:thiserror",
 ]
 log_prune = [
-  "p2panda-core/prune",
+  "ingest",
```
adzialocha: They dont need that feature to compile, or is something off here as well?

sandreae: I think something is off here as well, I get errors without having all features required for ingest in both `log_prune` and `orderer`

adzialocha (approved):

Issue #1133, opened 2026-04-24
Re-use traits from `p2panda-core` in `p2panda-auth`


sandreae:
`IdentityHandle` -> `Author` (from core)
`OperationId` -> `OperationId` (from core)

Issue #1065, opened 2026-03-09
Consider all store methods to use explicit SQL transactions


adzialocha:
Every query in SQLite is automatically wrapped inside a transaction (BEGIN & COMMIT), that's [Auto-Commit mode](https://sqlite.org/lang_transaction.html). We currently distinct between queries using `execute` of sqlx which runs it with auto-committing and queries using `tx` where we need to design the transaction ourselves.

I would propose only offering the latter and asking users of the API to handle transactions themselves with the cost of making use of these methods more cumbersome, but for a good reason. This would allow us to not distinct between these two different modes anymore (`get_latest_entry` vs. `get_latest_entry_tx`) / less code, the performance is the same.

adzialocha:
Related: https://github.com/p2panda/p2panda/pull/1060

PR #1174, opened 2026-05-20
Refactor the groups processor to use the `Processor` trait

mycognosist:
In this PR we remove the groups processor implementation in `p2panda-auth`
and replace it with an implementation using the `Processor` trait in
`p2panda-stream`.

Since the groups processor is designed to be used in a chain of
processors, including `Ingest` and `Orderer`, we no longer need to handle
operation ingestion and ordering in the `process` method.

## 📋 Checklist


p2panda-stream/src/groups/processor.rs
```diff
@@ -0,0 +1,729 @@
...
+                group_id = groups_operation.group_id().to_hex(),
+                "current group membership: {:?}",
+                y.members(groups_operation.group_id())
+            );
+
+            debug!(
+                id = operation_id.to_hex(),
+                "apply ready operation to groups state"
+            );
+            let groups_operation =
```
sandreae: Getting the groups operation from the store isn't required anymore as it is constructed earlier from the operation being processed. The reason it was in the previous groups processor version is because when the orderer released operations only their id was provided, so a database call was then required.

mycognosist: Thanks! I've removed this call and the related function.

p2panda-stream/src/groups/processor.rs
```diff
@@ -0,0 +1,684 @@
...
+            debug!(
+                group_id = groups_operation.group_id().to_hex(),
+                "current group membership: {:?}",
+                y.members(groups_operation.group_id())
+            );
+
+            debug!(
+                id = operation_id.to_hex(),
+                "apply ready operation to groups state"
+            );
```
adzialocha, line 129: There's another log in line 129 which seems to do the same

p2panda-stream/src/groups/processor.rs
```diff
@@ -0,0 +1,684 @@
...
+use std::fmt::Debug;
+use std::marker::PhantomData;
+
+use p2panda_auth::group;
+use p2panda_auth::traits::{Conditions, Operation as GroupsOperationTrait};
+use p2panda_core::{Extension, Extensions, Hash, LogId, Operation, VerifyingKey};
+use p2panda_store::groups::GroupsStore;
+use p2panda_store::{SqliteError, SqliteStore, Transaction};
+#[cfg(any(test, feature = "groups"))]
+use serde::{Deserialize, Serialize};
```
adzialocha: Could probably remove the compilation flag here as we're already in the groups mod and we need serde for non-test cases as well

p2panda-stream/src/groups/processor.rs
```diff
@@ -0,0 +1,684 @@
...
+            Self::Noop => false,
+        }
+    }
+}
+
+/// Processor for groups operations.
+pub struct GroupsProcessor<SID, T, E, L, C = ()> {
+    store: SqliteStore,
+    notify: Notify,
+    // TODO: Do we want to keep a set for all processed items?
```
adzialocha: What would that be used for?

p2panda-stream/Cargo.toml
```diff
@@ -21,25 +21,35 @@ workspace = true
...
 [features]
 default = [
+  "groups",
   "ingest",
   "log_prune",
   "orderer",
 ]
+groups = [
+  "p2panda-auth/serde",
+  "dep:tracing",
```
adzialocha: I'd be fine with making `tracing` a general dependency for the crate, so we don't need to gate it behind this feature

p2panda-auth/Cargo.toml
```diff
@@ -57,6 +47,7 @@ p2panda-core = { workspace = true, default-features = true, features = ["test_ut
 p2panda-net = { workspace = true, default-features = true }
 p2panda-store = { workspace = true, default-features = true, features = ["test_utils"] }
 p2panda-sync = { workspace = true, default-features = true }
+p2panda-stream = { workspace = true, features = ["ingest", "groups"] }
```
adzialocha, line 50: -stream is a dev-dependency, right?

adzialocha: Maybe -store as well and -core? Not sure

adzialocha (commented):
Just a few comments from first glance. It looks already really good I think.

p2panda-stream/src/groups/processor.rs
mycognosist: Can't remember why I was thinking about this. Will remove the comment now.

p2panda-auth/Cargo.toml
mycognosist: `-core` is the only one which should appear in both. I'll remove `-store` from `dependencies`.

p2panda-auth/examples/groups.rs
```diff
@@ -43,23 +42,32 @@ use p2panda_net::utils::ShortFormat;
...
+    StateId,
+    GroupsProcessorArgs<StateId, AppExtensions>,
+    AppExtensions,
+    LogId,
+>;
 
 /// This application maintains only one log per author, this is why we can hard-code it.
 const LOG_ID: LogId = 1;
 
 /// Identifier for the singleton group state used in this example.
```
adzialocha: Unrelated to the PR, just realized this during review.

https://en.wikipedia.org/wiki/Singleton_pattern is maybe a bit misleading. Maybe we can just call it "global" or remove the word altogether?

p2panda-auth/examples/groups.rs
```diff
@@ -163,57 +183,100 @@ async fn main() -> Result<()> {
...
+                        }
+                    };
+
+                    let y: GroupsState = tx_unwrap!(store, {
+                        store
+                            .get_groups_state(&GROUPS_STATE_ID)
+                            .await
+                            .unwrap()
+                            .unwrap_or_default()
+                    });
```
adzialocha, line 211: Interesting, I didn't realize that these "read" queries are actually using explicit transactions. So far we've named all of these methods with a `_tx` at the end. Let's add this for now for consistency, later we can remove that naming scheme when tackling this issue: https://github.com/p2panda/p2panda/issues/1065

adzialocha: Sorry, this is also unrelated to this PR, but maybe nice to clean up as part of this work piece (can be a separate commit)

p2panda-auth/examples/groups.rs
```diff
@@ -163,57 +183,100 @@ async fn main() -> Result<()> {
...
         });
+    }
 
-        rt.block_on(local);
-    });
+    // Contruct the ingest processor.
+    let ingest = Ingest::new(store.clone());
+
+    // Construct groups processor.
+    let groups = GroupsProcessor::new(store.clone());
```
adzialocha: This is not using the orderer here, do we want that as well? Otherwise operations will need to be strictly ordered.

p2panda-auth/examples/groups.rs
```diff
@@ -163,57 +183,100 @@ async fn main() -> Result<()> {
...
+        if let Err(err) = ingest
+            .process(IngestEvent {
+                operation: op.clone(),
+                args: IngestArgs {
+                    log_id: LOG_ID,
+                    topic,
+                    prune_flag: false,
+                },
+            })
+            .await
```
adzialocha, line 256: Makes sense to use the processor directly without the pipeline API here, however this can lead to undesirable behaviour as soon as out-of-order messages / buffering is involved.

Here an example:

```
We have a sequence of incoming inputs, like that:
A <- B <- C

To be processed: [A, B, C]

=== Processor Step 1 ===
Input: A
Ingest: Process A, Yield A
Orderer: Process A, Yield A
Groups: Process A, Yield A

=== Processor Step n ===
..
```

Here everything is fine, the while loop takes the next item from the sequence, every processor input is symmetric with the outputs. Nothing different from a standard Rust `Stream`.

Now, if we introduce buffering things will break:

```
Same input sequence but arriving in wrong order:
A <- B <- C, 

To be processed: [A, C, B]

=== Processor Step 1 ===
Input: A
Ingest: Process A, Yield A
Orderer: Process A, Yield A
Groups: Process A, Yield A

=== Processor Step 2 ===
Input: C
Ingest: Process C, Yield C
Orderer: Process C, Yield %

=== Processor Step 3 ===
Input: B
Ingest: Process B, Yield B
Orderer: Process B, Yield B
Groups: Process B, Yield B

The loop terminates, since all input items have been "processed" now. Item C will never be processed by the groups processor.
```

We want to await / poll `process` and `next` independently of any input stream. This is what the `Pipeline` is providing.

p2panda-auth/src/extension.rs
```diff
@@ -0,0 +1,17 @@
...
+use crate::group::GroupAction;
+
+/// Additional arguments which can be attached to a p2panda operation in their extensions.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(any(test, feature = "serde"), derive(Deserialize, Serialize))]
+pub struct GroupsExtensionArgs<C = ()> {
+    pub group_id: VerifyingKey,
+    pub action: GroupAction<VerifyingKey, C>,
+    pub dependencies: Vec<Hash>,
+}
```
adzialocha, line 17: Let's already consider more efficient serialization formats, think it's good practice to not make this an afterthought to prevent unnecessary breaking changes.

p2panda-auth/src/lib.rs
```diff
@@ -75,12 +75,20 @@
...
+mod operation;
 #[cfg(any(test, feature = "test_utils"))]
 pub mod test_utils;
 pub mod traits;
 
 pub use access::{Access, AccessError, AccessLevel};
+pub use extension::GroupsExtensionArgs;
+pub use operation::GroupsOperation;
+
+use crate::traits::{IdentityHandle, OperationId};
```
adzialocha: We can use the existing traits in `p2panda-core` instead

p2panda-auth/src/lib.rs
```diff
@@ -75,12 +75,20 @@
...
+use p2panda_core::{Hash, VerifyingKey};
```
adzialocha, line 90: Nitpick: Import grouping

p2panda-auth/src/lib.rs
adzialocha: They're already implemented for `Hash` and `VerifyingKey`

p2panda-stream/src/groups/args.rs
adzialocha: Oh yeah, I agree, that's a bit confusing. What's the difference to `GroupsExtensionArgs` defined in `extension.rs`?

p2panda-stream/src/groups/processor.rs
```diff
@@ -0,0 +1,808 @@
...
+pub struct GroupsProcessor<SID, T, E, L, C = ()> {
+    store: SqliteStore,
+    notify: Notify,
+    queue: RefCell<VecDeque<(T, GroupsProcessorResult)>>,
+    _marker: PhantomData<(SID, E, L, C)>,
+}
+
+impl<SID, T, E, L, C> GroupsProcessor<SID, T, E, L, C>
+where
+    E: Extensions + Extension<GroupsArgs<C>> + Extension<L>,
```
adzialocha: We've stopped using the `Extension` trait in `p2panda-stream` and `p2panda` as `Borrow` is doing the same job, it's not used anywhere right now, so probably something we want to phase-out soon?

adzialocha: So instead of requesting a specific `Extension` you put a bound on the concrete arguments you want to be able to borrow from `T` (see ingest / orderer, etc.)

adzialocha: What's nice with this approach is that it doesn't imply that it needs to come from an "extension", some of this information can be gathered from _outside_ the operation / extensions.

p2panda-stream/src/groups/processor.rs
```diff
@@ -0,0 +1,808 @@
...
+                    topic,
+                    prune_flag: false,
+                },
+            })
+            .await
+            .unwrap();
+
+        // TODO(glyph): Naming disparity between `Ingest` and `GroupsProcessor`. Maybe we could
+        // rename `Ingest` to `IngestProcessor`? Then `Orderer` would become `OrderProcessor` or
+        // `DependencyProcessor` or something similar.
```
adzialocha: So far we have the processors: Ingest, Orderer and LogPrune, then I'd stick to this naming scheme and call "GroupsProcessor" just "Groups"

adzialocha (changes requested):

p2panda-stream/src/groups/processor.rs
```diff
@@ -0,0 +1,808 @@
...
+            })
+            .await
+            .unwrap();
+
+        // Operations are processed by the orderer in reverse order.
+        orderer.process(op_02).await.unwrap();
+        orderer.process(op_01).await.unwrap();
+        orderer.process(op_00).await.unwrap();
+
+        // Each operation freed from the orderer is processed by the groups processor.
```
sandreae, line 442: Thanks for adding this test! What I wanted to have checked was that the processor chain / pipeline buffers out-of-order messages and then frees them for the groups processor to handle when they're ready. From @adzialocha comment it looks like we should use the `Pipeline` from p2panda for this both in tests and _the real word_.

sandreae: Which would mean that ingest / ordering / processing would all be handled in a single pipeline per test user, and this is where they would process each operation (some of which are out-of-order with missing dependencies).

p2panda-stream/src/groups/processor.rs
sandreae: I also like the `Borrow` approach 👍

p2panda-auth/src/lib.rs
sandreae: Think that's out of scope for this PR, there's an issue here https://github.com/p2panda/p2panda/issues/1133

p2panda-stream/src/groups/args.rs
sandreae: `GroupsExtensionArgs` isn't required anymore. We can define the arguments that the processor expects how we like in this refactor.

p2panda-auth/src/operation.rs
```diff
@@ -0,0 +1,45 @@
...
+#[cfg(any(test, feature = "serde"))]
+use serde::{Deserialize, Serialize};
+
+use crate::group::GroupAction;
+use crate::traits::{Conditions, Operation};
+
+/// Concrete groups operation type.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(any(test, feature = "serde"), derive(Deserialize, Serialize))]
+pub struct GroupsOperation<C = ()> {
```
sandreae, line 13: Now that the processor has moved to -stream this type isn't used in p2panda-auth anymore. I suggest we move it over there. It's possible that we want to want to rethink how the p2panda_auth::traits::Operation trait is satisfied with the different patterns introduced in -stream in any case (I'm not sure about that, just saying as it's a possibility. We don't strictly _need_ a `GroupsOperation` as it is defined here, only that the trait is implemented on a type) .

p2panda-auth/src/extension.rs
sandreae: Mentioned elsewhere but this type isn't required any more and can be removed.

p2panda-auth/src/operation.rs
mycognosist: I've opened an issue for this:

https://github.com/p2panda/p2panda/issues/1188

p2panda-stream/src/groups/processor.rs
mycognosist: Opened an issue for this: https://github.com/p2panda/p2panda/issues/1189

p2panda-stream/src/groups/processor.rs
mycognosist: Opened an issue for this: https://github.com/p2panda/p2panda/issues/1190

adzialocha (approved):

PR #1193, opened 2026-05-28
Replace `IdentityHandle` and `OperationId` with traits from `p2panda-core`

mycognosist:
Addresses: https://github.com/p2panda/p2panda/issues/1133

## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/group/member.rs
+++ b/p2panda-auth/src/group/member.rs
@@ -1,9 +1,8 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use p2panda_core::identity::Author;
 #[cfg(any(test, feature = "serde"))]
 use serde::{Deserialize, Serialize};
-
-use crate::traits::IdentityHandle;
 
 /// A group member which can be a single individual or another group.
 ///
@@ -42,4 +41,4 @@ where
     }
 }
 
-impl<ID> IdentityHandle for GroupMember<ID> where ID: IdentityHandle {}
+impl<ID> Author for GroupMember<ID> where ID: Author {}
--- a/p2panda-auth/src/group/member.rs
+++ b/p2panda-auth/src/group/member.rs
@@ -3,8 +3,6 @@
 #[cfg(any(test, feature = "serde"))]
 use serde::{Deserialize, Serialize};
 
-use crate::traits::IdentityHandle;
-
 /// A group member which can be a single individual or another group.
 ///
 /// The `Group` variant can be used to express nested group relations. In both cases, the member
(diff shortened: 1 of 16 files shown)


adzialocha:
What is left to get this merged @mycognosist ?

mycognosist:
@adzialocha Should be ready for review now. Thanks for your patience! ❤️

p2panda-core/src/identity.rs
```diff
@@ -41,10 +42,12 @@ pub const SIGNING_KEY_LEN: usize = ed25519_dalek::SECRET_KEY_LENGTH;
 pub const VERIFYING_KEY_LEN: usize = ed25519_dalek::PUBLIC_KEY_LENGTH;
 
 pub trait Author:
-    Clone + PartialEq + Ord + StdHash + Serialize + for<'de> Deserialize<'de>
+    Copy + Debug + Clone + PartialEq + Ord + StdHash + Serialize + for<'de> Deserialize<'de>
 {
 }
 
+impl Author for char {}
```
adzialocha, line 50: I think we only want to implement this for test environments, otherwise we would allow passing in these "unsafe" types in our official APIs

p2panda-core/src/traits.rs
```diff
@@ -1,11 +1,15 @@
...
 use std::hash::Hash as StdHash;
 
 use crate::identity::Author;
 
 /// Identifier of a single operation.
-pub trait OperationId: Copy + Clone + PartialEq + Eq + StdHash {}
+pub trait OperationId: Copy + Clone + Debug + PartialEq + Eq + Ord + StdHash {}
+
+impl OperationId for u32 {}
+impl OperationId for &str {}
```
adzialocha, line 14: Same here

p2panda-auth/src/group/member.rs
```diff
@@ -42,4 +42,5 @@ where
     }
 }
 
-impl<ID> IdentityHandle for GroupMember<ID> where ID: IdentityHandle {}
+#[cfg(any(test, feature = "serde"))]
+impl<ID> Author for GroupMember<ID> where ID: Author {}
```
adzialocha: Gating it behind the `serde` flag doesn't seem right, behind a test environment flag seems okay.

Before it wasn't gated at all, so I'm surprised it doesn't break anything. Where is this impl used?

adzialocha (changes requested):

p2panda-auth/src/group/member.rs
mycognosist: In the end this wasn't required (as far as I can tell) so I've removed it completely.

adzialocha (approved):

PR #1195, opened 2026-05-28
Remove timestamp from operation header

### LINKED ISSUE #1115: Consider removing timestamp from Header
It's not used in any low-level protocol

adzialocha:
Closes: #1115 

## 📋 Checklist


sandreae:
Just had a quick look at the failing test and the problem is that the timestamp in the header was providing uniqueness across operations that have the same body, even though the log ids are different. This is because we're just using blank `()` extensions in this test. If we put the log id into the extensions it'll pass again.

adzialocha:
> Just had a quick look at the failing test and the problem is that the timestamp in the header was providing uniqueness across operations that have the same body, even though the log ids are different. This is because we're just using blank `()` extensions in this test. If we put the log id into the extensions it'll pass again.

Thank you for the pointer!

I realized that the tests do ignore the log id when creating operations. I've changed the approach to actually organise operations in different logs and thus we don't need timestamps for uniqueness anymore.

sandreae (approved):

mycognosist (approved):

PR #1196, opened 2026-05-31
Skip encoding ZST extensions

### LINKED ISSUE #887: Use `mem::conjure_zst` to avoid `null` element when decoding `Header` without extension
When serializing `Header` currently without an `Extension` (rather an edge case), we're adding an extra "null" byte at the end for the extension which is redundant.

To avoid this we can wait for stable support of `mem::conjure_zst` which seems to be planned in Rust `1.92.0` 🥳 

Related discussion here: https://github.com/p2panda/p2panda/pull/811

adzialocha:
This is a proposal to get the best of both worlds: Ergonomic API around extensions as-is with unsafe code to ensure correctly encoding the header when using Zero-Sized Types.

Other safe-code approaches are:

1. Go back to `extensions: Option<E>`, cumbersome to use
2. Introduce `E: Default` bound, unsure how that works since extensions require deterministic topic -> log id mapping sometimes etc.
3. Change specification and make extensions non-optional, weird to make spec change because of Rust limitations

Closes: #887 

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Highlights are marked with a pancake 🥞
     - `payload_size` and `seq_num` from `u64` to `u32` [#1194](https://github.com/p2panda/p2panda/pull/1194)
     - `version` from `u64` to `u16` [#1194](https://github.com/p2panda/p2panda/pull/1194)
     - Remove `timestamp` [#1195](https://github.com/p2panda/p2panda/pull/1195)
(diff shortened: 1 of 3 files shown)


sandreae (approved):
Thanks for all your research into the possible options we had available. All things considered this is my preferred option and the implementation and docs are really clear about what's happening under the hood.

sandreae (approved):
Thanks for all your research into the possible options we had available. All things considered this is my preferred option and the implementation and docs are really clear about what's happening under the hood.

mycognosist (approved):

Issue #1018, opened 2026-02-19
Use `AnyOperation` type for use in places where `E` extensions type is not needed/known


sandreae:
This allows users to handle operations where the extensions are not concretely know without having to introduce an `E` generic everywhere. It could be achieved by holding the encoded extension bytes and decoding them only when needed.

adzialocha:
Things have improved now and we have this `AnyOperation` type in `p2panda-core`: https://github.com/p2panda/p2panda/pull/1200

The next step would be to:

- Remove `E` from all `p2panda-store` methods, we work with AnyHeader and AnyOperation now
- Remove `E` from all `p2panda-sync` code
- Decode into `Header::<E>` in `p2panda`

PR #1200, opened 2026-06-04
New forwards-compatible, safer and faster API to create, encode and decode operations

### LINKED ISSUE #1201: Detach serde from extensions encoding to assure forwards-compatibility
We want to not use serde to encode the header bytes with extensions inside as this can break forwards-compatibility. When encoding a header using serde, we use the concrete Rust types to determine the shape of the bytes. If that Rust type has changed in a future version, an invalid encoding will result and break the hash digest & signature as the header shape has changed now. We need to preserve the shape of the header extensions, even if future systems might interpret them differently.

We want to only parse a more abstract CBOR AST and not rely on concrete Rust types. The AST helps us to remove the signature to encode the bytes properly again to verify the integrity of an operation without changing it during the encoding process.

This is not a critical bug _right now_ but will be important to fix as soon as systems rely on p2panda in production and want to introduce forwards-compatible changes in their regarding extensions format.

Since we can't control how and what extensions will take shape in the future this needs to be fixed on `-core` level.

To decribe the issue better, here's an example. Let's say we have this extensions format:

```
(1, 0, timestamp)
```

.. and later, in 2 years we decide to extend this with another field:

```
(1, 0, timestamp, prune_flag)
```

That field is optional, if you set the flag you prune, if you don't have it you can't prune, but it doesn't break stuff.

An old system didn't get that extensions update yet, but it'll still break. It will use `serde` which derives the shape of the data from Rust types to decode the bytes and end up with this shape:

```
(1, 0, timestamp)
```

That's because the `Deserializer` is built to ignore superflous / unknown fields at the end of the sequence.

Now that node wants to verify the signature, it runs `Header::to_bytes` to get the right bytes which will now not include prune_flag (because that system doesn't know about that and uses Rust to get the concrete types for E via serde).

.. this will lead to different bytes than from the moment when that operation was forged by the new system (which serialized prune_flag) and thus lead to an invalid signature, the operation will be rejected, even though introducing the prune_flag field would be totally forwards-compatible. This is why we want to ignore the concrete type of E when checking the signature, same for the hashing digest.

adzialocha:
### Background

`Header::<E>` has been slightly too simple for what we need. It was used to create operations, validate them, encode and decode them all at once. That was really easy to use, but also ignored some important implementation details as we revealed now and in hindsight also makes dealing with operations actually harder .. and slower.

The main issue is: We can't rely only on `serde` to encode headers as this is one abstraction level too high (Rust type level), see related issue here: https://github.com/p2panda/p2panda/issues/1201. We want to be on CBOR level instead (with the help of an AST representation).

### New type to decode operations: `AnyHeader` and `AnyOperation`

This PR introduces an `AnyHeader` type. It doesn't know about an explicit extensions type `E` anymore (less generics, yay) and is designed to decode headers as efficiently as possible using `cbor_core`. It keeps the extension's CBOR AST around to always correctly encode it back into it's original shape while ensuring correct hash digest, size and signature check.

There's no serde support for encoding and decoding headers anymore, instead one needs to call `AnyHeader::decode(bytes)` or `AnyHeader::encode()` etc.

We can do already a lot of validation here and I focused on optimizing for reducing redundant, repetitive work and memory usage. Now we're computing the digest and check the header format AND the signature in one go with as little memory allocations (zero copy) and encodings as possible.

The hash digest and header size is stored inside AnyHeader so any call on it like `header.hash()` etc. returns an already computed result, no need to encode anything again.

Things like this were not possible before since `Header::<E>` didn't have any proper constructors, we always assembled it from using the fields directly, so that's a nice side-effect from this refactor.

Obviously there's an `AnyOperation` type as well now, converting between all of them is really easy and almost doesn't cost much.

For converting between `AnyHeader` -> `Header::<E>` or `AnyOperation` -> `Operation::<E>` we also have a lot in our hands already, the computed hash, mostly all fields of the header. The only thing left is to decode the extensions, now that we know the concrete Rust type `E`.

There's not really a need anymore for `Header::<E>` to check the signature or compute a hash or size, this type is now mostly representing an immutable version of a header. `Signature` is not an `Option` anymore and the pre-computed hash and size lives inside of it as well.

To decode bytes to headers the flow looks like this now:

```
Vec<u8> -> AnyHeader::decode(bytes)? -> Header::<E>::from_any(any_header)?

.. or

Vec<u8> -> Header::<E>::decode(bytes)?
```

### New traits to describe data-type guarantees

To check the log integrity we need to update our `validate_operation`, `validate_backlink` and `validate_header` methods to work with these new types.

Instead of asking for a concrete type here this PR introduces trait bounds. Here are the traits (naming etc. heavily WIP):

```rust
/// Identifier of a single operation.
pub trait OperationId: Copy + Clone + PartialEq + Eq + StdHash {}

/// Returns (unique) hash digest, which can be used as identifier of this published data type.
pub trait Digest<ID>
where
    ID: OperationId,
{
    fn hash(&self) -> ID;
}

/// Returns the author of this published data type and a method to verify the authenticity of it.
pub trait Provenance<A>
where
    A: Author,
{
    fn author(&self) -> A;

    fn verify(&self) -> bool;
}

/// Hash-chain structure with integrity guarantees and sequence numbers as a performance
/// optimization.
pub trait Chain<ID> {
    /// Pointer at previous entry in log which gives us the integrity guarantee of the "hash chain".
    /// The first entry in a log returns `None`.
    fn backlink(&self) -> Option<ID>;

    /// Sequence numbers are helpful to fastly detect forks and use the much faster and optimized
    /// diffing strategy when the local log is not forked.
    fn seq_num(&self) -> SeqNum;
}

/// Additional data which can be removed from the on-chain data-type.
pub trait Offchain<ID> {
    fn payload(&self) -> Option<&Body>;

    fn payload_hash(&self) -> Option<ID>;

    fn payload_size(&self) -> PayloadSize;
}
```

### New builder type to create & sign headers

Lastly we need a way now to create operations. Since we can't use `Header<E>` for that anymore we need .. a new builder struct!

Here is an example:

```rust
let header = Header::builder()
    .body(b"test")
    .build(&signing_key, extensions);
```

.. or:

```rust
let header = Header::builder()
    .chain(1, previous_header.hash())
    .build(&signing_key, extensions);
```

The encoding of everything doesn't use serde internally but `cbor_core` directly to skip the `Signature: None` workaround and also because we don't really have any serializer implemented for such a struct anymore.

Again it's optimized for encoding only twice (once for signature calculation and once for digest) and these things, plus header size are then moved straight into `Header<E>`, so we don't need to compute it ever again.

This gives us the path to publish operations as such:

```
Header::<E>::builder() -> Header::<E>::encode() -> Vec<u8>
```

Since `Header<E>` can only be created ever by ourselves (and never came from bytes) we know that the signature and operation format is always checked and correct.

### Use of `cbor_core` instead of `ciborium`

This PR replaced `ciborium` with [`cbor_core`](https://docs.rs/cbor-core/latest/cbor_core/). The reason is that `ciborium` doesn't have a Value and AST as powerful as `cbor_core` to work with for low-level messing with CBOR types. It only offers a fairly abstract serde API.

Closes https://github.com/p2panda/p2panda/issues/1201

## Todo

- [x] Rework flow to create, encode and decode operations
- [x] Allow low-level (dangerous!) access to manipulating headers in test environments

## Further work

- Remove `E` from all `p2panda-store` methods, we work with AnyHeader and AnyOperation now
- Remove `E` from all `p2panda-sync` code
- Decode into `Header::<E>` in `p2panda`

## 📋 Checklist

### DIFF
--- a/p2panda-core/examples/basic.rs
+++ b/p2panda-core/examples/basic.rs
@@ -10,22 +10,10 @@ fn main() {
     let signing_key = SigningKey::generate();
 
     // An operation body contains application data.
-    let body = Body::new("Hello, Sloth!".as_bytes());
+    let body = Body::from_bytes("Hello, Sloth!".as_bytes());
 
-    // Create a header.
-    let mut header = Header {
-        version: 1,
-        verifying_key: signing_key.verifying_key(),
-        signature: None,
-        payload_size: body.size(),
-        payload_hash: Some(body.hash()),
-        seq_num: 0,
-        backlink: None,
-        extensions: None::<()>,
-    };
-
-    // Sign the header.
-    header.sign(&signing_key);
+    // Create and sign a header.
+    let header = Header::builder().body(&body).build(&signing_key, ());
 
     // An operation containing the header hash (the operation id), the header itself and an optional body.
     let operation = Operation {
--- a/p2panda-core/src/extensions.rs
+++ b/p2panda-core/src/extensions.rs
@@ -61,20 +61,11 @@
 //! };
 //!
 //! let signing_key = SigningKey::generate();
-//! let body: Body = Body::new("Hello, Sloth!".as_bytes());
+//! let body: Body = Body::from_bytes("Hello, Sloth!".as_bytes());
 //!
-//! let mut header = Header {
-//!     version: 1,
-//!     verifying_key: signing_key.verifying_key(),
-//!     signature: None,
-//!     payload_size: body.size(),
-//!     payload_hash: Some(body.hash()),
-//!     seq_num: 0,
-//!     backlink: None,
-//!     extensions: extensions.clone(),
-//! };
-//!
-//! header.sign(&signing_key);
+//! let header = Header::builder()
+//!     .body("Hello, Sloth".as_bytes())
+//!     .build(&signing_key, extensions.clone());
 //!
 //! let log_id: LogId = header.extension().unwrap();
 //! let expiry: Expiry = header.extension().unwrap();
--- a/p2panda-core/src/extensions.rs
+++ b/p2panda-core/src/extensions.rs
@@ -61,20 +61,10 @@
 //! };
 //!
 //! let signing_key = SigningKey::generate();
-//! let body: Body = Body::new("Hello, Sloth!".as_bytes());
 //!
-//! let mut header = Header {
-//!     version: 1,
-//!     verifying_key: signing_key.verifying_key(),
-//!     signature: None,
-//!     payload_size: body.size(),
-//!     payload_hash: Some(body.hash()),
-//!     seq_num: 0,
-//!     backlink: None,
-//!     extensions: extensions.clone(),
-//! };
-//!
-//! header.sign(&signing_key);
+//! let header = Header::builder()
+//!     .body("Hello, Sloth".as_bytes())
+//!     .build(&signing_key, extensions.clone());
 //!
 //! let log_id: LogId = header.extension().unwrap();
 //! let expiry: Expiry = header.extension().unwrap();
--- a/p2panda-core/src/operation.rs
+++ b/p2panda-core/src/operation.rs
@@ -28,23 +28,13 @@
 //! ### Construct and sign a header
 //!
 //! ```
-//! use p2panda_core::{Body, Header, SigningKey};
+//! use p2panda_core::{Header, SigningKey};
 //!
 //! let signing_key = SigningKey::generate();
 //!
-//! let body = Body::new("Hello, Sloth!".as_bytes());
-//! let mut header = Header {
-//!     version: 1,
-//!     verifying_key: signing_key.verifying_key(),
-//!     signature: None,
-//!     payload_size: body.size(),
-//!     payload_hash: Some(body.hash()),
-//!     seq_num: 0,
-//!     backlink: None,
-//!     extensions: (),
-//! };
-//!
-//! header.sign(&signing_key);
+//! let header = Header::builder()
+//!     .body(b"Hello, Icebear!")
+//!     .build(&signing_key, ());
 //! ```
 //!
 //! ### Custom extensions
@@ -70,33 +60,26 @@
 //!     prune_flag: PruneFlag::new(true),
 //! };
 //!
-//! let body = Body::new("Prune from here please!".as_bytes());
-//! let mut header = Header {
-//!     version: 1,
-//!     verifying_key: signing_key.verifying_key(),
-//!     signature: None,
-//!     payload_size: body.size(),
-//!     payload_hash: Some(body.hash()),
-//!     seq_num: 0,
-//!     backlink: None,
-//!     extensions,
-//! };
-//!
-//! header.sign(&signing_key);
+//! let body = Body::from_bytes("Prune from here please!".as_bytes());
+//! let header = Header::builder()
+//!     .body(&body)
+//!     .build(&signing_key, extensions);
 //!
 //! let prune_flag: PruneFlag = header.extension().unwrap();
 //! assert!(prune_flag.is_set())
 //! ```
 use std::borrow::Borrow;
+use std::marker::PhantomData;
 
+use cbor_core::Value;
 use thiserror::Error;
 
-use crate::cbor::{DecodeError, decode_cbor, encode_cbor};
-use crate::extensions::{Extension, Extensions};
-use crate::hash::Hash;
-use crate::identity::{Signature, SigningKey, VerifyingKey};
+use crate::Extension;
+use crate::extensions::Extensions;
+use crate::hash::{HASH_LEN, Hash};
+use crate::identity::{SIGNATURE_LEN, Signature, SigningKey, VERIFYING_KEY_LEN, VerifyingKey};
 use crate::logs::SeqNum;
-use crate::traits::{Digest, Provenance};
+use crate::traits::{Chain, Digest, Offchain, Provenance};
 
 /// Encoded bytes of an operation header and optional body.
 pub type RawOperation = (Vec<u8>, Option<Vec<u8>>);
@@ -107,21 +90,6 @@ pub struct Operation<E = ()> {
     pub hash: Hash,
     pub header: Header<E>,
     pub body: Option<Body>,
-}
-
-impl<E> Operation<E>
-where
-    E: Extensions,
-{
-    /// Get a reference to the operation header.
-    pub fn header(&self) -> &Header<E> {
-        &self.header
-    }
-
-    /// Get a reference to the operation body.
-    pub fn body(&self) -> Option<&Body> {
-        self.body.as_ref()
-    }
 }
 
 impl<E> PartialEq for Operation<E> {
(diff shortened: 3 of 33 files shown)


p2panda-core/examples/basic.rs
```diff
@@ -10,22 +10,10 @@ fn main() {
...
-        payload_hash: Some(body.hash()),
-        seq_num: 0,
-        backlink: None,
-        extensions: None::<()>,
-    };
-
-    // Sign the header.
-    header.sign(&signing_key);
+    // Create and sign a header.
+    let header = Header::builder().body(&body).build(&signing_key, ());
```
sandreae, line 16: Nice and compact with the new builder API

p2panda-core/src/operation.rs
```diff
@@ -170,8 +150,234 @@ where
...
+
+    pub fn seq_num(mut self, seq_num: SeqNum) -> Self {
+        self.seq_num = seq_num;
+        self
+    }
+
+    pub fn backlink(mut self, backlink: Option<Hash>) -> Self {
+        self.backlink = backlink;
+        self
+    }
```
sandreae, line 324: When does the user need these separate setters for seq_num and backlink instead of using the (safer) chain method?

p2panda-core/src/operation.rs
```diff
@@ -170,8 +150,234 @@ where
...
+        let version = 1;
+
+        let verifying_key = signer.verifying_key();
+
+        let extensions_cbor = if Header::<E>::has_non_zero_sized_extensions() {
+            let extensions_cbor = Value::serialized(&extensions).expect("serializable extensions");
+            Some(extensions_cbor)
+        } else {
+            None
+        };
```
sandreae: Could extensions also be set via a method on the builder rather than being passed in to the `build(..)` method itself? I can see how this creates type issues when E is not specifically set by the user, but maybe there is a way around this by defining a default for E on Builder.

p2panda-core/src/operation.rs
```diff
@@ -262,46 +506,55 @@ impl<E> Header<E>
...
 
     /// BLAKE3 hash of the header bytes.
     ///
     /// This hash is used as the unique identifier of an operation, aka the Operation Id.
     pub fn hash(&self) -> Hash {
-        Hash::digest(self.to_bytes())
+        // Re-calculate hash and size in test environments.
+        if cfg!(any(test, feature = "test_utils")) {
+            return Hash::digest(self.encode());
+        }
```
sandreae, line 545: Why do we re-calculate the has in test environments?

p2panda-core/src/operation.rs
```diff
@@ -552,28 +1300,22 @@ where
...
 /// * Current headers seq number increments from the past one by exactly `1`
 /// * Backlink hash contained in the current header matches the hash of the past header
-pub fn validate_backlink<E>(
-    past_header: impl Borrow<Header<E>>,
-    header: impl Borrow<Header<E>>,
-) -> Result<(), OperationError>
+pub fn validate_backlink<T>(past_header: &T, header: &T) -> Result<(), OperationError>
 where
-    E: Extensions,
+    T: Provenance<VerifyingKey> + Digest<Hash> + Chain<Hash>,
```
sandreae, line 1311: Great we have all the traits needed to express generic APIs like this now.

p2panda-core/src/traits.rs
```diff
@@ -25,6 +27,27 @@ where
...
 }
 
+/// Hash-chain structure with integrity guarantees and sequence numbers as a performance
+/// optimization.
+pub trait Chain<ID> {
+    /// Pointer at previous entry in log which gives us the integrity guarantee of the "hash chain".
+    /// The first entry in a log returns `None`.
+    fn backlink(&self) -> Option<ID>;
+
+    /// Sequence numbers are helpful to fastly detect forks and use the much faster and optimized
```
sandreae, line 43:
```suggestion
    /// Sequence numbers are helpful to quickly detect forks and use the much faster and optimized
```

p2panda-core/src/traits.rs
```diff
@@ -25,6 +27,27 @@ where
     fn verify(&self) -> bool;
 }
 
+/// Hash-chain structure with integrity guarantees and sequence numbers as a performance
+/// optimization.
+pub trait Chain<ID> {
```
sandreae, line 38: Like the new traits naming.

p2panda-sync/src/protocols/log_sync.rs
sandreae, line 62: Can we remove `E` from `LogSync` now?

sandreae, line 97: My above point about removing `E` from `LogSync` would require removing it here in `LogStore` as well. Maybe that's why you didn't do it (yet?).

p2panda-sync/src/protocols/topic_log_sync.rs
```diff
@@ -109,7 +109,7 @@ where
     E: Extensions + Send + 'static,
 {
     type Error = TopicLogSyncError;
-    type Message = TopicLogSyncMessage<L, E>;
```
sandreae, line 112: 😍

p2panda-store/src/operations/sqlite.rs
sandreae, line 31: Would be cool to have an `AnyOperationStore` / API which could be used in places where the extensions aren't / don't need to be known (like in sync).

sandreae: Maybe that would even be the only store? The user has to take care of de-serializing the `AnyOperation` received from the store into it's concrete type.

sandreae (commented):
First pass comments from me, great work!

p2panda-core/src/operation.rs
adzialocha: We have a lot of places where the parameters seq num & backlink come from the database and I couldn't find a way to elegantly map that to the API, other than just passing them over.

Maybe an enum could help and we can rewrite the places where we use the store?

p2panda-core/src/operation.rs
adzialocha: Otherwise tests will fail because the hash-integrity is invalid (hash was not generated yet) etc. when constructing the `Header` manually

p2panda-sync/src/protocols/log_sync.rs
adzialocha: I didn't dive too deep into each crate, but I guess so?

adzialocha: Yeah, it's a rabbit hole 😄

p2panda-core/src/operation.rs
adzialocha: Let me try this out

adzialocha: Tried it out but couldn't get to a good / possible design here unfortunately!

sandreae: Thanks for trying :+1:

p2panda-core/src/operation.rs
sandreae: Yeah, an issue for that could be good, it's no biggie.

adzialocha: Done that here: https://github.com/p2panda/p2panda/issues/1363

p2panda-store/src/operations/sqlite.rs
adzialocha: Totally agree that this will be a good refactor in a future PR!

adzialocha: I guess there's an issue for this already: https://github.com/p2panda/p2panda/issues/1018

p2panda-core/src/operation.rs
mycognosist, line 1: Bit of a meta-note (not requiring changes in this PR): this module has grown very large. How do we feel about that generally? I know I lean towards small modules (`body.rs`, `header.rs` etc. inside `operation` module) but would be interesting to hear if it bothers you at all.

p2panda-core/src/extensions.rs
```diff
@@ -61,20 +61,11 @@
...
-//!     version: 1,
-//!     verifying_key: signing_key.verifying_key(),
-//!     signature: None,
-//!     payload_size: body.size(),
-//!     payload_hash: Some(body.hash()),
-//!     seq_num: 0,
-//!     backlink: None,
-//!     extensions: extensions.clone(),
-//! };
+//! let body: Body = Body::from_bytes("Hello, Sloth!".as_bytes());
```
mycognosist: I believe this `body` declaration is redundant; the `.body` method on L67 simply passes `"Hello, Sloth".as_bytes()`.

p2panda-core/src/operation.rs
```diff
@@ -327,73 +580,578 @@ impl<E> Header<E> {
...
-    pub fn to_hex(&self) -> String {
-        hex::encode(self.to_bytes())
+    type Error = HeaderError;
+
+    fn try_from(value: AnyHeader) -> Result<Self, Self::Error> {
+        let extensions = match value.extensions {
+            Some(ref cbor) => {
+                // For ZST extension types we don't expect the extensions field in the header to be
+                // set. Since we now know E we can assure that this is the case.
+                if !Header::<E>::has_non_zero_sized_extensions() {
```
mycognosist: Very minor nitpick: I'd consider renaming this method to `has_zero_sized_extensions` to make these kinds of statements easier to understand (currently we have a double negation here with `!` and `non` which tripped me up when reading this).

p2panda-core/src/operation.rs
```diff
@@ -327,73 +580,578 @@ impl<E> Header<E> {
...
+    UnexpectedHeaderType(cbor_core::Error),
+
+    #[error("missing \"{0}\" field in header")]
+    MissingField(&'static str),
+
+    #[error("unexpected \"{0}\" field type for \"{0}\"")]
+    UnexpectedFieldType(cbor_core::Error, &'static str),
+
+    #[error("invalid verifying key: {0}")]
+    InvalidVerifyingKey(crate::identity::IdentityError),
```
mycognosist, line 1123: Style nitpick: import `IdentityError` at the top of the module.

p2panda-core/src/serde.rs
```diff
@@ -407,16 +271,15 @@ mod tests {
 
         // For CBOR the bytes just get deserialized straight away as an array as it is not a human
         // readable encoding
```
mycognosist: Style nitpick: full stops missing from code comments in this module.

mycognosist (approved):
Great work! All of these small improvements are going to make a huge difference in the long-run.

sandreae (approved):
Nothing more to add on top of my previous comments. Great work! Beyond the core forwards-compatibility feature this PR introduces I'm especially happy to have `AnyOperation` type now 😄

p2panda-core/src/operation.rs
adzialocha: I was thinking the same .. this should definitely be part of a follow-up (didn't wanna make the diffs even harder to read with changing the files in this PR).

p2panda-core/src/operation.rs
adzialocha: I find this actually better, makes the import section above cleaner and you see straight away where the external error is coming from :-)

PR #1202, opened 2026-06-05
Encrypted spaces and groups management integration into high-level Node API

### LINKED ISSUE #1213: Integrate `p2panda-spaces` into `p2panda` Node API

adzialocha:
First PR towards E2EE for operations in p2panda's event streaming API with Group Management CRDT and DCGKA of `p2panda-spaces`.

The API roughly looks like this:

```rust
#[derive(Serialize, Deserialize)]
struct SecretData {
    title: String,
    content: String,
}

let node = p2panda::spawn().await?;

// Spaces behave like topic-streams, just that they're encrypted towards members.
let topic = Topic::random();

// Create a space with only us inside.
//
// An empty initial members array will insert us automatically.
let (space, _rx) = node.create_space::<SecretData>(topic, &[]).await?;

// We can manage (nested) groups (useful for multi-device, etc.)
let penguin_laptop = SigningKey::generate().verifying_key();
let penguin_mobile = SigningKey::generate().verifying_key();

let penguin = node
    .create_group(&[
        (penguin_laptop, AccessLevel::Read),
        (penguin_mobile, AccessLevel::Write),
    ])
    .await?;

// .. and add them to the space as well.
space.add(penguin, AccessLevel::Read).await?;

// Every message published into a space can be decrypted by it's members.
space
    .publish(SecretData {
        title: "My favorite things".to_string(),
        content: "Hello, everyone!".to_string(),
    })
    .await?;
```

Note that this is only the first step towards a proper `p2panda-spaces` integration into `p2panda` and mostly defines the initial high-level API plus main sites of integration. There's a bunch of open TODOs I've marked in the regarding places.

Closes: #1213 

## 📋 Checklist


p2panda/src/processor/event.rs
```diff
@@ -51,6 +63,7 @@ where
         log_id: L,
         topic: TP,
         prune_flag: PruneFlag,
+        spaces_args: Option<crate::spaces::types::SpacesArgs>,
```
mycognosist: Type import convention: can be moved to top of the file.

p2panda/src/processor/event.rs
```diff
@@ -69,7 +83,21 @@ where
                 LogPruneArgs::Ignore
             },
             log_prune: ProcessorStatus::Pending,
+            spaces_args: match spaces_args {
+                Some(args) => SpacesArgs::Process {
+                    msg: p2panda_store::spaces::SpacesMessage {
```
mycognosist: Type import convention: can be moved to top of the file.

p2panda/src/spaces/forge.rs
```diff
@@ -0,0 +1,203 @@
...
+
+const KEY_BUNDLE_LOG_ID: &[u8] = b"key_bundle/v1";
+
+const GROUP_CONTROL_MESSAGE: &[u8] = b"group_control/v1";
+
+const SPACE_CONTROL_MESSAGE: &[u8] = b"space_control/v1";
+
+const SPACE_APPLICATION_MESSAGE: &[u8] = b"space_application/v1";
+
+/// This forge maintains space, key-bundle and group logs which start are organised independently.
```
mycognosist:
```suggestion
/// This forge maintains space, key-bundle and group logs which are organised independently.
```

p2panda/src/processor/event.rs
```diff
@@ -69,7 +83,21 @@ where
...
+                        id: operation.hash,
+                        author: operation.header.verifying_key,
+                        args,
+                    }
+                    // TODO: Wrangling with all sorts of SpacesMessage and SpacesArgs variants is
```
sandreae: One thing that will improve this: p2panda-spaces can use  SpacesMessage from p2panda-store instead of introducing a wrapper.

p2panda/src/processor/pipeline.rs
```diff
@@ -107,6 +113,10 @@ where
                     // Prepare event processing pipeline.
                     let ingest = Ingest::<S, Event<L, E, TP>, L, E, TP>::new(store.clone());
                     let log_prune = LogPrune::<S, Event<L, E, TP>, L, E>::new(store);
+                    let spaces = SpacesProcessor::<Event<L, E, TP>>::new(spaces_manager);
+
+                    // TODO: Add orderer whenever it's ready (needs API adjustments). Who is
+                    // ordering for us currently? Is it still in the spaces processor?
```
sandreae: We already discussed this, but only additional ordering related to encryption (awaiting welcome messages) is performed in -spaces.

p2panda/src/processor/pipeline.rs
```diff
@@ -131,8 +141,23 @@ where
...
+                                event.spaces = ProcessorStatus::Completed(result);
+                                event
+                            }
+                            Err((mut event, err)) => {
+                                event.spaces = ProcessorStatus::Failed(Arc::new(err));
+                                event
+                            }
                         });
 
+                    // TODO: After processing we want to update our topic mapping (in TopicStore)
```
sandreae: Is this also where we'll perform spaces repair?

p2panda/src/spaces/forge.rs
sandreae: Great docs!

sandreae: And I agree with the topic -> log mappings.

p2panda/src/spaces/forge.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    fn verifying_key(&self) -> VerifyingKey {
+        crate::forge::Forge::verifying_key(self)
+    }
+
+    async fn forge(
+        &self,
+        args: p2panda_spaces::SpacesArgs<SpaceId, AuthCapabilities>,
+    ) -> Result<SpacesMessage, Self::Error> {
+        // TODO: Do we need to query graph tips here for causal ordering or is this taken care off
```
sandreae, line 90: In spaces dependencies related to "groups" and "spaces" are included in the spaces args. I think it makes sense to include both there as they are hard dependencies of the spaces system. But open to other approaches if you think it's important.

p2panda/src/spaces/forge.rs
```diff
@@ -0,0 +1,203 @@
...
+                let bytes = encode_cbor(&key_bundle).expect("serialisation of key bundle to CBOR");
+
+                // Every author maintains their own key bundle log.
+                let log_id = LogId::digest(KEY_BUNDLE_LOG_ID);
+
+                // TODO: The key bundle itself should not be in the header to allow deleting it
+                // after a while (without breaking the whole key bundle log.
+                let extensions = Extensions::builder(log_id).build_space(args);
+
+                // TODO: The associated topic should be the space_id but we don't have this info
```
sandreae, line 106: We discussed making the topic mappings after the event has passed through the processor, I assume that addresses this TODO as well.

p2panda/src/spaces/forge.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    async fn forge(
+        &self,
+        args: p2panda_spaces::SpacesArgs<SpaceId, AuthCapabilities>,
+    ) -> Result<SpacesMessage, Self::Error> {
+        // TODO: Do we need to query graph tips here for causal ordering or is this taken care off
+        // by -spaces? If yes, are declaring _all_ dependencies really it's concern or only the ones
+        // which are relevant to the spaces protocol?
+        //
+        // TODO: We need to apply a topic mapping so our TopicStore is aware of the relationship
```
sandreae: We discussed moving the topic mapping logic to after an event has been processed.

p2panda/src/spaces/forge.rs
```diff
@@ -0,0 +1,203 @@
...
+                    // The group id would be enough to indicate the log id, we hash it here together
+                    // with a constant value to prevent possible collisions with logs of same id but
+                    // different purpose.
+                    bytes.extend_from_slice(GROUP_CONTROL_MESSAGE);
+                    bytes
+                });
+
+                let extensions = Extensions::builder(log_id).build_space(args);
+
+                // TODO: The associated topic should be the space_id but we don't have this info
```
sandreae, line 129: Same as above.

p2panda/src/spaces/forge.rs
```diff
@@ -0,0 +1,203 @@
...
+                    let mut bytes = Vec::new();
+                    bytes.extend_from_slice(space_id.as_bytes());
+                    bytes.extend_from_slice(SPACE_CONTROL_MESSAGE);
+                    bytes
+                });
+
+                let extensions = Extensions::builder(log_id).build_space(args);
+
+                // Associate this log with the space id / topic.
+                let topic = Topic::from(space_id);
```
sandreae, line 153: Assume this will be removed after topic association moves.

p2panda/src/spaces/forge.rs
```diff
@@ -0,0 +1,203 @@
...
+                self.create_operation(topic, log_id, None, extensions)
+                    .await?
+            }
+
+            // 4. Application logs.
+            //
+            // TODO: We likely don't want to forge application messages here _at all_ and rather use
+            // an independent (new) method on `Manager` which allows us to encrypt anything (control
+            // messages & application messages) against the latest secret _without_ forging
+            // something.
```
sandreae: Do we need any changes to the fields on the application variant for spaces args to make this change? Currently they are: https://github.com/p2panda/p2panda/blob/d37d09ee70cb2982fa6aa03fd82036df4d82b4d8/p2panda-spaces/src/message.rs#L198-L213

p2panda/src/spaces/forge.rs
```diff
@@ -0,0 +1,203 @@
...
+                    let mut bytes = Vec::new();
+                    bytes.extend_from_slice(space_id.as_bytes());
+                    bytes.extend_from_slice(SPACE_APPLICATION_MESSAGE);
+                    bytes
+                });
+
+                let extensions = Extensions::builder(log_id).build_space(args);
+
+                // Associate this log with the space id / topic.
+                let topic = Topic::from(space_id);
```
sandreae, line 179: Same here (to be removed I assume)

p2panda/src/spaces/group.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    pub fn event_stream(&self) -> impl Stream<Item = GroupEvent> + Send + Unpin + 'static {
+        // Make sure we're not re-subscribing and thus dropping all events which might be still in
+        // the buffer of the first broadcast receiver instance.
+        let stream = self
+            .event_stream_rx
+            .replace_with(|stream| stream.resubscribe());
+
+        // TODO: Check if we really want to silence broadcast "lagged" errors here?
```
sandreae, line 83: What are the implications of silencing these errors?

p2panda/src/spaces/group.rs
```diff
@@ -0,0 +1,218 @@
...
+        let stream = BroadcastStream::new(stream).filter_map(|event| async { event.ok() });
+
+        Box::pin(stream)
+    }
+
+    pub async fn add(
+        &self,
+        actor: impl Into<ActorId>,
```
sandreae, line 91: I'm not 100% about the name "actor" here, maybe I prefer "member". I'm slightly on the fence. I like "actor" because it makes sense when using it in general to describe entities / identities which exist. But in the API it's very clearly a "member" being added/removed.

p2panda/src/spaces/group.rs
```diff
@@ -0,0 +1,218 @@
...
+    // TODO: "actors" method to return the _non-flattened_ actors in a group. This will help to
+    // build multi-device applications.
+}
+
+#[allow(clippy::from_over_into)]
+impl Into<ActorId> for Group {
+    fn into(self) -> ActorId {
+        self.inner.id()
+    }
+}
```
sandreae, line 154: Is this here so that a group struct can be used as an argument in the add(..), remove(..) API ?

p2panda/src/spaces/group.rs
```diff
@@ -0,0 +1,218 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
```
sandreae, line 156: What's the reason for using this new type instead of https://github.com/p2panda/p2panda/blob/d37d09ee70cb2982fa6aa03fd82036df4d82b4d8/p2panda-spaces/src/event.rs#L96 ?

p2panda/src/spaces/member.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+impl From<Member> for p2panda_spaces::member::Member {
+    fn from(value: Member) -> p2panda_spaces::member::Member {
+        value.inner
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct GroupActor {
```
sandreae, line 36: Is this new type needed or can we use the equivalent type from p2panda-spaces?

sandreae (commented):
Partial review 👍

p2panda/src/node.rs
```diff
@@ -353,6 +367,124 @@ impl Node {
...
     }
 
+    pub async fn register_member(&self, member: Member) -> Result<(), MemberError> {
+        let member: p2panda_spaces::member::Member = member.into();
+        self.spaces_manager.register_member(&member).await?;
+
+        Ok(())
+    }
+
+    pub async fn group(&self, actor_id: impl Into<ActorId>) -> Result<Option<Group>, GroupError> {
```
mycognosist: Already discussed but the `actor_id` parameter could potentially be renamed to `group_id`. The goal is to make it clear that this method returns a handle to the group represented by the provided identifier (it's not returning all members of the group, for example).

We also discussed clear feedback for the user. For example, if they provide an id which does not belong to a group but rather an individual. This work can happen later, just mentioning it here.

mycognosist (commented):
Ran out of time so submitting a partial review for now.

Another thing which came up on our call was the need to publicly export some of the types (such as events).

p2panda/src/spaces/space.rs
```diff
@@ -0,0 +1,247 @@
...
+    rx: StreamSubscription<M>,
+) -> (Space<M>, SpaceSubscription<M>) {
+    (Space { inner, tx }, SpaceSubscription { rx })
+}
+
+// TODO: We need a way to automatically publish key bundles (if configured, it should also be an
+// option to _not_ do that to only allow initial key agreement through side channels which is more
+// private).
+//
+// TODO: Automatically repair spaces (detect missing group changes and apply them to affected
```
sandreae, line 34: This is something else we discussed moving to inside/after the spaces processor.

p2panda/src/node.rs
sandreae: Agree that returning an error if the actor id passed belongs to an individual would be good. Probably this is best done by querying the key stores rather than via the groups API.

p2panda/src/operation.rs
```diff
@@ -400,44 +454,33 @@ impl BasicExtensions {
...
-#[allow(unused)]
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct CausalExtensions {
+pub struct SpaceExtensions {
     pub log_id: LogId,
     pub timestamp: Timestamp,
-    pub previous: HashSet<Hash>,
+
+    // TODO: We likely want to split spaces args into more extensions types. At least one for
+    // application messages & key bundles, potentially also auth / group messages.
```
sandreae, line 464: 👍

sandreae (approved):
All looking good to me, only some minor comments. Of course there is lots of follow-up work to be done, but that's outside of this PR.

p2panda/src/processor/pipeline.rs
adzialocha: Yeah! Why not, similar like how we associate the topics now in the processor we could also have the throttle to repair there, makes sense!

p2panda/src/spaces/forge.rs
adzialocha: I haven't thought about it much yet but I assume we want the space_id, group_secret_id and nonce (for header) and the plaintext (for body) here.

group_secret_id and nonce can then be used as an argument on Manager::encrypt to get the ciphertext.

adzialocha: We want to tie the operation to a fixed nonce & secret id because the type itself is immutable and we can't assign it a _different_ nonce or secret later.

p2panda/src/spaces/group.rs
adzialocha: Users will not notice that they are missing out on events

p2panda/src/processor/event.rs
adzialocha: I think this issue was solved but just as a note that sometimes one makes the path explicit to make it very clear where this type comes from, especially around args and message we have a lot of variants in different places, so this was intentional.

p2panda/src/spaces/group.rs
adzialocha: Good question! I'm not sure. I maybe would leave it as is, until we understand what a groups event stream from the perspective of the Node API is? Probably it'll be an enum which might wrap GroupEvent coming from -spaces.

PR #1216, opened 2026-06-09
Adjust spaces API to suit "command issuing" and "processing" separation

sandreae:
Several API changes required for integrating `p2panda-spaces` into the Node API.

- only emit events on Manager::process calls
- don't automatically sync groups state changes with all spaces (now this should be done by the user)
- don't persist state in public "command" issuing API (Manager::create_space, Space::add etc...)

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Highlights are marked with a pancake 🥞
 - Use framed postcard codec instead of CBOR for wire protocols [#1198](https://github.com/p2panda/p2panda/pull/1198)
 - Future-proof extensions format in Node API [#1155](https://github.com/p2panda/p2panda/pull/1155)
 - spaces: Only return events when calling Manager::process [#1216](https://github.com/p2panda/p2panda/pull/1216)
+- spaces: Don't sync all spaces when group membership changes
+  [#1216](https://github.com/p2panda/p2panda/pull/1216)
 
 ## [0.6.1] - 22/05/2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Highlights are marked with a pancake 🥞
 - spaces: Only return events when calling Manager::process [#1216](https://github.com/p2panda/p2panda/pull/1216)
 - spaces: Don't sync all spaces when group membership changes
   [#1216](https://github.com/p2panda/p2panda/pull/1216)
+- spaces: Adjust all "command" methods to not persist state locally [#1216](https://github.com/p2panda/p2panda/pull/1216)
 
 ## [0.6.1] - 22/05/2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Highlights are marked with a pancake 🥞
     - Fix header encoding for ZST extensions [#1196](https://github.com/p2panda/p2panda/pull/1196)
 - Use framed postcard codec instead of CBOR for wire protocols [#1198](https://github.com/p2panda/p2panda/pull/1198)
 - Future-proof extensions format in Node API [#1155](https://github.com/p2panda/p2panda/pull/1155)
+- spaces: Only return events when calling Manager::process [#1216](https://github.com/p2panda/p2panda/pull/1216)
 
 ## [0.6.1] - 22/05/2026
 
--- a/p2panda-spaces/src/group.rs
+++ b/p2panda-spaces/src/group.rs
@@ -81,7 +81,7 @@ where
     pub(crate) async fn create(
         manager_ref: Manager<ID, S, K, F, M, C, RS>,
         initial_members: Vec<(ActorId, Access<C>)>,
-    ) -> Result<(Self, Vec<M>, Event<ID, C>), GroupError<ID, S, K, F, M, C, RS>> {
+    ) -> Result<(Self, Vec<M>), GroupError<ID, S, K, F, M, C, RS>> {
         // Generate random group id.
         let group_id: ActorId = {
             let manager = manager_ref.inner.read().await;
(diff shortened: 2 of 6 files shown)


p2panda-spaces/src/manager.rs
```diff
@@ -194,20 +242,25 @@ where
     /// without manager rights. If this is done then after creation no further change of the group
     /// membership would be possible.
     ///
-    /// Returns messages for replication to other instances and events which inform users of any
-    /// state changes which occurred.
+    /// Returns resulting auth state, group id and message for processing.
     pub async fn create_group(
         &self,
+        auth_y: AuthGroupState<C>,
```
adzialocha: Is there a reason why the auth state needs to be explicitly passed in as an argument and returned?

From a users perspective (Node API), I would not really do anything with the `AuthGroupState` object.

sandreae: In the spaces API it needs to be passed in so that we can isolate changes to both the auth state and space state in all places where state mutations occur.

sandreae: I hadn't imagined this being exposed users of the Node API though (if that's what you're suggesting).

sandreae: You as a Node API developer does need to handle this of course, this would involve getting current state from the store and passing it into this method.

adzialocha: > You as a Node API developer does need to handle this of course, this would involve getting current state from the store and passing it into this method.

Do you have an example where I would need to do that? Right now I' calling "get state" manually and just pass it into the method without needing to do something with it (it's read-only), that's why I'm wondering why it is exposed to me at all as a high-level API user (someone integrating p2panda-spaces).

sandreae: Ok, I think I grok you :smile: on the manager we don't need to pass the auth state argument from the outside. I don't think there are any cases where we're intending to pass in anything except the current local state. I'll change this, thanks :+1:

sandreae: I'm wondering if we never have to pass the state in from the outside actually :thinking: maybe this was an anti-pattern :thinking: the main thing is that we don't persist the mutated state, but it's always fine just to take the latest state from the store when each method is called. Let me check this...

sandreae: Jah, this is the case, I'll update the API (which looks much nicer now)

p2panda-spaces/src/space.rs
```diff
@@ -547,29 +561,46 @@ where
...
-        let my_id = self.manager.id();
-        let is_reader = self
-            .members()
-            .await?
-            .iter()
-            .any(|(member, access)| *member == my_id && access > &Access::pull());
+        // @TODO: here we need to account for the new Groups::heads_filtered(..) approach to
+        // calculating dependencies and only include the ones strictly necessary for this space.
+        let operation_ids =
+            toposort(&global_auth_y.inner.graph, None).expect("auth graph does not contain cycles");
```
adzialocha, line 565: Is this part of a follow-up PR?

adzialocha (approved):

p2panda-spaces/src/manager.rs
```diff
@@ -316,6 +367,18 @@ where
...
 
+    /// Get the global auth state.
+    pub async fn auth(&self) -> Result<AuthGroupState<C>, <S as AuthStore<C>>::Error> {
+        let manager = self.inner.read().await;
+        manager.store.auth().await
+    }
+
+    /// Set the global auth state.
+    pub async fn set_auth(&self, y: &AuthGroupState<C>) -> Result<(), <S as AuthStore<C>>::Error> {
+        let manager = self.inner.read().await;
```
mycognosist: Might you rather want to gain exclusive access here by calling `.write()`?

mycognosist (approved):
Looking good! Nice to have the separate persisted method variants.

CHANGELOG.md
```diff
@@ -18,7 +18,10 @@ Highlights are marked with a pancake 🥞
     - `version` from `u64` to `u16` [#1194](https://github.com/p2panda/p2panda/pull/1194)
     - Remove `timestamp` [#1195](https://github.com/p2panda/p2panda/pull/1195)
     - Fix header encoding for ZST extensions [#1196](https://github.com/p2panda/p2panda/pull/1196)
-- Use framed postcard codec instead of CBOR for wire protocols [#1198](https://github.com/p2panda/p2panda/pull/1198)
+- spaces: Only return events when calling Manager::process [#1216](https://github.com/p2panda/p2panda/pull/1216)
```
mycognosist, line 23: We've never taken this approach of prefixing the `CHANGELOG` entry with the repo being affected before, as far as I can tell, but I quite like it.

adzialocha: I've started to also use it in commit messages!

p2panda-spaces/src/group.rs
```diff
@@ -76,103 +75,106 @@ where
...
+
+        Ok(message)
+    }
+
+    /// Add member to group with specified access level.
+    ///
+    /// Returns resulting state and message for processing.
+    pub async fn add(
+        &self,
+        y: AuthGroupState<C>,
```
adzialocha: Is it possible to remove passing in the state here as well?

p2panda-spaces/src/group.rs
```diff
@@ -76,103 +75,106 @@ where
...
-        Ok((messages, events))
+        Ok(message)
+    }
+
+    /// Remove member from group.
+    ///
+    /// Returns resulting state and message for processing.
+    pub async fn remove(
+        &self,
+        y: AuthGroupState<C>,
```
adzialocha: Is it possible to remove passing in the state here as well?

p2panda-spaces/src/group.rs
sandreae: Ah yes, missed this one.

p2panda-spaces/src/group.rs
sandreae: And here.

CHANGELOG.md
sandreae: A copied it from you @adzialocha cos I liked it :rofl:

Issue #1226, opened 2026-06-10
SQLite implementation of `KeySecretStore`


sandreae:
https://github.com/p2panda/p2panda/blob/d4aae547601be303c7306f00460c17fd8c184a60/p2panda-spaces/src/traits/store.rs#L64-L74

Issue #1225, opened 2026-06-10
SQLite implementation of `KeyRegistryStore`


sandreae:
https://github.com/p2panda/p2panda/blob/d4aae547601be303c7306f00460c17fd8c184a60/p2panda-spaces/src/traits/store.rs#L52-L62

PR #1230, opened 2026-06-11
Introduce SQLite implementations of `KeySecretsStore` and `KeyRegistryStore`

mycognosist:
In this PR we introduce trait definitions and SQLite implementations for
storing prekey bundles and key registry state.

An `id` column is added to the SQL table for both. For now this value is
constant for each table; in the future we may choose to provide a
user-defined value. A simple `INSERT OR REPLACE` query is used to
overwrite the previous state on each call to `set_`.

`PartialEq` and `Eq` have been implemented for `PreKeyBundlesState` and
`PreKeyBundle` to facilitate state comparison in tests.

https://github.com/p2panda/p2panda/issues/1226
https://github.com/p2panda/p2panda/issues/1225

## 📋 Checklist


p2panda-store/src/key_secrets/tests.rs
```diff
@@ -0,0 +1,53 @@
...
+    let new_state = key_manager.prekey_bundles();
+
+    // Ensure the prekey bundles are unique.
+    assert_ne!(state, new_state);
+
+    // Store the new prekey bundles.
+    tx_unwrap!(store, store.set_prekey_secrets(&new_state).await.unwrap());
+
+    // New prekey bundles have overwritten the previous state.
+    assert_eq!(
```
mycognosist, line 54: This is the source of failure in both tests. We expect that the value retrieved from the store will be the one which was most recently inserted (ie. `new_state`). It seems that the initial state is not being overwritten by the call in L46.

p2panda-store/src/key_registry/traits.rs
```diff
@@ -0,0 +1,20 @@
...
+use p2panda_encryption::key_registry::KeyRegistryState;
+use p2panda_spaces::ActorId;
+
+/// Interface for setting and getting key registry state.
+pub trait KeyRegistryStore {
+    type Error: Debug;
+
+    fn get_key_registry(
+        &self,
+    ) -> impl Future<Output = Result<Option<KeyRegistryState<ActorId>>, Self::Error>>;
```
mycognosist: I've kept the function signatures concrete here but then we run into an issue with circular dependencies if we want to remove these traits from `p2panda-spaces` (because then `p2panda-spaces` and `p2panda-store` depend on one another).

One option is to follow the example in the `groups` store and keep the traits completely generic; then we only need to depend on `p2panda-spaces` in our tests.

sandreae: > One option is to follow the example in the groups store and keep the traits completely generic; then we only need to depend on p2panda-spaces in our tests.

I believe this is the way to go, it's what I'm doing with the work on `SpacesMessageStore` as well.

p2panda-store/src/key_registry/sqlite.rs
```diff
@@ -0,0 +1,68 @@
...
+use p2panda_encryption::key_registry::KeyRegistryState;
+use p2panda_spaces::ActorId;
+use sqlx::{query, query_scalar};
+
+use crate::key_registry::traits::KeyRegistryStore;
+use crate::{SqliteError, SqliteStore};
+
+// Constant identifier used to provide a primary key for the database table.
+// This makes it possible to use INSERT OR REPLACE to update the key registry state.
+const ID: &str = "key_registry_state_id";
```
adzialocha: Maybe we can give this a more "meaningful" identifier (instead of ".. id") like "default" or "main"

p2panda-store/src/key_registry/sqlite.rs
```diff
@@ -0,0 +1,68 @@
...
+                    "
+                    SELECT
+                        state
+                    FROM
+                        key_registry_v1
+                    WHERE
+                        id = ?
+                    ",
+                )
+                .bind(encode_cbor(&ID).map_err(|err| SqliteError::Encode("id".to_string(), err))?)
```
adzialocha: You can change the type in the SQL table to TEXT and save the CBOR encoding step here just for the id, it is already a string

p2panda-store/src/key_registry/sqlite.rs
```diff
@@ -0,0 +1,68 @@
...
+                .fetch_optional(pool)
+                .await
+                .map_err(SqliteError::Sqlite)
+            })
+            .await?;
+
+        if let Some(bytes) = state_bytes {
+            let state: KeyRegistryState<ActorId> = decode_cbor(&bytes[..])
```
adzialocha: Thinking out loud: We want to ensure backwards-compatibility over the encoding format for these as well. Let's look into this when we make this store a bit more "efficient" (rather than dumping a whole blob into it)

p2panda-store/src/key_registry/traits.rs
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::fmt::Debug;
+
+use p2panda_encryption::key_registry::KeyRegistryState;
+use p2panda_spaces::ActorId;
+
+/// Interface for setting and getting key registry state.
+pub trait KeyRegistryStore {
+    type Error: Debug;
```
adzialocha: In other store trait definitions we require `std::error::Error` here

p2panda-store/src/lib.rs
```diff
@@ -121,6 +121,8 @@
 pub mod address_book;
 pub mod cursors;
 pub mod groups;
+pub mod key_registry;
+pub mod key_secrets;
```
adzialocha, line 125: Wondering if we should start moving these behind feature flags. Otherwise users will need to compile `p2panda-encryption` if they just need the address book store (see bootstrap node for example).

We might want to do that for other "non core" stores as well (groups, spaces, ..)

adzialocha (approved):
This looks good!

I only have nitpicks, nothing blocking a merge. Also very optional: You might want to split your commit into two and isolate the changes to `-encryption` (adding `Eq` etc.). You can use `git history split <commit_id>` for this.

p2panda-store/src/key_registry/traits.rs
adzialocha: > > One option is to follow the example in the groups store and keep the traits completely generic; then we only need to depend on p2panda-spaces in our tests.
> 
> I believe this is the way to go, it's what I'm doing with the work on `SpacesMessageStore` as well.

+1

p2panda-store/src/lib.rs
mycognosist: https://github.com/p2panda/p2panda/issues/1243

p2panda-store/src/key_registry/sqlite.rs
mycognosist: https://github.com/p2panda/p2panda/issues/1244

sandreae (approved):

PR #1241, opened 2026-06-16
Spaces store traits with SQLite implementations

### LINKED ISSUE #1224: SQlite implementations of all spaces stores

sandreae:
Introduces `SpacesStore`, `SpacesStoreWrite` and `SpacesMessageStore` traits in p2panda-store along with SQLite implementations on new concrete store type `SqliteSpacesStore`.

## 📋 Checklist

### DIFF
--- a/p2panda-store/src/operations/traits.rs
+++ b/p2panda-store/src/operations/traits.rs
@@ -2,22 +2,24 @@
 
 use std::error::Error;
 
+use p2panda_core::LogId;
+
 /// Interface for storing, deleting and querying operations.
 ///
 /// The concrete type of an "operation" is generic and implementors can use the same interface for
 /// different approaches: sets, append-only logs, hash-graphs (DAG) etc.
-pub trait OperationStore<T, ID, C> {
+pub trait OperationStore<T, ID> {
     type Error: Error;
 
     /// Insert an operation.
     ///
     /// Returns `true` when the insert occurred, or `false` when the operation already existed and
     /// no insertion occurred.
-    fn insert_operation(
+    fn insert_operation<L: LogId>(
         &self,
         id: &ID,
         operation: &T,
-        collection_id: &C,
+        collection_id: &L,
     ) -> impl Future<Output = Result<bool, Self::Error>>;
 
     /// Get an operation by id.
--- a/p2panda-store/src/spaces/mod.rs
+++ b/p2panda-store/src/spaces/mod.rs
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+mod sqlite;
+#[cfg(test)]
+mod tests;
+mod traits;
+
+pub use sqlite::SqliteSpacesStore;
+pub use traits::{SpacesMessageStore, SpacesStore, SpacesStoreWrite};
+
+use p2panda_core::{Hash, VerifyingKey};
+
+/// Spaces message type with generic parameter for additional arguments.
+// @TODO: this type is so generic that it could actually be used / defined somewhere else. It
+// represents a message with id and author, replacing the need for trait interfaces.
+#[derive(Debug, Clone)]
+pub struct SpacesMessage<T> {
+    pub id: Hash,
+    pub author: VerifyingKey,
+    pub args: T,
+}
--- a/p2panda-store/src/spaces/sqlite.rs
+++ b/p2panda-store/src/spaces/sqlite.rs
@@ -1,0 +1,220 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::borrow::Borrow;
+use std::marker::PhantomData;
+
+use p2panda_core::cbor::{decode_cbor, encode_cbor};
+use p2panda_core::{Extensions, Hash, Operation};
+use serde::{Deserialize, Serialize};
+use sqlx::{query, query_as};
+
+use crate::groups::GroupsStore;
+use crate::operations::OperationStore;
+use crate::spaces::traits::SpacesMessageStore;
+use crate::spaces::{SpacesMessage, SpacesStore, SpacesStoreWrite};
+use crate::sqlite::TransactionPermit;
+use crate::{SqliteError, SqliteStore};
+
+/// Concrete SqliteSpacesStore type which wraps the SqliteStore.
+pub struct SqliteSpacesStore<E> {
+    store: SqliteStore,
+    _phantom: PhantomData<E>,
+}
+
+impl<E> SqliteSpacesStore<E> {
+    pub fn new(store: SqliteStore) -> Self {
+        Self {
+            store,
+            _phantom: PhantomData,
+        }
+    }
+
+    pub fn inner(&self) -> SqliteStore {
+        self.store.clone()
+    }
+}
+
+impl<A, E> SpacesMessageStore<Hash, A> for SqliteSpacesStore<E>
+where
+    A: Clone,
+    E: Extensions + Borrow<A>,
+{
+    type Error = SqliteError;
+
+    async fn get_spaces_message(&self, id: &Hash) -> Result<Option<SpacesMessage<A>>, Self::Error> {
+        match <SqliteStore as OperationStore<Operation<E>, Hash>>::get_operation(&self.store, id)
+            .await?
+        {
+            Some(operation) => {
+                let args = operation.header().extensions.borrow().clone();
+                let message = SpacesMessage {
+                    id: operation.hash,
+                    author: operation.header().verifying_key,
+                    args,
+                };
+                Ok(Some(message))
+            }
+            None => Ok(None),
+        }
+    }
+}
+
+impl<ID, S, E> SpacesStore<ID, S> for SqliteSpacesStore<E>
+where
+    ID: for<'a> Deserialize<'a> + Serialize,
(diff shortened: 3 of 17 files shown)


p2panda-store/src/operations/traits.rs
```diff
@@ -2,22 +2,24 @@
...
     ///
     /// Returns `true` when the insert occurred, or `false` when the operation already existed and
     /// no insertion occurred.
-    fn insert_operation(
+    fn insert_operation<L: LogId>(
         &self,
         id: &ID,
         operation: &T,
-        collection_id: &C,
+        collection_id: &L,
```
sandreae, line 22: We kind-of move towards suggesting this "collection_id" is actually a log id here, however the LogId trait  itself is not really anything to do with logs but rather serializing and deserializing the type plus some handy trait bounds.

p2panda-store/src/spaces/sqlite.rs
```diff
@@ -0,0 +1,46 @@
...
+use p2panda_core::{Extensions, Hash, Operation};
+
+use crate::spaces::SpacesMessage;
+use crate::spaces::traits::{SpacesMessageStore};
+use crate::{SqliteError, SqliteStore, operations::OperationStore};
+
+/// Concrete SpacesStore type which wraps the SqliteStore and introduces a generic M which
+/// represents the underlying persisted message type. For integration with the OperationStore the
+/// M will be an Operation<E>.
+pub struct SpacesStore<M> {
```
sandreae: This is now the concrete store type which would be passed into spaces Manager, we would implement all required store on this type instead of SqliteStore.

p2panda-store/src/spaces/sqlite.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    pub fn inner(&self) -> SqliteStore {
+        self.store.clone()
+    }
+}
+
+impl<A, E> SpacesMessageStore<Hash, A> for SqliteSpacesStore<Operation<E>>
+where
+    A: Clone,
```
adzialocha, line 39: Does `A` need to be generic? Isn't there only one concrete args type for -spaces?

p2panda-store/src/spaces/sqlite.rs
```diff
@@ -0,0 +1,218 @@
...
+                Ok(Some(message))
+            }
+            None => Ok(None),
+        }
+    }
+}
+
+impl<ID, S, M> crate::spaces::SpacesStore<ID, S> for SqliteSpacesStore<M>
+where
+    ID: for<'a> Deserialize<'a> + Serialize,
```
adzialocha, line 64: Can use the `SpaceId` trait here, that would help readability

p2panda-store/src/spaces/sqlite.rs
```diff
@@ -0,0 +1,218 @@
...
+    E: Extensions + Borrow<A>,
+{
+    type Error = SqliteError;
+
+    async fn get_spaces_message(&self, id: &Hash) -> Result<Option<SpacesMessage<A>>, Self::Error> {
+        match <SqliteStore as OperationStore<Operation<E>, Hash>>::get_operation(&self.store, id)
+            .await?
+        {
+            Some(operation) => {
+                let args = operation.header().extensions.borrow().clone();
```
mycognosist, line 49: We chatted about this already but just to mention here in text that this is a good example of a place where we could implement `Into<A>` and require `E: Extensions + Into<A>` (rather than `Borrow`).

`let args = operation.header().extensions.into();`

p2panda-store/src/spaces/sqlite.rs
```diff
@@ -0,0 +1,218 @@
...
+                    args,
+                };
+                Ok(Some(message))
+            }
+            None => Ok(None),
+        }
+    }
+}
+
+impl<ID, S, M> crate::spaces::SpacesStore<ID, S> for SqliteSpacesStore<M>
```
mycognosist: There are a few places in this module where imports can rather be defined up top (`SpacesStore`, `SpacesStoreWrite`, `Transaction`).

mycognosist (approved):

p2panda-store/src/spaces/sqlite.rs
sandreae: Ah yeah, right! I wonder if I actually make that change. It would be nicer I think.

p2panda-store/src/spaces/sqlite.rs
sandreae: It doesn't need to be generic except that if we use a concrete type from -spaces here it creates a circular dependency.

p2panda-store/src/spaces/sqlite.rs
sandreae: Yeah, if we move SpaceId into p2panda-store or p2panda-core we can do that.

p2panda-store/src/spaces/sqlite.rs
sandreae: I re-define and export the traits with concrete types here as super-traits and provide blanket implementations.

sandreae: https://github.com/p2panda/p2panda/blob/7e02f36c03688addd375cfd37960edc62ad25a2d/p2panda-spaces/src/types.rs#L97-L112

p2panda-store/src/spaces/sqlite.rs
adzialocha: Or we make it a Topic straight away?

p2panda-store/src/spaces/sqlite.rs
```diff
@@ -0,0 +1,218 @@
+use std::{borrow::Borrow, marker::PhantomData};
```
adzialocha: Nit: License header missing + import style

p2panda-store/src/spaces/sqlite.rs
```diff
@@ -0,0 +1,218 @@
...
+use crate::groups::GroupsStore;
+use crate::spaces::SpacesMessage;
+use crate::spaces::traits::SpacesMessageStore;
+use crate::sqlite::TransactionPermit;
+use crate::{SqliteError, SqliteStore, operations::OperationStore};
+
+/// Concrete SqliteSpacesStore type which wraps the SqliteStore and introduces a generic M which
+/// represents the underlying persisted message type. For integration with the OperationStore the
+/// M will be an Operation<E>.
+pub struct SqliteSpacesStore<M> {
```
adzialocha: As an API user I'm not sure what `M` stands for: It's not used anywhere else and other impls define a concrete `Operation<E>` type, I suggest we just do the same here?

p2panda-store/src/spaces/sqlite.rs
```diff
@@ -0,0 +1,218 @@
...
+use p2panda_core::cbor::{decode_cbor, encode_cbor};
+use p2panda_core::{Extensions, Hash, Operation};
+use serde::{Deserialize, Serialize};
+use sqlx::{query, query_as};
+
+use crate::groups::GroupsStore;
+use crate::spaces::SpacesMessage;
+use crate::spaces::traits::SpacesMessageStore;
+use crate::sqlite::TransactionPermit;
+use crate::{SqliteError, SqliteStore, operations::OperationStore};
```
adzialocha: nit: Same here

p2panda-store/src/spaces/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Spaces message type with generic parameter for additional arguments.
+// @TODO: this type is so generic that it could actually be used / defined somewhere else. It
+// represents a message with id and author, replacing the need for trait interfaces.
+#[derive(Debug, Clone)]
+pub struct SpacesMessage<T> {
+    pub id: Hash,
+    pub author: VerifyingKey,
+    pub args: T,
+}
```
adzialocha, line 21: I count multiple places where `SpacesMessage` is currently defined, it's indeed a bit confusing.

adzialocha (changes requested):

adzialocha:
Needs a CHANGELOG.md entry 👍🏻

p2panda-store/src/spaces/sqlite.rs
adzialocha: In my rework I've opted for Hash, it's a bit more neutral

p2panda-store/src/spaces/sqlite.rs
sandreae: I'm gunna leave it like this for now until we're sure this is the move we want. For me it's still a little unclear if there are implications elsewhere we haven't considered.

p2panda-store/src/spaces/sqlite.rs
sandreae: I'll leave it generic for now as that's then consistent with other stores (operation store). Happy to change across the board if we want.

p2panda-store/src/spaces/mod.rs
sandreae: Yeah, it's not ideal. We need a definition here in -store (or at least not in -spaces) so the type can be used in the spaces store API. This would change if we move the the store trait definitions and impls themselves into p2panda-spaces. Raised this related issue: https://github.com/p2panda/p2panda/issues/1252

PR #1245, opened 2026-06-17
Use `p2panda-store` traits and SQLite implementations in `p2panda-spaces `

### LINKED ISSUE #1124: Avoid leaking `ActorId` type in `AuthoredMessage` trait

### LINKED ISSUE #1221: Use `GroupStore` from `p2panda-store` in `p2panda-spaces`
This would replace the `AuthStore` https://github.com/p2panda/p2panda/blob/d4aae547601be303c7306f00460c17fd8c184a60/p2panda-spaces/src/traits/store.rs#L41-L50

### LINKED ISSUE #1246: Return computed state rather than persisting inside `Manager::process`

### LINKED ISSUE #1248: Use SQLite versions of key stores in p2panda-spaces test utils

sandreae:
- [x] use spaces store traits defined in `p2panda-store`  and remove old ones
- [x] use SQLite implementations from `p2panda-store` in testing, remove old memory store
- [x] replace `ActorId` and `OperationId` wrappers with core types `VerifyingKey` and `Hash`
- [x] return computed states from calls to `Manager::process` instead of persisting internally
- [x] do not require `SpacesStoreWrite` in any of the public APIs, only used behind testing feature flags 
- [X] use transactions in `IdentityManager` API

## Related next steps

- changes implemented in this branch `sqlite-spaces-adz-changes`
- also separate `GroupsStore` into read/write halves https://github.com/p2panda/p2panda/issues/1247
- message stores for auth and encryption messages

## 📋 Checklist


p2panda-encryption/src/key_manager.rs
```diff
@@ -21,7 +21,10 @@ use crate::traits::{IdentityManager, PreKeyManager};
 
 /// Key manager to maintain secret key material (like our identity key) and to generate signed
 /// public pre-key bundles.
-#[derive(Clone, Debug)]
+// @TODO(sam): the serde derive traits are not actually required (KeyManager is never serialized)
+// however we're forced to include them to satisfy inferred trait bounds of generic parameters on
+// GroupState. I'm sure there is a trick to get around this...
```
adzialocha: I believe this can be avoided by implementing the traits manually on the type and not using `derive` on `SpaceState`

sandreae: Ah right, that was it. I can do that :+1:

p2panda-spaces/src/encryption/orderer.rs
```diff
@@ -17,7 +18,7 @@ use crate::types::{ActorId, EncryptionControlMessage, EncryptionDirectMessage, O
 /// dependencies for encryption messages and performs some internal buffering. It does _not_ take
 /// care of ordering of control and application messages; p2panda-spaces expects messages to be
 /// orderer before being processed.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
adzialocha, line 21: Is it the same here? I assume the orderer is also never serialized? This is a bit confusing to read

sandreae: Yes! Same here.

p2panda-spaces/Cargo.toml
```diff
@@ -21,21 +21,26 @@ workspace = true
...
 ]
 
 [dependencies]
+anyhow = { version = "1.0.102", optional = true }
 p2panda-auth = { workspace = true, features = ["serde"], default-features = false }
 p2panda-core = { workspace = true, default-features = false }
 p2panda-encryption = { workspace = true, features = ["data_scheme"], default-features = false }
-p2panda-store = { workspace = true, default-features = false }
+# @TODO: we're using tx_unwrap (which is behind test_utils flag) to fix an issue with unhandled 
+# transaction error return types, remove when we don't need that anymore.
```
adzialocha: Seems to be only used in a test util though? Then we can enable that flag when in test env or test_utils feature is active

adzialocha: And what is the issue you are refering to?

p2panda-encryption/src/key_manager.rs
sandreae: I raised an issue for this, let me know if that isn't actually is the solution your suggesting: https://github.com/p2panda/p2panda/issues/1253

p2panda-spaces/Cargo.toml
sandreae: The issue was handling the transaction errors, but now that we don't use the macro it's possible to do that so no issue.

p2panda-encryption/src/key_manager.rs
adzialocha: I'll tackle that part of this PR directly :+1:

PR #1249, opened 2026-06-17
Return input with error in orderer processor

mycognosist:


p2panda-stream/src/orderer/processor.rs
```diff
@@ -44,27 +44,26 @@ where
 {
     type Output = T;
 
-    type Error = OrdererError<T, ID, S>;
+    type Error = (Option<T>, OrdererError<T, ID, S>);
```
mycognosist, line 47: We don't always have access to `T` in the case of an error so I've wrapped it in `Option` for now.

adzialocha: Oh dammit! This is unfortunately a hard requirement for the pipeline. Let's have it as an Option for now and I see what I can do, it makes sense that we can't keep this around in the buffer.

Issue #965, opened 2026-01-28
Confidential exchange of network id


adzialocha:
The ALPN in TLS 1.3 is transmitted in clear text to another node. Even if we would encrypt it towards the remote party it would still be visible to them and they can from now on use the ALPN.

Ideally we want some sort of private equality testing approach on that level as well, similar to our discovery protocol to allow secure partition of the network via it's network identifier.

adzialocha:
We could hash the ALPN + NetworkId + Our Node Id + Remote Node Id before establishing connection, on the reverse - when accepting the incoming connection - we could re-hash our list of ALPNs and see if one of them is matching to finally forward the connection to the right protocol handler (or reject).

PR #1257, opened 2026-06-22
Add `accept_unmixed` function for registering a protocol under a raw known ALPN

maackle:
Standard `accept` mixes the NetworkId in with the ALPN. In some cases it is desirable to share an ALPN with nodes outside of the NetworkId.

In our use case we need to use `iroh_blobs::Downloader`, which uses a fixed ALPN, and this was a convenient way to hook that up.

If this is unmergeable, I'm happy for this to be a way of uncovering a better way to do this (including getting p2panda-blobs working?), or a moment of decision that we need to maintain this small fork

## 📋 Checklist

### DIFF
--- a/p2panda-net/src/iroh_endpoint/actors/endpoint.rs
+++ b/p2panda-net/src/iroh_endpoint/actors/endpoint.rs
@@ -51,6 +51,12 @@ pub enum ToIrohEndpoint {
     /// Register protocol handler for a given ALPN (protocol identifier).
     RegisterProtocol(ProtocolId, Box<dyn DynProtocolHandler>),
 
+    /// Register a protocol handler under its raw ALPN, *without* mixing in the
+    /// network id. Use this only for protocols that intentionally interoperate
+    /// with vanilla iroh peers (e.g. iroh-blobs' `Downloader`, which dials the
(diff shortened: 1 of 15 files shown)


adzialocha:
Thank you @maackle !

I can see how these methods allow more "hackability" of p2panda which is very useful.

It of course blurs the meaning of a network id and can introduce pitfalls / security issues, especially when we introduce that network id is confidential, see: https://github.com/p2panda/p2panda/issues/965

Maybe we keep this behind the test_utils flag for now? I can also see how the iroh-blobs custom ALPN is easy to change, so maybe this is anyhow of a more "testing" nature for now?

maackle:
I wish that it were easy to change the Downloader ALPN, but alas it is hard-coded https://github.com/n0-computer/iroh-blobs/blob/7cc92c7cf84d4f4dab32704e328551212977d2a2/src/api/downloader.rs#L86

Maybe this should rather be a PR to iroh-blobs? Then I would only need p2panda to expose the mixing function https://github.com/p2panda/p2panda/blob/5d2085592299650d6cad53f159efbf4883d36a24/p2panda-net/src/lib.rs#L318

maackle:
The iroh-blobs fork solution was much more straightforward so I'll close this!

maackle:
see https://github.com/n0-computer/iroh-blobs/pull/242

PR #1258, opened 2026-06-23
Fix faulty configuration for in-memory SQLite which caused dropped databases

### LINKED ISSUE #1255: In-memory SQLite database "deletes" itself after a while, causing "no such table" errors
Initially we've observed this bug in Reflection, when running in Builder: https://github.com/p2panda/reflection/issues/149

.. but now we're also observing it on our bootstrap node which runs for many days and crashes at one point:

```
Jun 19 10:57:14 ubuntu-s-1vcpu-512mb-10gb-fra1-01 bootstrap-node-api[21341]: 2026-06-19T10:57:14.309549Z ERROR bootstrap: calling remove_older_than failed: error returned from database: (code: 1) no such table: node_infos_v1
```

I'm not sure what's going on but it seems like SQLite "rolls back" all (migration) changes to the database and "forgets" about everything?

This only happens when the database runs *in-memory*, this is why we only observed it in Builder and in the bootstrap.

sandreae:
If there are no active connections it is possible that an in-memory SQLite database will be dropped, causing any further queries to fail.

This issue can be avoided if idle_timeout and max_lifetime values are both set to None. This PR offers a recommended configuration for in-memory databases via a new SqliteBuilder::memory() method.

Related issue: https://github.com/p2panda/p2panda/issues/1255

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Highlights are marked with a pancake 🥞
 - spaces: Fix generics in MessageStore and MemoryStore [#1229](https://github.com/p2panda/p2panda/pull/1229)
 - ci: Check for unused mut and async, fix affected methods [#1163](https://github.com/p2panda/p2panda/pull/1163)
 - net: Return existing rx on call to sync manager subscribe [#1269](https://github.com/p2panda/p2panda/pull/1269)
+- store: Faulty configuration for in-memory SQLite which caused dropped databases [#1258](https://github.com/p2panda/p2panda/pull/1258)
 
 ## [0.6.1] - 22/05/2026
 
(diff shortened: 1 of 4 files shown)


adzialocha:
This looks good! I like the API as it's a bit weird that the default is in-memory but with  faulty values, but I think having `SqliteStoreBuilder::memory` as a constructor is the good call so one is not tempted to use `::default`.

I've removed "random_memory_url" as it wouldn't compile and I think we don't need the shared cache anymore (it's been an artefact from aquadoggo days where we used the `AnyPool` and there seems to be bugs around it which this prevented).

Also I've deployed this branch on our future bootstrap node for Reflection, let's see if it does a difference. If it does I think this is ready to merge!

adzialocha (approved):

PR #1261, opened 2026-06-23
Topic reverse lookup

maackle:
In TopicStore:

- Add resolve_topic
- rename resolve to resolve_associations

## 📋 Checklist

### DIFF
--- a/p2panda-store/Cargo.toml
+++ b/p2panda-store/Cargo.toml
@@ -50,6 +50,7 @@ test_utils = [
   "dep:rand",
   "dep:rand_chacha",
   "sqlite",
+  "p2panda-encryption/test_utils",
 ]
 
 [dependencies]
--- a/p2panda-store/src/topics/sqlite.rs
+++ b/p2panda-store/src/topics/sqlite.rs
@@ -5,7 +5,7 @@ use std::collections::BTreeMap;
 use p2panda_core::cbor::{decode_cbor, encode_cbor};
 use p2panda_core::{LogId, VerifyingKey};
 use serde::{Deserialize, Serialize};
-use sqlx::{query, query_as};
+use sqlx::{query, query_as, query_scalar};
 
 use crate::sqlite::{DecodeError, SqliteError, SqliteStore};
 use crate::topics::TopicStore;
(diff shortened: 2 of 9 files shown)


sandreae (changes requested):
Thank you! This looks great, nice query to add. My only request would be to leave `resolve` named as it currently is. I _think_ this will still be the most commonly used API method alongside `associate`, and so keeping the symmetry there is maybe nice. What do you think?

maackle:
I think I agree. I was about to say that association itself is symmetrical (a two-way connection between two things), so it makes sense to distinguish the two directions of resolution. But, the way it's currently named, one of those is `resolve_associations`, which already implies that association is one-way, which is weird. Also, less breaking changes is better. So yeah, I'll change it when I get a moment.

sandreae:
@maackle is this still something you need?

maackle:
@sandreae I just made the update you requested a long time ago. `resolve_topic` is currently in our fork in two places that we may be able to remove with some future refactors, but for now yes we are using it.

p2panda-store/src/topics/sqlite.rs
```diff
@@ -135,4 +135,44 @@ where
 
         Ok(result)
     }
+
+    /// Given a prior association, return the associated topic.
+    async fn resolve_topic(
+        &self,
+        author: &VerifyingKey,
+        data_id: &L,
+    ) -> Result<Option<T>, Self::Error> {
```
sandreae: This method needs to return `Vec<T>` as one author log can be associated with many topics. This is already the case in the spaces integration, where a group log can be required in multiple spaces, which each have their own topic.

sandreae (changes requested):

p2panda-store/Cargo.toml
```diff
@@ -50,6 +50,7 @@ test_utils = [
   "dep:rand",
   "dep:rand_chacha",
   "sqlite",
+  "p2panda-encryption/test_utils",
```
adzialocha, line 53: Oh! Curious why our CI didn't catch this (we check & compile against all feature combinations)

adzialocha (commented):
Looks good!

I would squash this into 2-3 commits for a nice history and prefix each commit message with a tag of the affected crate if you can.

We have fresh contributor guidelines with more details if you're curious: https://github.com/p2panda/.github/blob/main/CONTRIBUTING.md

p2panda-store/Cargo.toml
maackle: FYI the command that surfaced it was:

    cargo test --package p2panda-store --lib --all-features -- topics::tests::resolve_topics_from_association --exact --show-output --nocapture --include-ignored

maackle:
Cleaned it up!

adzialocha (approved):

sandreae (approved):

PR #1262, opened 2026-06-24
Implement pipeline orderer with store for event metadata

mycognosist:
This PR is mainly concerned with introducing ordering to the `p2panda` pipeline.

Here we introduce an orderer over p2panda pipeline events which
implements the `Processor` trait. We've chosen to create a
p2panda-specific orderer to keep the implementation in `p2panda-stream`
generic - especially over the `T` input type.

A new `EventMetadata` type has been introduced; this is the data
(excluding `operation`) which is required to construct an `Event`. This
new type is inserted into the newly-introduced `ProcessorStore` when an
event is passed into the orderer. The metadata is later retrieved from
the store in the `process()` method of the orderer if an operation has
been released. The metadata can then be used to reconstruct the `Event`,
which is expected by the next processor in the pipeline.

We also introduce an internal queue for the orderer processor
which is used to temporarily store events which failed to be processed
by the ingest processor (due to validation failure, for example). These
events should not be inserted into the causal orderer but nonetheless
need to continue flowing through the pipeline. Any queued events are
returned when `next()` is called on the orderer processor.

`Serialize` and `Deserialize` have been derived in a couple of places to
allow the `EventMetadata` to be stored. The signature of the `Ordering`
trait has also be changed to return a `Vec<Hash>` instead of a
`&[Hash]`.

## 📋 Checklist

### DIFF
--- a/p2panda-store/src/lib.rs
+++ b/p2panda-store/src/lib.rs
@@ -131,6 +131,8 @@ pub mod logs;
 mod macros;
 pub mod operations;
 pub mod orderer;
+#[cfg(feature = "processor")]
+pub mod processor;
 #[cfg(feature = "spaces")]
 pub mod spaces;
 #[cfg(feature = "sqlite")]
--- a/p2panda-store/src/processor/traits.rs
+++ b/p2panda-store/src/processor/traits.rs
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::error::Error;
+
+use p2panda_core::Hash;
+
+pub trait ProcessorStore<T> {
+    type Error: Error;
+
+    // TODO: Are we happy for `id` to be a concrete type here?
+    fn get_event(&self, id: &Hash) -> impl Future<Output = Result<Option<T>, Self::Error>>;
+
+    fn set_event(&self, id: &Hash, event: &T) -> impl Future<Output = Result<(), Self::Error>>;
+}
--- a/p2panda-store/src/processor/traits.rs
+++ b/p2panda-store/src/processor/traits.rs
@@ -4,10 +4,11 @@ use std::error::Error;
 
 use p2panda_core::Hash;
 
+// TODO: Is `ProcessorStore` the best name? Maybe `EventStore` or `EventMetadataStore` is more
+// specific and thus clearer?
 pub trait ProcessorStore<T> {
     type Error: Error;
 
-    // TODO: Are we happy for `id` to be a concrete type here?
     fn get_event(&self, id: &Hash) -> impl Future<Output = Result<Option<T>, Self::Error>>;
 
     fn set_event(&self, id: &Hash, event: &T) -> impl Future<Output = Result<(), Self::Error>>;
--- a/p2panda/src/processor/event.rs
+++ b/p2panda/src/processor/event.rs
@@ -6,13 +6,16 @@ use p2panda_core::traits::Digest;
 use p2panda_core::{Body, Hash, Header, LogId, Operation, PruneFlag, SeqNum, VerifyingKey};
 use p2panda_stream::ingest::{IngestArgs, IngestError, IngestResult};
 use p2panda_stream::log_prune::{LogPruneArgs, LogPruneError, LogPruneResult};
+use p2panda_stream::orderer::Ordering;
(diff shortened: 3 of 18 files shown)


p2panda/src/processor/event.rs
```diff
@@ -137,6 +173,12 @@ where
     }
 }
 
+impl<L, E, TP> Ordering<Hash> for Event<L, E, TP> {
+    fn dependencies(&self) -> &[Hash] {
+        todo!()
```
mycognosist: @sandreae Could you point me to where I can retrieve these dependencies? I think you previously mentioned something about `SpacesArgs`.

p2panda-store/src/processor/traits.rs
```diff
@@ -0,0 +1,16 @@
...
+use std::error::Error;
+
+use p2panda_core::Hash;
+
+// TODO: Is `ProcessorStore` the best name? Maybe `EventStore` or `EventMetadataStore` is more
+// specific and thus clearer?
+pub trait ProcessorStore<T> {
+    type Error: Error;
+
+    // TODO: Are we happy for `id` to be a concrete type here?
```
sandreae: Jah, happy 👍

p2panda-store/src/lib.rs
```diff
@@ -131,6 +131,9 @@ pub mod logs;
 mod macros;
 pub mod operations;
 pub mod orderer;
+// TODO: Feature flag; do we want to bundle this with the `spaces` flag or create a separate
```
sandreae: I don't think it should be bundled with spaces as it's not strictly related. Maybe it's a new "processor" flag.

p2panda/src/processor/event.rs
sandreae: Here: https://github.com/p2panda/p2panda/blob/4a4999c90e75415979b8c55db4613021de42117d/p2panda-spaces/src/message.rs#L229

p2panda/src/processor/event.rs
```diff
@@ -37,6 +69,9 @@ pub struct Event<L, E, TP> {
     /// Status of the "ingest" processor.
     pub(crate) ingest: ProcessorStatus<IngestResult, IngestError>,
 
+    /// Status of the "orderer" processor.
+    pub(crate) orderer: ProcessorStatus<(), OrdererError>,
```
sandreae: Why is there no result in this status?

p2panda/src/processor/pipeline.rs
```diff
@@ -130,6 +133,25 @@ where
...
+                            Ok(mut event) => {
+                                event.orderer = ProcessorStatus::Completed(());
+                                event
+                            }
+                            Err((event, err)) => {
+                                if let Some(mut event) = event {
+                                    event.orderer = ProcessorStatus::Failed(err);
+                                    event
+                                } else {
+                                    // TODO: Properly handle error case. I'm not entirely sure what
```
sandreae, line 147: Yeah, this is tricky. The only errors which can occur are critical store or inconsistency errors right? With these maybe it's ok if none of the remaining processors get a chance to perform their task. We would then just need the pipeline to "complete" so I guess we need to pass _something_ to further layers. Even if it's a no-op. We would need later processors to be able to interpret this correctly and just pass the event on silently. The spaces processor already has something like this (the Ignore variant). Not sure how to make that generic over any processor though 🤔

mycognosist: The root issue is that in this failure branch we do not have an `event` to pass along. We could construct a new event (with `event.orderer = ProcessorStatus::Failed(err)`) and set the `log_prune` and `spaces` flags to ensure that those processors ignore the event (`log_prune` would return `LogPruneResult::Noop` and `spaces` would return `SpacesResult::Ignored`). That would result in the silent passing-on you suggested.

mycognosist: This probably requires some changes to `Event` or some kind of wrapper type...because we probably want a way to create an `Event` which only has an `operation_id` (this is potentially the only relevant data we have access to if the store returns `None` for either the operation or the associated event metadata).

mycognosist:
I've opened an issue concerning a possible refactor of the way we're handling events in the orderer processor:

https://github.com/p2panda/p2panda/issues/1263

PR #1264, opened 2026-06-26
p2panda: Topic -> Log associations

sandreae:
This PR introduces associating logs with space topics (as there are now many logs per topic).

## 📋 Checklist

### DIFF
--- a/p2panda-spaces/src/message.rs
+++ b/p2panda-spaces/src/message.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use std::borrow::Borrow;
-use std::fmt::Debug;
+use std::fmt::{Debug};
 
 use p2panda_auth::group::GroupAction;
 use p2panda_auth::traits::Conditions;
@@ -247,6 +247,16 @@ impl<C> SpacesArgs<C> {
             SpacesArgs::Application {
                 space_dependencies, ..
             } => space_dependencies.to_owned(),
+        }
+    }
+
+    pub fn variant(&self) -> String {
+        match self {
+            SpacesArgs::KeyBundle { .. } => "key bundle".to_string(),
+            SpacesArgs::Auth { .. } => "auth group".to_string(),
+            SpacesArgs::SpaceMembership { .. } => "space membership".to_string(),
+            SpacesArgs::SpaceUpdate { .. } => "space update".to_string(),
+            SpacesArgs::Application { .. } => "application".to_string(),
         }
     }
 }
--- a/p2panda-spaces/src/message.rs
+++ b/p2panda-spaces/src/message.rs
@@ -250,7 +250,7 @@ impl<C> SpacesArgs<C> {
         }
     }
 
-    pub fn variant(&self) -> String {
+    pub fn variant_str(&self) -> String {
         match self {
             SpacesArgs::KeyBundle { .. } => "key bundle".to_string(),
             SpacesArgs::Auth { .. } => "auth group".to_string(),
--- a/p2panda-spaces/src/message.rs
+++ b/p2panda-spaces/src/message.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use std::borrow::Borrow;
-use std::fmt::{Debug};
+use std::fmt::Debug;
 
 use p2panda_auth::group::GroupAction;
 use p2panda_auth::traits::Conditions;
--- a/p2panda-spaces/src/utils.rs
+++ b/p2panda-spaces/src/utils.rs
@@ -3,7 +3,6 @@
 use p2panda_auth::Access;
 use p2panda_auth::group::GroupMember;
 use p2panda_auth::traits::Conditions;
-use p2panda_core::{Hash, VerifyingKey};
 
 use crate::types::AuthGroupState;
 use crate::{ActorId, MemberId};
@@ -70,23 +69,4 @@ pub(crate) fn removed_members(
         .collect::<Vec<_>>();
     members.sort();
     members
-}
-
-// TODO: we define and this trait in many crates, let's move it to -core.
-/// Returns a displayable string representing the underlying value in a short format, easy to read
-/// during debugging and logging.
-pub trait ShortFormat {
-    fn fmt_short(&self) -> String;
-}
-
-impl ShortFormat for VerifyingKey {
-    fn fmt_short(&self) -> String {
-        self.to_hex()[0..10].to_string()
-    }
-}
-
-impl ShortFormat for Hash {
-    fn fmt_short(&self) -> String {
-        self.to_hex()[0..5].to_string()
-    }
 }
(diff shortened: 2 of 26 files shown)


p2panda-spaces/src/utils.rs
```diff
@@ -70,3 +71,22 @@ pub(crate) fn removed_members(
     members.sort();
     members
 }
+
+// TODO: we define and this trait in many crates, let's move it to -core.
```
adzialocha: Instead of doing the effort to copy it here I suggest we directly move this to `p2panda-core`

p2panda-spaces/src/message.rs
```diff
@@ -249,4 +249,14 @@ impl<C> SpacesArgs<C> {
             } => space_dependencies.to_owned(),
         }
     }
+
+    pub fn variant(&self) -> String {
```
adzialocha: Maybe name this `variant_display` or `variant_str`, it suggests a more strict type otherwise (enum etc.) but this is more human facing.

p2panda/src/processor/event.rs
```diff
@@ -160,24 +160,25 @@ where
...
+            reasons.push(err.to_owned().into());
         }
 
         if let ProcessorStatus::Failed(err) = &self.spaces {
-            return Some(err.to_owned().into());
+            reasons.push(err.to_owned().into());
         }
 
-        None
+        reasons
```
adzialocha: I don't think we'll ever have more than one error variant soon in a processed event since we're "ejecting" the event on first failure.

This is not implemented yet but since it'll be soon I think we can already account for this "future" here

p2panda/src/processor/pipeline.rs
```diff
@@ -130,6 +131,9 @@ where
                             }
                             Err((mut event, err)) => {
                                 event.ingest = ProcessorStatus::Failed(err);
+                                // As processing the operation failed here we should ignore it in
+                                // the spaces processor.
+                                event.spaces_args = SpacesProcessorArgs::Ignore;
```
adzialocha, line 136: Nice! In my PR I'll have this as a full method which automatically sets all args to "ignore" when setting an error.

p2panda/src/spaces/mod.rs
```diff
@@ -61,3 +64,15 @@ pub(crate) fn to_initial_members(
...
+    LogId::digest(&{
+        let mut bytes = Vec::new();
+        bytes.extend_from_slice(group_id.as_bytes());
+        // The group id would be enough to indicate the log id, we hash it here together
+        // with a constant value to prevent possible collisions with logs of same id but
+        // different purpose.
+        bytes.extend_from_slice(GROUP_CONTROL_MESSAGE);
+        bytes
+    })
+}
```
adzialocha: One .._log_id method is defined here, the others somewhere else (forge), I'd pick one place for all of them (and their constants).

p2panda/src/streams/acked.rs
```diff
@@ -119,10 +119,10 @@ impl Acked {
...
         let header = header.borrow();
 
-        // Make sure we're only acking operations for the given topic.
-        if LogId::from_topic(self.topic) != header.extensions.log_id() {
-            return Err(AckedError::InvalidTopic(self.topic));
-        }
+        // @TODO: what does this check mean now that topic does not directly map to a log id? //
+        // Make sure we're only acking operations for the given topic. if
+        // LogId::from_topic(self.topic) != header.extensions.log_id() { return
+        //     Err(AckedError::InvalidTopic(self.topic)); }
```
adzialocha: Yeah we can remove this, this was mostly a defensive approach to make users of `Acked` aware of potential mistakes.

p2panda/src/streams/stream.rs
```diff
@@ -295,15 +312,15 @@ where
                 //
                 // If channel stopped working because the subscriber got dropped, ignore it as we
                 // still might want to process locally published operations.
-                let _ = app_tx.send(event).await;
+                if let Err(err) = app_tx.send(event).await {
+                    warn!("error sending event to application layer: {err:?}");
+                };
             }
+
+            warn!("stream processing task exited");
```
adzialocha: These warnings might become annoying since they'll be shown on every app shutdown

p2panda/src/streams/stream.rs
```diff
@@ -273,6 +279,17 @@ where
...
+                                // Try pushing operation to other nodes if we have an active and "live" sync session with
+                                // them. This allows disseminating new messages quickly in the network.
+                                //
+                                // If no active live session exists, nodes will pick up the operation later when running the
+                                // sync protocol.
+                                if sync_handle
+                                    .publish(*operation.clone())
+                                    .await.is_err() {
+                                        warn!(session_id = session_id, operation_id = %operation.hash(), "failed sending imported operation on sync channel")
+                                    };
```
adzialocha: Two similar flows (calling publish or import) have slightly different code-paths now for the same functionality (publish to live-mode peers & process), this might need refactoring to avoid bugs in the future, I'd at least add a TODO here.

p2panda/src/streams/stream.rs
```diff
@@ -357,7 +373,8 @@ where
         .await;
 
     if event.is_failed() {
-        let error = event.failure_reason().expect("event has failed");
+        // @TODO: refactor to return all errors here.
+        let error = event.failure_reasons().remove(0);
```
adzialocha: Let's roll back to only having one error type ever, then this TODO (and others) can go

p2panda/src/node.rs
```diff
@@ -424,45 +430,149 @@ impl Node {
...
-        match self.spaces_manager.space(space_id).await? {
-            Some(inner) => {
-                let topic = space_id;
-                let (tx, rx) = self.stream_from::<M>(topic, from).await?;
+        // Associate the space topic with the key bundle logs.
+        //
+        // This does _not_ happen during ingest as at that point there is no topic which could be
+        // used to perform this association. The same key bundle log can be associated with many
+        // spaces.
+        tx!(&self.store, {
```
adzialocha, line 442: Let's look at it again later for transactions and readability, there's some detailed spaces things going on which clutter the general "node" code and are not atomic

p2panda/src/node.rs
```diff
@@ -535,6 +647,11 @@ impl Node {
     pub fn store(&self) -> SqliteStore {
         self.store.clone()
     }
+
+    /// Access the inner spaces manager.
+    pub fn spaces_manager(&self) -> SpacesManager {
+        self.spaces_manager.clone()
+    }
```
adzialocha, line 652: Feels dangerous to expose this to the user, can we put this behind a test_utils flag?

adzialocha (changes requested):
Looking good!

I think nothing is blocking a merge, except reverting the change to introduce more than one error on `Event` as there's no path forwards with it.

I'd also suggest to change the commit tags from "p2panda" to "node", simply because it's consistent with the git history and "p2panda" reads confusing (too broad). Clippy & fmt things can also be merged with the regarding commits. Aaand there seems to be an issue with the feature flags (see CI).

p2panda/src/node.rs
sandreae: It is behind the test_utils flag :+1:

PR #1267, opened 2026-06-29
Unblock task tracker by introducing "pass-through" events coming from orderer

adzialocha:
The current pipeline is designed in a strictly symmetric manner: One input event causes exactly one output event. With the introduction of an orderer and -spaces this changes significantly: One single event can result in nothing (because it got buffered by the orderer) or into many events (because buffered operations got freed, etc.).

The task tracker is the one enforcing this strict 1:1 rule. On top it's enforcing strict ordering of events, things are not allowed to be processed in parallel.

This design is blocking because the task tracker is waiting for a task to finish before it allows the new task to come in. In this PR we're fixing this simply by introducing a "noop" event coming from the orderer. So _something_ is informing the task tracker that it is "done".

The other scenario requires a larger refactoring of the task tracker and is handled in a second PR.

## 📋 Checklist

### DIFF
--- a/p2panda/src/processor/orderer.rs
+++ b/p2panda/src/processor/orderer.rs
@@ -1,22 +1,41 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use std::borrow::Borrow;
 use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::marker::PhantomData;
 
 use p2panda_core::traits::Digest;
-use p2panda_core::{Hash, LogId, Operation};
+use p2panda_core::{Extensions, Hash, Operation};
 use p2panda_store::Transaction;
 use p2panda_store::operations::OperationStore;
 use p2panda_store::orderer::OrdererStore;
 use p2panda_store::processor::ProcessorStore;
 use p2panda_stream::Processor;
-use p2panda_stream::orderer::{CausalOrderer, Ordering};
+use p2panda_stream::orderer::CausalOrderer;
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio::sync::{Mutex, Notify};
 
-use crate::processor::ProcessorStatus;
-use crate::processor::event::{Event, EventMetadata};
+pub trait OrdererMetadata<E> {
+    /// Metadata attached to an input event we want to persist in database next to ordering info.
+    type Metadata: Serialize + for<'de> Deserialize<'de>;
+
+    /// Extract metadata from input.
+    fn metadata(&self) -> Self::Metadata;
+
+    /// Re-construct input from operation and persisted metadata.
+    fn from_operation(operation: Operation<E>, meta: Self::Metadata) -> Self;
+}
+
+#[derive(Clone, Default, Debug)]
+pub enum OrdererArgs {
+    Process {
+        dependencies: Vec<Hash>,
+    },
+    #[default]
+    Ignore,
+}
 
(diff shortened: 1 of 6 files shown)


p2panda/src/processor/orderer.rs
```diff
@@ -1,39 +1,58 @@
...
 use p2panda_stream::Processor;
-use p2panda_stream::orderer::{CausalOrderer, Ordering};
+use p2panda_stream::orderer::CausalOrderer;
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio::sync::{Mutex, Notify};
 
-use crate::processor::ProcessorStatus;
-use crate::processor::event::{Event, EventMetadata};
+pub trait OrdererMetadata<E> {
```
mycognosist, line 20: Nice touch to separate this into a trait. It makes things neater in `process()` and `next()`.

p2panda/src/processor/orderer.rs
```diff
@@ -135,29 +161,22 @@ where
...
-                return Ok((event, OrdererResult::Processed));
+                return Ok((T::from_operation(operation, metadata), OrdererResult::Ready));
             }
 
             self.store
                 .commit(permit)
                 .await
                 .map_err(|err| (None, OrdererError::Transaction(err.to_string())))?;
 
+            drop(inner);
```
mycognosist, line 178: Was the lack of this `drop` causing some additional blocking behaviour?

mycognosist (approved):

p2panda/src/processor/orderer.rs
adzialocha: It was actually fine without it because we assure an eventual return (which will drop the lock) in the loop. I just thought to add it to be a bit more "cooperative" in the runtime in case the loop is working hard.

p2panda/src/processor/pipeline.rs
```diff
@@ -131,25 +123,26 @@ where
...
-                                event
+                                event.noop()
                             }
                         })
                         .layer(orderer)
                         .map(|result| match result {
                             Ok((mut event, result)) => {
                                 event.orderer = ProcessorStatus::Completed(result);
-                                event
+
```
sandreae, line 133: Just so I'm clear about what this change means, let me sketch a scenario from the perspective of the orderer processor:

process: Event 3 dependencies Event 2
next: Event 3 (no-op)
process: Event 2 dependencies Event 1
next: Event 2 (no-op)
process: Event 1
next: Event 1 (ready)
next: Event 2 (ready) <- will next be polled to release this event?
next: Event 3 (ready) <- will next be polled to release this event?

adzialocha: > next: Event 2 (ready) <- will next be polled to release this event?
> next: Event 3 (ready) <- will next be polled to release this event?

Yes, they will be polled. This is taken care of here in `p2panda-stream`: https://github.com/p2panda/p2panda/blob/main/p2panda-stream/src/processors/buffered.rs#L30-L46 and the reason why the test passes.

The problem is still that the task tracker can't account for one event freeing more than one, so event 2 and 3 will never be forwarded to the topic stream. It's a different issue, independent from the pipeline. This is what I'll fix in the next PR ✊🏻

sandreae: Ok, great, that maps with how I understood the processor stream worked, thanks for clarifying.

sandreae (approved):

Issue #1263, opened 2026-06-25
Consider using internal queue for events in orderer processor


mycognosist:
In https://github.com/p2panda/p2panda/pull/1262 we introduce a `ProcessorStore` which is used to store and retrieve metadata associated with events which flow through the `p2panda` processor pipeline.

Can we do away with the store and rather use a queue on `Orderer` to push the events in `process()` and pop them in `next()`? What are the implications for crash scenarios and replays? How does this interact with the inner `CausalOrderer` and any stores?

adzialocha:
> What are the implications for crash scenarios and replays?

Good point! I think there's two general design paths we can go with this and both have implications for replays:

The general flow is:

1. Pipeline receives out-of-order event A
2. Event is moved into buffer because it is ooo
3. Pipeline receives in-order event B, freeing event A
4. Event B is finished by pipeline, application acks Event B
5. Event A is finished by pipeline, application acks Event A

If the application restarts / crashes after step 2, it depends on the nature of the buffer how we can recover from that situation:

a) **In-memory buffer for events:** Pipeline needs to receive event A _again_, so some external replay logic would need to check which event was un-acked and then give it to the pipeline again
b) **Persisted buffer for events:** Event A is already in the buffer, we can just continue as Event B will free it and everything will move on as usual. We need to prevent now that Event A is being replayed as it wasn't acked yet, this will need some check if it's already in the buffer.

An in-memory version might be simpler to reason about & implement but it'll imply that _every_ out-of-order event needs to be replayed on restart. Depending on the ooo-ness of the data and amount this can cause a big lag before we're getting to actually interesting events for the application layer. Another problem is that all events need to arrive first (via calling `process`) _before_ we can start calling `next` to populate the "id -> events map", otherwise the original event is missing. Currently this is not possible as we're [calling both of them in any order](https://github.com/p2panda/p2panda/blob/main/p2panda-stream/src/processors/buffered.rs#L30-L46).

adzialocha:
Some updates based on what I've learned doing the implementation now and revisions of my comment above:

> Another problem is that all events need to arrive first (via calling process) before we can start calling next to populate the "id -> events map", otherwise the original event is missing. Currently this is not possible as we're [calling both of them in any order](https://github.com/p2panda/p2panda/blob/main/p2panda-stream/src/processors/buffered.rs#L30-L46).

[...]

PR #1271, opened 2026-06-30
Allow event processing to handle out-of-order buffering by separating i/o streams and preserve input ordering

adzialocha:
Related PR: https://github.com/p2panda/p2panda/pull/1267

Closes: #1263 

## 📋 Checklist

### DIFF
--- a/p2panda/src/processor/orderer.rs
+++ b/p2panda/src/processor/orderer.rs
@@ -39,8 +39,14 @@ pub enum OrdererArgs {
 
 #[derive(Copy, Clone, Debug)]
 pub enum OrdererResult {
-    Ready,
+    /// Item was buffered and is now in "pending" state.
     Pending,
+
+    /// Item was buffered by orderer and is now marked as "ready" to be finally forwarded in correct
+    /// order.
+    Ready,
+
+    /// Item was ignored as specified in input arguments.
     Ignored,
 }
 
@@ -127,7 +133,7 @@ where
 
     async fn next(&self) -> Result<Self::Output, Self::Error> {
         loop {
-            // First check to see if there are any ignored events which can be returned.
+            // First check to see if there are any ignored or pending events which can be returned.
             if let Some((event, result)) = self.queue.borrow_mut().pop_front() {
                 return Ok((event, result));
             }
--- a/p2panda/src/processor/orderer.rs
+++ b/p2panda/src/processor/orderer.rs
@@ -37,14 +37,19 @@ pub enum OrdererArgs {
     Ignore,
 }
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Clone, Debug)]
 pub enum OrdererResult {
-    /// Item was buffered and is now in "pending" state.
+    /// Item has unmet dependencies, is buffered and now in "pending" state.
     Pending,
 
-    /// Item was buffered by orderer and is now marked as "ready" to be finally forwarded in correct
-    /// order.
-    Ready,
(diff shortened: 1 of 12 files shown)


mycognosist (approved):

p2panda/src/processor/pipeline.rs
```diff
@@ -58,14 +58,22 @@ const PUBLISH_BUFFER_SIZE: usize = 128;
...
 pub struct Pipeline<L, E, TP> {
     to_pipeline_tx: mpsc::Sender<Event<L, E, TP>>,
     from_pipeline_queue: Arc<Mutex<VecDeque<Event<L, E, TP>>>>,
     from_pipeline_notify: Arc<Notify>,
     tasks: TaskTracker<Event<L, E, TP>, Hash>,
+    // Re-using a pipeline across streams can lead to undesirable effects such as a) receiving
+    // unwanted output events which were not intended for the topic stream b) broadcast channel
+    // designs dropping events when running full. This !Clone marker should hopefully make you think
+    // twice about re-using pipelines.
+    _marker: NotClone,
```
mycognosist, line 71: That's a neat solution 😊

adzialocha: Unfortunately had to make the whole thing `Clone` again in the end to allow re-use of the pipeline within the same topic stream :-o

But yeah, markers are useful for many things (not only marking generics), can use them to make types !Send or !Clone, etc. ✊🏻

p2panda/src/processor/orderer.rs
```diff
@@ -56,10 +61,20 @@ impl OrdererResult {
     }
 }
 
+// TODO: Move this back to p2panda-stream.
 pub struct Orderer<S, T, E> {
     inner: Mutex<CausalOrderer<Hash, S>>,
     store: S,
     notify: Notify,
+    // We're keeping an unbound, in-memory buffer of all "freed" events here. This means that if an
```
sandreae, line 69: There's a case relating to crash resilience replays I was considering when looking into this last week and I wonder if it's addressed here. It's not totally clear to me yet so I'll sketch it out to be sure. 

If we have operations in a dependency chain like so A <- B <- C and the following occurs:

process C (not ready)
orderer stores C in db
process B (not ready)
orderer stores B in db
process A (ready)
orderer stores A in db and marks A, B & C as "dependencies met" 
push A to out queue 
push B to out queue
push C to out queue
==== CRASH ====

=== RESTART ===
retrieve not acked ops from store
process C (ready?) <-- the orderer thinks this item has it's dependencies met

sandreae: The issue I can see is that from the perspective of the causal orderer (backed by SQLite database) then item C has had it's dependencies met, but actually A & C were never forwarded to the user yet.

sandreae: Let me know if I'm missing something 👍

sandreae (commented):
Great PR, thanks for the clear commit messages. The solution looks excellent, and addresses a lot of points I was scratching my head over last week. There's one last situation which I'd like to discuss, I commented about this and lets see how it is. Apart from that then it's all good to go.

p2panda/src/processor/orderer.rs
adzialocha: > .. if it's addressed here. 

I haven't addressed any crash resiliance related work here in this PR (this fixes mostly the blocking pipeline issue due to missing i/o separation).

Let's move your question into an GH issue?

sandreae (approved):

PR #1276, opened 2026-07-01
Handle spaces events when processing pipeline event

mycognosist:
Here we modify the `process_operation()` function to handle spaces events which are emitted from the processor pipeline. Since the processing of a single spaces operation may yield multiple events, the function now returns a `Vec<StreamEvent<M>>`. The `processed_stream()` function has been updated to send each of the returned events up to the application layer.

Application messages which were encrypted (via spaces) or unencrypted are both returned as the same `StreamEvent::Processed` variant. System-level spaces operations which create or modify a space are returned as the `StreamEvent::Space` variant.

Addresses: https://github.com/p2panda/p2panda/issues/1256

## 📋 Checklist

### DIFF
--- a/p2panda/src/spaces/space.rs
+++ b/p2panda/src/spaces/space.rs
@@ -223,7 +223,12 @@ where
     }
 }
 
-#[derive(Debug)]
+// TODO: Can we remove this type in favour of `SpaceEvent` from p2panda-spaces?
+//
+// The only difference is that the `SpaceEvent` does not have a `Processed` variant. In the stream
+// we use `StreamEvent::Processed` for returning operations which came from spaces _and_ non-spaces
+// contexts.
+#[derive(Clone, Debug)]
 pub enum SpaceEvent<M> {
     Processed {
         operation: Box<ProcessedOperation<M>>,
--- a/p2panda/src/streams/stream.rs
+++ b/p2panda/src/streams/stream.rs
@@ -15,7 +15,10 @@ use p2panda_core::traits::Digest;
 use p2panda_core::{Hash, Topic, VerifyingKey};
 use p2panda_net::NodeId;
 use p2panda_net::sync::SyncHandle;
+// TODO: Replace with ShortFormat from p2panda-core.
+// See: https://github.com/p2panda/p2panda/issues/1270
 use p2panda_net::utils::ShortFormat;
+use p2panda_spaces::{GroupEvent, SpaceEvent};
 use p2panda_store::SqliteStore;
 use p2panda_store::operations::OperationStore;
 use p2panda_stream::spaces::SpacesResult;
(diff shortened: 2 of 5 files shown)


p2panda/src/streams/stream.rs
```diff
@@ -382,85 +388,124 @@ where
...
+                                source: source.clone(),
+                            });
+                        }
+                        Err(error) => forward_events.push(StreamEvent::DecodeFailed {
+                            event: event.clone(),
+                            error,
+                        }),
+                    }
+                }
+                p2panda_spaces::Event::KeyBundle { author: _ } => {
```
sandreae: Is it possible to already forward key bundle events? I think this would be useful already for testing and the user.

p2panda/src/streams/stream.rs
```diff
@@ -382,85 +388,124 @@ where
...
+                    // For now we choose to ignore key bundle events; the user is not notified of
+                    // these.
+                }
+                // A group has been created or modified.
+                p2panda_spaces::Event::Group(_group_event) => {
+                    // For now choose to ignore group events; the user is not notified of these.
```
sandreae: I know we said we wouldn't forward these for now (just to keep this PR simple) but is it much work to do that? If it's just a case of adding a variant and forwarding the whole event, maybe we go for it straight away. Again, very good for testing and actually maybe quite useful for the users (then they can wait for a group to be _known_ to the space before then adding it).

sandreae: The "harder" thing related to group events is that one can await an event stream on Group itself, I don't think it's clear how this should be handled yet (and we shouldn't tackle it for now).

p2panda/src/streams/stream.rs
```diff
@@ -962,6 +1007,9 @@ pub enum StreamEvent<M> {
         event: Event<LogId, Extensions, Topic>,
         error: Arc<AckedError>,
     },
+
```
sandreae, line 956: I think we could just add a `KeyBundle` and `Group` variant here already maybe?

sandreae (approved):
This looks really good to me 👍 

I had a couple of questions about adding other event variants already, I really don't mind if that happens in this PR (if it's super easy) or a follow-up.

p2panda/src/spaces/space.rs
```diff
@@ -223,7 +223,12 @@ where
     }
 }
 
-#[derive(Debug)]
+// TODO: Can we remove this type in favour of `SpaceEvent` from p2panda-spaces?
+//
+// The only difference is that the `SpaceEvent` does not have a `Processed` variant. In the stream
+// we use `StreamEvent::Processed` for returning operations which came from spaces _and_ non-spaces
+// contexts.
```
adzialocha, line 230: If we decide to nest space events into stream events, then we could just use the type from p2panda-spaces. The downside here is a "spaces" variant which is not used by regular topic streams.

If we decide to flatten all stream event types (operation, sync started, failures, etc.) into `SpaceEvent` then it makes sense to just have our own, to avoid the nesting.

I'm open for any of these two directions.

p2panda/src/streams/stream.rs
```diff
@@ -962,6 +1007,9 @@ pub enum StreamEvent<M> {
...
+    /// Space has been created or modified.
+    Space { event: SpaceEvent },
```
adzialocha: Maybe use positional variant and not named if its just one field

adzialocha (approved):

PR #1277, opened 2026-07-01
Task for repairing spaces

### LINKED ISSUE #1279: Forward missing groups messages when repair occured in space repair task
Currently we only forward missing groups messages when repair didn't occur.

sandreae:
This change introduces a task which is dedicated to "repairing" spaces which we are subscribed to. Repairing a space involves incorporating missing groups operations observed on the global groups context but not yet published into the space.

There are 3 steps to this process:

1) re-publish missing groups operations into the space topic
2) create and publish space membership operations for each missing groups operation (only read members can do this)
3) associate missing groups logs with the space topic

All new messages will be sent into the topic stream to be processed and forwarded to other peers.

## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/group/crdt/mod.rs
+++ b/p2panda-auth/src/group/crdt/mod.rs
@@ -403,8 +403,11 @@ where
         self.inner.members(group_id)
     }
 
-    /// All groups which have ever been created on this groups state.
-    pub fn seen_groups(&self) -> Vec<ID> {
+    /// All groups which exist in the groups context.
+    /// 
+    /// This includes groups which have no parent -> child relation to each other, as opposed to
+    /// the groups() method which traverses all sub-groups of a single parent.
+    pub fn groups_global(&self) -> Vec<ID> {
         self.inner
             .operations
             .values()
--- a/p2panda-auth/src/group/crdt/mod.rs
+++ b/p2panda-auth/src/group/crdt/mod.rs
@@ -403,6 +403,21 @@ where
         self.inner.members(group_id)
     }
 
+    /// All groups which have ever been created on this groups state.
+    pub fn seen_groups(&self) -> Vec<ID> {
+        self.inner
+            .operations
+            .iter()
+            .filter_map(|(_, message)| {
+                if let GroupAction::Create { .. } = message.action() {
+                    Some(message.group_id())
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
     /// Get all transitive groups inside a group.
     ///
     /// This method recurses into all sub-groups and returns a resolved list of nested group members
--- a/p2panda-auth/src/group/crdt/mod.rs
+++ b/p2panda-auth/src/group/crdt/mod.rs
@@ -407,8 +407,8 @@ where
     pub fn seen_groups(&self) -> Vec<ID> {
         self.inner
             .operations
-            .iter()
-            .filter_map(|(_, message)| {
+            .values()
+            .filter_map(|message| {
                 if let GroupAction::Create { .. } = message.action() {
                     Some(message.group_id())
                 } else {
--- a/p2panda-auth/src/group/crdt/mod.rs
+++ b/p2panda-auth/src/group/crdt/mod.rs
@@ -404,7 +404,7 @@ where
     }
 
     /// All groups which exist in the groups context.
-    /// 
+    ///
     /// This includes groups which have no parent -> child relation to each other, as opposed to
     /// the groups() method which traverses all sub-groups of a single parent.
     pub fn groups_global(&self) -> Vec<ID> {
--- a/p2panda-spaces/src/group.rs
+++ b/p2panda-spaces/src/group.rs
@@ -65,7 +65,7 @@ where
     C: Conditions,
     RS: AuthResolver<C>,
 {
-    pub(crate) fn new(manager_ref: Manager<S, F, C, RS>, id: GroupId) -> Self {
+    pub fn new(manager_ref: Manager<S, F, C, RS>, id: GroupId) -> Self {
         Self {
             manager: manager_ref,
             id,
--- a/p2panda-spaces/src/group.rs
+++ b/p2panda-spaces/src/group.rs
@@ -65,7 +65,7 @@ where
     C: Conditions,
     RS: AuthResolver<C>,
 {
-    pub fn new(manager_ref: Manager<S, F, C, RS>, id: GroupId) -> Self {
+    pub(crate) fn new(manager_ref: Manager<S, F, C, RS>, id: GroupId) -> Self {
         Self {
             manager: manager_ref,
             id,
--- a/p2panda/src/processor/pipeline.rs
+++ b/p2panda/src/processor/pipeline.rs
@@ -207,10 +207,8 @@ where
                         LogPrune::<SqliteStore, Event<L, E, TP>, L, E>::new(store.clone());
 
                     let spaces_store = SqliteSpacesStore::new(store);
-                    let spaces = SpacesProcessor::<Event<L, E, TP>>::new(
-                        spaces_store.clone(),
-                        spaces_manager.clone(),
-                    );
+                    let spaces =
+                        SpacesProcessor::<Event<L, E, TP>>::new(spaces_store, spaces_manager);
 
                     // Receive incoming events through mpsc channel.
                     let pipeline = ReceiverStream::new(to_pipeline_rx)
--- a/p2panda/src/processor/pipeline.rs
+++ b/p2panda/src/processor/pipeline.rs
@@ -198,15 +198,18 @@ where
                 let local = LocalSet::new();
 
                 local.spawn_local(async move {
+                    let me = spaces_manager.id();
                     // Prepare event processing pipeline.
                     let ingest =
                         Ingest::<SqliteStore, Event<L, E, TP>, L, E, TP>::new(store.clone());
                     let orderer = Orderer::<SqliteStore, Event<L, E, TP>, E>::new(store.clone());
                     let log_prune =
                         LogPrune::<SqliteStore, Event<L, E, TP>, L, E>::new(store.clone());
+
+                    let spaces_store = SqliteSpacesStore::new(store);
                     let spaces = SpacesProcessor::<Event<L, E, TP>>::new(
-                        SqliteSpacesStore::new(store),
-                        spaces_manager,
+                        spaces_store.clone(),
+                        spaces_manager.clone(),
                     );
 
                     // Receive incoming events through mpsc channel.
@@ -274,6 +277,7 @@ where
                         // stuck here forever.
                         if let Some(err) = output_event.failure_reason() {
                             warn!(
+                                me = me.fmt_short(),
                                 id = %output_event.hash().fmt_short(),
                                 "failed processing event: {}",
                                 err
--- a/p2panda/src/spaces/forge.rs
+++ b/p2panda/src/spaces/forge.rs
@@ -2,7 +2,6 @@
 
 use p2panda_auth::group::GroupAction;
 use p2panda_core::cbor::encode_cbor;
-use p2panda_core::traits::Digest;
 use p2panda_core::{Hash, Topic, VerifyingKey};
 use p2panda_store::operations::OperationStore;
 use p2panda_store::topics::TopicStore;
@@ -236,38 +235,12 @@ pub(crate) async fn make_space_group_log_associations(
     };
 
     // Associate all group logs for members introduced by this operation.
-    let Some(SpacesArgs::Auth { group_action, .. }) =
-        groups_operation.header.extensions.spaces_args()
-    else {
-        return Err(SpacesForgeError::MissingGroupsArgs(groups_operation.hash()));
-    };
-
-    // @TODO: Maybe it's enough to only associate the group log on group creation?
-    let sub_groups = match group_action {
-        GroupAction::Create { initial_members } => initial_members
-            .into_iter()
-            .filter_map(|(member, _)| {
-                if member.is_group() {
-                    Some(member.id())
-                } else {
-                    None
-                }
-            })
-            .collect(),
-        GroupAction::Add { member, .. } => {
-            if member.is_group() {
-                vec![member.id()]
-            } else {
-                vec![]
-            }
-        }
-        // We don't handle removals here, this is a concern of a higher layer which may even
-        // require consensus.
-        _ => vec![],
-    };
-
-    // For every new sub-group in a space group associate the logs with this space.
-    for group_id in sub_groups {
+    if let Some(SpacesArgs::Auth {
+        group_id,
+        group_action: GroupAction::Create { .. },
+        ..
+    }) = groups_operation.header.extensions.spaces_args()
+    {
         // Every author maintains their own log of control messages _per_ group.
         let log_id = group_log_id(group_id);
 
@@ -288,7 +261,7 @@ pub(crate) async fn make_space_group_log_associations(
         store
             .associate(&Topic::from(space_id), &me, &log_id)
             .await?;
-    }
+    };
 
     let log_id = group_log_id(space_group_id);
     debug!(
--- a/p2panda/src/spaces/forge.rs
+++ b/p2panda/src/spaces/forge.rs
@@ -98,7 +98,7 @@ impl p2panda_spaces::Forge<AuthCapabilities> for OperationForge {
         // TODO: Do we need to query graph tips here for causal ordering or is this taken care off
         // by -spaces? If yes, are declaring _all_ dependencies really it's concern or only the ones
         // which are relevant to the spaces protocol?
-        let operation = match args.clone() {
+        let operation = match args {
             // 1. Key Bundle logs.
             p2panda_spaces::SpacesArgs::KeyBundle { ref key_bundle } => {
                 // TODO: Check actual encoding format of key bundle. Will require versioning.
@@ -274,7 +274,7 @@ pub(crate) async fn make_space_group_log_associations(
         debug!(
             topic = space_id.fmt_short(),
             group_id = group_id.fmt_short(),
-            log_ig = Hash::from(log_id.as_bytes()).fmt_short(),
+            log_id = Hash::from(log_id.as_bytes()).fmt_short(),
             "associate group log with space topic"
         );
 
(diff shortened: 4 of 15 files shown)


sandreae:
Some tests still intermittently blocking / failing because of missing ordering features for both spaces messages and key bundles.

p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -403,6 +403,21 @@ where
...
+            .values()
+            .filter_map(|message| {
+                if let GroupAction::Create { .. } = message.action() {
+                    Some(message.group_id())
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
```
adzialocha: What is the difference between this and the `groups` method?

p2panda-spaces/src/group.rs
```diff
@@ -65,7 +65,7 @@ where
     C: Conditions,
     RS: AuthResolver<C>,
 {
-    pub(crate) fn new(manager_ref: Manager<S, F, C, RS>, id: GroupId) -> Self {
```
adzialocha, line 68: Exposing this method on this level shouldn't be necessary (and is pretty low-level). We can retreive the group instance by looking it up ourselves in `p2panda-spaces`

See: https://github.com/p2panda/p2panda/pull/1271/changes/a938992182eef0fbe84971e10effba5146ebe34b

p2panda/src/spaces/space.rs
```diff
@@ -178,6 +171,36 @@ where
         Ok(result)
     }
 
+    pub async fn repair(&self) -> Result<SpaceFuture, SpaceError> {
```
adzialocha: I'd be up for not exposing this method to the high-level API

p2panda/src/spaces/space.rs
```diff
@@ -281,6 +304,9 @@ pub enum SpaceError {
 
     #[error(transparent)]
     Store(#[from] SqliteError),
+
+    #[error(transparent)]
+    Recv(#[from] RecvError),
```
adzialocha: Could make that error more meaningful for the user ("couldn't process spaces change due to broken channel")

p2panda/src/processor/pipeline.rs
```diff
@@ -102,15 +102,18 @@ where
...
                     let orderer = Orderer::<SqliteStore, Event<L, E, TP>, E>::new(store.clone());
                     let log_prune =
                         LogPrune::<SqliteStore, Event<L, E, TP>, L, E>::new(store.clone());
+
+                    let spaces_store = SqliteSpacesStore::new(store);
                     let spaces = SpacesProcessor::<Event<L, E, TP>>::new(
-                        SqliteSpacesStore::new(store),
-                        spaces_manager,
+                        spaces_store.clone(),
+                        spaces_manager.clone(),
```
adzialocha: Redundant clones

p2panda/src/spaces/forge.rs
```diff
@@ -97,7 +98,7 @@ impl p2panda_spaces::Forge<AuthCapabilities> for OperationForge {
         // TODO: Do we need to query graph tips here for causal ordering or is this taken care off
         // by -spaces? If yes, are declaring _all_ dependencies really it's concern or only the ones
         // which are relevant to the spaces protocol?
-        let operation = match args {
+        let operation = match args.clone() {
```
adzialocha: Is it possible to remove this clone here?

p2panda/src/spaces/forge.rs
```diff
@@ -224,39 +223,26 @@ fn group_log_id(group_id: VerifyingKey) -> LogId {
...
+    // Associate all group logs for members introduced by this operation.
     let Some(SpacesArgs::Auth { group_action, .. }) =
-        auth_operation.header.extensions.spaces_args()
+        groups_operation.header.extensions.spaces_args()
     else {
-        return Err(SpacesForgeError::MissingGroupsArgs(groups_message_id));
+        return Err(SpacesForgeError::MissingGroupsArgs(groups_operation.hash()));
     };
 
+    // @TODO: Maybe it's enough to only associate the group log on group creation?
```
adzialocha: For data encryption without removal of secrets this should be sufficient

p2panda/src/spaces/forge.rs
```diff
@@ -304,10 +290,33 @@ pub(crate) async fn make_group_topic_associations(
             .await?;
     }
 
+    let log_id = group_log_id(space_group_id);
+    debug!(
+        topic = space_id.fmt_short(),
+        group_id = space_group_id.fmt_short(),
+        log_ig = Hash::from(log_id.as_bytes()).fmt_short(),
```
adzialocha: typo: `log_ig` -> `log_id`

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,211 @@
...
+
+use crate::operation::{Extensions, Operation};
+use crate::spaces::group_log_id;
+use crate::spaces::types::{AuthCapabilities, SpacesArgs};
+use crate::spaces::{SpacesManagerError, types::SpacesManager};
+use crate::streams::ExternalStreamFuture;
+
+const REPAIR_DELAY_SECS: u64 = 1;
+
+// Repairing a space involves incorporating missing groups operations observed on the global groups
```
adzialocha: Could be doc-string -> `///`

p2panda/src/streams/stream.rs
```diff
@@ -321,6 +330,9 @@ where
                 // still might want to process locally published operations.
                 let _ = app_tx.send(event).await;
             }
+
+            // Abort the repair task.
+            repair_task_handle.abort();
```
adzialocha, line 376: Good you've prepared this already.

Just noting that it'll not be called for now until https://github.com/p2panda/p2panda/issues/1272 is fixed and if you decide to use `abort` (instead of a graceful shutdown) then it might never be called.

p2panda/src/streams/stream.rs
```diff
@@ -142,8 +144,15 @@ where
         .await
         .map_err(|err| CreateStreamError(err.to_string()))?;
 
+    // Task concerned with repairing a space.
+    let repair_task_handle = spawn_repair_task(
+        topic,
+        spaces_manager.clone(),
+        store.clone(),
+        import_tx.clone(),
```
adzialocha, line 159: We really love our import function 😆 

We definitely need a "silent" publish method next.

p2panda/tests/spaces.rs
```diff
@@ -116,6 +186,7 @@ async fn spaces_sync() -> Result<(), Box<dyn std::error::Error>> {
 
     // penguin also receives the message.
     // @TODO: currently fails because of operation decoding bug.
+    println!("await penguin receive");
```
adzialocha: Remove println's

p2panda/tests/spaces.rs
```diff
@@ -63,13 +137,6 @@ async fn spaces_sync() -> Result<(), Box<dyn std::error::Error>> {
 
     use p2panda::Topic;
     use p2panda_auth::AccessLevel;
-    use serde::{Deserialize, Serialize};
-
-    #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-    struct SecretData {
-        title: String,
-        content: String,
-    }
```
adzialocha, line 72: We can also use other data-types for `M` 😆

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,211 @@
...
+// There are 3 steps to this process:
+//
+// 1) re-publish missing groups operations into the space topic
+// 2) create and publish space membership operations for each missing groups operation (only read
+//    members can do this)
+// 3) associate missing groups logs with the space topic
+//
+// All new messages will be sent into the topic stream to be processed and forwarded to other peers.
+pub(crate) async fn repair_space(
+    topic: Topic,
```
adzialocha: Could pass in a space id directly here, because this method is concerned with "spaces"

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,211 @@
...
+        return Ok(false);
+    }
+
+    // Attempt to repair the space. As we pass in an array containing a single space id there will
+    // be only ever max one result returned.
+    //
+    // @TODO: This method uses transactions internally (eg. in the Forge) and so we can't make
+    // everything part of one transaction on this level yet. It isn't a source of bugs though so
+    // for now this is ok.
+    let mut repaired = manager.repair_spaces(&[space_id]).await?;
```
adzialocha: Yes, this should be possible to fix by making the spaces forge independent of the operation forge underneath and have it's own way of doing transactions

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    // Send the resulting operations into the stream.
+    let stream = Box::pin(futures_util::stream::iter(operations));
+    let (ready_tx, _ready_rx) = oneshot::channel::<ExternalStreamFuture>();
+    import_tx
+        .send((stream, ready_tx))
+        .await
+        .map_err(|err| RepairError::SendToProcessor(err.to_string()))?;
+
+    // Don't await processing otherwise we'd block the stream.
```
adzialocha: You can import the events after you commit the transaction I believe. As long as all operations have been associated, they'll be replayed on a crash.

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,211 @@
...
+        };
+
+        // Ignore non-groups operations.
+        let Some(SpacesArgs::Auth {
+            group_id,
+            group_action,
+            ..
+        }) = operation.header.extensions.spaces_args()
+        else {
+            warn!("expected auth groups operation");
```
adzialocha, line 100: Can this ever happen that we observe non-auth messages coming from the groups state?

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,211 @@
...
+            node_id = manager.id().fmt_short(),
+            space_id = space_id.fmt_short(),
+            "space not yet materialised"
+        );
+        return Ok(false);
+    };
+
+    let permit = store.begin().await?;
+
+    // Collect all groups operations in our global context that the space is missing.
```
adzialocha: I'm not sure what the difference between this and the previous code is. Maybe more comments can help to understand.

I see `repair_spaces` doing the seemingly same thing (identify missing group operations, forge spaces membership messages, associate groups).

p2panda/src/spaces/repair.rs
adzialocha: Maybe we're fine here though: The spaces membership messages forged here are already correctly associated inside the spaces forge (there's tx's bugs there as well but we ignore them for now on this layer). If the application crashes afterwards we'd replay them and still apply all the changes.

If the application crashes _before_ calling repair_spaces that's also okay, the task will repeat.

adzialocha: I'm not 100% sure what the other steps are doing, if they are crucial then we need tx's of course.

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,211 @@
...
+    )>,
+) -> JoinHandle<()> {
+    tokio::spawn(async move {
+        loop {
+            sleep(Duration::from_secs(REPAIR_DELAY_SECS)).await;
+            if let Err(err) = repair_space(topic, &manager, &store, &import_tx).await {
+                warn!("failed to repair spaces: {}", err);
+            }
+        }
+    })
```
adzialocha, line 228: Cool, this should do the job already. Later we can add a throttle to make it snappier / react faster to changes.

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,211 @@
...
+    Store(#[from] SqliteError),
+
+    #[error(transparent)]
+    SpacesManager(#[from] SpacesManagerError),
+
+    #[error("could not send to processor pipeline: {0}")]
+    SendToProcessor(String),
+
+    #[error(transparent)]
+    Recv(#[from] RecvError),
```
adzialocha: Similar here that error type could say more

adzialocha (changes requested):
This is mostly ready to merge, I have some minor comments, mostly clean ups and some understanding questions which might be solved by adding more comments.

p2panda-auth/src/group/crdt/mod.rs
sandreae: `groups` gives you all sub-groups of a particular group, whereas `seen_groups` gives you all groups ever seen in the whole context. It's useful for typing a raw actor id into a `GroupMember`.

p2panda/src/spaces/repair.rs
sandreae: Yeah, I agree that the guarantees we need here around txs are maybe lower than other places. Will still consider it more closely when we firm up our transaction approach in general.

p2panda/src/spaces/repair.rs
sandreae: In a previous version I had repair being triggered by SyncEnd or ImportEnd events which was also a pretty nice approach.

p2panda/src/spaces/repair.rs
sandreae: Yeah, they are indeed doing the same thing, the reason it needs to be called here is to avoid race conditions where the repair task hasn't picked up some required changes yet. An alternative solution which would be neater/clearer would be to maybe signal the task to perform it's work and await the response. This would just need some additional channel plumbing.

p2panda/src/spaces/repair.rs
adzialocha: I believe we want to repair instantly on every local action and throttle on remotly incoming ones.

p2panda/src/spaces/repair.rs
adzialocha: > An alternative solution which would be neater/clearer would be to maybe signal the task to perform it's work and await the response. This would just need some additional channel plumbing.

Let's do this!

p2panda/tests/spaces.rs
```diff
@@ -171,3 +242,67 @@ async fn encode_decode() -> Result<(), Box<dyn std::error::Error>> {
...
+        .await?;
+
+    sleep(Duration::from_secs(3)).await;
+
+    // We expect panda to be able to add penguin group to the space.
+    let ready = panda_space
+        .add(penguin_group.id(), AccessLevel::Read)
+        .await?;
+    assert!(ready.await.is_ok());
+
```
mycognosist, line 384: Would we maybe also want to assert that penguin's space state eventually includes the group that was added to the space by panda? Perhaps the same is also true for the previous test?

sandreae: If the group was not added the test would already fail as `panda` would interpret the new member to be an individual and therefore require a key bundle (which does not exist).

p2panda-auth/src/group/crdt/mod.rs
adzialocha: Hmn, so one method is traversing down from a user-defined starting point while this method starts from the root? These methods are quite similar, it's just that `seen_groups` takes the whole graph into account?

Maybe we can rename `seen_groups` to `groups` and `groups` to `groups_from`? Or `seen_groups` to `groups_global` and keep the name of the other method? Or something in this direction.

p2panda/src/spaces/forge.rs
adzialocha: Not resolved 😆

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,247 @@
...
+                        Some(reply_tx) => Some(reply_tx),
+                        None => {
+                            // If the repair_tx is dropped then we can exit this task.
+                            return;
+                        }
+                    }
+                }
+                // Scheduled repair triggered.
+                _ = sleep(Duration::from_secs(REPAIR_DELAY_SECS)) => None,
+            };
```
adzialocha, line 216: Ah, this is nice, very close to the throttle logic!

I think "frequency" is a more correct than "REPAIR_DELAY_SECS", since this is a re-occurring cron-like job.

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,247 @@
...
+
+            // Repair the space.
```
adzialocha: mini nit: this comment is adding no information

p2panda/src/spaces/repair.rs
```diff
@@ -0,0 +1,247 @@
...
+            let result = repair_space(topic.into(), &manager, &store, &import_tx).await;
+
+            if let Err(ref err) = result {
+                warn!("failed to repair spaces: {}", err);
+            }
+
+            // Return the result.
```
adzialocha: Same here :-)

p2panda/src/spaces/repair.rs
adzialocha: @sandreae I'd like to understand here if this becomes a very spammy WARN log or not. If we should never see non-auth messages at this point and this is a bug then I think it's fine to keep it here, otherwise it should maybe be on debug / trace level or go.

adzialocha (changes requested):

p2panda-auth/src/group/crdt/mod.rs
sandreae: It's a bit different from that. There's not necessarily a "root" group, as not all not all groups have a parent -> child relationship. The context can contain many completely unrelated groups, and what this method does is look through the entire context and return all group ids. So it's not a traversal of any one sub-section of the context.

sandreae: I like `groups_global` though if that expresses the function more accurately for you.

p2panda/src/spaces/repair.rs
sandreae: Yeah, we should never see a non-auth message, that would mean we have a bug in p2panda-spaces. In fact, maybe we should even error here.

p2panda-auth/src/group/crdt/mod.rs
adzialocha: Great, do you mind clarifying this in the method names and docs?

p2panda/src/spaces/repair.rs
adzialocha: > Yeah, we should never see a non-auth message, that would mean we have a bug in p2panda-spaces. In fact, maybe we should even error here. 

Oh yes, that sounds good as well. Or an unreachable if we assume that this is a coding error.

adzialocha (approved):
Thank you for adressing all the change requests! Let's merge it!

sandreae:
The tests are so solid with the event waiting now :star2:

PR #1290, opened 2026-07-08
Emit spaces events resulting from local actions

sandreae:
This PR corrects an issue where space membership changes resulting from calling API methods locally did not result in events being sent on the stream subscription stream. To fix this events are forwarded directly onto the stream in the node API when Node::create_space, Space::add or Space::remove are called.

The issue itself is a result of the fact that locally forged operations cannot be re-processed by the spaces manager without significant changes to the underlying p2panda-spaces APIs. This may rather signal that changes to the event sourcing design are required. Any more involved solution will be tackled in further follow-up tasks.

## 📋 Checklist

### DIFF
--- a/p2panda/src/node.rs
+++ b/p2panda/src/node.rs
@@ -32,8 +32,9 @@ use crate::spaces::{
     to_initial_members,
 };
 use crate::streams::{
-    EphemeralStreamPublisher, EphemeralStreamSubscription, Pipeline, StreamFrom, StreamPublisher,
-    StreamSubscription, SystemEvent, TaskTracker, ephemeral_stream, event_stream, processed_stream,
+    EphemeralStreamPublisher, EphemeralStreamSubscription, Pipeline, StreamEvent, StreamFrom,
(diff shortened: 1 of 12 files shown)


p2panda/src/node.rs
```diff
@@ -409,7 +410,12 @@ impl Node {
         // state here, the processor would detect that we already processed this control message
         // and therefore not emit any events.
         let initial_members = to_initial_members(initial_members);
-        let (_, group_id, message) = self.spaces_manager.create_group(&initial_members).await?;
+
+        // @TODO: Group events are currently not forwarded to the user. It's not clear which
```
mycognosist, line 414: Hmm yeah that's an interesting one. My gut feel is that they should be on a global groups stream but I'd need to think about it carefully.

Made an issue for this:

https://github.com/p2panda/p2panda/issues/1293

mycognosist (approved):

p2panda/src/node.rs
sandreae: Thank you!

PR #1291, opened 2026-07-08
Filter application events authored by concurrently removed members

mycognosist:
Here we modify the spaces processor to ensure that application-related
events authored by concurrently removed members are not forwarded to the
user.

Concretely, the flow which is addressed is as follows:

- Alice removes Bob from Space A
- Bob publishes a message to the Space A <-- concurrent to the remove
- Claire receives the remove and correctly removes Bob as a member of Space A
- Claire receives Bob's message, by looking at the "claimed" proof/dependencies Bob is authorized, but we know that they have been concurrently removed by Alice

## 📋 Checklist

### DIFF
--- a/p2panda-spaces/src/message.rs
+++ b/p2panda-spaces/src/message.rs
@@ -27,6 +27,12 @@ pub struct SpacesMessage<C> {
     pub args: SpacesArgs<C>,
 }
 
+impl<C> SpacesMessage<C> {
+    pub fn is_application_message(&self) -> bool {
+        matches!(self.args, SpacesArgs::Application { .. })
+    }
+}
+
 impl<C> Borrow<SpacesArgs<C>> for SpacesMessage<C> {
     fn borrow(&self) -> &SpacesArgs<C> {
         &self.args
--- a/p2panda-spaces/src/message.rs
+++ b/p2panda-spaces/src/message.rs
@@ -28,6 +28,12 @@ pub struct SpacesMessage<C> {
     pub args: SpacesArgs<C>,
 }
 
+impl<C> SpacesMessage<C> {
+    pub fn is_application_message(&self) -> bool {
+        matches!(self.args, SpacesArgs::Application { .. })
+    }
+}
+
 impl<C> Borrow<SpacesArgs<C>> for SpacesMessage<C> {
     fn borrow(&self) -> &SpacesArgs<C> {
         &self.args
--- a/p2panda-stream/src/spaces/processor.rs
+++ b/p2panda-stream/src/spaces/processor.rs
@@ -66,7 +66,7 @@ where
 
         let result = if let SpacesProcessorArgs::Process { msg } = input_args {
             // Process incoming event.
-            let (groups_y, space_y, events) = match self.manager.process(msg).await {
+            let (groups_y, space_y, mut events) = match self.manager.process(msg).await {
                 Ok(result) => result,
                 Err(err) => return Err((input, SpacesError::SpacesManager(err.to_string()))),
             };
@@ -92,6 +92,29 @@ where
 
             if let Some(y) = space_y {
                 let space_id = y.space_id;
+
+                // Filter out application events from removed members who previously had write or
+                // manage access.
+                //
+                // Messages authored by removed members are still applied to the local state by
+                // this processor; all we are doing is ensuring that the application-related events
+                // are not forwarded up to the user.
+                //
+                // This check serves to catch any messages which were already being processed after
+                // a member lost their write or manage access.
+                if msg.is_application_message() {
+                    let mut is_member = false;
+
+                    for (member, access) in y.groups_y.members(y.group_id) {
+                        if member == msg.author && (access.is_write() || access.is_manage()) {
+                            is_member = true;
+                        }
+                    }
+
+                    if !is_member {
+                        events.retain(|event| !matches!(event, Event::Application { .. }));
+                    };
+                }
 
                 if let Err(err) = self
                     .store
--- a/p2panda-stream/src/spaces/processor.rs
+++ b/p2panda-stream/src/spaces/processor.rs
@@ -66,7 +66,7 @@ where
 
         let result = if let SpacesProcessorArgs::Process { msg } = input_args {
             // Process incoming event.
-            let (groups_y, space_y, events) = match self.manager.process(msg).await {
+            let (groups_y, space_y, mut events) = match self.manager.process(msg).await {
                 Ok(result) => result,
                 Err(err) => return Err((input, SpacesError::SpacesManager(err.to_string()))),
             };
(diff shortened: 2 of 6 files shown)


p2panda-spaces/src/message.rs
```diff
@@ -27,6 +27,12 @@ pub struct SpacesMessage<C> {
     pub args: SpacesArgs<C>,
 }
 
+impl<C> SpacesMessage<C> {
+    pub fn is_application_message(&self) -> bool {
```
mycognosist, line 32: The match is so simple that this method might not be necessary. I don't have a strong opinion either way.

p2panda-stream/src/spaces/processor.rs
```diff
@@ -93,6 +93,28 @@ where
...
+                // Messages authored by removed members are still applied to the local state by
+                // this processor; all we are doing is ensuring that the application-related events
+                // are not forwarded up to the user.
+                //
+                // This is an interim solution. In the future we will have access control in sync
+                // which will prevent these operations from ever arriving in the first place.
+                if msg.is_application_message() {
+                    let mut is_member = false;
+
+                    for (member, _access) in y.groups_y.members(y.group_id) {
```
mycognosist: I assume we only want to filter based on individual members. Is this correct?

p2panda-stream/src/spaces/processor.rs
```diff
@@ -93,6 +93,28 @@ where
             if let Some(y) = space_y {
                 let space_id = y.space_id;
 
+                // Filter out application events from removed members.
```
mycognosist: I chose to filter here so that we'd:

1) Be within the context of a transaction
2) Have the `space_id` readily at hand

p2panda-spaces/src/message.rs
sandreae: No harm having it 🤷

p2panda-stream/src/spaces/processor.rs
sandreae: That's correct.

sandreae: We didn't discuss this earlier, but actually want to filter based on if the member has `AccessLevel::Write` access or greater, rather than if they are not a member.

p2panda-stream/src/spaces/processor.rs
```diff
@@ -93,6 +93,28 @@ where
...
+                //
+                // Messages authored by removed members are still applied to the local state by
+                // this processor; all we are doing is ensuring that the application-related events
+                // are not forwarded up to the user.
+                //
+                // This is an interim solution. In the future we will have access control in sync
```
sandreae: I don't actually think this is an interim feature, even when access controls are integrated into sync we may still want to have a "last resort" filtering of events here. Maybe this could be re-phrased? Maybe something like "in order to catch any messages which were already being processed after a member lost their write access".

sandreae: In the future we want to inform the user of these kinds of messages so they can decide what to do, for now filtering them is fine.

p2panda/tests/spaces.rs
```diff
@@ -384,3 +384,104 @@ async fn live_repair_space() -> Result<(), Box<dyn std::error::Error>> {
...
+
+#[tokio::test]
+async fn application_messages_from_removed_members_are_filtered()
+-> Result<(), Box<dyn std::error::Error>> {
+    // Panda creates a space and Penguin joins it.
+    // Panda adds Penguin to the space.
+    // Penguin publishes a message into the space; Panda receives it.
+    // Panda removes Penguin from the space.
+    // Penguin publishes a second message into the space.
+    // Panda receives stream events related to Penguin's second message, proving that it has been
```
sandreae: Are you referring to the "import started" and "import ended" events here? I don't _think_ these are related to ~pandas~ penguins second message, it would have arrived via the sync session (in live-mode).

sandreae (changes requested):
Thanks! Looking good. I just requested some re-framing of the comments and a change to the validation logic.

p2panda/tests/spaces.rs
mycognosist: When penguin publishes a second message (after being removed by panda), panda receives three events:

`StreamEvent::ImportStarted { .. }`
`StreamEvent::Group(GroupEvent::Removed { .. }) { .. }`
`StreamEvent::ImportEnded { .. }`

These events are _only_ emitted if penguin was added with `Write` or `Manage` access. If not, no events are emitted.

Is this the expected behaviour?

p2panda-stream/src/spaces/processor.rs
mycognosist: Thanks! I've pushed a change.

p2panda-stream/src/spaces/processor.rs
mycognosist: Should be fixed now.

p2panda/tests/spaces.rs
sandreae: I think this is expected, but I don't believe they're related to penguin publishing anything (if that's what you're suggesting). The import occurs on panda's side, as a result of panda themselves calling `space.remove(penguin)`.

Are we just saying the same thing in a different way? 😅

mycognosist: Yes, you're correct. I'm with you now. Sorry for the confusion 😆

sandreae:
@mycognosist after working on the validation logic elsewhere in the API the issue this PR is solving became a little clearer. There's actually validation in two places we were missing. 

1) in p2panda-spaces when processing application messages we need to validate that the member's _claimed_ access contains "write" permission
2) after an application message has been successfully processed and events emitted, check if the member has been concurrently removed

We can see the former as "objective" validation which will consistently reject messages where the author had insufficient access at the time of publishing. The later can be understood as "subjective", meaning that if you detect the concurrent loss of write access depends on which other operations you've processed.

To be clear, the former case is _always_ buggy or malicious behaviour, the later is ambiguous, it could be a member publishing concurrently when they were removed and they didn't receive this control message yet, or it could be a member maliciously referring to a point in the past where they _did_ have write access.   

In any case, both of these were validation steps are required, the latter is what you worked on here. I'll open a PR for the former. It doesn't change the nature of the assertion you introduced, but it maybe does change how we want to react when a potentially invalid message is detected. There are two options:

1) filter the operations out, don't surface them to the user (maybe just warn log as this becomes quite uncommon now)
2) still forward the operation to the user, but add a flag to show the publisher has since (concurrently) lost their write access. It's then up to the user to decide what to do with it.

sandreae:
It's not clear to me yet which option we take (even just at first), I'd like to chat about it a bit before deciding.

adzialocha:
Just to clarify the title of this PR: Is this really handling _concurrent_ cases? I read the code as filtering out unauthorized write events for application messages but it doesn't look like it's handling any cases such as writes who have been authored and removed _after_ a concurrent removal has been detected (this would need to be handled with an application-level event and the user would need to decide to take action and replay from the offending point on, I don't think we need to do that right now as it's a very advanced feature).

adzialocha:
> It's not clear to me yet which option we take (even just at first), I'd like to chat about it a bit before deciding.

I'm a bit confused by the wording here, probably just missing context .. I'd say we can't really decide right now as concurrency handling is not possible and this is already sufficient - here is my take on how to prioritize things:

>  1. in p2panda-spaces when processing application messages we need to validate that the member's _claimed_ access contains "write" permission

From the perspective of the verifying node we will only ever process spaces messages which have been dependency checked, so the claimed access should have been given or maliciously omitted. In the latter case we will still filter them out as you say.

> 2. after an application message has been successfully processed and events emitted, check if the member has been concurrently removed

This point is where I stumble a bit. The thing we _can't_ do is the actual check if something got concurrently removed by just looking at an _application_ message (that's my previous comment, I think we should remove any mentions of "concurrent" as this is not handled here). We can only check for concurrent writes looking at an _control message_ (removal).

So in terms of priorities I would suggest this is already enough and we do the interfacing of concurrent removals later. I believe it'll involve something like:

1. User observes removal event, we query if there's been concurrent writes and add that to the event as meta-data with the concrete hashes of the "offending" operations.
2. User can remove these events by calling a removal method we don't have yet
3. User can re-play stream from the offending point on, something which is broken right now for spaces

sandreae:
This PR took a drastic scope turn along the way, so I'm not surprised the current scope was hard to figure out :sweat_smile: 

What we address in this PR is the case is this flow:

- Alice removes Bob from Space A
- Bob publishes a message to the Space A <-- concurrent to the remove
- Claire receives the remove and correctly removes Bob as a member of Space A
- Claire receives Bob's message, by looking at the "claimed" proof/dependencies Bob is authorized, but we know that they have been concurrently removed by Alice

In this case we decide to _not_ forward Bob's message to the user. 

> This point is where I stumble a bit. The thing we can't do is the actual check if something got concurrently removed by just looking at an application message (that's my previous comment, I think we should remove any mentions of "concurrent" as this is not handled here). We can only check for concurrent writes looking at an control message (removal).

I think the above point explains how this is detectable concurrent activity. You're correct that we don't yet handle the "reverse" case (which you describe above).

sandreae:
The ci failures are due to an unrelated flaky test `tests::process_operation_from_expired_member`

sandreae (approved):

adzialocha (approved):
Looking good!

PR #1292, opened 2026-07-08
Space membership change validation

### LINKED ISSUE #1286: Validate membership change actions in `Space` API
We should validate that the local actor has "manage" access to the space in the Space API before forging a control message.

sandreae:
Validation for local space methods `Space::add`, `Space::remove` and `Space::publish`.

## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/validation.rs
+++ b/p2panda-auth/src/validation.rs
@@ -108,7 +108,7 @@ where
     }
 
     if !is_member(removed, members) {
-        return Err(RemoveMemberError::NonMemberRemove);
+        return Err(RemoveMemberError::NonMember);
     }
 
     Ok(())
@@ -144,7 +144,7 @@ pub enum RemoveMemberError {
     InsufficientAccess,
 
     #[error("attempted to remove non-member actor")]
-    NonMemberRemove,
+    NonMember,
 }
 
 #[cfg(test)]
@@ -209,6 +209,6 @@ mod tests {
         let err = can_remove_member(2, 3, &members()).unwrap_err();
         assert!(matches!(err, RemoveMemberError::InsufficientAccess));
         let err = can_remove_member(1, 99, &members()).unwrap_err();
-        assert!(matches!(err, RemoveMemberError::NonMemberRemove));
+        assert!(matches!(err, RemoveMemberError::NonMember));
     }
 }
--- a/p2panda-auth/src/validation.rs
+++ b/p2panda-auth/src/validation.rs
@@ -1,0 +1,204 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+use std::fmt::Debug;
+
+use thiserror::Error;
+
+use crate::AccessLevel;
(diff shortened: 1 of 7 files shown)


p2panda-auth/src/validation.rs
```diff
@@ -0,0 +1,214 @@
...
+#[derive(Debug, Error)]
+pub enum RemoveMemberError {
+    #[error("actor is not a member")]
+    UnrecognisedActor,
+
+    #[error("actor lacks sufficient access to remove a member")]
+    InsufficientAccess,
+
+    #[error("attempted to remove non-member actor")]
+    NonMemberRemove,
```
mycognosist: I wonder if `UnrecognisedActor` could be used instead of `NonMemberRemove`? Then I'd probably add the id of the member onto the `UnrecognisedActor` variant. In both cases the "actor is not a member" - you just need to differentiate whether the error pertains to the remover or the removed.

If you kept the `NonMemberRemove` variant, it could maybe be changed to `NonMember` (it's already a variant on the `RemoveMemberError` enum).

Nothing I feel strongly about, just musing about naming.

p2panda/tests/spaces.rs
```diff
@@ -459,3 +451,29 @@ async fn live_repair_space() -> Result<(), Box<dyn std::error::Error>> {
...
+async fn api_validation() -> Result<(), Box<dyn std::error::Error>> {
+    setup_logging();
+
+    let topic = Topic::random();
+
+    let panda = p2panda::spawn().await?;
+
+    let (panda_space, _panda_rx) = panda.create_space::<String>(topic).await?;
+    let result = panda_space.add(panda.id(), AccessLevel::Write).await;
+    assert!(result.is_err());
```
mycognosist: Might be a nice touch to match on the specific error variant that's returned (same for L470 and 476).

mycognosist (approved):
Looks good. I've made a suggestion relating to the tests.

p2panda/tests/spaces.rs
sandreae: True, true, I did that now and added another test case.

p2panda-auth/src/validation.rs
sandreae: I like that there is an own error variant for "the actor performing the action is not known" and "the member being added is not known". I'll leave them as they are if that's ok but rename the `NonMember` variant as you suggest.

sandreae: A general point about the error design in this module. I followed the suggestion that came up recently to try and keep errors very custom to the method scope, I like this. And also I decided to _not_ log any values in the validation errors themselves so as to leave it up to the user (us in p2panda...) to do this themselves as they like. 

The reasoning here is that how you want to log these values is quite context specific (we start using `ShortFormat::fmt_short` in many places, but not everywhere). You can see how this influenced the validation error design in `p2panda`.

It's an experiment anyway, shout if you don't like it.

PR #1299, opened 2026-07-15
Support partial merge of groups graph into space

### LINKED ISSUE #1227: Support partitioning of groups dependencies in `p2panda-spaces`
This is required so that only the required sub-section of the groups operation dependency graph can be integrated into a space (operations are pinned, encrypted and synchronized). Will mean using... https://github.com/p2panda/p2panda/blob/d4aae547601be303c7306f00460c17fd8c184a60/p2panda-auth/src/group/crdt/mod.rs#L128-L149 ...to calculate the dependencies.

sandreae:
The shared auth state contains operations relating to all known groups. Until now by default they
were interleaved into one graph which resulted in potentially unnecessary cross-group operation
dependencies. This change makes the default behaviour in -spaces to be      
that group operations are only dependant on each other if one becomes a member of another. For
example, operations for group A and group B will be appended to their own unconnected graphs until
A becomes a member of B. From that point on, all operations of A will also become a dependency of
B. It does not mean however that B is a dependency of A, it is still possible to replicate
operations for group A without any knowledge of group B.

Example

```
           ADD A
     ______/ |
    /        v
 ADD ..    ADD ..
   |         |
   v         v
CREATE A  CREATE B
```

Additionally changes were required in p2panda Node to _optionally_ chose to prefer merging all
known groups into a space forcefully. This is required for improved group discovery until there is
an invite flow which accounts for manually registering a new group (and processing any required
operations) on a space.

Some minor related refactorings were done as well to move reusable graph queries onto the inner CRDT state.

## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/group/action.rs
+++ b/p2panda-auth/src/group/action.rs
@@ -38,4 +38,27 @@ where
     pub fn is_create(&self) -> bool {
         matches!(self, GroupAction::Create { .. })
     }
+
+    /// Returns groups which are dependant on being able to process this action.
+    ///
+    /// This calculation looks into the members being added / removed by the action and returns
+    /// only the members which are groups.
+    pub fn required_groups(&self) -> Vec<ID> {
+        let members = match self {
+            GroupAction::Create { initial_members } => {
+                initial_members.iter().map(|(member, _)| member).collect()
+            }
+            GroupAction::Add { member, .. }
+            | GroupAction::Remove { member }
+            | GroupAction::Promote { member, .. }
+            | GroupAction::Demote { member, .. } => vec![member],
+        };
+        members
+            .iter()
+            .filter_map(|member| match member {
+                crate::group::GroupMember::Individual(_) => None,
+                crate::group::GroupMember::Group(id) => Some(*id),
+            })
+            .collect::<Vec<_>>()
+    }
 }
--- a/p2panda-auth/src/group/crdt/mod.rs
+++ b/p2panda-auth/src/group/crdt/mod.rs
@@ -6,8 +6,9 @@ use std::collections::{HashMap, HashSet};
 use std::fmt::Debug;
 use std::marker::PhantomData;
 
+use petgraph::algo::toposort;
 use petgraph::prelude::DiGraphMap;
-use petgraph::visit::{Bfs, DfsPostOrder, IntoNodeIdentifiers, NodeIndexable, Reversed};
+use petgraph::visit::{DfsPostOrder, IntoNodeIdentifiers, NodeIndexable, Reversed};
 #[cfg(any(test, feature = "serde"))]
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
@@ -46,6 +47,9 @@ where
     #[error("missing dependency {0} for operation {1}")]
     MissingDependencies(OP, OP),
 
+    #[error("non-create operation received for unknown group: {0}")]
+    UnknownGroup(ID),
+
     #[error("group cycle detected adding {0} to {1} operation={2}")]
     GroupCycle(ID, ID, OP),
 
@@ -63,6 +67,9 @@ pub(crate) type GroupStates<ID, C> = HashMap<ID, GroupMembersState<GroupMember<I
 
 /// Inner state object for `GroupCrdt` which contains the actual groups state,
 /// including operation graph and membership snapshots.
+///
+/// @TODO: We no longer need a separation between the inner and outer state objects, they can be
+/// merged into one.
 #[derive(Clone, Debug)]
 #[cfg_attr(any(test, feature = "serde"), derive(Deserialize, Serialize))]
 pub struct GroupCrdtInnerState<ID, OP, M, C>
@@ -109,8 +116,8 @@ where
     M: Operation<ID, OP, C>,
     C: Conditions,
 {
-    /// Current tips for the groups operation graph.
-    pub fn heads(&self) -> HashSet<OP> {
+    /// Current tips for all groups in the operation graph.
+    pub(crate) fn heads_global(&self) -> HashSet<OP> {
         self.graph
             // TODO: clone required here when converting the GraphMap into a Graph. We do this
             // because the GraphMap api does not include the "externals" method, where as the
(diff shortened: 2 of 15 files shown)


p2panda-auth/src/group/action.rs
```diff
@@ -38,4 +38,23 @@ where
     pub fn is_create(&self) -> bool {
         matches!(self, GroupAction::Create { .. })
     }
+
+    pub fn required_groups(&self) -> Vec<ID> {
```
sandreae, line 46: Needs doc string

sandreae: This method is for getting the groups which are a dependency of this action, so that we can append it to the correct sub-section of the auth graph.

p2panda-auth/src/traits/operation.rs
```diff
@@ -22,4 +26,10 @@ pub trait Operation<ID, OP, C = ()> {
 
     /// The group action.
     fn action(&self) -> GroupAction<ID, C>;
+
+    fn required_groups(&self) -> Vec<ID> {
```
sandreae, line 34: Same as previous method, except now on operation, the group being effected by the action is also included in the calculation here.

p2panda-spaces/src/manager.rs
```diff
@@ -413,31 +413,21 @@ where
         Ok(y)
     }
 
-    /// Returns a list of all spaces which are "out-of-sync" with the global shared auth state.
-    pub async fn spaces_repair_required(&self) -> Result<Vec<SpaceId>, ManagerError<F, C>> {
+    pub async fn space_repair_required(
```
sandreae, line 418: Doc string

p2panda-spaces/src/manager.rs
```diff
@@ -504,13 +494,41 @@ where
...
+                .repair(&[space.group_id().await?])
+                .await
+                .map_err(ManagerError::Space)?;
             results.push(result);
         }
 
         Ok(results)
     }
 
+    pub async fn repair_space(
```
sandreae, line 476: Doc string

p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -126,35 +133,78 @@ where
...
+        graph
+            .clone()
+            .into_graph::<usize>()
+            .externals(petgraph::Direction::Outgoing)
+            .map(|idx| graph.from_index(idx.index()))
+            .collect::<HashSet<_>>()
+    }
+
+    /// Returns a vector of operation ids in their topologically sorted order where only
+    /// operations relating to the passed groups (and their dependencies) are been included.
```
mycognosist:
```suggestion
    /// operations relating to the passed groups (and their dependencies) are included.
```

p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -126,35 +133,78 @@ where
...
+    pub fn toposort(&self, groups: &[ID]) -> Vec<OP> {
+        let graph = self.filtered_graph(groups);
+        toposort(&graph, None).expect("graph has a cycle")
```
mycognosist, line 152: What are the implications of a cycle here? Would we need to drop some state? Are there currently higher-level checks in place to prevent cycles in the first place?

Mostly wondering how likely it is that this occurs and then we end up with the application repeatedly panicking every time this `toposort` method is called.

p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -627,14 +686,27 @@ where
             ));
         }
 
+        let heads = HashSet::from_iter(operation.dependencies().clone());
+
+        // If this operation is not being appended directly to our current local state we need to
+        // remove any operations from the graph which are concurrent or sequentially later to the
+        // the graph to the operations' claimed dependencies.
```
mycognosist: Bit of a sentence mix-up here...not exactly sure how it should read.

p2panda/src/spaces/repair.rs
```diff
@@ -29,8 +29,38 @@ use crate::streams::{LocalStreamFuture, StreamEvent};
...
+/// space yet. This results in improved discoverability (new groups are "automatically"
+/// discovered) at the expense of privacy (even if a group is not added to a space it is
+/// replicated on the space topic).
+///
+/// ## Partial
+///
+/// Only operations for groups added to a space (via a local action or by explicit association)
+/// are merged into the space. This results in improved privacy as there is no group "leakage"
+/// from the shared state into the space, however it means the initial "discovery" of a new group
+/// to-be-added group must be solved via another channel (side-channel, dedicated topic, etc..).
```
mycognosist:
```suggestion
/// from the shared state into the space, however it means the initial "discovery" of a new
/// to-be-added group must be solved via another channel (side-channel, dedicated topic, etc..).
```

p2panda/src/spaces/repair.rs
```diff
@@ -29,8 +29,38 @@ use crate::streams::{LocalStreamFuture, StreamEvent};
 
 const REPAIR_FREQUENCY_SECS: u64 = 1;
 
-/// Repairing a space involves incorporating missing groups operations observed on the global groups
-/// context but not yet published into the space.
+/// Strategy by which a space should be repaired.
```
mycognosist, line 32: Great description of the problem space.

mycognosist (approved):
I can't claim to understand all the graph manipulation code but the overall logic and approach looks sound to me.

I also really like the docs you added to the high-level methods.

There are a couple of docs-related changes I've suggested, along with a request for clarification about some graph cycle behaviour.

p2panda-auth/src/group/crdt/mod.rs
sandreae: There are high-level checks for a cycles when an operation is appended to the graph, so I believe we should never encounter a cycle at this level.

mycognosist: Perfect, thanks!

p2panda/src/spaces/repair.rs
sandreae: Thanks! I was really happy when I figured out we needed this type and then that it was the best place to introduce the problem space.

sandreae:
> I can't claim to understand all the graph manipulation code ....

@mycognosist yeah, sorry about that... I should have said that most of the graph traversal code was already there but a bit all over the place. In this PR I refactored it all into one place and exposed an API for better re-use.

PR #1300, opened 2026-07-16
Cancel processed stream tasks on drop of stream handles

### LINKED ISSUE #1275: Gracefully wind down pipeline task on drop
It currently stays running, even if the regarding `Pipeline` struct is dropped. This is not an issue in main as we have one pipeline only which runs as long as the whole program, but in our spaces integration pipelines live shorter.

mycognosist:
Here we introduce a `StreamDropGuard` to ensure that the input and output event processing tasks in `processed_stream()` are broken out of when all instances of a topic stream handle are dropped.

The commit messages tell the detailed story.

Addresses: https://github.com/p2panda/p2panda/issues/1272

## 📋 Checklist

### DIFF
--- a/p2panda/src/streams/stream.rs
+++ b/p2panda/src/streams/stream.rs
@@ -2,14 +2,11 @@
 
 use std::borrow::Borrow;
 use std::fmt::Debug;
-use std::marker::PhantomData;
-use std::pin::Pin;
 use std::sync::Arc;
-use std::task::{Context, Poll};
 
+use futures_util::StreamExt;
 use futures_util::stream::BoxStream;
-use futures_util::{FutureExt, Stream, StreamExt};
-use p2panda_core::cbor::{DecodeError, EncodeError, decode_cbor, encode_cbor};
+use p2panda_core::cbor::{DecodeError, decode_cbor};
 use p2panda_core::traits::Digest;
 use p2panda_core::{Hash, Topic, VerifyingKey};
 use p2panda_net::NodeId;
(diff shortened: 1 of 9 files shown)


p2panda/src/streams/stream.rs
```diff
@@ -612,303 +626,6 @@ where
     Some(forward_events)
 }
 
-/// Publish messages into a topic stream.
```
sandreae, line 615: Very nice to have these in their own modules now 👍

p2panda/src/streams/stream.rs
sandreae, line 170: I believe we should also close down the repair task in the same way now.

sandreae (changes requested):
This looks great to me, nice work :-)

I have two requests, one is regarding the repair task (I assume we should close this down using the cancellation token now too, correct me if that's not needed for any reason though) and the other is about tests. It would be great to have a test which asserts that the user receives a `SyncStarted` event after the tx + rx for a topic get dropped and then the same topic is subscribed to again. This would assert that the same sync sessions aren't just still running.

p2panda/src/streams/stream.rs
mycognosist: I looked into this (even quickly implemented the cancellation in the repair task) but then I noticed that we already abort the task via the handle as the input event task winds-down. The cancellation token breaks us out of the loop and the final thing we do before exiting the task is call `repair_task_handle.abort()`. So that one's all covered.

I've also added the test you suggested. Thanks for the recommendation!

sandreae (approved):
Great! Thanks for the extra test 🙏

PR #1306, opened 2026-07-19
Make group sendsync

matthiasbeyer:
Ensure that the `Group` type is `Send + Sync`, so it can be used in multi-threaded environments.

I am not 100% sure on whether this alters functionality and I did not execute the testsuite for now.

## 📋 Checklist


adzialocha (approved):
Looks good!

Nice touch with the `static_assertions`, we should use this in other places as well.

Issue #1274, opened 2026-06-30
Re-creating sync manager immediately after dropping it fails with "error occurred in internal actor: no stream exists for the given topic"


adzialocha:
When adding a sleep in-between dropping and re-creating it it works.

PR #1307, opened 2026-07-20
Allow graceful closure of sync sessions

mycognosist:
In this PR we introduce the ability to close a topic sync session and await cleanup. This complements the previous fire-and-forget approach to closing / dropping sync sessions.

Changes have been made to the sync manager and topic sync manager; a oneshot sender is supplied along with the `ToSyncManager::Close` and `ToTopicManger::Close` actor messages. A reply can be awaited on the oneshot receiver. The termination signal is sent once all state cleanup has been completed.

The new graceful termination functionality is exposed at the high-level API via the `StreamPublisher::close()` method.

Deduplication logic was necessary to ensure that unexpected behaviour doesn't occur as a result of two `Close` messages being sent: one by an explicit call to `StreamPublisher::close()` and one via the implicit `Drop` behaviour of the internal `SyncHandle`.

Addresses: https://github.com/p2panda/p2panda/issues/1274

## 📋 Checklist

### DIFF
--- a/p2panda-net/src/sync/actors/topic_manager.rs
+++ b/p2panda-net/src/sync/actors/topic_manager.rs
@@ -1,5 +1,13 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+//! Topic manager actor.
+//!
+//! The topic manager holds state for all sync sessions associated with a single topic. It provides
+//! a means of initating new sync sessions, accepting inbound sync sessions and publishing messages
+//! to all active sync sessions for the associated topic.
+//!
+//! A separate topic manager actor is spawned by the sync manager for each topic of interest.
+
 use std::collections::{HashMap, HashSet};
 use std::error::Error as StdError;
 use std::fmt::Debug;
@@ -14,7 +22,7 @@ use p2panda_sync::traits::Manager as SyncManagerTrait;
 use p2panda_sync::{FromSync, SessionConfig, ToSync};
 use ractor::thread_local::{ThreadLocalActor, ThreadLocalActorSpawner};
 use ractor::{ActorId, ActorProcessingErr, ActorRef, SupervisionEvent};
-use tokio::sync::broadcast;
+use tokio::sync::{broadcast, oneshot};
 use tokio::time::Duration;
 use tracing::{debug, warn};
 
@@ -60,10 +68,13 @@ pub enum ToTopicManager<T> {
 
     /// Close all active sync sessions running over the given topic. This essentially shuts down
     /// the whole manager.
-    CloseAll,
+    CloseAll(oneshot::Sender<()>),
 
     /// Close all active sync sessions running with the given node id.
-    Close { node_id: NodeId },
+    Close {
+        node_id: NodeId,
+        reply: oneshot::Sender<()>,
+    },
 }
 
 pub struct TopicManagerState<M>
@@ -311,7 +322,7 @@ where
                     let _ = handle.send(ToSync::Payload(data.clone())).await;
                 }
             }
-            ToTopicManager::CloseAll => {
+            ToTopicManager::CloseAll(reply) => {
                 // Get a handle onto any sync sessions running over the subscription topic and send
                 // a Close message. The session will send a close message to the remote then
                 // immediately drop the session.
@@ -332,8 +343,12 @@ where
                         node_id.fmt_short()
                     );
                 }
+
+                // The receiver may have been dropped immediately if the caller is not interested
+                // in awaiting the termination signal, so we ignore any potential error here.
+                let _ = reply.send(());
             }
-            ToTopicManager::Close { node_id } => {
+            ToTopicManager::Close { node_id, reply } => {
                 if state.active_sync_set.remove(&node_id) {
                     debug!(
                         topic = state.topic.fmt_short(),
@@ -363,6 +378,8 @@ where
                         let _ = handle.send(ToSync::Close).await;
                     }
                 };
+
+                let _ = reply.send(());
             }
         }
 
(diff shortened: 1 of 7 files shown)


p2panda-net/src/sync/actors/topic_manager.rs
```diff
@@ -1,5 +1,13 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+//! Topic manager actor.
```
mycognosist, line 3: Added a bit of document here to help the reader distinguish between the sync manager and topic manager actors.

p2panda/src/streams/publisher.rs
```diff
@@ -154,6 +161,43 @@ where
         self.publish_inner(Some(message), false).await
     }
 
+    async fn publish_inner(
```
mycognosist, line 164: I moved this method so that it would come directly after the `publish()` method.

No actual logic has been changed.

p2panda/src/streams/publisher.rs
```diff
@@ -120,6 +125,7 @@ where
     pub fn new(
         topic: Topic,
         forge: OperationForge,
+        sync_handle: Arc<SyncHandle<Operation, TopicLogSyncEvent<Extensions>>>,
```
mycognosist, line 128: I had to reintroduce the sync handle to be able to call `close()` on it.

p2panda-net/src/sync/actors/topic_manager.rs
sandreae: Thank you! :pray:

p2panda-net/src/sync/handle.rs
```diff
@@ -92,10 +112,27 @@ where
     E: Clone + Send + 'static,
 {
     fn drop(&mut self) {
-        // Ignore error here as the actor might already be dropped.
-        let _ = self
-            .manager_ref
-            .send_message(ToSyncManager::Close(self.topic));
+        let (reply, _reply_rx) = oneshot::channel();
+
+        // Sending `ToSyncManager::Close` twice for the same topic in quick succession can cause
```
sandreae, line 117: Really clear solution and docs, thanks!

p2panda/tests/api.rs
```diff
@@ -613,3 +601,40 @@ async fn sync_is_closed_on_drop() {
...
+    // Briefly sleep to await cleanup.
+    sleep(Duration::from_millis(50)).await;
+
+    let (icebear_tx, icebear_rx) = icebear.stream::<String>(chat_id).await.unwrap();
+
+    // Gracefully close the publisher, awaiting termination.
+    let _ = icebear_tx.close().await;
+
+    drop(icebear_tx);
+    drop(icebear_rx);
```
sandreae, line 636: What is it demonstrating in the test to call drop here after close()? I would expect this test path to just be 1) the user calls tx.close().await 2) the user subscribes again successfully.

p2panda-net/src/sync/handle.rs
```diff
@@ -84,6 +87,23 @@ where
             .send_message(ToSyncManager::InitiateSync(self.topic, node_id))
             .unwrap();
     }
+
+    /// Close the associated sync session gracefully.
+    ///
+    /// This method can be awaited to ensure that all sync-related state has been cleaned up.
+    pub async fn close(&self) -> Result<(), SyncHandleError<M, E>> {
```
sandreae, line 94: What do we expect to happen if the user still publishes to a clone of this tx or is awaiting message on the rx? Am I right thinking the tx would error with some channel close error and the rx might terminate?

sandreae (approved):
These changes look great to me, very clear and simple solution to the problem, thanks. 

After this change I'm wondering what UX we expect when a user calls `tx.close()` but they are either still holding and try to publish to clones of the `tx` and/or they still have the `rx` and are awaiting events. I mentioned this in the comments, would be good to know what happens now and if we need any adjustments there in follow-up PRs.

p2panda/tests/api.rs
mycognosist: I can see why this seems a bit strange.

It serves to illustrate that there is nothing problematic about explicitly closing the publisher and then dropping the stream before attempting to create the stream again; the underlying sync handle implementation deduplicates the `Close` message that is sent to the sync session actor.

Prior to the deduplication logic in `SyncHandle`, the (re)creation of the stream would have returned an error because the dropping would have resulted in a second `Close` being sent (thus creating a state-wiping race condition).

I'm going to add a clarifying comment and will also add the test path you expected (ie. close and then resubscribe successfully).

p2panda-net/src/sync/handle.rs
mycognosist: That's a very good question. The current behaviour as I understand it is as follows:

Only sync is affected if the tx (`StreamPublisher`) is closed; the user can still publish to the tx and the operation will be processed but _not_ sent via sync. This is illustrated in the logs (I threw together a quick test to observe the behaviour):

// Sync fails.
`WARN p2panda::streams::stream: failed sending operation on sync handle operation_id=785453fec66004bde9b388689785a959c2812daa02c4e5b62a09d2facc93d432`

// Operation processing succeeds.
`DEBUG p2panda::streams::stream: processing application message id=785453fec66004bde9b388689785a959c2812daa02c4e5b62a09d2facc93d432` 

Since the sync session(s) have been closed for the associated topic, calling `.next()` on the rx (`StreamSubscription`) will not return any messages from sync. However, the rx is not terminated because the inner stream is a `ReceiverStream<StreamEvent<M>>` and is therefore decoupled from sync. I haven't verified this but the rx should still emit external stream events, for example.

I do think this warrants a deeper discussion so we can arrive on consensus over the desired behaviour and ensure we are clearly communicating to API users.

p2panda/tests/api.rs
sandreae: Thanks!

p2panda-net/src/sync/handle.rs
sandreae: Cool, thanks for the explanation, I'll raise an issue for this follow-up task :+1:

PR #1311, opened 2026-07-21
API for promoting and demoting space and group members

sandreae:
## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/group/crdt/mod.rs
+++ b/p2panda-auth/src/group/crdt/mod.rs
@@ -1967,4 +1967,101 @@ pub(crate) mod tests {
         sorted.sort();
         assert_eq!(sorted, vec![0, 1, 2, 3]);
     }
+
+    #[test]
+    fn promote_demote_group() {
+        let y = TestGroupState::new();
+
+        let op1 = create_group(
+            ALICE,
+            0,
+            G1,
+            vec![
+                (GroupMember::Individual(ALICE), Access::manage()),
+                (GroupMember::Individual(BOB), Access::write()),
+                (GroupMember::Individual(CLAIRE), Access::read()),
+            ],
+            vec![],
+        );
+
+        let y_i = TestGroup::process(y, &op1).unwrap();
+        let mut members = y_i.members(G1);
+        members.sort();
+        assert_eq!(
+            members,
+            vec![
+                (ALICE, Access::manage()),
+                (BOB, Access::write()),
+                (CLAIRE, Access::read()),
+            ]
+        );
+
+        let op2 = create_group(
+            DAN,
+            1,
+            G2,
+            vec![
+                (GroupMember::Individual(DAN), Access::manage()),
+                (GroupMember::Group(G1), Access::read()),
+            ],
+            vec![op1.id()],
+        );
+
+        let y_ii = TestGroup::process(y_i, &op2).unwrap();
+        let mut members = y_ii.members(G2);
+        members.sort();
+        assert_eq!(
+            members,
+            vec![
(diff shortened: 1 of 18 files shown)


p2panda-spaces/src/space.rs
```diff
@@ -398,106 +408,114 @@ where
...
-            );
+        Ok(Some((y, events)))
+    }
 
-            // Insert membership event at front of vec.
-            events.insert(0, membership_event);
-            events
-        } else {
-            vec![]
+    /// Generate spaces events based on previous and next space members.
```
sandreae, line 472: This is mostly existing code which was duplicated in two places and I refactored into a helper method.

p2panda-spaces/src/space.rs
```diff
@@ -630,6 +647,58 @@ where
         Ok((space_y, messages, events))
     }
 
+    /// Instantiate space state from existing global auth state.
```
sandreae, line 650: Just moved this method lower in the file for better method order.

p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -1967,4 +1967,101 @@ pub(crate) mod tests {
         sorted.sort();
         assert_eq!(sorted, vec![0, 1, 2, 3]);
     }
+
+    #[test]
+    fn promote_demote_group() {
```
mycognosist, line 1972: This would be a great test to point at when developers aren't completely clear on the nested group / permissions interaction; it spells things out really clearly.

p2panda-auth/src/validation.rs
```diff
@@ -137,6 +165,70 @@ where
     Ok(())
 }
 
+/// Validate if an existing group member member can be promoted to a higher access level.
```
mycognosist:
```suggestion
/// Validate if an existing group member can be promoted to a higher access level.
```

p2panda-auth/src/validation.rs
```diff
@@ -137,6 +165,70 @@ where
...
+    }
+
+    if has_gte_access(promoted, access, members) {
+        return Err(PromoteMemberError::HasGreaterOrEqualAccess);
+    }
+
+    Ok(())
+}
+
+/// Validate if an existing group member member can be demoted to a lower access level.
```
mycognosist:
```suggestion
/// Validate if an existing group member can be demoted to a lower access level.
```

mycognosist (approved):
Looks good to me.

There are a couple of small typos.

My one clarifying question is around events. I seem to remember us discussing this but have lost hold of the details. It seems we don't emit `SpaceEvent` variants specifically labelled as `Promote` or `Demote`, but rather as `Added` and `Removed`. Could you give me a brief summary of your thoughts around this?

sandreae:
@mycognosist 

> My one clarifying question is around events. I seem to remember us discussing this but have lost hold of the details. It seems we don't emit SpaceEvent variants specifically labelled as Promote or Demote, but rather as Added and Removed. Could you give me a brief summary of your thoughts around this?

Yeah, this was the case, but after considering the reason I realised it didn't really hold up as the desired UX. My reasoning _was_ that the most critical thing from the perspective of a space is if a user has "read" access or not as this means they get the group secret. This resulted in me only exposing events to the user when a member passed in or out of "read" access, which is where the no "promote" or "demote" events came from. However, even though this is true ("at least read" being a critical member state) still when you observe a space, you want to see the relevant auth changes, and not a separate groups events, but just directly in space events. And so now all auth actions which apply to a space are forwarded up to the user.

I think you must have started your review just before I pushed this change, sorry about that :pray:

PR #1312, opened 2026-07-23
Move orderer processor from node to stream

mycognosist:
Here we remove the orderer processor implementation (along with the associated types) from `p2panda` and replace the existing implementation in `p2panda-stream` with it. This requires changes to several tests, since the previous implementation returned an operation but we now return a tuple of an event and result.

## 📋 Checklist


mycognosist:
Most changes have to do with updating the tests; this required test implementations of event and event metadata types.

p2panda/src/processor/event.rs
```diff
@@ -8,11 +8,12 @@ use p2panda_core::{
 };
 use p2panda_stream::ingest::{IngestArgs, IngestError, IngestResult};
 use p2panda_stream::log_prune::{LogPruneArgs, LogPruneError, LogPruneResult};
+use p2panda_stream::orderer::{OrdererArgs, OrdererError, OrdererMetadata, OrdererResult};
 use p2panda_stream::spaces::{SpacesError, SpacesProcessorArgs, SpacesResult};
+
```
adzialocha: They should actually be grouped together here (`p2panda_stream` is an "external" dependency for `p2panda`, similar as `serde`), I know it's a bit of a gray zone

adzialocha (approved):

sandreae (approved):

Issue #1282, opened 2026-07-03
Sync control messages of groups we want to be added to a space


sandreae:
A group is created via a call to `node.create_group()` which creates the group locally but does not bring it into any spaces context where groups control messages would then be encrypted and replicated to other subscribed peers. This would happen when calling `space.add(group_id)` however the creator of a group is not necessarily a manager of the space.

The underlying issue here is that the user who created the spaces doesn't have have the control messages yet describing the group and the group itself has not been associated to any space yet, so the user will never sync the messages.

adzialocha:
> This would happen when calling space.add(group_id) however the creator of a group is not necessarily a manager of the space.

I might miss the important bit here. Why would a non-manager be allowed to add somebody to a space?

sandreae:
They're not adding someone to the space, they're registering their group (eg. replicate groups control messages) so that someone else (a manager) could then add them to the space if they want. The scenario is:

- Alice creates a space
- Alice shares the topic id with Bob on a side channel
- Bob subscribes to the space
- Bob creates a device group
=== Alice wants to add Bob's device group but doesn't have the control messages ===
- Bob registers the device group on the space
=== Alice receives Bob's device group control messages ===
- Alice adds Bob's device group to the space

sandreae:
We need a way to register a group to a space actually adding it as a member. This effectively involves calling `store.associate(space_id, group_log_id)` somewhere. 

Suggested APIs:
- `space.create_group()`
- `space.register_group(group_id)`

Under the hood `space.register_group(group_id)` would be calling `store.associate(space_id, group_log_id)`.

adzialocha:
> Under the hood space.register_group(group_id) would be calling store.associate(space_id, group_log_id).

As you say, the actual underlying mechanism here is a log association to a topic, we don't really associate any group to any space yet, I've changed the title of the issue to highlight this a bit more.

My suggestion here would not be to offer additional methods on the API (as I don't think they are very easy to understand) but automate the process internally.

We could delay the "add to group" action when we don't have any materialised data on the actor (key bundle of member or create control message for group) and do the association automatically instead.

There's multiple options on how to go about this "delaying":

1. Keep a wishlist for each add action which couldn't be handled yet, automatically execute it when missing actor messages got finally synced, don't know if we want to offer cancelling items in the wishlist. API-wise it could be part of the future which is returned to the user when calling `add` and offer a way to call `cancel` on it.
[...]

PR #1313, opened 2026-07-24
Emit per-group and global groups events

### LINKED ISSUE #1293: Forward group events to user
Group events are currently not forwarded to the user. It's not clear which channel these should be sent on: the spaces stream, a stream for the specific group, or a global groups stream.

sandreae:
This change wires up the event streams so that events containing group membership changes are forwarded to the correct subscribers. There are two routes a group event can take:

1) to the global system events stream
2) to any per-group event streams

For the former all events are forwarded. For the latter we only want to forward events which effect membership of the subscribed-to group, this includes events targeting the subscribed-to group or any child groups (current or removed). This filtering required adding some additional contextual information to the groups events in p2panda-spaces.

Additional changes were made to make the structure of group and space events consistent, and to present the user with the most pertinent information up front. In this case that would be the current membership of a group or space, rather than the actual action which resulted in the group membership change.

## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/group/crdt/mod.rs
+++ b/p2panda-auth/src/group/crdt/mod.rs
@@ -303,6 +303,22 @@ where
         }
     }
 
+    /// Get all ancestor groups of the passed target group.
+    ///
+    /// A ancestor group is any group that the target group is a transitive child of. Said another
+    /// way, if the membership of the target group changed, a ancestor is any group which would be
+    /// effected by this change.
+    pub fn ancestors(&self, target: ID) -> Vec<ID> {
+        let mut ancestors = vec![];
+        for (id, _) in self.current_state() {
+            let children = self.groups(id);
+            if children.iter().any(|(id, _)| id == &target) {
+                ancestors.push(id)
+            }
+        }
+        ancestors
+    }
+
     /// Get all current individual members of a group.
     pub fn members(&self, group_id: ID) -> Vec<(ID, Access<C>)> {
         self.traverse_members(group_id, 0)
--- a/p2panda-spaces/src/event.rs
+++ b/p2panda-spaces/src/event.rs
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use std::collections::{HashMap, HashSet};
+
+use p2panda_core::VerifyingKey;
 use serde::{Deserialize, Serialize};
 
 use p2panda_auth::Access;
@@ -52,24 +55,50 @@ impl GroupActor {
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[allow(clippy::large_enum_variant)]
 pub enum Event<C> {
-    Group(GroupEvent<C>),
+    /// A group membership change occurred in the shared groups state.
+    ///
+    /// This event does _not_ signify that any space has incorporated this change yet. The
+    /// Event::Spaces variant is emitted on space membership changes.
+    Groups(GroupEvent<C>),
+
+    /// An application message was decrypted.
+    ///
+    /// Encrypted application messages are buffered until the local member is welcomed into a
+    /// space with a "create" or "add" message.
     Application { space_id: SpaceId, data: Vec<u8> },
+
     // @TODO: Could maybe add field to show when the bundle is valid until?
+    /// An actors' pre-key bundle has been received.
     KeyBundle { author: MemberId },
-    Space(SpaceEvent<C>),
+
+    /// A membership change occurred on a space.
+    ///
+    /// This event is emitted every time the membership of a space changes. Events are silently
+    /// dropped if the local member is not (yet) a member of the space. For every Event::Groups
+    /// event a Event::Spaces event will be emitted for every effected space.
+    Spaces(SpaceEvent<C>),
 }
 
 /// Additional context attached to group events.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-pub struct GroupsContext<C> {
-    /// The actor who authored this action.
+pub struct GroupContext<C> {
+    /// The actor who authored the associated group action.
     pub author: ActorId,
 
     /// Root group actors, can be individuals or groups.
-    pub group_actors: Vec<(GroupActor, Access<C>)>,
+    pub actors: Vec<(GroupActor, Access<C>)>,
 
-    /// Members of this group.
+    /// Members of the group.
     pub members: Vec<(ActorId, Access<C>)>,
+
+    /// All groups for which the group is a child (direct or transitive).
+    pub ancestors: Vec<ActorId>,
+
+    /// All groups effected by the associated group change and their members.
+    pub effected_group_members: HashMap<ActorId, Vec<(ActorId, Access<C>)>>,
+
+    /// All groups effected by the associated group change and their direct actor members.
+    pub effected_group_actors: HashMap<ActorId, Vec<(GroupActor, Access<C>)>>,
 }
 
 /// Additional context attached to space events.
@@ -81,13 +110,13 @@ pub struct SpaceContext<C> {
     /// auth changes which effect a space are applied later by other members.
     pub author: MemberId,
 
-    /// Id of the group associated with this space.
+    /// Id of the group associated with the space.
     pub group_id: GroupId,
 
-    /// Members in the spaces' space.
+    /// Current members of the space.
     pub members: Vec<(MemberId, Access<C>)>,
 
-    /// Members in the spaces' space.
+    /// Current direct actor members of the space.
     pub actors: Vec<(GroupActor, Access<C>)>,
 }
 
@@ -103,7 +132,7 @@ pub enum GroupEvent<C> {
         initial_members: Vec<(GroupActor, Access<C>)>,
 
         /// Additional event context and group state after any change occurred.
-        context: GroupsContext<C>,
+        context: GroupContext<C>,
     },
 
     /// A member was added to a group.
@@ -118,7 +147,7 @@ pub enum GroupEvent<C> {
         access: Access<C>,
 
         /// Additional event context and group state after any change occurred.
-        context: GroupsContext<C>,
+        context: GroupContext<C>,
     },
 
     /// A member was removed from a group.
@@ -130,7 +159,7 @@ pub enum GroupEvent<C> {
         removed: GroupActor,
 
         /// Additional event context and group state after any change occurred.
-        context: GroupsContext<C>,
+        context: GroupContext<C>,
     },
 
     /// An existing group member was promoted.
@@ -145,7 +174,7 @@ pub enum GroupEvent<C> {
         access: Access<C>,
 
         /// Additional event context and group state after any change occurred.
-        context: GroupsContext<C>,
+        context: GroupContext<C>,
     },
 
     /// An existing group member was demoted.
@@ -160,11 +189,12 @@ pub enum GroupEvent<C> {
         access: Access<C>,
 
         /// Additional event context and group state after any change occurred.
-        context: GroupsContext<C>,
+        context: GroupContext<C>,
     },
 }
 
 impl<C> GroupEvent<C> {
+    /// The target group of this event.
     pub fn group_id(&self) -> GroupId {
         match self {
             GroupEvent::Created { group_id, .. } => *group_id,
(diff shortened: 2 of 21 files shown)


p2panda-spaces/src/event.rs
```diff
@@ -66,10 +69,19 @@ pub struct GroupsContext<C> {
...
-    pub group_actors: Vec<(GroupActor, Access<C>)>,
+    pub actors: Vec<(GroupActor, Access<C>)>,
 
     /// Members of this group.
     pub members: Vec<(ActorId, Access<C>)>,
+
+    /// All groups for which this group is a child (direct or transitive).
+    pub parents: Vec<ActorId>,
+
+    /// All effected groups and their members.
```
sandreae: In the case of a remove action, this includes all groups from which the member was removed.

p2panda-spaces/src/event.rs
```diff
@@ -66,10 +69,19 @@ pub struct GroupsContext<C> {
...
+    pub effected_members: HashMap<ActorId, Vec<(ActorId, Access<C>)>>,
```
sandreae: Better naming needed. This is a map of all effected groups and their resulting members (after the change was applied).

p2panda-spaces/src/event.rs
```diff
@@ -174,6 +186,26 @@ impl<C> GroupEvent<C> {
...
+        match self {
+            GroupEvent::Created { context, .. }
+            | GroupEvent::Added { context, .. }
+            | GroupEvent::Removed { context, .. }
+            | GroupEvent::Promoted { context, .. }
+            | GroupEvent::Demoted { context, .. } => context,
+        }
+    }
+
+    pub fn effects(&self, group_id: GroupId) -> bool {
```
sandreae: Again, naming difficult, docs will help. This checks if the passed group was affected by (member added, removed, promoted or demoted) the action which caused this event to be emitted.

p2panda/src/spaces/group.rs
```diff
@@ -219,3 +201,52 @@ pub enum GroupError {
     #[error(transparent)]
     CreateStream(#[from] CreateStreamError),
 }
+
+#[derive(Clone, Debug, PartialEq)]
```
sandreae, line 205: Needs thorough docs.

p2panda/src/streams/stream.rs
```diff
@@ -754,18 +797,87 @@ pub enum StreamEvent<M> {
...
 
     /// Space has been created or modified.
-    Space(SpaceEvent),
-
-    // @TODO: It's not clear where group events should be forwarded so we don't send any for now
-    // and therefore this variant is commented out.
-    //
-    /// Group has been created or modified.
-    // Group(GroupEvent<AuthCapabilities>),
+    Space {
```
sandreae, line 779: Docs

p2panda/tests/spaces.rs
```diff
@@ -542,7 +581,11 @@ async fn api_validation() -> Result<(), Box<dyn std::error::Error>> {
...
     while let Some(event) = tiger_rx.next().await {
-        if let StreamEvent::Space(SpaceEvent::Added { .. }) = event {
+        if let StreamEvent::Space {
+            inner: SpaceEvent::Added { .. },
+            ..
+        } = event
+        {
             break;
         };
     }
```
sandreae, line 609: Need tests for group events on Groups and the system event stream.

p2panda-auth/src/group/crdt/mod.rs
```diff
@@ -303,6 +303,22 @@ where
         }
     }
 
+    /// Get all parent groups of the passed target group.
+    ///
+    /// A parent group is any group that the target group is a transitive child of. Said another
+    /// way, if the membership of the target group changed, a parent is any group which would be
+    /// effected by this change.
+    pub fn parents(&self, target: ID) -> Vec<ID> {
```
adzialocha: I believe the graph theory term would be "ancestors" here

p2panda-spaces/src/event.rs
```diff
@@ -52,42 +55,68 @@ impl GroupActor {
...
     // @TODO: Could maybe add field to show when the bundle is valid until?
+    /// A key bundle has
     KeyBundle { author: MemberId },
+
+    /// A membership change occurred on a space.
+    ///
+    /// This event is emitted every the membership of a space changes. Events are silently dropped
+    /// if the local member is not (yet) a member of the space. For every Event::Groups event a
+    /// Event::Space event will be emitted for every effected space.
     Space(SpaceEvent<C>),
```
adzialocha: Maybe plural Spaces here as well, consistent with the plural Groups variant?

p2panda-spaces/src/event.rs
```diff
@@ -52,42 +55,68 @@ impl GroupActor {
...
+    /// A group membership change occurred in the shared groups state.
+    ///
+    /// This event does _not_ signify that any space has incorporated this change yet. The
+    /// Event::Space variant is emitted on space membership changes.
+    Groups(GroupEvent<C>),
+
+    /// An application message was decrypted.
+    ///
+    /// Application messages are buffered until the local member is welcomed into a space with a
+    /// "create" or "add" message.
```
adzialocha: Maybe you add that these buffered messages are not decrypted yet, just to be crystal clear.

p2panda-spaces/src/event.rs
```diff
@@ -52,42 +55,68 @@ impl GroupActor {
...
     Application { space_id: SpaceId, data: Vec<u8> },
+
     // @TODO: Could maybe add field to show when the bundle is valid until?
+    /// A key bundle has
```
adzialocha: Incomplete doc string

p2panda-spaces/src/event.rs
```diff
@@ -52,42 +55,68 @@ impl GroupActor {
...
     KeyBundle { author: MemberId },
+
+    /// A membership change occurred on a space.
+    ///
+    /// This event is emitted every the membership of a space changes. Events are silently dropped
```
adzialocha: Missing word after "every"

p2panda-spaces/src/event.rs
```diff
@@ -52,42 +55,68 @@ impl GroupActor {
...
+    /// if the local member is not (yet) a member of the space. For every Event::Groups event a
```
adzialocha: Slightly offtopic, just came into my mind: Will these events be repeated when we become part of the space?

p2panda-spaces/src/event.rs
```diff
@@ -52,42 +55,68 @@ impl GroupActor {
...
     /// Root group actors, can be individuals or groups.
-    pub group_actors: Vec<(GroupActor, Access<C>)>,
+    pub actors: Vec<(GroupActor, Access<C>)>,
 
-    /// Members of this group.
+    /// Members of the group.
     pub members: Vec<(ActorId, Access<C>)>,
+
+    /// All groups for which the group is a child (direct or transitive).
+    pub parents: Vec<ActorId>,
```
adzialocha: parents -> ancestors

p2panda-spaces/src/event.rs
```diff
@@ -52,42 +55,68 @@ impl GroupActor {
...
+
+    /// All groups effected by the associated group change and their members.
+    pub effected_group_members: HashMap<ActorId, Vec<(ActorId, Access<C>)>>,
```
adzialocha, line 98: Not sure: is it affected or effected? Or is both ok?

p2panda-spaces/src/event.rs
```diff
@@ -52,42 +55,68 @@ impl GroupActor {
...
+
+    /// All groups effected by the associated group change and their direct actor members.
+    pub effected_group_actors: HashMap<ActorId, Vec<(GroupActor, Access<C>)>>,
 }
 
 /// Additional context attached to space events.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct SpaceContext<C> {
-    /// The actor who applied this action to the spaces state.
+    /// The actor who applied space action to the spaces state.
```
adzialocha: Maybe a word missing here

p2panda/src/node.rs
```diff
@@ -377,7 +383,9 @@ impl Node {
             .await
             .map_err(|err| CreateStreamError(err.to_string()))?;
 
-        Ok(event_stream(discovery_events))
+        let system_events = self.events_tx.subscribe();
```
adzialocha: All current events in the channel will be ignored here when creating a new subscription, we probably want to keep the original rx around and return it first.

p2panda/src/node.rs
adzialocha: Doc string needs update since its not only networking related events anymore

p2panda/src/node.rs
```diff
@@ -391,9 +399,9 @@ impl Node {
         match self.spaces_manager.group(group_id.into()).await? {
             Some(inner) => {
                 let topic = actor_to_topic(inner.id());
-                let (tx, rx) = self.stream::<NoBody>(topic).await?;
+                let (tx, _rx) = self.stream::<NoBody>(topic).await?;
 
-                Ok(Some(Group::new(inner, tx, rx)))
+                Ok(Some(Group::new(inner, tx, self.events_tx.subscribe())))
```
adzialocha: Same here (missing events in buffer)

p2panda/src/node.rs
```diff
@@ -627,6 +637,8 @@ impl Node {
             .await
             .map_err(|_| CreateSpaceError::AppSend)?;
 
+        // self.events_tx.send(value)
```
adzialocha: Dead code

p2panda/src/streams/event_stream.rs
```diff
@@ -1,33 +1,65 @@
...
+use crate::spaces::GroupActor;
+use crate::spaces::types::InnerGroupEvent;
+
 /// System event.
 ///
 /// System events encompass all network-related events which are not directly associated with a
 /// topic.
 #[derive(Clone, Debug, PartialEq)]
+// @TODO: GroupEvent is a large type and I believe it will remain that way, we might want to box
+// here, but then this leaks out in the public API.
```
adzialocha: The Box should be okay, for most public use the compiler will automatically de-reference it.

p2panda/src/streams/stream.rs
```diff
@@ -754,18 +768,121 @@ pub enum StreamEvent<M> {
...
+        inner: event.clone(),
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum ForwardEvent<M> {
+    System(Box<SystemEvent>),
+    Stream(Box<StreamEvent<M>>),
+}
+
```
adzialocha, line 878: Stream naming is tricky 😹

Another angle could be to see these stream events in two different "scopes", one is global / system level, another is local / topic level.

p2panda/tests/spaces.rs
```diff
@@ -573,3 +620,152 @@ async fn api_validation() -> Result<(), Box<dyn std::error::Error>> {
...
+
+    let panda = p2panda::spawn().await?;
+    let mut panda_system_rx = panda.event_stream().await?;
+
+    let topic = Topic::random();
+
+    // Create a space with only us inside.
+    //
+    // NOTE: Having a space in this test is only required to sync the group operations. When there
+    // is a dedicated API for sharing group operations then this can be replaced.
```
adzialocha: Curious: What is an API for sharing group operations?

adzialocha (approved):
Great work and cool functionality!

I don't have much to add, mostly naming, typos, doc string stuff.

p2panda-spaces/src/event.rs
sandreae: No, the first event from a control message the user will receive is the one (create or add) which brought them into the space.

p2panda-spaces/src/event.rs
sandreae: I've never been sure what the correct approach is here and consistency has suffered as a result. Would be good to take this chance to nail it down. I feel like plural is correct for "Groups" because events of this type relate to the shared groups state, which is made up of _many_ groups. Whereas spaces are isolated from each other and a "Space" stream will only ever contain events for a single space.

It's very possible I'm overcomplicating this, and the plurality doesn't actually matter or impart any additional meaning, it should just be consistent...

sandreae: Do you have any opinion here?

p2panda-spaces/src/event.rs
sandreae: I'm not sure either.... I even looked it up and am still not sure....

p2panda-spaces/src/event.rs
adzialocha: > I've never been sure what the correct approach is here and consistency has suffered as a result. Would be good to take this chance to nail it down. I feel like plural is correct for "Groups" because events of this type relate to the shared groups state, which is made up of _many_ groups. Whereas spaces are isolated from each other and a "Space" stream will only ever contain events for a single space.

Yeah, this stuff is never straight-forward 🤣 

Random thoughts:

The most inner type is called a "GroupEvent" which says in the docs it's related to a single group change.

The enum is more a "switch" between different categories before reaching the most inner thing, so I'd read it as "this event is concerned with x (spaces, groups, dogs, ..)" and I think it's fine to use plural on this level.

p2panda/src/streams/stream.rs
sandreae: I like this naming idea, raised an issue: https://github.com/p2panda/p2panda/issues/1325

p2panda/tests/spaces.rs
sandreae: It doesn't exist yet :sweat_smile: and its exact form is still TBD.

This issue is related: https://github.com/p2panda/p2panda/issues/1282

It might also be addressed by whatever invite token flow we arrive at.

p2panda-spaces/src/event.rs
sandreae: Cool, so all event variants become plural, but the event itself is singular? I can live with that.

p2panda/src/streams/stream.rs
adzialocha: I think this is too small to be made an issue 😅 I'd say we better change it directly or I'll close this issue instantly after this PR is merged 😂

p2panda/src/streams/event_stream.rs
sandreae: I raised an issue for this as boxing the inner value caused the discovery event to now be considered too large by clippy, and then some more changes were required which I think can be taken in a follow-up.

p2panda/tests/spaces.rs
adzialocha: > It doesn't exist yet 😅 and its exact form is still TBD.

I'm still not sure why it would be necessary. The feature we are building is for spaces, not for groups so of course one would need to use a space to use the spaces feature :-D I understand you need to sync to test the groups events API because that's the nature of the spaces system we are building.

I'd write "Having a space in this test is required to sync the group operations" and leave out the second half maybe?

> This issue is related: #1282

Added some rough ideas here as well!

sandreae: Thanks, I replied in the issue. Happy to shorten this comment in any case though.

PR #1288, opened 2026-07-07
Group chat example

sandreae:
Example E2EE group chat CLI app using the spaces API.

## 📋 Checklist

- [x] Add this PR to the _Unreleased_ section in `CHANGELOG.md`
- [x] New files contain a SPDX license header

PR #1317, opened 2026-07-29
Wrote profile example app

ethanfrey:
This was requested on the rocket chat group, so I thought it would be a nice simple example to add, showing how to allow world-readable and user-writable data.

Each node-id / VerifyingKey has one slot which only it can write to. All peers materialize the profiles of all other peers, indexed by author.

This does not effect any code, only an example, so I added no tests. I did test manually and did `cargo fmt` and `cargo clippy`.

Describe your changes here.

## 📋 Checklist

### DIFF
--- a/p2panda/examples/profile.rs
+++ b/p2panda/examples/profile.rs
@@ -1,0 +1,287 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Publically readable profiles with write control.
+//! Simple auth example - approach 1.
+//! This approach lets many users share one topic, but we merge each user's ops to a separate object - they can only edit their own state.
+//! It is very simplistic, not using p2panda-auth groups as we should for multi-device support, but done intentionally to expose the core power of the merge function.
+//!
+//! ## Usage
+//!
+//! ```text
+//! # Start a new todo list, a random id will be generated
+//! cargo run --example profile
+//!
+//! # Join an existing profile list by entering the id
+//! cargo run --example profile -- <id>
+//!
+//! # Type /set-name <name> to set your profile name
+//! /set-name Lothar
+//!
+//! # Type /set-bio <bio> to set your bio
+//! /set-bio King of the Hill People
+//!
+//! # Print current profile list
+//! /show
+//! ```
+//!
+//! ## How does this work?
+//!
+//! This builds on todo and is another example of an "event sourcing" approach: We are creating "events" triggered by
+//! "commands" (see `set-name` and `set-bio` methods) which are then processed (see `process`
+//! method). Every processed event changes our internal state, this is also called "materialisation".
+//!
+//! In distinction from the todo example, we do not store one object, but rather one per device identifier (public key).
+//! This allows us to provide world-readable data, but grant authorization on each "object" to only one key.
+//! Beyond this specific example of auth, it also shows the power of the "merge function" and how it can be customized
+//!
+//! ```plain
+//! [Command: "set-name Bob"] ..
+//!     |                                      |       [Event Processor / Merge Function]
+//!     v                                      |
+//!  [Event: "SetName"]                     => | Database w. materialised state:
+//!                                            |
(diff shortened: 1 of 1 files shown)


p2panda/examples/profile.rs
```diff
@@ -0,0 +1,287 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Publically readable profiles with write control.
+//! Simple auth example - approach 1.
+//! This approach lets many users share one topic, but we merge each user's ops to a separate object - they can only edit their own state.
+//! It is very simplistic, not using p2panda-auth groups as we should for multi-device support, but done intentionally to expose the core power of the merge function.
```
sandreae, line 6: This description could more clearly state what the example is demonstrating, from reading the code I would describe it as something like "maintain a set of per-user profiles". I don't know if it's necessary to state what it _doesn't_ do (multi-device groups etc..), the feature set demonstrated is it's own use case (no multi-writer).

p2panda/examples/profile.rs
```diff
@@ -0,0 +1,287 @@
...
+//!
+//! ## Usage
+//!
+//! ```text
+//! # Start a new todo list, a random id will be generated
```
sandreae, line 11: It's not a todo list anymore.

p2panda/examples/profile.rs
```diff
@@ -0,0 +1,287 @@
...
+    let node = p2panda::spawn().await?;
+
+    println!("PROFILE");
+    println!("⎯⎯⎯⎯⎯");
+    println!("★ topic id: {}", topic);
+    println!("★ my node id: {}", node.id());
+    println!("⎯⎯⎯⎯⎯\n");
+
+    // Establish a publish/subscribe topic stream which will help us to find nodes who are also
+    // interested in the same todo list. We will automatically sync all `TodoEvent` messages with
```
sandreae, line 206: Not a todo list and no TodoEvents.

sandreae (commented):
Thanks for this example @ethanfrey 👋 I added some comments, mostly about the framing of the example in the description and some stale code comments.

I understand you wanted to address a specific use case tied to world-readable data with basic single-writer access controls with this example, and I think it does that well. I would be quite interested in an example which uses the spaces API as well, which although is not world-readable, could demonstrate powerful design patterns as then group management, e2e encryption and partially-ordered multi-author event streams could be leveraged. 

For example I'd be very excited to see something like the chat example here https://github.com/p2panda/p2panda/pull/1288 with the addition that:
- every space user is represented by a device-group
- any member of the device group with "manage" access can write to a "device-group profile" with LWW decided by timestamp

I know this wouldn't satisfy the "world-readable" requirement for this example though, just floating it as potential follow-up.

adzialocha:
Thank you for this PR @ethanfrey 🥳 

I think it's cool to have more examples highlighting different use-cases one can already achieve even with the "simpler" methods or using bare, single-writer append-only logs.

The advantages of logs are that they are totally ordered (no need to use timestamp for enforcing a total order) and single-writer so only one author is authorized to maintain one log which gives us a simple authorization primitive.

For me that would already be enough to show in an example but to make it a bit than "another" LWW-CRDT example (like todo.rs) I'd suggest to maybe extend this to a simple SSB (Secure Scuttlebutt) clone? You already have a lot down with the profiles.

Maybe you are not familiar with SSB but it roughly goes like this:

1. Every member maintains their own, single-writer append-only log to publish "posts" or updates to their "profile". This can be either expressed as two logs or one log with different message types inside. The SSB-way would be to have one single log
2. Optional: Other members "follow" each other. SSB had a more special approach here but for the example it could be enough to use the verifying key of each member as a topic. So you "follow" somebody by subscribing to their topic stream where they publish their log.

So the app is something like a very simple community "social media" group where everyone maintains their own log of posts.

This is just an idea, let me know what you think and if it matches with what you had in mind (also refering to @sandreae's comments).

PR #1319, opened 2026-07-30
Register hooks on Endpoint

mycognosist:
Here we add a `hooks()` method on the `Builder` which allows the user to register custom hooks on the endpoint (for example, to monitor connections or reject them based on access control lists).

We also introduce a custom trait to make iroh's `EndpointHooks` trait `Dyn` compatible.

A minimal connection monitor hook is implemented and used in the `establish_connection()` test.

## 📋 Checklist

- ~~[ ] Link this PR to any issues it closes~~

### DIFF
--- a/p2panda-net/src/iroh_endpoint/builder.rs
+++ b/p2panda-net/src/iroh_endpoint/builder.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashSet;
 
+use iroh::endpoint::EndpointHooks;
 use p2panda_core::SigningKey;
 use ractor::thread_local::{ThreadLocalActor, ThreadLocalActorSpawner};
 
@@ -9,6 +10,7 @@ use crate::address_book::AddressBook;
 use crate::iroh_endpoint::actors::{IrohEndpoint, IrohEndpointArgs};
 use crate::iroh_endpoint::api::{Endpoint, EndpointError};
 use crate::iroh_endpoint::config::IrohConfig;
+use crate::iroh_endpoint::hooks::EndpointHooksList;
 use crate::{DEFAULT_NETWORK_ID, NetworkId};
 
 pub struct Builder {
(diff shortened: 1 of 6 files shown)


p2panda-net/src/iroh_endpoint/hooks.rs
```diff
@@ -0,0 +1,91 @@
...
+impl EndpointHooks for EndpointHooksList {
+    async fn before_connect(
+        &self,
+        remote_addr: &EndpointAddr,
+        alpn: &[u8],
+    ) -> BeforeConnectOutcome {
+        for hook in self.inner.iter() {
+            match hook.before_connect(remote_addr, alpn).await {
+                BeforeConnectOutcome::Accept => continue,
+                reject @ BeforeConnectOutcome::Reject => {
```
adzialocha, line 72: Woah, I assume this comes from the iroh code-base, never seen an [Identifier Pattern](https://doc.rust-lang.org/reference/patterns.html#identifier-patterns), didn't know this exists 🙀

p2panda-net/src/iroh_endpoint/builder.rs
```diff
@@ -68,17 +72,35 @@ impl Builder {
...
+    /// Endpoint hooks intercept the connection establishment process of an iroh `Endpoint`.
+    ///
+    /// Multiple hooks can be registered with the `Endpoint` and will be called in their order of
+    /// registration. If any of the hooks result in a connection rejection, all further processing
+    /// is aborted and subsequent hooks will not be called.
+    ///
+    /// See iroh's [`EndpointHooks`] documentation for further details.
+    ///
+    /// [`EndpointHooks`]: https://docs.rs/iroh/latest/iroh/endpoint/trait.EndpointHooks.html
+    pub fn hooks(mut self, hook: impl EndpointHooks + 'static + Clone) -> Self {
```
adzialocha, line 86: Doesn't necessarily need changing now since it's the only hook we have, but I wonder if the `hooks` naming is too generic. In the context of iroh it makes sense as their endpoint only does one  thing, but for the general p2panda stack we will have different hooks all over the place.

We should consider naming this `on_connection` etc., so there's possibility for hooks in other places, such as `on_pre_ingest`, `on_sync_started` etc.

adzialocha (approved):
Amazing work, well done figuring out the dyn-compatibility!

PR #1321, opened 2026-07-30
Introduce `ConnectionAuthoriser` and integrate it into the `Node`

mycognosist:
This first iteration of the `ConnectionAuthoriser` maintains an allowlist and
blocklist which are used to accept or reject connections before they are
established. The `ConnectionAuthoriser` implements the `EndpointHooks` trait to
facilitate these connection interceptions. It also exposes an event stream.
A basic API is exposed for allowing or blocking a node, as well as allowing or
blocking a node-topic combination.

`ConnectionAuthoriserEvents` are emitted on the `SystemEvents` stream, alongside
`DiscoveryEvents`. The `ConnectionAuthoriser` has been integrated into the `Node`.
Two basic tests showcases the blocking functionality.

## 📋 Checklist

~~- [ ] Link this PR to any issues it closes~~

### DIFF
--- a/p2panda-net/src/connection_authoriser.rs
+++ b/p2panda-net/src/connection_authoriser.rs
@@ -5,6 +5,7 @@ use std::collections::HashSet;
 use std::fmt::Display;
 use std::sync::Arc;
 
+use iroh::endpoint::Side;
 use p2panda_core::{Topic, VerifyingKey};
 use thiserror::Error;
 use tokio::sync::RwLock;
@@ -12,7 +13,9 @@ use tokio::sync::broadcast::{self, Receiver, Sender};
 use tracing::warn;
 
 use crate::NetworkId;
-use crate::iroh_endpoint::{BeforeConnectOutcome, EndpointAddr, EndpointHooks};
+use crate::iroh_endpoint::{
+    AfterHandshakeOutcome, BeforeConnectOutcome, EndpointAddr, EndpointHooks,
+};
 use crate::utils::{ShortFormat, to_verifying_key};
 
 /// Connection authoriser mode for determining how connections are accepted and rejected.
@@ -25,18 +28,50 @@ enum ConnectionAuthoriserMode {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ConnectionRole {
+    Acceptor,
+    Initiator,
+}
+
+impl Display for ConnectionRole {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ConnectionRole::Acceptor => write!(f, "inbound"),
+            ConnectionRole::Initiator => write!(f, "outbound"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub enum ConnectionAuthoriserEvent {
-    ConnectionBlocked { node: VerifyingKey },
-    ConnectionAllowed { node: VerifyingKey },
-    TopicBlocked { topic: Topic, node: VerifyingKey },
-    TopicAllowed { topic: Topic, node: VerifyingKey },
+    Blocked {
+        node: VerifyingKey,
+        role: ConnectionRole,
+    },
+    Allowed {
+        node: VerifyingKey,
+        role: ConnectionRole,
+    },
+    TopicBlocked {
+        topic: Topic,
+        node: VerifyingKey,
+    },
+    TopicAllowed {
+        topic: Topic,
+        node: VerifyingKey,
+    },
 }
 
 impl Display for ConnectionAuthoriserEvent {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            ConnectionAuthoriserEvent::ConnectionBlocked { node } => {
-                write!(f, "blocked connection attempt to {}", node.fmt_short(),)
+            ConnectionAuthoriserEvent::Blocked { node, role } => {
+                write!(
+                    f,
+                    "blocked {} connection attempt to {}",
+                    role,
+                    node.fmt_short(),
+                )
             }
             ConnectionAuthoriserEvent::TopicBlocked { topic, node } => {
                 write!(
@@ -46,8 +81,13 @@ impl Display for ConnectionAuthoriserEvent {
                     topic.fmt_short()
                 )
             }
-            ConnectionAuthoriserEvent::ConnectionAllowed { node } => {
-                write!(f, "allowed connection attempt to {}", node.fmt_short(),)
+            ConnectionAuthoriserEvent::Allowed { node, role } => {
+                write!(
+                    f,
+                    "allowed {} connection attempt to {}",
+                    role,
+                    node.fmt_short(),
+                )
             }
             ConnectionAuthoriserEvent::TopicAllowed { topic, node } => {
                 write!(
@@ -125,12 +165,10 @@ impl ConnectionAuthoriser {
         //
         // This is primarily to prevent flooding the logs with warnings when events are emitted but
         // no event stream subscription exists.
-        if connection_authoriser.tx.receiver_count() > 0 {
-            if let Err(err) = connection_authoriser.tx.send(event) {
-                warn!("failed to send authoriser event: {}", err)
-            } else {
-                println!("successfully sent authoriser event")
-            }
+        if connection_authoriser.tx.receiver_count() > 0
+            && let Err(err) = connection_authoriser.tx.send(event)
+        {
+            warn!("failed to send authoriser event: {}", err)
         }
     }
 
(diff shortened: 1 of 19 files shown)


p2panda/src/authoriser.rs
```diff
@@ -0,0 +1,136 @@
...
+        match self {
+            AuthoriserEvent::Blocked(node) => {
+                write!(f, "blocked connection attempt to {}", node.fmt_short(),)
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Authoriser {
```
adzialocha: Early thought: I find the `Authoriser` naming a bit too generic maybe. _What_ is being authorised and on which layer?

We will have a lot of different authorization concepts in different places and even if it's just one for now it's unclear where this one belongs.

I would suggest something like `EndpointConnectionAuthoriser` or `ConnectionAuthoriser`, I think this implies enough that this is about a point-to-point connection, authorizing it based on the remote parties verifying key / node id.

mycognosist: @mycognosist Are you OK with keeping the `Authoriser` naming now that it has moved into `p2panda-net`? This could always be imported as `ConnectionAuthoriser` in `p2panda` to differentiate it from other (future) authorisers.

adzialocha: I would already call it `ConnectionAuthoriser` in `p2panda-net`, just because it describes what it is and what it does directly

p2panda-net/src/sync/actors/manager.rs
```diff
@@ -558,6 +590,23 @@ where
             .await
             .map_err(|err| iroh::protocol::AcceptError::from_err(err))?;
 
+        let allow = self
+            .connection_authoriser
+            .can_connect_on_topic(node_id, topic)
+            .await;
+
+        if !allow {
```
sandreae, line 598: Should we also emit events here as we do when authorising connections we initiated ourselves above?

p2panda-net/src/connection_authoriser.rs
```diff
@@ -0,0 +1,320 @@
...
+enum ConnectionAuthoriserMode {
+    /// Allow all connections except for nodes which have been explicitly blocked.
+    Permissive,
+    /// Block all connections except for nodes which have been explicitly allowed.
+    Restrictive,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ConnectionAuthoriserEvent {
+    ConnectionBlocked { node: VerifyingKey },
```
sandreae: What do you think of renaming these two variants with the "Global" prefix instead of "Connection"?

p2panda-net/src/connection_authoriser.rs
```diff
@@ -0,0 +1,320 @@
...
+    ConnectionAllowed { node: VerifyingKey },
+    TopicBlocked { topic: Topic, node: VerifyingKey },
+    TopicAllowed { topic: Topic, node: VerifyingKey },
+}
+
+impl Display for ConnectionAuthoriserEvent {
```
sandreae, line 65: Do we want to differentiate between connection attempts initiated by us or a remote in the events? Could be useful for debugging and testing. I'd be fine with just raising an issue.

p2panda/src/node.rs
```diff
@@ -91,10 +93,14 @@ impl Node {
     ) -> Result<Self, SpawnError> {
         let forge = OperationForge::new(credentials.clone(), store.clone());
 
+        let connection_authoriser = ConnectionAuthoriser::new();
+        connection_authoriser.permissive().await;
```
sandreae, line 97: Was wondering about the API for configuring the permissive / restrictive modes. Also if this should be globally set or per-topic. Again, probably just something for an issue.

p2panda-net/src/connection_authoriser.rs
```diff
@@ -0,0 +1,320 @@
...
+            ConnectionAuthoriserMode::Restrictive => {
+                connection_authoriser.global_allowlist.contains(&node)
+            }
+        }
+    }
+}
+
+impl EndpointHooks for ConnectionAuthoriser {
+    // Runs before an outgoing connection begins.
+    async fn before_connect(
```
sandreae, line 261: Is there another method we need to implement in order to run the hook before an incoming connection is established?

sandreae (changes requested):
Looking great 😃 I have a few points, some of which can just be raised as issues for later.

p2panda/src/node.rs
```diff
@@ -706,6 +719,25 @@ impl Node {
     ) -> Result<(), NetworkError> {
         self.network.insert_bootstrap(node_id, relay_url).await
     }
+
+    /// Allows all connection attempts with the given node.
```
sandreae, line 748: We should note somewhere (probably in the method docs here) that the allow/block lists are not persisted now, so they need to be re-populated between process restarts.

sandreae (changes requested):

p2panda/src/node.rs
```diff
@@ -706,6 +719,25 @@ impl Node {
...
+    pub async fn allow(&self, node_id: NodeId) {
+        self.connection_authoriser.allow(node_id).await;
+    }
+
+    /// Allows all connection attempts with the given node for a single topic.
```
sandreae, line 756: And actually I wonder if these methods should actually live on the `StreamPublisher` and `Space` objects? That would be quite a sweet API. It would also give us a place to explain the "default" behavior which will come with using a space (removed members blocked etc...).

mycognosist: That's an interesting question. Do you imagine having the topic-node block / allow exposed on the `StreamPublisher` and `Space`, with a separate global block / allow API on `Node`? Then the global state provides a default which can be overruled on a per-stream / space basis.

p2panda-net/src/connection_authoriser.rs
mycognosist: Yes, good catch! I forgot to implement the `after_handshake()` method. I'll include that in this PR.

p2panda-net/src/lib.rs
```diff
@@ -315,6 +317,7 @@ pub const DEFAULT_NETWORK_ID: NetworkId = [
 ];
 
 /// Hash the concatenation of the given protocol- and network identifiers.
+#[cfg(any(feature = "gossip", feature = "iroh_endpoint"))]
```
adzialocha: It's a private method and not very complex, so likely not giving much benefit to gate this behind a feature flag? Maybe the compiler would even factor it out automatically.

adzialocha (changes requested):
I would like to suggest seperating out the sync authorisation from the connection authorisation.

They are clearly two different layers (one is part of the sync protocol, above QUIC + TLS, another one is on the QUIC + TLS layer) and also code-wise I think it shows: the sync protocol is merely doing a lookup into the struct to determine if it should close the connection and not much more. Additionally I find it confusing to call a "connection authoriser" on the `LogSync` builder if it's not related to connections. Something like `EndpointBuilder::hooks` and `LogSync::hooks` is cool.

It shouldn't be too hard to keep them separated? I suggest to do that in an own PR and keep this one only around the `IrohEndpoint` module.

Otherwise this looks good to me, good work!

mycognosist:
> I would like to suggest seperating out the sync authorisation from the connection authorisation.

This sounds good to me; your justification for the separation makes sense.

> It shouldn't be too hard to keep them separated? I suggest to do that in an own PR and keep this one only around the `IrohEndpoint` module.

Are you suggesting removing the sync authorisation aspect from this PR? Or keeping it here for now and then separating them in the next PR?

p2panda/src/node.rs
sandreae: Yeah, that's what I was imagining. It feels like a nice API and consistent with the per-topic/space APIs (publish, import, etc...) in general.

adzialocha:
> Are you suggesting removing the sync authorisation aspect from this PR? Or keeping it here for now and then separating them in the next PR?

The first, have one PR for adding connection authorisation functionality to `Endpoint` and then a second PR adding an authorisation layer to `LogSync` 👍🏻 

But no worries if that's too complicated, you can handle it whatever works best for ya of course.

mycognosist:
> But no worries if that's too complicated, you can handle it whatever works best for ya of course.

Wrangling git history will really slow me down so I'll open issues for outstanding comments / work and perform the split in a separate PR.

p2panda-net/src/connection_authoriser.rs
mycognosist: These events will be split into different types in a follow-up PR (entirely separating connection and sync authorisation events).

p2panda-net/src/sync/actors/manager.rs
mycognosist: Good idea. I'll add that in the next PR which will separate the connection and sync authorisers.

p2panda-net/src/connection_authoriser.rs
mycognosist: I've added a `ConnectionRole` to the `Allow` and `Block` events (I'll apply the same treatment to the sync-specific authoriser events in a follow-up PR).

p2panda-net/src/connection_authoriser.rs
mycognosist: The `after_handshake()` method has now been implemented as part of this PR; this ensures we catch the inbound connections which we want to block / drop.

adzialocha (approved):

sandreae (approved):

PR #1322, opened 2026-07-30
Rewrite queries in p2panda-store to use safer QueryBuilder, update sqlx

matthiasbeyer:
Closes https://github.com/p2panda/p2panda/issues/1237

This patchset rewrites some of the queries in the `p2panda-store` crate to use the safe(r) `QueryBuilder` instead of doing unsafe string interpolation for building queries.

The patches run successfully through the tests, the patch for updating the dependency is added on top.

## 📋 Checklist

### DIFF
--- a/p2panda-store/Cargo.toml
+++ b/p2panda-store/Cargo.toml
@@ -62,7 +62,7 @@ p2panda-encryption = { workspace = true, default-features = true, optional = tru
 rand = { workspace = true, optional = true }
 rand_chacha = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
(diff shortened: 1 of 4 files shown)


matthiasbeyer:
Updated because one query had an issue that was not found via the testsuite. (I accidentally kept a `"` inside the query where it did not belong to).

adzialocha (approved):
Thank you so much @matthiasbeyer ! This is another really helpful contribution from our TODO list. Really cool we can use sqlx 0.9.0 now.

I don't think that we had any real security issues using string interpolation as the types used are all strictly validated (public keys, BLAKE3 hashes, etc.) but this sets a good baseline for future store work.

PR #1323, opened 2026-07-30
Use helper type to simplify sqlx query building in p2panda_store::address_book

matthiasbeyer:
This PR _includes_ the patches from https://github.com/p2panda/p2panda/pull/1322 and builds upon them.

Thus, #1322 should be looked at first (IMHO).

---

This patch creates a helper type within `p2panda-store/src/address_book/sqlite.rs` to simplify the query construction. For more details please have a look at the commit message.

This patch is mainly for demonstration purposes how helper types and `sqlx::Type`/`sqlx::Encode`/`sqlx::Decode` can be used to simplify the sqlx interfacing (IMHO of course), as I got the impression that there's room for improvement while I had a look at the codebase.

I expect a _small_ improvement in runtime as well, as we do not have to build a list twice in most queries (one for the return value of the query, the second one for parsing the CBOR in the first list). I would suspect that the unwrapping of the helper type is likely optimized to a no-op by LLVM :laughing: .

---

Created as draft because #1322 is a prerequisite for this and I am not sure whether changes like this are wanted - and if they are I could look for other places where stuff like this could be optimized before merging.

## 📋 Checklist

### DIFF
--- a/p2panda-store/Cargo.toml
+++ b/p2panda-store/Cargo.toml
@@ -62,7 +62,7 @@ p2panda-encryption = { workspace = true, default-features = true, optional = tru
 rand = { workspace = true, optional = true }
 rand_chacha = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
-sqlx = { version = "0.8.6", features = ["runtime-tokio", "sqlite"], optional = true }
+sqlx = { version = "0.9.0", features = ["runtime-tokio", "sqlite"], optional = true }
 thiserror = { workspace = true, optional = true }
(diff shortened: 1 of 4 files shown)


adzialocha:
I've merged https://github.com/p2panda/p2panda/pull/1322 ✨

adzialocha (approved):
> This patch is mainly for demonstration purposes how helper type.

Thank you! We really like places where we can demonstrate better patterns to follow in the future and using the `QueryBuilder` seems like one of them. Didn't know about this one yet!

PR #1332, opened 2026-08-04
Safe key bundle registration by cross-signing X3DH identity- and verifying-keys

### LINKED ISSUE #1210: Safe type to deal with member information coming from a side-channel (QR code etc.)
Registering members via a side-channel (QR code) should be _safe_ in the Node API (identity- and signing key combination is authentic), we should provide a type for that.

adzialocha:
This PR changes the notion from key-bundles towards "members" objects (tbd if this is still a good name).

Member objects contain all information required to identify a member and verify that the associated X3DH key-bundle is checked and was authored / claimed by that member. The latter is now provided through a cross-signing process.

Some additional naming changes have been made on the more public methods, using the Member naming instead of KeyBundle.

Closes: https://github.com/p2panda/p2panda/issues/1210

## 📋 Checklist

### DIFF
--- a/p2panda-spaces/src/member.rs
+++ b/p2panda-spaces/src/member.rs
@@ -1,27 +1,81 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+use p2panda_core::Signature;
+use p2panda_core::cbor::encode_cbor;
+use p2panda_encryption::Rng;
+use p2panda_encryption::crypto::xeddsa::{XSignature, xeddsa_sign, xeddsa_verify};
 use p2panda_encryption::key_bundle::{KeyBundleError, LongTermKeyBundle};
 use p2panda_encryption::traits::KeyBundle;
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
 
-use crate::MemberId;
+use crate::{Credentials, MemberId};
 
-/// A group member and their long-term key bundle.
-#[derive(Debug)]
+/// A group member and their associated long-term key bundle.
+///
+/// Authenticity between the verifying key and X3DH identity key is assured through cross-signing.
+/// Through this we assure that the associated key bundle belongs to the verifying key as well.
+/// Additional checks are performed against the key bundle's pre-key and it's lifetime.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Member {
+    /// Verifying key of this space member, used as an identifier.
     id: MemberId,
+
+    /// Associated X3DH key bundle with used identity key and pre-key.
+    ///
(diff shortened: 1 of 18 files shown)


p2panda-spaces/src/member.rs
```diff
@@ -1,41 +1,227 @@
...
+    /// [Identity Key] --> [Pre-Key + Member-Id]
+    /// ```
+    cross_signature_1: XSignature,
+
+    /// Signature proves that the X3DH identity key and associated key-bundle is authentic (cross
+    /// signed).
+    ///
+    /// ```text
+    /// [Verifying Key] --> [Pre-Key + Identity Key]
+    /// ```
```
sandreae, line 47: In the code where signatures are generated the bundle bytes are what is signed over, that's not clear from the doc string here where it looks like both cross signatures are both signing over the pre-key bytes.

p2panda-spaces/src/message.rs
```diff
@@ -152,13 +153,11 @@ where
...
-    ///
-    /// Note: Applications should check if the key bundle was authored by the sender.
-    KeyBundle { key_bundle: LongTermKeyBundle },
+    /// System message, contains key bundle of the given member.
+    Member(Member),
 
-    /// System message containing an auth control message.
-    Auth {
+    /// System message containing group control message.
+    Group {
```
sandreae, line 160: 🙏 approaching consistent naming scheme....

p2panda/src/streams/stream.rs
```diff
@@ -576,9 +576,8 @@ where
                     }
                 }
 
-                p2panda_spaces::Event::KeyBundle { author } => {
-                    forward_events.push(ForwardEvent::topic_stream(StreamEvent::KeyBundle(*author)))
-                }
+                p2panda_spaces::Event::Member(member) => forward_events
+                    .push(ForwardEvent::topic_stream(StreamEvent::Member(member.id()))),
```
sandreae: Do you think these should be forwarded on the system event stream? As a key bundle can be used in any space (doesn't matter which space it arrived on).

sandreae (changes requested):
Great work, only a couple of small comments from me 👍

p2panda-spaces/src/member.rs
adzialocha: I tried to focus here on the underlying thing which get authenticated, not the implementation details. Signing the key bundle is essentially authenticating the pre-key + identity-key. Signing the pre-key signature doesn't add anything anymore.

p2panda/src/streams/stream.rs
adzialocha: Good idea! Let's do that.

Issue #1314, opened 2026-07-25
Validate membership change actions in `Group` API


sandreae:
Same as the validation present in the `Space` API.

PR #1336, opened 2026-08-04
Membership change validation in `Group` API

sandreae:
closes: #1314 

## 📋 Checklist

- [X] Add tests that cover your changes

### DIFF
--- a/p2panda-spaces/src/group.rs
+++ b/p2panda-spaces/src/group.rs
@@ -8,6 +8,7 @@ use std::fmt::Debug;
 use p2panda_auth::Access;
 use p2panda_auth::group::GroupAction;
 use p2panda_auth::traits::{Conditions, Operation};
+use p2panda_core::VerifyingKey;
 use p2panda_core::traits::ShortFormat;
 use p2panda_store::Transaction;
 use p2panda_store::groups::GroupsStore;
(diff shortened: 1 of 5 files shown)


p2panda-spaces/src/group.rs
```diff
@@ -71,6 +72,11 @@ where
         }
     }
 
+    /// Verifying key of the local actor.
+    pub fn me(&self) -> VerifyingKey {
+        self.manager.id()
+    }
```
adzialocha, line 78: We have another `me` method in the high-level `Manager` API which returns a key bundle / member info struct.

Maybe we rename this to `my_id` or `member_id`?

p2panda/src/spaces/group.rs
```diff
@@ -282,3 +322,35 @@ pub enum GroupError {
     #[error(transparent)]
     CreateStream(#[from] CreateStreamError),
 }
+
+#[derive(Debug, Error)]
+pub enum AddGroupMemberError {
+    #[error("failed validation adding {actor} to group {group_id}: {err}", actor = actor.fmt_short(), group_id = group_id.fmt_short())]
```
adzialocha: These exceed the max. line length of the Rust formatted but not get caught since some macros mess with it. Let's format them manually then.

adzialocha (approved):
Very nice! Looking all good

PR #1337, opened 2026-08-04
Automatically renew expired key-bundles and publish member messages more efficiently

adzialocha:
This PR improves the efficiency of published "member" messages which contain X3DH key bundles, used for initial key-agreement.

Messages are only published when key bundles expire or do not exist yet. We introduce a background task here which gets spawned on application start to assure that this is happening, even when the applications runs for many weeks.

Nodes who are offline will pick up the latest key bundles during sync. Online nodes receive them via "live-mode".

Another important aspect is the association of member logs with a space. This is done when a space is created or joined. Additionally we need to associate _other members_ logs by observing spaces events with a new processor hook, in particular CREATE and ADD events.

## 📋 Checklist

### DIFF
--- a/p2panda/src/node.rs
+++ b/p2panda/src/node.rs
@@ -551,25 +551,6 @@ impl Node {
             .unwrap_or(InnerSpace::new(self.spaces_manager.clone(), space_id));
         let (tx, rx) = self.stream_from::<M>(topic, from).await?;
 
-        // Publish one key bundle whenever we subscribe to a space.
-        //
-        // @TODO: this is a rather naive approach, we likely want some (configurable?) service
-        // that periodically publishes key bundles.
-        let message = self.spaces_manager.key_bundle_message().await?;
-
-        let operation = message.into_operation();
-        let processed = tx
-            .import_local(futures_util::stream::once(async { operation }))
-            .await?;
-
-        // Wait until processing the events has finished.
-
-        // TODO: Would be good to get an error / report here if processing the imported operations
-        // failed. This error so far only tells us that the channel broke down.
-        if processed.await.is_err() {
-            panic!();
-        }
-
         // Spawn per-space repair background task.
         let repair_task = RepairTask::spawn(
             inner.id(),
@@ -619,9 +600,6 @@ impl Node {
         let (tx, rx) = self.stream::<M>(topic).await?;
 
         // Publish one key bundle whenever we create a space.
-        //
-        // @TODO: this is a rather naive approach, we likely want some service that periodically
-        // publishes key bundles.
         let key_bundle_message = self.spaces_manager.key_bundle_message().await?;
 
         // Issue the events to create a space.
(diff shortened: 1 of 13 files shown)


p2panda/src/node.rs
```diff
@@ -485,47 +511,39 @@ impl Node {
...
-            .await?
-            .unwrap_or_default();
-
-        for group_id in y.groups_global() {
-            debug!(
-                group_id = group_id.fmt_short(),
-                space_id = space_id.fmt_short(),
-                "associate group log with space topic"
-            );
+            // Associate the space topic with our own member / key bundle logs.
```
sandreae, line 549: Can this association be taken care of in the key bundle task now?

sandreae (approved):
Looks great to me, loving the hooks. I have one question but if it doesn't lead to anything I'm happy to accept this as is. Well done wrapping you're head around the requirements a(nother) particularly knotty part of of the system!

p2panda/src/node.rs
adzialocha: Hmn. Good question.

The key bundle task .. is for key bundles 😅 so that would require changing it to something more powerful. More like a general "member service" you can talk to doing different things. It currently also runs only every x minutes and would need to be triggered on space start / task instead. I think the current "repair" task is a little bit like that right now.

But if you mean the new association hook, yeah, that could indeed do that work for us here already as it's just blindly associating all initial members and the added ones, which includes us. We would then need to add pushing into live-mode sessions as well ..

I'm imagining something like a) a dedicated "import-like" method on `Node` which allows me to process messages & push them into live-mode sessions for every active topic stream - or filtered to only go into active spaces streams, or only one single stream (can be an enum where I select the "mode"). We are close to this, it's just a bit too entangled with non-related parts of the stack. b) Consolidated logic of associating member AND group logs to a space. Could just be a unified "spaces hook" which has access to that import handle.

adzialocha: > We would then need to add pushing into live-mode sessions as well ..

Oh, actually need to do that anyhow as my failing E2E test shows 🤣

adzialocha:
I believe what's still missing here and might cause unexpected behaviour is that we need to push the operations of newly associated member logs into live mode sessions as well.

This is not breaking anything, it'll be a worse experience in some cases, so I think we can do that in a separate PR. On top I'd like to look at it again _after_ all group association changes are merged as well.

p2panda/src/node.rs
adzialocha: Actually, that's an important one as well (see my comment below), but wasn't the reason. The reason was missing associations when I'm adding someone locally. Please have another look!

sandreae: Really nice observations!

I've been reflecting a bit on what the repair task is currently doing (obviously too much) and how to pull it apart a bit. Think it's related to the changes you're describing too. Essentially it is doing two things:

- sync: making sure all subscribed nodes receive the operations they need (by log association and operation forwarding)
- repair: incorporating "late arriving" group changes into the space state

This is a really clear separation which could be implemented (don't know if it's tasks or hooks yet) and the former is what you describe above in point `b)`.

sandreae (approved):

PR #1342, opened 2026-08-05
Add AccessLevel helper functions

matthiasbeyer:
Adds some utility functions for the `AccessLevel` type

## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/access.rs
+++ b/p2panda-auth/src/access.rs
@@ -28,6 +28,30 @@ pub enum AccessLevel {
     Manage,
 }
 
+impl AccessLevel {
+    #[inline]
+    pub fn can_pull(&self) -> bool {
+        true
+    }
+
+    #[inline]
+    pub fn can_read(&self) -> bool {
+        matches!(self, AccessLevel::Read)
+            || matches!(self, AccessLevel::Write)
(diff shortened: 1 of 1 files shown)


p2panda-auth/src/access.rs
```diff
@@ -28,6 +28,30 @@ pub enum AccessLevel {
...
 
+impl AccessLevel {
+    #[inline]
+    pub fn can_pull(&self) -> bool {
+        true
+    }
+
+    #[inline]
+    pub fn can_read(&self) -> bool {
+        matches!(self, AccessLevel::Read)
```
sandreae, line 39: As `AccessLevel` implements `PartialOrd` all of these assertions can be written as so:

```rust
// Access level allows read.
self >= AccessLevel::Read
```

sandreae: This would better demonstrate that a higher access level also contains all lower ones.

matthiasbeyer: So you would rather not have these fns implemented? I am of course ok with that! :smile:

sandreae: Ah, I didn't actually mean that, just that then inner assertion statements could use >= comparators. Happy for this API to be introduced :+1:

Issue #1344, opened 2026-08-10
Inform user when a space member removal "invalidates" concurrently published application messages


sandreae:
This involves looking up any application messages which were published by a removed member and are:
- concurrent to the space membership operation incorporating the groups change into the space
- the `proof` they carry is referring to a concurrent part of the groups operation graph

PR #1345, opened 2026-08-10
Inform users of application messages "invalidated" by concurrent removal

sandreae:
This change adds an `effected_application_messages` field to remove and demote events which
contains the ids of all application messages by the removed author which were published
concurrently to the remove message itself.

There are two possible concurrency cases to detect, one can be inferred by inspecting the "proof"
(a concrete point in the groups operation graph) an application message carries, the other by
observing concurrency in the space operation graph itself.

## Case 1: proof is concurrent

The application message is referring to a proof which is entirely concurrent to the target
remove operation.

```
   Remove A                              ...          
      │                                   │           
      │                                   ▼           
      │   Add B ◄─────────────────────── App          
      │    │            proof             │           
      │    ▼                              ▼           
      └─► Add A                          ...          
           │                              │           
           ▼                              ▼           
         Group                          Space         
                                                                                                         
```

## Case 2: space operation is concurrent                                                    
                                    
The application message is published concurrently to the membership message which incorporated
the target remove into the space.
```         
    Remove A ◄─────────────────────── Membership         
      │                                   │              
      │                                   ▼              
      │   Add B ◄──────────────────── Membership      App
      │    │                              │            │
      │    ▼                              ▼            │
      └─► Add A ◄───────────────────  Membership ◄─────┘
           │                              │              
           ▼                              ▼              
         Group                          Space      
```      

closes: #1344 

## 📋 Checklist

### DIFF
--- a/p2panda-auth/src/graph.rs
+++ b/p2panda-auth/src/graph.rs
@@ -57,7 +57,7 @@ where
 ///
 /// Operations are considered concurrent if they are neither predecessors nor successors of the
 /// target operation.
-fn concurrent_operations<OP>(graph: &DiGraphMap<OP, ()>, target: OP) -> HashSet<OP>
+pub fn concurrent_operations<OP>(graph: &DiGraphMap<OP, ()>, target: OP) -> HashSet<OP>
 where
     OP: OperationId + Ord,
 {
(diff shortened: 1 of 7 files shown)


sandreae:
Realised this approach doesn't take account of members which may have become a member again due to a concurrent removes arriving (and invalidating the remove targeting then). To handle this it would be better to provide users with message IDs for all concurrent invalidated messages (for all authors) when a remove is processed, rather than just for the author being removed.

PR #1355, opened 2026-08-15
p2panda: optimize building LogId from SpaceId/VerifyingKey on stack

matthiasbeyer:
~~Please read the commit message, it should explain everything in detail and if not, I am happy to expand it!~~ :heart: 

---

Edit: Please read the message below as well as the commit message(s) :wink:

### DIFF
--- a/p2panda/src/spaces/forge.rs
+++ b/p2panda/src/spaces/forge.rs
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
+use std::io::Write;
 
 use p2panda_auth::group::GroupAction;
 use p2panda_core::cbor::encode_cbor;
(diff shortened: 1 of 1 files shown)


p2panda/src/spaces/forge.rs
```diff
@@ -207,22 +207,28 @@ impl p2panda_spaces::Forge<AuthCapabilities> for OperationForge {
 
 fn space_log_id(space_id: SpaceId) -> LogId {
     LogId::digest(&{
-        let mut bytes = Vec::new();
-        bytes.extend_from_slice(space_id.as_bytes());
+        let mut bytes =
+            [0; p2panda_core::identity::VERIFYING_KEY_LEN + SPACE_CONTROL_MESSAGE.len()];
+        debug_assert!(bytes.len() == space_id.as_bytes().len() + GROUP_CONTROL_MESSAGE.len());
+
+        bytes.clone_from_slice(space_id.as_bytes());
```
matthiasbeyer: WAIT.

This does not work as expected. I am looking into that right now.

matthiasbeyer: Nice that CI found it btw :laughing:

matthiasbeyer:
Converted to draft because I can do a bit more in this PR.

matthiasbeyer:
Okay here we go.

I found some stuff that needed to be fixed with my first patch, most importantly that I missed the opportunity to also optimize the `SPACE_APPLICATION_MESSAGE` bits, but also that I used the wrong constant in one of the `debug_assert!()`s.

I added another patch to reduce the code size with a helper macro. I first tried to use a `const fn`, but I found that its simpler to use a `macro_rules!()`.

If that second patch is wanted (I can understand if not), I am more than happy to squash everything into one patch!

I am of course also here to squash away these `fixup!` commits before merge!

Tell me what you think, marking this as ready for review now.

adzialocha:
> Okay here we go.
> 
> I found some stuff that needed to be fixed with my first patch, most importantly that I missed the opportunity to also optimize the `SPACE_APPLICATION_MESSAGE` bits, but also that I used the wrong constant in one of the `debug_assert!()`s.
> 
> I added another patch to reduce the code size with a helper macro. I first tried to use a `const fn`, but I found that its simpler to use a `macro_rules!()`.
> 
> If that second patch is wanted (I can understand if not), I am more than happy to squash everything into one patch!
> 
> I am of course also here to squash away these `fixup!` commits before merge!
> 
> Tell me what you think, marking this as ready for review now.

Thank you for the descriptive commit message!

This is very interesting, haven't seen the godbolt.org tool yet, fun way to learn more about the compiler. 😍

I suggest we squash the two commits. I can also see how we might want to move the macro one day to `p2panda-core`, there's more places where this will come in handy. Can already think of:

1. Mix ALPN with network id

https://github.com/p2panda/p2panda/blob/0489d6095ccd7f25463fb5531772818fd49abaf5/p2panda-net/src/lib.rs#L318

2. Mix gossip overlay topic with constant to prevent collision with second overlay using the same topic for peer sampling

https://github.com/p2panda/p2panda/blob/0489d6095ccd7f25463fb5531772818fd49abaf5/p2panda-net/src/sync/actors/manager.rs#L143

Happy to merge when you're ready @matthiasbeyer

matthiasbeyer:
Squashed. :rocket:

matthiasbeyer:
Oh, I have to edit a bit of the commit message again... holdon

matthiasbeyer:
There you go.
The "debug assert" bit was removed from the patch, since it really is a no-op comparing `x` to `x` :laughing: so I removed that bit from the commit message as well!

adzialocha (approved):
Let's go!

PR #1356, opened 2026-08-15
p2panda: Put group operation on group topic

matthiasbeyer:
As @adzialocha suggested [here](https://wald.liebechaos.org/channel/p2panda?msg=R7vmlcnOPRkc3EPIn) - but I am still not sure whether this is really all it takes :laughing: 

Lets talk about it!

## 📋 Checklist

### DIFF
--- a/p2panda/src/spaces/forge.rs
+++ b/p2panda/src/spaces/forge.rs
@@ -122,10 +122,11 @@ impl p2panda_spaces::Forge<AuthCapabilities> for OperationForge {
             p2panda_spaces::SpacesArgs::Group { group_id, .. } => {
                 // Every author maintains their own log of control messages _per_ group.
                 let log_id = group_log_id(group_id);
+                let topic = Topic::from(group_id);
 
                 let extensions = Extensions::builder(log_id).build_space(args);
 
-                self.create_operation(None, log_id, None, extensions)
+                self.create_operation(Some(topic), log_id, None, extensions)
                     .await?
             }
 


matthiasbeyer:
What I still cannot wrap my head around... other members of the group must receive and process every group operation, right?
That must be implemented somewhere as well, right?

sandreae:
> What I still cannot wrap my head around... other members of the group must receive and process every group operation, right? That must be implemented somewhere as well, right?

Unfortunately, as I think you suspect, sharing groups on a topic is not quite this simple. 

I think the implementation requirements can be split into two broad categories:

1) sync group logs: associate _all_ logs for a group with the topic
2) process group operations: when both publishing and receiving group control messages

These same requirements exist for spaces as well, and it's part of what we've now implemented. What I need to give a little more thought to is really what _isn't_ needed from that work, as groups on their own have less complex system than spaces (no encryption via p2panda-spaces and p2panda-encryption). Important to note that point 1) is not only about associating the "root" groups logs, but also any "sub-groups" which are themselves a member of that group.

There's actually a privacy issue regarding the unencrypted nature of topics related to this feature. The current design of spaces is that group operations are _only_ synced via a space, in fact they are synced via all spaces that need them. The reason for this is that control messages are encrypted, and so membership of the groups inside a space will only be readable by space members. If we offer syncing groups on a topic, then we risk users leaking group membership unintentionally. I understand that supporting "world readable" topics with group based access-control is actually a desirable feature though, and being able to share groups on a topic is a step towards this. Just want to flag that we'd need to account for these two privacy cases, in a way where the user can't accidentally leak groups they thought were private, if this feature were to be supported.

sandreae:
Some code references which might be helpful:

- associating group logs to a space topic in forge: https://github.com/p2panda/p2panda/blob/cfb90f641125a15071e9551571f8bd88127464fb/p2panda/src/spaces/forge.rs#L230
- spaces and groups operations received via sync being processed https://github.com/p2panda/p2panda/blob/cfb90f641125a15071e9551571f8bd88127464fb/p2panda/src/processor/pipeline.rs#L255
- the actual spaces processor https://github.com/p2panda/p2panda/blob/development/p2panda-stream/src/spaces/processor.rs 
- task for "pushing" missing group operations into a space https://github.com/p2panda/p2panda/blob/development/p2panda/src/spaces/repair.rs

sandreae:
The patterns for associating group logs with a topic will be slightly different from how we do it in spaces I expect, there are different operation types there which "pull" a group change into a space, these don't exist in groups.

sandreae:
Something like this would get you quite far with the association part:

```rust
let y: AuthGroupState<AuthCapabilities> = self
    .store
    .get_groups_state_tx(Hash::digest(GLOBAL_GROUPS_CONTEXT_ID))
    .await?
    .unwrap_or_default();

for sub_group_id in y.groups(group_id) {
    self.store
        .associate(&Topic::from(space_id), &self.id(), &group_log_id(sub_group_id))
        .await?;
}
```

matthiasbeyer:
I tried to look into it, but I feel like I am not able to implement this, as I do not understand the details here... :disappointed:

sandreae:
That's understandable, the patterns we've introduced now for spaces are still settling which makes generalising to other areas still a bit hard. I'd be up for revisiting this with some more concrete advise in a few weeks 👍

adzialocha:
> I tried to look into it, but I feel like I am not able to implement this, as I do not understand the details here... 😞

@matthiasbeyer If you want to move ahead with your experiments you can express sync through a space as a workaround, even when spaces is not what you're looking for, since there we have all the groups log association figured out. You can see a test which is should be close to what you need:

https://github.com/p2panda/p2panda/blob/development/p2panda/tests/spaces.rs#L758

matthiasbeyer:
Ahh, sure, I have not thought of that :laughing: for a first prototype of my application, that's what I can do of course! :smile: 

And, if abstracted correctly, I can switch the underlying implementation from spaces to groups later on without hassle :laughing: 

I think you just saved my weekend (I got something to do now!)

PR #1358, opened 2026-08-18
Expose InnerSpaceError

matthiasbeyer:
The type is returned by Space::members() for example, so users of that function should be able to see the type.

## 📋 Checklist


matthiasbeyer:
I am really confused as to why `rustc` or `clippy` do not report this when running `cargo check` or `cargo clippy` :eyes: 

I suspect that because it is a `pub type InnerSpaceError =` and the used type is public in `p2panda_spaces` ... though its generic arguments are not apparently, which makes it also unusable by consumers of the API :thinking:

PR #1359, opened 2026-08-18
Pipeline hook managing per-space allow- and block-lists 

### LINKED ISSUE #1351: Emit authorisation events when accepting sync sessions
https://github.com/p2panda/p2panda/pull/1321#discussion_r3766395331

sandreae:
This changes introduces a ConnectionAuthoriserHook which holds the global ConnectionAuthoriser. It
watches the event stream and reacts to membership change events ("create", "add" and "remove").
Current members are always added to the allow-list for the space topic, removed members are added
to the block-list. The former handles re-admitting a previously removed member.

## 📋 Checklist

### DIFF
--- a/p2panda-spaces/src/space.rs
+++ b/p2panda-spaces/src/space.rs
@@ -38,7 +38,8 @@ use crate::types::{
     EncryptionGroup, EncryptionGroupError, EncryptionGroupState,
 };
 use crate::utils::{
-    added_secret_members, removed_secret_members, secret_members, sort_members, typed_members,
+    added_secret_members, removed_members, removed_secret_members, secret_members, sort_members,
+    typed_members,
 };
 use crate::{ActorId, GroupId, MemberId, OperationId, SpaceEvent, SpaceId, StrongRemoveResolver};
 
@@ -473,6 +474,16 @@ where
             &current_ancestors,
         );
 
+        // Record any individuals that were removed from the space.
+        for event in events.iter() {
+            if let Event::Spaces(SpaceEvent::Removed { context, .. }) = &event {
+                let removed = removed_members(&current_members, &context.members)
+                    .into_iter()
+                    .map(|(member, _)| member);
+                y.removed.extend(removed);
+            }
+        }
+
         events.extend(application_events);
 
         Ok(Some((y, events)))
@@ -778,6 +789,7 @@ where
                     group_id,
                     groups_y,
                     encryption_y,
+                    removed: Default::default(),
                 }
             }
         };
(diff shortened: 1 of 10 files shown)


p2panda/src/spaces/authoriser.rs
```diff
@@ -0,0 +1,78 @@
+use p2panda_net::connection_authoriser::ConnectionAuthoriser;
```
adzialocha, line 3: License header missing

p2panda-spaces/src/space.rs
```diff
@@ -997,6 +1021,10 @@ pub struct SpacesState<C> {
     pub group_id: VerifyingKey,
     pub groups_y: AuthGroupState<C>,
     pub encryption_y: EncryptionGroupState,
+
+    /// The set of all individuals who have ever been removed from the space, either via. direct
```
adzialocha: Nit: period after "via"

p2panda-spaces/src/space.rs
```diff
@@ -473,6 +474,16 @@ where
             &current_ancestors,
         );
 
+        // Record any individuals that were removed from the space.
+        for event in events.iter() {
+            if let Event::Spaces(SpaceEvent::Removed { context, .. }) = &event {
+                let removed = removed_members(&current_members, &context.members)
+                    .into_iter()
+                    .map(|(member, _)| member);
+                y.removed.extend(removed);
```
adzialocha, line 488: Shouldn't the current state be overwritten by the new one instead of extending it? I wonder if otherwise re-adds will be broken (the previously removed user stays in the removed set even when they were added again)?

sandreae: This field is for persisting the list of members who have ever been removed from the group, so it should actually grow. When managing the allow/block lists one can add current members to the allow list (what we do now) so that re-adds are accounted for. 

We need the list of all removed members for the moment as the authoriser state is not persisted, if it was just the diff here then on restart previously removed members would not get blocked.

adzialocha: Thanks! That makes sense. Maybe we should call it `once_removed` or add a comment?

sandreae: There's a comment on the struct field and accessor method (which already filters out re-added members) but I'll add a comment here too :+1:

adzialocha (approved):

p2panda/src/spaces/authoriser.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl ConnectionAuthoriserHook {
+    pub fn new(inner: ConnectionAuthoriser) -> Self {
+        Self { inner }
+    }
+}
+
+impl ProcessorHook<Event> for ConnectionAuthoriserHook {
+    async fn on_input(&self, input: &Event) {
+        let crate::processor::ProcessorStatus::Completed(ref result) = input.spaces else {
```
mycognosist: There are a few type inputs here which could be cleaned up. `ProcessorStatus`, `SpacesResult`, `SpaceEvent`.

mycognosist (approved):
Awesome! Great work. I just left one small comment about type imports.

It's going to be super cool when we no longer have to manually call `update_authoriser()`.

PR #1361, opened 2026-08-19
p2panda-net: support bearer authentication for iroh relays

orangetin:
`p2panda-net` does not currently support passing in `auth_token` when connecting to iroh relays. This PR adds a `relay_url_with_token` method to the iroh_endpoint builder to configure an optional auth token per relay.

Also added tests to verify each relay gets it's own token in relay_config and that the old `relay_url` method still works. I've verified this works on a private relay.

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Highlights are marked with a pancake 🥞
 
 ### Added
 
-- net: Support bearer authentication for iroh relays
+- net: Support bearer authentication for iroh relays [#1361](https://github.com/p2panda/p2panda/pull/1361)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


adzialocha (changes requested):
Hello @orangetin ! Thank you for this PR and useful addition!

Since one can register multiple relays with potentially a different token each, I'd suggest to change the new builder method to something like:

```rust
fn relay_url_with_token(mut self, url: iroh::RelayUrl, token: impl Into<String>) -> Self
```

You would need to accordingly also change how the relay urls are managed by the builder (some with or without token).

orangetin:
Ah that's a good point thanks! I've updated `relay_urls` from a hashset to `relays` a hashmap with an optional string auth token val.

Let me know if there's anything else needed here!

adzialocha (approved):
Perfect! Thank you @orangetin :-)

PR #1381, opened 2026-08-29
Defer out-of-order live operations

comradelemoncake:
Describe your changes here.

## 📋 Checklist

### DIFF
--- a/p2panda-stream/src/ingest/processor.rs
+++ b/p2panda-stream/src/ingest/processor.rs
@@ -5,7 +5,7 @@ use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::marker::PhantomData;
 
-use p2panda_core::{Extensions, Hash, LogId, Operation, SeqNum, VerifyingKey};
+use p2panda_core::{Extensions, Hash, LogId, Operation, OperationError, SeqNum, VerifyingKey};
 use p2panda_store::Transaction;
 use p2panda_store::logs::LogStore;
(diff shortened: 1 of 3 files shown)


adzialocha:
Oh, hi @comradelemoncake 😊

Thank you so much for your PR!

It's miraculously also something we are planning to build in the next two weeks, so good timing!

Strictly speaking the default sync protocol (using log sync and iroh in `p2panda-net`) doesn't need an out-of-order (ooo) buffer since operations are coming always in-order via that path. But we are improving our API right now to allow people hacking alternative sync protocols for different transports and networks which will require them to have such buffer in some cases. I assume this is where you are right now!

Since not everyone will require an ooo-buffer, we probably want to make it more modular - on the other hand it can also be good to have it always there, by default, on the high-level `p2panda` API.

The approach to get the best of both worlds could be:

* Introduce a standalone ooo-buffer solution in `p2panda-sync`, this will allow developers to pull that in - wherever they need it - if they hack their own networking and sync layers.
* Have an ooo-Processor in `p2panda-stream` wrapping that thing. This will allow developers to add that functionality when they want it in their custom processor pipeline in `p2panda`. Note that building custom pipelines is not possible yet, but will be.
* Have that processor in the default pipeline in `p2panda`, potentially redundant but avoiding surprises.

btw, we also will need this when we deal with fully encrypted operations, it'll be part of a decryption buffer. Another reason to keep it reusable and modular.

PR #1382, opened 2026-09-01
Sneakernet sync via USB-stick example

adzialocha:
This example shows how one can not only drop p2panda operations onto an USB-stick to deliver them somewhere but also how we can use it to run an _interactive_, delay-tolerant sync protocol where every node shares its current state with the network, using the USB-stick as a "very slow" transport and "store and forward" buffer.

## 📋 Checklist

### DIFF
--- a/p2panda-sync/examples/sneakernet.rs
+++ b/p2panda-sync/examples/sneakernet.rs
@@ -1,0 +1,563 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Sync via USB-sticks (sneakernet).
+//!
+//! This example shows how one can not only drop p2panda operations onto an USB-stick to deliver
+//! them somewhere but also how we can use it to run an _interactive_, **delay-tolerant sync
+//! protocol** where every node shares its current state with the network, using the USB-stick as a
+//! "very slow" transport.
+//!
+//! With this approach we efficiently only write the requested ranges to the USB-stick after we've
+//! considered every node's "needs" when reading the stick. One could also describe this as sync in
+//! a **broadcast** network.
+//!
+//! The USB-stick functions as a **store and forward** buffer where data is persisted from other
+//! participants even if the current node reading or writing to the USB-stick is not interested in
+//! it. The USB-stick helps to eventually dissiminate data to everyone in the network.
+//!
+//! Note that sync via USB-stick is possible to do in both encrypted & unencrypted scenarios. For
+//! fully encrypted data on the stick, we would attempt decrypting all ciphertexts first or leave a
+//! "hint" for which ciphertexts are likely interesting for us by coming up with a file name or path
+//! scheme.
+//!
+//! ## Protocol
+//!
+//! 1. Check what's inside the USB stick, compute diff, load delta of everything what we are
+//!    interested in and don't have yet.
+//!
+//!    We don't necessarily need to compute a diff, could also just forward whatever there is.
+//!    Applications would see potentially more duplicates then. Having an idempotent processing
+//!    logic is key (probably should have that in any case).
+//!
+//!    Steps:
+//!
+//!    - We need awareness of topics we are interested in
+//!    - Resolve log ids + authors for all topics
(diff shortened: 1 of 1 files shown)


mycognosist (approved):
I like the details you added in the doc strings; there's are lots of trails to follow there.

p2panda-sync/examples/sneakernet.rs
```diff
@@ -0,0 +1,563 @@
...
+            let their_log_heights = {
+                stick
+                    .announcements
+                    .iter()
+                    .find(|announcement| announcement.topic == topic)
+                    .map(|announcement| announcement.log_heights.clone())
+                    .unwrap_or_default()
+            };
+
+            let log_ranges: LogRanges<VerifyingKey, LogId> =
```
sandreae, line 412: For this situation where we've compared log heights of remote against our own (or visa-versa) and derived LogRanges, a nice API would be a method like this on OperationStore:

```rust
impl OperationStore {
    async fn get_log_ranges(
        &self,
        ranges: LogRanges,
    ) -> Pin<Box<dyn Stream<Item = Result<Operation<E>, SqliteError>>>> {
        // TODO
    }
}
```

Benefits of this method are a) the user doesn't need to iterate over each author+log pair manually and b) it bounds the operations which will be kept in memory and c) still we "lock" the operations which are retrieved to only those in the exact LogRanges we pre-computed.

It could be implemented quite easily with the `fetch()` method provided by sqlx https://docs.rs/sqlx/latest/sqlx/query/struct.Query.html#method.fetch

p2panda-sync/examples/sneakernet.rs
```diff
@@ -0,0 +1,563 @@
...
+
+    // Export data.
+
+    let mut announcements: Vec<Announcement> = Vec::new();
+    let mut operations: HashMap<Topic, Logs> = HashMap::new();
+
+    for topic in &topics_a {
+        let all_log_heights = get_topic_log_heights(&store_a, &topic).await?;
+
+        // TODO: This is ugly.
```
sandreae, line 184: See comment below about suggestion for a new store method which I believe could tidy up this query as well.

p2panda-sync/examples/sneakernet.rs
sandreae: This is more of a comment on possible API improvements which would benefit all use cases / examples, rather than a review of this PR :+1:

p2panda-sync/examples/sneakernet.rs
```diff
@@ -0,0 +1,563 @@
...
+pub type LogIds = BTreeMap<VerifyingKey, Vec<LogId>>;
+
+type Logs = HashMap<(VerifyingKey, LogId), Vec<AnyOperation>>;
+
+#[derive(Debug)]
+struct UsbStickContent {
+    announcements: HashSet<Announcement>,
+    // NOTE: Could be ring-buffer (we reserve capacity in a config), or support-node-like k/v store.
+    // In both cases this would allow us to _not_ mention the topic & work with encryption.
+    operations: HashMap<Topic, Logs>,
```
sandreae, line 127: In the examples current form this could be a sqlite database, which would make the query logic symmetrical which would be kind of neat. But I see that for demonstrating the benefits of having alternative store structures (which support full encryption and capacity reservation) then maybe this approach is better.

p2panda-sync/examples/sneakernet.rs
```diff
@@ -0,0 +1,563 @@
...
+                let logs: LogIds = store_b.resolve(&topic).await?;
+                get_log_heights(&store_b, &logs).await?
+            };
+
+            // Determine the operations we need from the stick.
+            let log_ranges: LogRanges<VerifyingKey, LogId> =
+                // NOTE: we reverse the roles here, "their" and "our", because we are determining
+                // what they should "send" to us...not what we should send to them.
+                //
+                // The docs for `compare()` could maybe be updated to reflect this bidirectional
```
sandreae, line 347: 👍

PR #1385, opened 2026-09-02
Point-to-point (unicast) sync session with live mode example

mycognosist:
Here we introduce a sync example which utilises a similar protocol to the one currently offered in `p2panda-sync`. The example operates in the context of a single sync session between two peers; the topic is assumed to have been learned _a priori_.

Node A sends an announcement based on their current log heights for the topic.
Node B queries their local log heights and responds with any operations needed by node A.
Node B then sends an announcement based on their current log heights.
Node A ingests the operations from node B.
Node A queries their local log heights and responds with any operations needed by node B.
Node B ingests the operations from node A.
_The "catch up" phase of the session is thus concluded and both peers move into eager push / live mode._
Node B sends a new operation to node A.

Example output from running the example:

```text
a25e0b34a3: node a
7d48226bc8: node b
--------
7d48226bc8: process remote announcement 6f143
--------
a25e0b34a3: ingest remote operation 5793a
a25e0b34a3: ingest remote operation 6d30e
a25e0b34a3: ingest remote operation b9052
a25e0b34a3: ingest remote operation 0b398
a25e0b34a3: ingest remote operation dec74
a25e0b34a3: process remote announcement 94dd8
--------
7d48226bc8: ingest remote operation f9f3c
7d48226bc8: ingest remote operation 35c30
7d48226bc8: ingest remote operation b24a8
7d48226bc8: ingest remote operation 76b1e
7d48226bc8: ingest remote operation eca93
--------
a25e0b34a3: ingest remote live operation 53b66
``` 

## 📋 Checklist

### DIFF
--- a/p2panda-sync/examples/sync_w_live_mode.rs
+++ b/p2panda-sync/examples/sync_w_live_mode.rs
@@ -1,0 +1,334 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Session-based sync example (w. eager push as "live-mode").
+//!
+//! Node A sends an announcement.
+//! Node B responds with operations and their own announcement.
+//! Node A ingests the operations and responds with operations.
+//! Node B creates a new operation and sends it immediately.
+//!
+//! To keep the example simple, messages (in the forms of announcements and operations) are passed
+//! directly between nodes. In a real-world context the messages would be sent over an underlying
+//! connection of some kind (e.g. iroh's QUIC streams).
+use std::collections::BTreeMap;
+
+use p2panda_core::logs::{LogHeights, LogRanges, compare};
+use p2panda_core::test_utils::TestLog;
+use p2panda_core::traits::Digest;
+use p2panda_core::{AnyOperation, Hash, Operation, SeqNum, SigningKey, Topic, VerifyingKey};
+use p2panda_store::logs::LogStore;
+use p2panda_store::operations::OperationStore;
+use p2panda_store::topics::TopicStore;
+use p2panda_store::{SqliteStore, tx};
+use p2panda_sync::protocols::ShortFormat;
+use serde::{Deserialize, Serialize};
+
+type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
+
+type LogId = Hash;
+type LogIds = BTreeMap<VerifyingKey, Vec<LogId>>;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+struct CustomExtensions {
+    log_id: LogId,
+}
+
+// We assume the topic is known as a result of the sync session context.
+#[derive(Debug, PartialEq, Eq)]
+struct Announcement {
+    log_heights: BTreeMap<VerifyingKey, BTreeMap<LogId, SeqNum>>,
+}
+
+impl Digest<Hash> for Announcement {
+    fn hash(&self) -> Hash {
+        Hash::digest({
+            let mut bytes = Vec::new();
+
+            for (author, log_heights) in &self.log_heights {
+                for (log_id, seq_num) in log_heights {
+                    bytes.extend_from_slice(author.as_bytes());
+                    bytes.extend_from_slice(log_id.as_bytes());
+                    bytes.extend_from_slice(&seq_num.to_be_bytes());
+                }
+            }
+            bytes
+        })
+    }
+}
+
+#[derive(Debug)]
+struct Node {
+    signing_key: SigningKey,
+    store: SqliteStore,
+}
+
+impl Node {
+    async fn new() -> Self {
+        Self {
+            signing_key: SigningKey::generate(),
+            store: SqliteStore::temporary().await,
+        }
+    }
+
+    fn id(&self) -> VerifyingKey {
+        self.signing_key.verifying_key()
+    }
+
+    /// Create a new log, populate it with five operations and associate it with the topic.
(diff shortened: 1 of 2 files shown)


p2panda-sync/examples/sync_w_live_mode.rs
```diff
@@ -0,0 +1,370 @@
...
+                for (log_id, seq_num) in log_heights {
+                    bytes.extend_from_slice(author.as_bytes());
+                    bytes.extend_from_slice(log_id.as_bytes());
+                    bytes.extend_from_slice(&seq_num.to_be_bytes());
+                }
+            }
+            bytes
+        })
+    }
+}
```
adzialocha, line 57: Not important, but the hashing of announcements it's only used for the println output, maybe nice to omit this just to keep the example smaller, otherwise it could look like something "important"?

p2panda-sync/examples/sync_w_live_mode.rs
```diff
@@ -0,0 +1,370 @@
...
+enum Message {
+    Announcement(Announcement),
+    Operation(AnyOperation),
+}
+
+// Shared mailbox where messages are sent and received by both peers.
+#[derive(Debug)]
+struct Mailbox {
+    messages: VecDeque<Message>,
+}
```
adzialocha: Could also have a `type Mailbox = VecDeque<Message>` instead

p2panda-sync/examples/sync_w_live_mode.rs
```diff
@@ -0,0 +1,370 @@
...
+                let body = (op_i as usize).to_be_bytes();
+                let operation = log.operation(&body, CustomExtensions { log_id });
+
+                <SqliteStore as OperationStore<Operation<CustomExtensions>, Hash>>::insert_operation(
+                &self.store,
+                &operation.hash,
+                &operation,
+                &log_id,
+            )
+            .await?;
```
adzialocha: Formatting

p2panda-sync/examples/sync_w_live_mode.rs
```diff
@@ -0,0 +1,370 @@
...
+        messages: VecDeque::new(),
+    };
+
+    let topic = Topic::random();
+
+    // Node A and B populate their logs with operations.
+
+    let node_a = Node::new().await;
+    let _log_a = node_a.populate_log(topic).await?;
+    println!("{}: node a", node_a.signing_key.verifying_key().fmt_short());
```
adzialocha: Could be an `Node::id` method instead returning the verifying key

p2panda-sync/examples/sync_w_live_mode.rs
adzialocha: .. or skip the mailbox altogether. With the name I'd expect the mailbox to be one per node, but if you just use it to pass over messages to each other, you could skip the pushing and popping and just give them directly.

p2panda-sync/examples/sync_w_live_mode.rs
```diff
@@ -0,0 +1,370 @@
...
+
+    // Node A sends an announcement to node B.
+    // Node B processes the announcement then sends operations and an announcement.
+
+    let announcement = node_a.generate_announcement(topic).await?;
+    mailbox
+        .messages
+        .push_back(Message::Announcement(announcement));
+
+    if let Some(Message::Announcement(announcement)) = mailbox.messages.pop_front() {
```
adzialocha: Could skip these checks if you just give the announcement over directly (see comment above)

p2panda-sync/examples/sync_w_live_mode.rs
```diff
@@ -0,0 +1,370 @@
...
+                }
+                Message::Announcement(announcement) => {
+                    println!(
+                        "{}: process remote announcement {}",
+                        node_a.signing_key.verifying_key().fmt_short(),
+                        announcement.hash().fmt_short()
+                    );
+
+                    let log_operations = node_a.process_announcement(announcement, topic).await?;
+                    mailbox.messages.extend(log_operations);
```
adzialocha: This seems to work because of the LIFO logic of the stack and that announcements are always followed by operations

adzialocha (commented):
Looks all good to me!

I didn't find anything which needs changing, but wondering a bit about the shared stack (LIFO) abstraction to "communicate" between two nodes.

It definitely works for this example but is a bit "unnatural" for any communication protocol? Usually it's a bi-directional, ordered channel of some kind (FIFO) so I found it a bit hard to read. Removing the "mailbox" altogether and passing messages over directly could just be fine as an alternative and would be more "realistic" (as I've written in the comments). Just a thought, I'm also fine with merging it as such.

ps.
I've mentioned previously that it was confusing that A sends live messages to B but I see now that your sync protocol is bi-directional, aka A syncs with B and B syncs with A. Then of course it doesn't matter.

p2panda-sync/examples/sync_w_live_mode.rs
```diff
@@ -0,0 +1,370 @@
...
+            .await?;
+        });
+
+        Ok(())
+    }
+}
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> Result<()> {
+    let mut mailbox = Mailbox {
```
sandreae: Maybe the naming is a little misleading here, we have associations with Mailboxes being a remote store of some kind, calling this MessageBus or MockChannel or something would be good.

mycognosist: Got rid of it completely.

PR #1388, opened 2026-09-03
Add method on `TopicStore` to get all topics

### LINKED ISSUE #1396: Method on `TopicStore` to get all topics

mycognosist:
Here we add a `topics()` method to the `TopicStore` which returns a deduplicated list of all topics for which active log associations exist. Topics for previously removed associations are not returned.

This query can be helpful for understanding which topics to sync over when you only have access to the SQLite store (for example, in the context of sneakernet sync).

Closes: https://github.com/p2panda/p2panda/issues/1396

## 📋 Checklist

### DIFF
--- a/p2panda-store/src/topics/sqlite.rs
+++ b/p2panda-store/src/topics/sqlite.rs
@@ -176,4 +176,33 @@ where
 
         Ok(topics)
     }
+
+    /// Retrieve all topics for which active associations exist.
+    async fn topics(&self) -> Result<Vec<T>, Self::Error> {
+        let topics = self
+            .tx(async |tx| {
+                query_as::<_, (Vec<u8>,)>(
+                    "
+                    SELECT DISTINCT
+                        topic
+                    FROM
+                        topics_v1
+                    ",
+                )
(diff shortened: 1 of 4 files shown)


sandreae (approved):

p2panda-store/src/topics/sqlite.rs
```diff
@@ -135,4 +135,33 @@ where
...
+                    SELECT DISTINCT
+                        topic
+                    FROM
+                        topics_v1
+                    ",
+                )
+                .fetch_all(pool)
+                .await
+                .map_err(SqliteError::Sqlite)
+            })
```
adzialocha, line 154: Should we do this with an transaction instead for composability?

adzialocha (approved):

p2panda-store/src/topics/sqlite.rs
adzialocha: .. also related: #1065

mycognosist: Good idea. Done.

PR #1391, opened 2026-09-04
Use  `AnyOperation` on `LogStore`

### LINKED ISSUE #1390: Use `AnyOperation` in `LogStore`
Towards #1018

sandreae:
Use `AnyOperation` on `LogStore` in place of `Operation<E>` and update all affected call sites. 

closes: #1390 

## 📋 Checklist

### DIFF
--- a/p2panda-core/src/operation/any.rs
+++ b/p2panda-core/src/operation/any.rs
@@ -20,7 +20,7 @@ use crate::traits::{Chain, Digest, Offchain, Provenance};
 /// Applications usually want to attach custom extensions to the operation, if you need to know the
 /// type you can easily convert from `AnyOperation` to [`Operation`](crate::Operation) with an
 /// explicit `E` extensions type.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct AnyOperation {
     pub hash: Hash,
     pub header: AnyHeader,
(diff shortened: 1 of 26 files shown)


p2panda-core/src/operation/any.rs
```diff
@@ -20,7 +20,7 @@ use crate::traits::{Chain, Digest, Offchain, Provenance};
 /// Applications usually want to attach custom extensions to the operation, if you need to know the
 /// type you can easily convert from `AnyOperation` to [`Operation`](crate::Operation) with an
 /// explicit `E` extensions type.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
```
adzialocha, line 23: On `Operation<E>` we have a custom `PartialEq` implementation which looks only at the hash for efficiency. We could do that here as well ..

However, it implies that the hash is correct (which is not checked here) ..

sandreae: What do you think? I wouldn't mind having the optimised one here as well.

sandreae: I've raised an issue for this: https://github.com/p2panda/p2panda/issues/1401

mycognosist (approved):
Generics be ~~gone~~ reduced! ✨

PR #1393, opened 2026-09-04
Add method for streaming operation ranges in new API module

### LINKED ISSUE #1387: Helper method to stream log ranges directly from database

sandreae:
closes: #1387

Additionally introduces new `api` module in p2panda-sync and moves some generally useful types/methods there.

## 📋 Checklist


p2panda-sync/src/api.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use std::pin::Pin;
+
+use futures_util::{Stream, TryStreamExt, stream};
+use p2panda_core::logs::{LogHeights, LogRanges, Logs};
+use p2panda_core::{AnyOperation, Hash, LogId, SeqNum, VerifyingKey};
+use p2panda_store::logs::LogStore;
+
+pub type OperationStream<L, Error> =
+    Pin<Box<dyn Stream<Item = Result<(AnyOperation, L, Vec<u8>), Error>> + Send>>;
```
adzialocha: Could be nice to simplify this type a bit more to make it more readable, for example by using https://docs.rs/futures/latest/futures/stream/type.BoxStream.html or defining it yourself

p2panda-sync/src/api.rs
```diff
@@ -0,0 +1,82 @@
...
+        };
+        result.insert(*verifying_key, log_heights);
+    }
+
+    Ok(result)
+}
+
+/// Construct a stream which returns all operations in the provided log ranges.
+///
+/// Operations are queried per-log so as to avoid holding all operations in memory at one time.
```
adzialocha: If the log is long then this would still hold a lot of data in memory, so optimizing the query to be paginated (using LIMIT) would be great as well some day. I assume that's what you wanted to write in the issue already.

p2panda-sync/src/api.rs
```diff
@@ -0,0 +1,82 @@
...
+pub fn stream_log_ranges<S, L>(
+    store: &S,
+    ranges: LogRanges<VerifyingKey, L>,
+) -> OperationStream<L, S::Error>
+where
+    S: LogStore<AnyOperation, VerifyingKey, L, SeqNum, Hash> + Clone + Send + 'static,
+    L: LogId + Clone + Send + 'static,
```
adzialocha: LogId is already Clone

p2panda-sync/src/api.rs
```diff
@@ -0,0 +1,82 @@
...
+    // Query each log range in a stream iterator which produces the query result when polled.
+    let store = store.clone();
+    let stream = stream::try_unfold(
+        (0usize, flattened_ranges, store),
+        |(index, ranges, store)| async move {
+            if index >= ranges.len() {
+                return Ok::<_, S::Error>(None);
+            }
+
+            let (author, log_id, after, until) = ranges[index].clone();
```
adzialocha: Only log_id needs cloning here

p2panda-sync/src/api.rs
```diff
@@ -0,0 +1,82 @@
...
+    );
+
+    // Flatten the query stream so that an AnyOperation is produced on each call.
+    let stream = stream
+        .map_ok(|(rows, log_id)| {
+            stream::iter(rows.into_iter().map(move |(operation, header_bytes)| {
+                Ok((operation, log_id.clone(), header_bytes))
+            }))
+        })
+        .try_flatten();
```
adzialocha: If an error would be returned from `get_log_entries` then it would show up as part of the item in the Stream (looking at the OperationStream type). I somehow would expect the whole stream_log_ranges method to fail and not find the error for a whole log query on a specific item.

p2panda-sync/src/api.rs
```diff
@@ -0,0 +1,82 @@
...
+        result.insert(*verifying_key, log_heights);
+    }
+
+    Ok(result)
+}
+
+/// Construct a stream which returns all operations in the provided log ranges.
+///
+/// Operations are queried per-log so as to avoid holding all operations in memory at one time.
+pub fn stream_log_ranges<S, L>(
```
adzialocha: I'd be fine with just calling this method `log_ranges`, that it gives me a Stream I can see in the return type (similar to sqlx "fetch").

p2panda-sync/src/api.rs
```diff
@@ -0,0 +1,82 @@
...
+use futures_util::{Stream, TryStreamExt, stream};
+use p2panda_core::logs::{LogHeights, LogRanges, Logs};
+use p2panda_core::{AnyOperation, Hash, LogId, SeqNum, VerifyingKey};
+use p2panda_store::logs::LogStore;
+
+pub type OperationStream<L, Error> =
+    Pin<Box<dyn Stream<Item = Result<(AnyOperation, L, Vec<u8>), Error>> + Send>>;
+
+/// Compute log heights of all passed author logs based on what is known in the local store.
+pub async fn get_log_heights<L, S>(
```
adzialocha: Maybe remove the "get_" prefix

p2panda-sync/src/api.rs
```diff
@@ -0,0 +1,82 @@
...
+    store: &S,
+    logs: &Logs<VerifyingKey, L>,
+) -> Result<LogHeights<VerifyingKey, L>, S::Error>
+where
+    L: LogId,
+    S: LogStore<AnyOperation, VerifyingKey, L, SeqNum, Hash> + Clone + Send + 'static,
```
adzialocha: Clone + Send + 'static are not required here

p2panda/src/streams/replay.rs
```diff
@@ -69,26 +70,20 @@ where
...
-                    .try_into()
-                    .expect("values from the database are valid");
-                process_operation_in(operation, Source::LocalStore, topic, pipeline, sync_handle)
-                    .await;
-            }
-        }
+    let mut operations = stream_log_ranges(store, log_ranges);
+
+    while let Some(result) = operations.next().await {
+        let (operation, _, _) = result?;
```
adzialocha: The triple returned here is a bit hard to read, maybe we make it a struct with named fields?

CHANGELOG.md
```diff
@@ -38,6 +38,8 @@ Highlights are marked with a pancake 🥞
 - stream: Move orderer processor from node to stream [#1312](https://github.com/p2panda/p2panda/pull/1312)
 - node: ConnectionAuthoriserHook updating authoriser based on space membership [#1359](https://github.com/p2panda/p2panda/pull/1359)
 - node: E2EE CLI chat example using spaces API [#1288](https://github.com/p2panda/p2panda/pull/1288)
+- sync: `api` module containing generally useful types [#1393](https://github.com/p2panda/p2panda/pull/1393)
```
adzialocha: Maybe something like: `api` module with useful methods to hack your own sync protocols

adzialocha (changes requested):

p2panda-store/src/logs/sqlite/mod.rs
```diff
@@ -31,7 +31,7 @@ const GET_LATEST_ENTRY: &str = "
 
 impl<L> LogStore<AnyOperation, VerifyingKey, L, SeqNum, Hash> for SqliteStore
 where
-    L: LogId,
+    L: LogId + Send + Sync + 'static,
```
adzialocha: Send + 'static are not required here

p2panda-sync/src/api.rs
sandreae: Related to this point, I've asked a question on the SQLx github to see if anyone knows where the behaviour of Query::fetch is documented: https://github.com/transact-rs/sqlx/discussions/4405

I'd still like to use this method if possible.

sandreae: I've implemented this query to use SQLx Query::fetch which we believe (TBC) only keeps one item in memory at a time.

p2panda-sync/src/api.rs
sandreae: I introduced an OperationStreamError which differentiates between "fetch" and "item" errors. How's that? Do you see a better way to surface these two types of error?

p2panda/src/streams/replay.rs
sandreae: I introduced a StreamItem type, in order for the field and method naming used elsewhere in LogStore I had to name the operation field `entry` which is a bit ugly. If we want to change this to `operation` we probably need to do it along with all method names (renaming "entry" -> "operation")

p2panda-sync/src/api.rs
adzialocha: > I introduced an OperationStreamError which differentiates between "fetch" and "item" errors. How's that? Do you see a better way to surface these two types of error?

What are these two kinds of errors? How can they happen and why?

sandreae: "fetch" can occur when `store.stream_log_entries` is _called_ and means that composing the query and stream failed, the only error which can occur here is when a log id fails encoding into bytes.

"item" can occur when the next row is pulled from the stream (caused by a call to stream.next()) and that call fails or the returned row fails converting into AnyOperation etc...

adzialocha: Thank you for clarifying! I was not very clear in my first message, let me try to expand. On top the nested stream handling and macro-magic is a bit hard to read (but surely powerful).

I think the error types could be a bit more clearer maybe and I'm still not sure about the stream termination.

Lemme repeat for myself, just to go through the code one more time:

1. sqlx's `fetch` can fail when a database error occurred, see `try_next` when iterating over the rows: https://github.com/p2panda/p2panda/blob/streaming-operation-ranges-query/p2panda-store/src/logs/sqlite/mod.rs#L274
2. Additionally, as you write, it can fail when conversion from row into `AnyOperation` fails (decode)
3. .. and it can fail when encoding the log id

All errors are critical and I expect the whole stream to terminate from this point on (there's no use continuing to forward broken logs with gaps to the user).

For me it's a bit hard to understand what is going on with `try_stream` and how it behaves in such as case (it's not well documented). I did some experiments with `try_next` and it doesn't handle this behaviour automatically as you can see, not sure if `try_stream` makes sure the stream is terminated:

```rust
use futures_util::stream::{self, TryStreamExt};

let mut stream = stream::iter(vec![Ok(()), Err(()), Ok(())]);

assert_eq!(stream.try_next().await, Ok(Some(())));
assert_eq!(stream.try_next().await, Err(()));

// Would expect Ok(None) here instead: Stream termination after error.
assert_eq!(stream.try_next().await, Ok(Some(())));
```

It would be good to verify the error behaviour here in terms of stream termination on critical fail.

Also I'd suggest that error type can be renamed to a `SqliteError` (or the generic equivalent) and distincting between "fetch" and "item" is not necessary. I believe we can also use that as the internal error type instead of `String` as the `AnyOperation` and `LogId` conversion is part of this error case in `p2panda-store`. "fetch" is the name of the sqlx method, on this high-level we don't really know anymore about this. I suggest we call this an SqliteError as it will fail when something is wrong with the database.

p2panda-sync/src/api.rs
adzialocha: Awesome!

p2panda-store/src/logs/sqlite/mod.rs
```diff
@@ -238,6 +244,43 @@ where
         }
     }
 
+    /// Stream all entries in a log after an optional starting point. This is the memory efficient
+    /// equivalent to `get_log_entries` and should only keep one entry in memory at a time.
+    fn stream_log_entries(
```
adzialocha: Do we still want to keep `log_entries` (the non streamable one) around? I only see it being used in tests.

It could be good to remove it and just name `stream_log_entries` to `log_entries`?

p2panda/src/streams/replay.rs
adzialocha: > If we want to change this to operation we probably need to do it along with all method names (renaming "entry" -> "operation")

Yeah, I like that!

p2panda-sync/examples/sneakernet.rs
```diff
@@ -395,32 +392,22 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                    log_id,
+                    ..
+                } = result?;
+                let logs = operations.entry(topic).or_default();
+
+                // NOTE: Appending only the "latest" operations to the log allows us to
+                // build some ring-buffer logic here where we would drop old operations when
+                // running full.
+                logs.entry((operation.author(), log_id))
+                    .and_modify(|entry| entry.push(operation));
```
mycognosist: There's a small logical error here. We are modifying an existing entry but not creating a new one when required. This leads to the example failing because each topic remains empty (none of the created operations are appended).

We need to add `or_insert()` to handle the first-operation-insertion case:

```rust
logs.entry((operation.author(), log_id))
    .and_modify(|entry| entry.push(operation.clone()))
    .or_insert(vec![operation]);
```

mycognosist (changes requested):

p2panda-sync/src/api.rs
sandreae: Ok, thanks, that's clear, and I agree on the desired stream behaviour.

I'll re-work the error type and adjust the implementation so the stream behaviour on an error occurring: if any error is observed on the "flattened" stream it immediately closes

p2panda/src/streams/replay.rs
sandreae: I'll raise an issue for this :+1:

p2panda-store/src/logs/sqlite/mod.rs
sandreae: I think they are, but Sync is not.

p2panda-store/src/logs/sqlite/mod.rs
sandreae: I'll remove the old query :+1:

p2panda-sync/examples/sneakernet.rs
sandreae: :pray:

sandreae: I did this instead:

```rust
logs.entry((operation.author(), log_id))
    .or_default()
    .push(operation);
```

adzialocha (approved):
This looks great! Good work!

PR #1402, opened 2026-09-07
Out-of-order buffer for validating & inserting operations

### LINKED ISSUE #1399: Out-of-order buffer as a building block for alternative sync strategies

adzialocha:
Closes: #1399 

## 📋 Checklist

### DIFF
--- a/p2panda-stream/src/ingest/ooo.rs
+++ b/p2panda-stream/src/ingest/ooo.rs
@@ -98,6 +98,18 @@ where
         }
     }
 
+    pub async fn len(&self) -> usize {
+        self.buffer.lock().await.len()
+    }
+
+    pub async fn is_empty(&self) -> bool {
+        self.buffer.lock().await.is_empty()
+    }
+
+    pub async fn clear(&self) {
+        self.buffer.lock().await.clear();
+    }
+
     // TODO: Use AnyOperation when OperationStore is ready.
     pub async fn process<'a>(
         &self,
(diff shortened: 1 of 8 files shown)


p2panda-stream/src/ingest/ooo.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    pub async fn clear(&self) {
+        self.buffer.lock().await.clear();
+    }
+
+    // TODO: Use AnyOperation when OperationStore is ready.
+    pub async fn process<'a>(
+        &self,
+        operation: &'a Operation<E>,
+        past_header: Option<&AnyHeader>,
```
mycognosist: Did you make a conscious decision to name this `past_` rather than `previous_` or `latest_`?

I assume we're always comparing the received operation with the header of the latest / most recent operation we have in our local store. Is that correct?

p2panda-stream/src/ingest/operation.rs
```diff
@@ -11,55 +11,154 @@ use p2panda_store::Transaction;
...
+
+    /// Out-of-order operation which was moved to internal buffer.
+    OutOfOrder,
+
+    /// Operation was from before a pruning point and was ignored.
+    Outdated,
+}
+
+/// Checks an incoming operation for duplicates, it's correct format and log integrity and persists
+/// it into the store when valid.
```
mycognosist:
```suggestion
/// Checks an incoming operation to ensure correct formatting and log integrity before persisting
/// it into the store when valid. This function is idempotent; duplicate operations are ignored.
```

mycognosist (approved):

p2panda-stream/src/ingest/ooo.rs
adzialocha: Good question. I think I've just took the name which was there already, but I agree `latest_header` is better. I'll change that.

PR #1408, opened 2026-09-09
Add a custom_cursor_name param to stream_from, which overrides the default ack cursor

### LINKED ISSUE #1407: Expose the ability to use customer cursor storage in stream_from
In the node's stream_from function, when it streams from a certain start position, it records it's cursor using the `Acked` object.

It always creates this using Acked::new, which uses the topic name as the cursor name.

In order to be able to replay old operations without changing the main topic cursor, it's nice to be able to use a custom cursor_name. Given this, it may also make sense to provide a custom SqliteStore, for this to be placed in, so possibly providing an entire Acked object is the right approach. 

PR to follow.

jadehopepunk:
The param gets passed through, allowing independent cursor tracking.

I considered passing in the entire Acked object, but that created complexity. It wouldn't make sense to have a topic different from the one being streamed on. The semaphore is an implementation detail and shouldn't be left to the caller. Custom cursor storage might be nice, but the SqliteStore used as the store param isn't exposed from this crate at present. 

I've changed a couple of existing tests to pass in None for this new param, to get the default behaviour.

I've added one new API test that shows using a custom cursor and getting different results from the default topic cursor.

Closes: https://github.com/p2panda/p2panda/issues/1407

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Highlights are marked with a pancake 🥞
 
 ### Added
 
+- node: Add custom_cursor_name optional param to Node::stream_from [1408](https://github.com/p2panda/p2panda/pull/1408)
 - net: Introduce an `Authoriser` for maintaining and enforcing allow- and blocklists [#1321](https://github.com/p2panda/p2panda/pull/1321)
 - node: API for promoting and demoting space members [#1311](https://github.com/p2panda/p2panda/pull/1311)
 - node: Allow graceful closure of sync sessions [#1307](https://github.com/p2panda/p2panda/pull/1307)
--- a/p2panda/src/node.rs
+++ b/p2panda/src/node.rs
@@ -37,7 +37,7 @@ use crate::spaces::{
     to_initial_members,
 };
 use crate::streams::{
-    EphemeralStreamPublisher, EphemeralStreamSubscription, Event, ImportError, Pipeline,
+    Acked, EphemeralStreamPublisher, EphemeralStreamSubscription, Event, ImportError, Pipeline,
     StreamFrom, StreamPublisher, StreamSubscription, SystemEvent, TaskTracker, ephemeral_stream,
     event_stream, processed_stream, to_stream_event, to_system_event,
 };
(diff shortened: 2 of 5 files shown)


jadehopepunk:
The failing CI test , `tests::process_operation_from_expired_member`, seems unrelated to these changes. Any chance it's a flaky test?

adzialocha:
> The failing CI test , `tests::process_operation_from_expired_member`, seems unrelated to these changes. Any chance it's a flaky test?

Yes! There's a lot of parts moving right now in `main`. I've restarted the CI and it's passing now.

adzialocha (approved):
Thank you @jadehopepunk! This looks good. I'm curious to hear if it also does the job for your project on your end?

I think the API turns out to be suboptimal slowly, also `processed_stream` is way too powerful now but that's all nothing todo with this PR, just a note to myself for a refactoring in the future.

jadehopepunk:
@adzialocha I can't totally verify that it does the job, because I can't easily run on this version of the code. I need https://github.com/p2panda/p2panda/pull/1419 in first so that I can run it :)

So, I'd be keen to keep this one open until after #1419 or some other fix for that is in.

jadehopepunk:
@adzialocha we're now using this PR in dashchat too, and it does what we need, so I should probably go ahead and get it applied. But, before we do that, I noticed something:

The default topic cursor uses the topic as the cursor name:

```rust
pub fn new(store: SqliteStore, topic: impl Into<Topic>) -> Self {
    let topic = topic.into();
    Self::from_name(store, topic, topic.to_string())
}
```

When I use a custom cursor in this PR, I pass that straight into `from_name`. That would appear to mean that the name is not scoped by topic. For example, if as an app developer, I decide I want a custom cursor called "replay", and thus I call `from_name(store, some_topic, "replay".to_string())`, is it true that if I call `from_name` with a different topic, but the same string "replay", that the names will clash?

I think the intention of such a developer would be that it's a unique name but scoped by topic. So perhaps implemented as a custom prefix.

So we did use this in the current DashChat release, but just to be safe I supplied it a name that was "custom prefix + topic".

So the question is? Does `from_name` create a cursor with a globally scoped cursor name, or topic scoped. And, if it's globally scoped, perhaps I should change this PR to expose only a method to supply a custom prefix, not a custom name.

PR #1412, opened 2026-09-09
Doc-strings and clean-ups for new `api` module

adzialocha:
Whole bunch of doc-strings for the new `api` module in `p2panda-sync` with some minor API changes (mostly renames of types).

### DIFF
--- a/p2panda-core/src/cursor.rs
+++ b/p2panda-core/src/cursor.rs
@@ -3,7 +3,7 @@
 //! State vector to track and compare logs.
 use std::hash::Hash as StdHash;
 
-use crate::logs::{LogHeights, LogId, LogRanges, SeqNum, compare};
+use crate::logs::{LogHeights, LogId, LogRanges, SeqNum, compare_logs};
 use crate::traits::Author;
 
 /// Cursor to track log heights (state vector).
@@ -45,7 +45,7 @@ where
 
     /// Calculates the difference between two state vectors.
     pub fn compare(&self, other: &LogHeights<A, L>) -> LogRanges<A, L> {
-        compare(other, &self.state)
+        compare_logs(other, &self.state)
     }
 
     /// Advances the state of a specific log.
--- a/p2panda-core/src/logs.rs
+++ b/p2panda-core/src/logs.rs
@@ -37,15 +37,56 @@ pub trait LogId: Clone + Eq + Ord + StdHash + Serialize + for<'de> Deserialize<'
 impl<T> LogId for T where T: Clone + Eq + Ord + StdHash + Serialize + for<'de> Deserialize<'de> {}
 
 /// Sequence number of an entry in an append-only log.
+///
+/// ```text
+/// [0] <- [1] <- [2]
+///                ^
+///            log height
+/// ```
+///
+/// To express the state vector of a log we can use the _log height_ (also HEAD or "frontier") which
+/// is a `SeqNum`.
 pub type SeqNum = u32;
 
 /// Author logs mapping.
+///
+/// Each log is identified with `L` and grouped by author `A`.
+///
+/// ```text
+/// Author "panda":
+/// - "trees" log
+/// - "animals" log
+///
+/// Author "icebear":
+/// - "trees" log
+/// ```
 pub type Logs<A, L> = BTreeMap<A, Vec<L>>;
 
-/// Map of log heights grouped by author.
+/// Map of [`SeqNum`] log heights grouped by author `A`. Each log is identified with `L`.
+///
+/// ```text
+/// Author "panda":
+/// - "trees" log: 6
+/// - "animals" log: 12
+///
+/// Author "icebear":
+/// - "trees" log: 4
+/// ```
 pub type LogHeights<A, L> = BTreeMap<A, BTreeMap<L, SeqNum>>;
 
-/// Map of log ranges grouped by author.
+/// Map of log ranges, grouped by author `A` and each log identified with `L`.
+///
+/// Log ranges are used to express the _difference_ between two [`LogHeights`] state vectors. Use
+/// the [`compare_logs`] method to compute it.
+///
+/// ```text
+/// Author "panda":
+/// - "trees" log: [0..6]
+/// - "animals" log: [4..12]
+///
+/// Author "icebear":
+/// - "trees" log: [3..4]
+/// ```
 pub type LogRanges<A, L> = BTreeMap<A, BTreeMap<L, (Option<SeqNum>, Option<SeqNum>)>>;
 
 /// Compare two sets of logs (local and remote) and calculate the "diff" representing ranges of
--- a/p2panda-core/src/logs.rs
+++ b/p2panda-core/src/logs.rs
@@ -66,7 +66,7 @@ pub type LogRanges<A, L> = BTreeMap<A, BTreeMap<L, (Option<SeqNum>, Option<SeqNu
 /// them to the remote. If both local and remote replicas do this then they will arrive at the
 /// same state. If pruned logs are being replicated and a range has been returned from this
 /// method, then it is expected only the remaining "frontier" will be replicated for each log.
-pub fn compare<A, L>(local: &LogHeights<A, L>, remote: &LogHeights<A, L>) -> LogRanges<A, L>
+pub fn compare_logs<A, L>(local: &LogHeights<A, L>, remote: &LogHeights<A, L>) -> LogRanges<A, L>
 where
     A: Author,
     L: LogId,
@@ -124,7 +124,7 @@ where
 mod tests {
     use std::collections::BTreeMap;
 
-    use crate::logs::compare;
+    use crate::logs::compare_logs;
 
     type Author = u8;
 
@@ -137,7 +137,7 @@ mod tests {
     fn both_empty() {
         let local: BTreeMap<Author, BTreeMap<u32, u32>> = BTreeMap::new();
         let remote: BTreeMap<Author, BTreeMap<u32, u32>> = BTreeMap::new();
-        let result = compare(&local, &remote);
+        let result = compare_logs(&local, &remote);
         assert!(result.is_empty());
     }
 
@@ -149,7 +149,7 @@ mod tests {
 
         let remote: BTreeMap<Author, BTreeMap<u32, u32>> = BTreeMap::new();
 
-        let result = compare(&local, &remote);
+        let result = compare_logs(&local, &remote);
         let needs = result.get(&ALICE).unwrap();
 
         assert_eq!(needs.get(&1), Some(&(None, Some(5))));
@@ -164,7 +164,7 @@ mod tests {
         let mut remote = BTreeMap::new();
         remote.insert(ALICE, BTreeMap::from([(1, 5)]));
 
-        let result = compare(&local, &remote);
+        let result = compare_logs(&local, &remote);
         let needs = result.get(&ALICE).unwrap();
 
         assert_eq!(needs.get(&2), Some(&(None, Some(10))));
@@ -179,7 +179,7 @@ mod tests {
         let mut remote = BTreeMap::new();
         remote.insert(ALICE, BTreeMap::from([(1, 10)]));
 
-        let result = compare(&local, &remote);
+        let result = compare_logs(&local, &remote);
         let needs = result.get(&ALICE).unwrap();
 
         assert_eq!(needs.get(&1), Some(&(Some(10), Some(20))));
@@ -193,7 +193,7 @@ mod tests {
         let mut remote = BTreeMap::new();
         remote.insert(ALICE, BTreeMap::from([(1, 30)]));
 
-        let result = compare(&local, &remote);
+        let result = compare_logs(&local, &remote);
         assert!(result.is_empty());
     }
 
@@ -205,7 +205,7 @@ mod tests {
         let mut remote = BTreeMap::new();
         remote.insert(ALICE, BTreeMap::from([(1, 20)]));
 
-        let result = compare(&local, &remote);
+        let result = compare_logs(&local, &remote);
         assert!(result.is_empty());
     }
 
@@ -217,7 +217,7 @@ mod tests {
         let mut remote = BTreeMap::new();
         remote.insert(ALICE, BTreeMap::from([(1, 5)]));
 
-        let result = compare(&local, &remote);
+        let result = compare_logs(&local, &remote);
         let needs = result.get(&ALICE).unwrap();
 
         assert_eq!(needs.get(&2), Some(&(None, Some(10))));
@@ -234,7 +234,7 @@ mod tests {
         let mut remote = BTreeMap::new();
         remote.insert(ALICE, BTreeMap::from([(1, 5)]));
 
-        let result = compare(&local, &remote);
+        let result = compare_logs(&local, &remote);
         let needs = result.get(&BOB).unwrap();
 
         assert_eq!(needs.get(&1), Some(&(None, Some(5))));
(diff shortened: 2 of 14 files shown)


adzialocha:
2026-09-10-133906_873x219_scrot

Rust doc spits out broken links for SeqNum and LogId across the docs for me and I'm not sure how to fix that.

p2panda-core/src/logs.rs
```diff
@@ -37,15 +37,56 @@ pub trait LogId: Clone + Eq + Ord + StdHash + Serialize + for<'de> Deserialize<'
...
+/// - "trees" log: 6
+/// - "animals" log: 12
+///
+/// Author "icebear":
+/// - "trees" log: 4
+/// ```
 pub type LogHeights<A, L> = BTreeMap<A, BTreeMap<L, SeqNum>>;
 
-/// Map of log ranges grouped by author.
+/// Map of log ranges, grouped by author `A` and each log identified with `L`.
```
mycognosist:
```suggestion
/// Map of log ranges grouped by author `A`. Each log is identified with `L`.
```

Small change to match the formatting of `LogHeights`.

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
-use futures_util::stream::{self, BoxStream};
+//! Methods for building sync protocols using p2panda's [append-only log](p2panda_core::operation)
+//! for a wide range of network topologies or transports.
```
mycognosist:
```suggestion
//! for a wide range of network topologies and transports.
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//!
+//! This module contains all essential methods, types and helpers for querying local log state from
+//! the store, computing and comparing log state vectors, validating operation log integrity,
+//! out-of-order buffering, reading and writing operations and more.
```
mycognosist:
```suggestion
//! buffering out-of-order operations, reading and writing operations and more.
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//!
+//!               [Sync Protocol ends]
+//!
+//! => Node A & B have the same state now: [A] [B] [C] [D] [E]
+//! ```
+//!
+//! The local state of a node is called a _replica_. It can be seen as a subset of "all data" in the
+//! network. With a sync protocol replicas want to converge to the same state. This underlying
+//! characteristic we call _eventual consistency_ which essentially means: All replicas will
+//! converge to the same state after "some" time, of course given that they had a chance to exchange
```
mycognosist:
```suggestion
//! converge to the same state after "some" time, of course given that they have a chance to exchange
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! data at some point. Eventual consistency is one of the core guarantees of any local-first system
+//! and therefore we always need a sync protocol.
+//!
+//! A naive, very spammy sync protocol would send everything all the time and eventually everyone
+//! would have received everything. To make things more efficient, replicas usually compute and
+//! exchange so called _state vectors_ first, informing others about their local replica state. The
```
mycognosist, line 66: Not super important for this PR but do you have a source for this usage of "state vector"? I've not been able to find any reference but I haven't looked at the computer science / distributed systems literature very closely.

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! receiving nodes compute the difference between both replica's sets and send over only the
+//! missing data. If all replicas reached the same state, no data is exchanged anymore.
```
mycognosist:
```suggestion
//! missing data. Once all replicas have reached the same state, no data is exchanged anymore.
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! ordering](https://docs.rs/p2panda-stream/latest/p2panda_stream/orderer/index.html) across
+//! multiple logs, [log-prefix
+//! pruning](https://docs.rs/p2panda-stream/latest/p2panda_stream/log_prune/index.html) or
+//! [Byzantine-Fault tolerance](https://arxiv.org/abs/2307.08381).
+//!
+//! We've chosen this to minimise messaging overhead (one message to exchange the state vector),
+//! enable efficient state comparison (constant time), and support sync protocols that do not assume
+//! a point-to-point (unicast) connection, especially where message throughput may be limited.
+//!
+//! Additionally logs represent well the event sourcing aspects of peer-to-peer applications due to
```
mycognosist:
```suggestion
//! Additionally, logs neatly represent the event sourcing aspects of peer-to-peer applications due to
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//!                      vs.
+//!
+//! Sync + Transport     -->     Insertion, etc.
+//!   + Validation
+//! ```
+//!
+//! This allows us to clearly separate these concerns for maximum flexibility. If data is fully
+//! encrypted, for example in support nodes or store-and-forward buffers, this separation becomes
+//! necessary, since the sync layer might not be aware of any encryption secrets or the process runs
+//! on a separate device. However this separation implies that any higher processing layer needs to
```
mycognosist:
```suggestion
//! on a separate device. However, this separation implies that any higher processing layer needs to
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! inform the sync protocol about potential violations or errors on-the-fly.
+//!
+//! Broker nodes might want to validate already during sync if the data is readable to them, so they
+//! can be sure to only stream validated data further to clients.
+//!
+//! ### Fast-push vs. sync
+//!
+//! Sync can be seen as a strategy of _repairing_ messages we've missed due to being offline or
```
mycognosist: Hmm I'm not sure about repairing here. Maybe if you extend the sentence to read: "Sync can be seen as a strategy of _repairing_ logs by requesting and receiving message we've missed...". But it still feels a bit off to me. I think I lean more in favour of the catching-up analogy.

mycognosist: To extend this a bit: repair is used when something has been damaged or broken. In the case of sync, an out-of-date log is not broken; it is fine just as it is (perfectly valid and with integrity), but it may just not contain the latest data.

mycognosist: I could see repair being used in a case where you have out-of-order operations _in addition to_ an intact log; in that case you have a log with holes in it and you want to make it complete by filling-in the missing operations. But to me this is not the same as simply getting the latest operations to advance the state of an intact log.

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! In reliable multicast networking protocols we can understand
+//! [NACK](https://en.wikipedia.org/wiki/NACK-Oriented_Reliable_Multicast) as a request to
+//! retransmit packets which have been missed and _not_ acknowledged yet. This is very similar to a
+//! sync protocol.
+//!
+//! NACKing comes with its own issues: It can lead to spamming the network when multiple
+//! participants start announcing what they've missed and everyone starts streaming back the same
+//! missing operations until the network breaks down. This is what we call a _NACK-implosion_.
+//!
+//! For some protocols we want to mitigate these issues with _not_ responding based on a density
```
mycognosist:
```suggestion
//! For some protocols we want to mitigate these issues by _not_ responding based on a density
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! factor `d` of network participants or similar, however this implies that sync can become slower.
```
mycognosist: Maybe this paragraph could be expanded slightly:

For some protocols we want to mitigate these issue by _not_ always responding. One way of achieving this is to calculate a response probability in relation to the number of nearby network participants (ie. a density factor `d`); the greater the number of neighbours, the lower the chance of rebroadcasting a received message. Any chosen mitigation strategy must take into account the trade-off between the robustness, availability and speed of message propagation.

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//!
+//! ### Out-of-order buffering
+//!
+//! Not every transport will guarantee an ordered connection (such as TCP) and data might arrive
+//! out-of-order. This can also happen if we eagerly publish operations without syncing first (see
+//! _Fast-push_).
+//!
+//! For this we want to buffer operations until the log-integrity is intact again. Usually we want
+//! to keep this buffer only in-memory with a fixed bound to prevent attacks of maliciously
```
mycognosist:
```suggestion
//! to keep this buffer only in-memory with a fixed bound to prevent attacks which maliciously
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! exhausting system resources with out-of-order data. In some scenarios with extreme
```
mycognosist:
```suggestion
//! exhaust system resources with out-of-order data. In some scenarios with extreme
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! This means that we need a form of authenticity guarantee (digital signature) for each message
+//! delivered over the network so we can reliably verify the authorship and origin of them.
+//!
+//! Mitigating spam comes as a trade-off of metadata leakage as we have to introduce some sort of
+//! device id (or similar) and signing scheme. Sync protocol designers should try to separate these
+//! identifiers from the ones used to identify operation authors.
+//!
+//! ### Store-and-forward
+//!
+//! Sometimes nodes need to sync data they are _not interested_ in, meaning that they don't know the
```
mycognosist: This paragraph could benefit from adding the "why" aspect. Why do nodes sometimes need to sync data they're not interested in?

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! need to help syncing data for others. This can be achieved with a _store-and-forward_ approach,
+//! while it is up to the implementer to decide how much and how long (ring-buffer, expiry date, ..)
+//! and for whom they want to store (per topic, device, access via capability, ..) unknown data.
+//!
+//! Store-and-forward buffers do not always need to reside next to the node but can also be
+//! external, such as USB-sticks or p2panda support nodes which behave more like external _mailbox_
+//! servers.
+//!
+//! Ideally all data is fully encrypted and its shape and contents oblivious to anyone. Usually we
+//! can't leverage any "data type tricks" such as log-height sync in these cases or need to build
```
mycognosist:
```suggestion
//! can't leverage any "data type tricks", such as log-height sync, in these cases or we need to build
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! hybrids where some information about the base data type (backlink, etc.) is still available in
+//! plaintext and readable by everyone.
+//!
+//! As an example we've implemented a data-type-agnostic, generic store-and-forward solution named
+//! [MemoryLAN](https://github.com/p2panda/memorylan) based on a bounded, probabilistic
+//! Cuckoo-Filter sync protocol for unknown data which can be integrated as a base layer for any
```
mycognosist:
```suggestion
//! Cuckoo-Filter sync protocol for unknown data. This can be integrated as a base layer for any
```

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! sync protocol on top of a mesh.
+//!
+//! ### Bandwidth & message throughput
+//!
+//! Transports come with limitations, for example a maximum packet size of 255 bytes while
```
mycognosist: Suggested change:

Some transports enforce limitations in the form of maximum packet size and / or broadcast rate. For example, a radio-based transport may restrict us to packets smaller than 255 bytes and may take several seconds to transmit.

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! transmitting these might take a couple of seconds. In these cases we want to be careful about
+//! the number of topics or logs we want to sync over.
+//!
+//! Some higher-layer protocol might be required to negotiate network channel capacity for a certain
```
mycognosist: Suggested change:

In order to deal with these limitations, some higher-level protocol(s) may be required. This could take the form of a negotiation protocol allowing channel capacity to be assigned on a per-topic basis. Another protocol might ensure that sync happens lazily (on demand) to ensure a minimum of unnecessary messages being broadcast. Finally, one could compact the payload size of state vector messages with the help of probabilistic filters (ie. bloom-, cuckoo-filters etc.).

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,540 @@
...
+//! topic, or only sync lazily on demand, only sending messages when it is really requested (rather
+//! than eagerly pushing into the void) or compacting state vectors with the help of probabilistic
+//! filters (bloom-, cuckoo-filters, etc.).
+//!
+//! Transports will require some sort of message framing in case the limits have been reached with
```
mycognosist:
```suggestion
//! Transports might require some sort of message framing in case the limits have been reached with
```

mycognosist (approved):
Brilliant summary of the whole sync topic (see what I did there?)! I've left some suggested changes and a question or two but those are all very minor things. It could be merged as-is and I'd still be fine with it.

p2panda-sync/src/api.rs
adzialocha: I don't have any sources either, mostly just from picking up conversations with other p2p folks.

The term though describes precicely what's going on: It _is_ a vector describing a state :-)

adzialocha:
Thank you @mycognosist for the corrections and improvements, I've applied all of them or combined them at least with the existing sentences.

adzialocha:
> Rust doc spits out broken links for SeqNum and LogId across the docs for me and I'm not sure how to fix that.

I've manually added the links and supressed the clippy warnings, it works now.

p2panda-core/src/cursor.rs
```diff
@@ -3,7 +3,7 @@
 //! State vector to track and compare logs.
 use std::hash::Hash as StdHash;
 
-use crate::logs::{LogHeights, LogId, LogRanges, SeqNum, compare};
+use crate::logs::{LogHeights, LogId, LogRanges, SeqNum, compare_logs};
```
sandreae, line 6: Like the renaming 👍

p2panda-sync/src/api.rs
```diff
@@ -1,18 +1,562 @@
...
+//!
+//! In order to deal with these limitations, some higher-level protocols may be required. This could
+//! take the form of a negotiation protocol allowing channel capacity to be allocated on a per-topic
+//! basis. Another protocol might ensure that sync happens lazily (on demand) to ensure a minimum of
+//! unnecessary messages being broadcast. Finally, one could compact the payload size of state
+//! vectors with the help of probabilistic filters (ie. bloom-, cuckoo-filters etc.).
+//!
+//! Transports might require some sort of message framing in case the limits have been reached with
+//! one single message, however this is out of scope for a sync protocol and rather a transport
+//! concern.
```
sandreae, line 542: Wow, quite amazing docs! Reads really well 👏 it could be nice to add one additional section about access controls, pointing out that the boundaries can be established at various points (in the sync protocol vs. in the processor) and with whichever approach is suitable for the context (tokens exchanged before sync, ACL-CRDT maintained locally, hard-coded allow list).

sandreae (approved):
Amazing docs 😄 well done! I had a suggestion for adding one more section, but it's not a blocker. See what you think.

p2panda-sync/src/api.rs
adzialocha: Good idea! I'll add this

PR #1413, opened 2026-09-09
Update `p2panda-sync` examples

sandreae:
- [x] Use methods from p2panda-sync API module
- [x] Use methods and types from `common` module to reduce LOC in examples
- [x] Rename `sync_w_live_mode.rs` to `unicast_w_live_mode.rs` to better relate to `broadcast` example
- [x] Expand `unicast` doc strings to provide context and describe protocol
- [x] Change `node_a` and `node_b` to `alpaca` and `bobcat` in `unicast` example to improve readability
- [x] Change `node_a` and `node_b` to `panda` and `sloth` in `sneakernet` example to improve readability


p2panda-sync/examples/broadcast.rs
```diff
@@ -370,13 +366,26 @@ impl Node {
...
+        // means associating a log (in this case our own) with the topic we wish to subscribe to.
+        self.store
+            .associate(
+                &topic,
+                &self.signing_key.verifying_key(),
+                &Hash::digest(topic.as_bytes()),
+            )
+            .await?;
+
+        self.store.commit(permit).await?;
```
adzialocha, line 382: I'm not sure that's a pattern we want to recommend. The TopicStore and Topic -> Log association is to associate topics with existing (!) logs, not to express interest in a topic.

sandreae: I think we do actually associate non-existing logs in `p2panda` (key bundle and group logs for other authors), but still I agree that the pattern used in this example is not something we want to recommend.

I think then we should go back to the original approach of having a separate "topics of interest" set maintained out of the store. This is because some of the subscribing peers subscribe before publishing or ingesting, so there's no "automatic" mechanism there to make the required association in the topic store for store.topics() to work for us.

adzialocha:
Since the revert refers to the commit before of the same branch I believe we can just squash them into one.

PR #1417, opened 2026-09-10
Add `validate_operation` method checking against store

adzialocha:
## 📋 Checklist

### DIFF
--- a/p2panda-stream/src/ingest/operation.rs
+++ b/p2panda-stream/src/ingest/operation.rs
@@ -1,11 +1,9 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 //! Methods to handle p2panda operations.
-use std::borrow::Borrow;
-
 use p2panda_core::prune::validate_prunable_backlink;
 use p2panda_core::{
-    AnyOperation, Extensions, Hash, LogId, Operation, SeqNum, VerifyingKey, validate_operation,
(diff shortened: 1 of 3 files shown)


p2panda-stream/src/ingest/operation.rs
```diff
@@ -38,13 +36,18 @@ pub enum IngestResult<E> {
 /// Checks an incoming operation to ensure correct formatting and log integrity before persisting it
 /// into the store when valid. This function is idempotent; duplicate operations are ignored.
 ///
+/// See [`validate_operation`] for an alternative method to validate an operation without
```
mycognosist, line 39: This links to `validate_operation` from `p2panda-stream` but I believe we want to link to the version from `p2panda-core` (without persistence).

mycognosist (approved):
One small documentation change required but otherwise looks good!

p2panda-stream/src/ingest/operation.rs
adzialocha: Oh, no, it should link to the one in the same module. This one doesn't persist (write), it just looks up in the database (read). I can rephrase that if that was confusing.

mycognosist: Ah of course. I see it now. I think it's just my brain that's confused; the docs are fine.

PR #1419, opened 2026-09-14
bugfix: when decoding node-info row, don't fail the whole query if it doesn't decode

### LINKED ISSUE #1418: 0.7.1 introduces breaking change in node-info decoding that is very hard to debug
There's a slight change in how node info is decoded in the address book. It was introduced in commit `e1a634d1`. This replaced ciborium with cbore-core. Although these still encode cbor the same way, the old ciborium implementation was more lenient. The new approach is raises an `Error::NonDeterministic` in certain circumstances.

This doesn't effect the decoding of node info data that is encoded from this point on, but address books from before this change will have the key order sorted by encoded key length, rather than lexicographically.

This is a breaking change, which is a bit of a surprise in a patch release. However, even if the breaking chance was desirable, the real problem is that it manifests in really surprising ways. I initially thought that it was an instance of https://github.com/p2panda/p2panda/issues/957 because it seems to **usually** (non-deterministically) cause node spawn to panic.

My comment on that issue includes my demonstration repo, at https://github.com/jadehopepunk/p2panda-071-spawn-repro which minimally demonstrates this bug.

The problem is somewhat hidden, because the decoding occurs in an sqlx row handler function, so a single node info with data that fails to decode will break the entire address-book loading. 

So, **even if** we wanted a breaking change here (and I certainly don't), having the break occur as a non-deterministic node spawn panic, that gives very little indication of where the error is, is a pretty bad experience. I've gone ahead and addressed this with a PR, and my primary goal was the make the error clear and deterministic, rather than to make any changes to encoding.

jadehopepunk:
Closes https://github.com/p2panda/p2panda/issues/1418

In this change, the goal was the surface the node info decode error clearly and avoid a panic, rather than to fix it. The sqlx::Decode function does return a result object, and so conceptually an error from that would be nice, but since it's called from within the sqlx query that gives us very little control (everything will fail). 

Instead, I have opted to surface the error in two ways:

1. I emit a `tracing::warn!`. This introduces tracing as a dependency of the sqlite feature on the p2panda-store crate, but that seems to be a blessed path that's true all through p2panda.
2. I return `Ok(NodeInfoDecode::Undecodable)` (a success result but with an enum) to allow the query to continue. This enum is only a boolean state. I tried it with `<Option>` but it became confusing in calling code telling which optional element meant what. Also, this enum allows us to differentiate between failure and missing if needed. It's only used fairly internally, as these node infos that failed to decide are stripped out for the public api.

I *think* that not only will this give us a clear error message and avoid a panic, but it will actually be somewhat self-healing. I'm not totally sure about the latter, but certainly it wont return a node info from the address book that it can't decode, so the show can go on. 

## 📋 Checklist


adzialocha:
I'd suggest instead of ignoring rows we could adjust the cbor decoder config to be less strict, at least for non-operation data when writing / reading from the database.

jadehopepunk:
@adzialocha that might be a good idea, and would fix this specific problem. However, it wouldn't address failures during cbor decoding in the sqlx row in general. There could be other kinds of failures, and while we don't necessarily need graceful continuation, we certainly need a clear and deterministic error.

Another option would be to move cbor decoding out of the sqlx decode function. Have the query return the undececoded cbor bytes, and keep the cbor decoding nice and seperate. That conceptually makes a lot of sense (but at first glance seems to make the code a bit uglier).

And finally, I can see that the deterministic encode/decode could be nice for tests. So perhaps there's an upside to this strictness.

adzialocha:
> @adzialocha that might be a good idea, and would fix this specific problem. However, it wouldn't address failures during cbor decoding in the sqlx row in general. There could be other kinds of failures, and while we don't necessarily need graceful continuation, we certainly need a clear and deterministic error.

Yeah, I can see how this is a separate issue and we should improve the error reporting. Will think about your suggestion todetach decoding from sqlx. I believe we do this already but it got "optimized" to be one method in this case.

> And finally, I can see that the deterministic encode/decode could be nice for tests. So perhaps there's an upside to this strictness. 

Yes, for operations having deterministic hashes is great, in general having unambiguous encoding is helpful.

But I can see how that bug sneaked in here since for just moving state into the database we don't need this strictness.

Users of our encoder will still get the nice determinism, but we don't want to break necessarily here because of outdated contents since it's just store related (nobody consumes this but us).

My suggestion is:

1. I'll make a PR fixing the bug by making store decoding generally less strict (operation format stays strict)
2. We continue in this PR to improve error reporting

What do you think?

jadehopepunk:
@adzialocha that sounds like a good approach. I'm happy with the less strict approach. Once that's in this immediate problem will go away, but I'm very happy to continue development on this PR to improve error reporting.

jadehopepunk:
@adzialocha I know you said you'd do the "less strict decoding" PR... but also I've done it in case it's a help: https://github.com/p2panda/p2panda/pull/1428

